PR #10, opened 2021-03-19
Add extended ID validity check

ileasile:
Here we generate enums containing keywords.
We use Kotlin compiler artifact as a source of keywords and kotlinpoet library for generation.

Note that dataframe artifact remains leightweight after all, only several enums are added to it, and no additional production dependencies.


ileasile:
Also, maybe it's better to use KotlinPoet for code generation?

PR #11, opened 2021-04-08
Fixed converter and added class check for wirteClass

devcrocod:


nikitinas:
All changes were applied in master

PR #22, opened 2021-05-10
Add display configuration

ileasile:
Works like this:
image

@nikitinas @UnlikeMars, you may suggest another naming for `dataFrameConfig` if you wish


ileasile:
Was merged manually: 4d126c87e6273d52ca1a7c99110683e8ea44f19b

PR #93, opened 2022-03-06
Support reading Arrow `.feather` file using Apache Arrow

### LINKED ISSUE #78: Arrow Support
Hi, I can't find that `dataframe` supports Arrow as internal serialization / backend.

Is this something which you're working on?

Jimexist:
Related to #78


koperagen:
Hi. Thank you for the PR! I will review parts related to the code generation. It's holidays, so we'll have to wait a little for @nikitinas to review the rest before merge.

src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/DefaultReadDfMethods.kt
```diff
@@ -9,6 +9,28 @@ private const val cast = "cast"
...
+        return buildString {
+            if (path != null) {
+                append("""        ${visibility}const val defaultPath: String = "${path.escapeStringLiteral()}"""")
+                appendLine()
+            }
+            val defaultPathClause = if (path != null) " = defaultPath" else ""
+            append(
+                """
+                |        ${visibility}fun $readArrow(path: String$defaultPathClause, $verify: Boolean? = null): DataFrame<$markerName> { 
+                |            val df = DataFrame.$readArrow(path)
```
koperagen, line 25: Add import for readArrow method here:
https://github.com/Kotlin/dataframe/blob/9d28735112c034b7d56791660a19b8746fcf44c3/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt#L291
It probably should be that each read method declares its own imports, but i think supporting it would require a lot of changes. So let's leave it out of scope of this PR :)

PR #94, opened 2022-03-09
Support automatic TSV format detection in DataFrame.read

PoslavskySV:


koperagen:
Hi! There are two compilation errors:
1. In line SupportedFormats.TSV -> readTSV(url). `readTSV` overload for url doesn't exist. Let's add it
2. In file SchemaReader.kt, in method `getReadDfMethod` `SupportedFormat.TSV` branch is missing from when. We need an implementation of `DefaultReadDfMethod` that is similar to `DefaultReadJsonMethod`, but calls `readTSV` instead of `readJson`

PoslavskySV:
Huh, sorry guys! I tried to add changes in a "blind" mode, since I couldn't manage to load the project in Intellij or compile it from gradle - seems I need to install some plugins or something. I'll try again and will fix the problems you mentioned!

koperagen:
It should work without plugins or anything, so don't hesitate to ask if the problem persists

PR #110, opened 2022-05-18
Explicit ArrowFormat parameter

Kopilov:
Because of Arrow has two encoding types (Random Access aka Feather and IPC Stream) both of which can be accessed via File, URL and other ways, including raw existing Channel (functions were converted to public), and we should have methods for reading any of them.

Also this patch includes Boolean (BitVector) support and Arrow version upgrade

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -17,7 +17,7 @@ kotlinDatetime = "0.3.1"
 junit = "4.13.2"
 kotestAsserions = "4.6.3"
 jsoup = "1.14.3"
-arrow = "7.0.0"
+arrow = "8.0.0"
 
 [libraries]
 ksp-gradle = { group = "com.google.devtools.ksp", name = "symbol-processing-gradle-plugin", version.ref = "ksp" }
--- a/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrow.kt
+++ b/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrow.kt
@@ -2,6 +2,7 @@ package org.jetbrains.kotlinx.dataframe.io
 
 import org.apache.arrow.memory.RootAllocator
 import org.apache.arrow.vector.BigIntVector
+import org.apache.arrow.vector.BitVector
 import org.apache.arrow.vector.Decimal256Vector
 import org.apache.arrow.vector.DecimalVector
 import org.apache.arrow.vector.DurationVector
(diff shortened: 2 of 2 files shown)


koperagen:
Hi! Thank you for the contributions :) I'm currently moving arrow support to separate module and re-working generic `read` a little. I expect no conflicts with your changes, so let's merge this a bit later.  
I want to understand to understand choice of defaults for `readArrow(url: URL, format: ArrowFormat = ArrowFormat.IPC)`, `readArrow(path: String, format: ArrowFormat = ArrowFormat.IPC)`. Is it common to use IPC with URLs? What will happen if someone tries to `readArrow("/path/to/local/file")` with default format?

Kopilov:
Hello @koperagen,

For URL I had used IPC as default because the original code used IPC for URL (`url.openStream().use` -> `readArrow(stream: InputStream)` ->  `readArrow(channel: ReadableByteChannel` -> `ArrowStreamReader`).

If someone tries to readArrow("/path/to/local/file") with default format, the result depends on type of that file :) Was it created with `ArrowStreamWriter` or `ArrowFileWriter` (both of them use `Channel` as sink).

Actually, IPC does not have metadata (that Feather has) and allow to read and write different dataframes one by one. IPC is better for real-time communication (if URL provides not static result but server-provided data stream) and  it's reading in future should support lambda subscriber pattern. Nevertheless, local file can also be encoded in IPC.

For example, in [contribution](https://github.com/jbytecode/rcaller/blob/master/RCaller/src/main/java/com/github/rcaller/io/arrow/ArrowImpl.java) to RCaller project I used IPC for supporting multiple values returning but the data was actually stored in the file (accessed by URI for flexibility).

In R `arrow::read_arrow` manual we can find that it is
> a wrapper around ‘read_ipc_stream()’ and ‘read_feather()’, is deprecated. You should explicitly choose the function that will read the desired IPC format (stream or file) since a file or ‘InputStream’ may contain either

According to this recomendation, we can remove default parameters at all. I left them to make contribution more backward-compatible.

koperagen:
I didn't know that original URL version is for IPC streams :) I tried to read feather with it and got an OutOfMemoryError, but thanks to your comment i quickly understood why. To fix this problem i've included this commit in https://github.com/Kotlin/dataframe/pull/115. Hope it's okay. 
About readArrow: it would be indeed better to follow this recomendation and have explicit readArrowIPC and readArrowFeather, so i removed readArrow

Kopilov:
@koperagen, nice!

PR #111, opened 2022-05-19
Excel IO improvements

Kopilov:
* Numeric header support
 * Save date and datetime
 * Allow using opened Workbook in-memory

Also this patch includes assemble fixes (same as in  #110, moved to separate commit).


koperagen:
Hi! Do you have some use case for writing to in-memory WorkBook using function that returns `Sheet` in mind? It would be nice to add an example to the documentation. Can you do it, please? :) It's a little bit tricky:
1. Go to `tests` module, in `package org.jetbrains.kotlinx.dataframe.samples.api`, `Write.kt`
2. Add a test with an arbitrary name, let's say `writeExcelInMemory`, wrap sample code in `// SampleStart` and `// SampleEnd`, like `writeJsonStr` does
3. Go to `write.md`, add this in `Writing spreadsheets` paragraph at the end:
```
<!---FUN writeExcelInMemory-->
<!---END-->
```
4. Launch `./gradlew korro`. You'll see that sample code from the test is now pasted in `write.md`

Kopilov:
Take a look :)

Issue #9, opened 2021-03-18
Support digit grouping separators on import


belovrv:
Support numbers like 1,593.23 or 10 233 000 etc
See: https://docs.oracle.com/cd/E19455-01/806-0169/overview-9/index.html

Jolanrensen:
We have support for `java.text.NumberFormat` locales. In my tests, it appears locales seem to work fine as long as there are no spaces in the number. This appears to be a limitation of `NumberFormat` as of now but might be fixed in newer Java versions.

Example:
```kotlin
val df = dataFrameOf("a")("4.294.967.295,000", "4.294.967.295,000", "4.294.967.295,000")
    .convert { column<String>("a") }.to {
        it.convertToDouble(Locale.GERMAN)
    }

df.print(columnTypes = true, title = true, borders = true)
//    ⌌--------------⌍
//    |  |   a:Double|
//    |--|-----------|
//    | 0| 4294967295|
//    | 1| 4294967295|
//    | 2| 4294967295|
//    ⌎--------------⌏
```

Locale is also an option anytime `ParseOptions` are available, which appears to be in TSV, CSV reading etc.

Closing for now :)

PR #133, opened 2022-08-01
More Converting operations

Kopilov:
When working with specific local settings, we still can get data in international format. And it should be parsed rather than rejected. Here we have fallback on parsing doubles with POSIX number format. (This was made for `convertTo` function but also works on primary parsing. For example, `readCsvWithFloats` test now passes on my computer without switching to empty locale manually.) May be, we should have similar logic in date, time and some other parsers.

Also some new converters (to Boolean and LocalTime) have been added.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
@@ -30,7 +30,7 @@ import org.jetbrains.kotlinx.dataframe.io.toDataFrame
 import java.math.BigDecimal
 import java.net.URL
 import java.time.LocalTime
-import java.util.*
+import java.util.Locale
 import kotlin.reflect.KProperty
 import kotlin.reflect.KType
 import kotlin.reflect.typeOf
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
@@ -16,6 +16,7 @@ import org.jetbrains.kotlinx.dataframe.RowValueExpression
 import org.jetbrains.kotlinx.dataframe.columns.ColumnReference
 import org.jetbrains.kotlinx.dataframe.dataTypes.IFRAME
 import org.jetbrains.kotlinx.dataframe.dataTypes.IMG
+import org.jetbrains.kotlinx.dataframe.exceptions.TypeConversionException
 import org.jetbrains.kotlinx.dataframe.impl.api.Parsers
 import org.jetbrains.kotlinx.dataframe.impl.api.convertRowColumnImpl
 import org.jetbrains.kotlinx.dataframe.impl.api.convertToTypeImpl
@@ -124,6 +125,48 @@ public fun <T : Any> DataColumn<T?>.convertToString(): DataColumn<String?> = con
 @JvmName("convertToDoubleFromT")
 public fun <T : Any> DataColumn<T>.convertToDouble(): DataColumn<Double> = convertTo()
 public fun <T : Any> DataColumn<T?>.convertToDouble(): DataColumn<Double?> = convertTo()
+
+/**
+ * Parse String column to Double considering locale (number format).
+ * If [locale] parameter is defined, it's number format is used for parsing.
+ * If [locale] parameter is null, the current system locale is used. If column can not be parsed, then POSIX format is used.
+ */
+@JvmName("convertToDoubleFromString")
+public fun DataColumn<String>.convertToDouble(locale: Locale? = null): DataColumn<Double> {
+    if (locale is Locale) {
+        val explicitConverter = Parsers.getDoubleConverter(locale) as (String) -> Double?
+        return map { explicitConverter(it.trim()) ?: error("Can't convert `$it` to Double") }
+    } else {
+        return try {
+            val defaultConverter = Parsers.getDoubleConverter() as (String) -> Double?
+            map { defaultConverter(it.trim()) ?: error("Can't convert `$it` to Double") }
+        } catch (e: TypeConversionException) {
+            val posixConverter = Parsers.getDoubleConverter(Locale.forLanguageTag("C.UTF-8")) as (String) -> Double?
+            map { posixConverter(it.trim()) ?: error("Can't convert `$it` to Double") }
+        }
+    }
+}
+
+/**
+ * Parse String column to Double considering locale (number format).
+ * If [locale] parameter is defined, it's number format is used for parsing.
+ * If [locale] parameter is null, the current system locale is used. If column can not be parsed, then POSIX format is used.
+ */
+@JvmName("convertToDoubleFromStringNullable")
+public fun DataColumn<String?>.convertToDouble(locale: Locale? = null): DataColumn<Double?> {
+    if (locale is Locale) {
+        val explicitConverter = Parsers.getDoubleConverter(locale) as (String) -> Double?
+        return map { it?.let { explicitConverter(it.trim()) ?: error("Can't convert `$it` to Double") } }
+    } else {
+        return try {
+            val defaultConverter = Parsers.getDoubleConverter() as (String) -> Double?
+            map { it?.let { defaultConverter(it.trim()) ?: error("Can't convert `$it` to Double") } }
+        } catch (e: IllegalStateException) {
+            val posixConverter = Parsers.getDoubleConverter(Locale.forLanguageTag("C.UTF-8")) as (String) -> Double?
+            map { it?.let { posixConverter(it.trim()) ?: error("Can't convert `$it` to Double") } }
+        }
+    }
+}
 
 @JvmName("convertToFloatFromT")
 public fun <T : Any> DataColumn<T>.convertToFloat(): DataColumn<Float> = convertTo()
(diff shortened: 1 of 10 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/parse.kt
```diff
@@ -171,18 +171,23 @@ internal object Parsers : GlobalParserOptions {
...
-                val result: Double? = format.parse(this, parsePosition)?.toDouble()
-                if (parsePosition.index != this.length) null
-                else result
+                fun parseWithFormat(format: NumberFormat): Double? {
+                    val parsePosition = ParsePosition(0)
+                    val result: Double? = format.parse(this, parsePosition)?.toDouble()
+                    return if (parsePosition.index != this.length) null
+                    else result
+                }
+                parseWithFormat(userNumberFormat) ?: parseWithFormat(posixNumberFormat)
```
koperagen, line 190: I wonder if it's possible to misinterpret some value this way?
Right now, in `parse`, we use one parser for a whole column. If a parser fails to produce a value, then we go to the next one. But here you try to parse first with the locale-specific one, then with POSIX.
row 1 = can parse with locale specific
row 2 = cannot parse with locale specific, fallback to POSIX
But then, maybe the whole column is in the POSIX format and we should use fallback parser for row 1 as well? Can it be that row 1 can be parsed with first and second parser but yield different results?

koperagen:
Hi! Thank you for the PR. Parsing numbers is quite an old problem here :) https://github.com/Kotlin/dataframe/issues/9

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/parse.kt
Kopilov: Do we consider that both formats are mixed up in one column? If yes, than it can.

For example (as it is shown in unit test), string like `"12,345"` would mean `12.345` in ru-RU locale and `12_345` in POSIX. On other side, string like `"12.345"` would be illegal for ru-RU parser but is often used in real life.

So if we have column like `columnOf("12.345", "67,890")` and Russian locale by default, it would be converted to values `12.345`, `67.89` in one case (parser per row) and to values `12.345`, `67_890` in another (parser per column).

Kopilov: But actually, if we have column with comma as thousands group delimiter (as in POSIX), like `columnOf("12,345", "67,890")` but ru-RU locale by default, it would be misinterpret anyway.

I do not have experience with other locales. There can be other kinds of controversies, may be.

koperagen: Hm, i think we should consider such columns if they exist in real world data, but I have too little experience. 
For example,  
`columnOf("12.345", "67,890")` is `12.345, 67_890` or `12.345, 67.89`? 

> But actually, if we have column with comma as thousands group delimiter (as in POSIX), like columnOf("12,345", "67,890") but ru-RU locale by default, it would be misinterpret anyway.
Valid point. So people would have to specify the format explicitly

Kopilov: > For example,
> `columnOf("12.345", "67,890")` is `12.345, 67_890` or `12.345, 67.89`?

In my case (dirty data or manual input without ple-validation) it should be parsed as `12.345`, `67.89`. Current dataframe version throws `TypeConversionException: Failed to convert '12.345' from kotlin.String to kotlin.Double` on Russian locale. And I tried to avoid this.

> Valid point. So people would have to specify the format explicitly

`DataFrame.readCSV` already has ParserOptions parameter. Should we add the same to `DataColumn.convertTo()`?

koperagen: Maybe `map { it.replace(",", ".").toDouble() }` could work for you? For now, i still think that per row parsers could create confusion. 

I like the general idea of a fallback to POSIX though. But with per column parsing, it won't solve your problem, right? So we need a use case where it's useful :)  

And yes, probably we need to have `locale` parameter in `DataColumn<String>.convetToDouble` and `Convert<>.toDouble` overloads, like `DataColumn<String>.convertToLocalDate` does

Kopilov:
Review current version, please.

Also do you think we should make rendering test locale-independent as was suggested [here](https://github.com/holgerbrandl/krangl/pull/144/files)?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/parse.kt
Kopilov: I have not found how to make `Convert<>.toDouble` with custom parameter. Code like 
```
public fun DataColumn<String>.convertTo<Double>(locale: Locale? = null): DataColumn<Double> {
    return this.convertToDouble(locale)
}
```
does not work...

Other requested features have been implemented.

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ParserTests.kt
```diff
@@ -77,18 +79,73 @@ class ParserTests {
...
+            columnMixed.convertToDouble(parsingLocaleUsesDot).shouldBe(columnOf(12.345, 67890.0))
+
+            shouldThrow<TypeConversionException> { columnDot.convertToDouble(parsingLocaleUsesComma) }
+            columnComma.convertToDouble(parsingLocaleUsesComma).shouldBe(columnOf(12.345, 67.89))
+            shouldThrow<TypeConversionException> { columnMixed.convertToDouble(parsingLocaleUsesComma) }
+
+            Locale.setDefault(Locale.forLanguageTag("ru-RU"))
+
+            columnDot.convertTo<Double>().shouldBe(columnOf(12.345, 67.89))
+            columnComma.convertTo<Double>().shouldBe(columnOf(12345.0, 67890.0))
```
koperagen, line 135: Shouldn't it be 12.345, 67.890? Looks suspicious that default convertTo works identically for en_US and ru_RU.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/parse.kt
```diff
@@ -271,6 +272,13 @@ internal object Parsers : GlobalParserOptions {
         ) else null
         return parser.applyOptions(options)
     }
+
+    internal fun getDoubleConverter(locale: Locale? = null): TypeConverter {
+        val options = if (locale != null) ParserOptions(
+            locale = locale
+        ) else null
+        return parserToDoubleWithOptions.toConverter(options)
```
koperagen, line 280: You can use `parserToDoubleWithOptions.applyOptions` here, i think

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
```diff
@@ -125,6 +126,48 @@ public fun <T : Any> DataColumn<T?>.convertToString(): DataColumn<String?> = con
...
+}
+
+/**
+ * Parse String column to Double considering locale (number format).
+ * If [locale] parameter is defined, it's number format is used for parsing.
+ * If [locale] parameter is null, the current system locale is used. If column can not be parsed, then POSIX format is used.
+ */
+@JvmName("convertToDoubleFromStringNullable")
+public fun DataColumn<String?>.convertToDouble(locale: Locale? = null): DataColumn<Double?> {
+    if (locale is Locale) {
```
koperagen, line 157: What do you think? 
```kotlin    
val converter = if (locale != null) {
    Parsers.getDoubleConverter(locale)
} else {
    Parsers.getDoubleConverter()
}
```

koperagen:
> Review current version, please.
> 
> Also do you think we should make rendering test locale-independent as was suggested [here](https://github.com/holgerbrandl/krangl/pull/144/files)?

Sure. I guess right now it's impossible to run all tests on a PC with locale that uses comma? Would be nice to fix it

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ParserTests.kt
Kopilov: Fixed, thanks! But I have not found a better way than calling `DataColumn<String>.convertToDouble` directly from universal `convertTo<>`.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/parse.kt
Kopilov: Done. Originally I used `toConverter` for getting TypeConversionException  to be thrown. Now it is thrown in `.convertToDouble` function and this is renamed to `getDoubleParser`.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Kopilov: And then simply call `map { converter(it) }`? This it doesn't seem to work because we wanted the scalar converter not to use fallback logic. And it is implemented here now…

Kopilov:
> Sure. I guess right now it's impossible to run all tests on a PC with locale that uses comma?

Until now I used `export LANG=C.UTF-8` command in bash session to avoid problems.

> Would be nice to fix it

Done. `Double` printing is unified by `DecimalFormatSymbols.getInstance().decimalSeparator` and `LocalDate` parsing is forced by using `en_US` explicitly.

Issue #143, opened 2022-08-08
Generation of Data Schemas from Json could be better


Jolanrensen:
Let's say I have the following sample data from the [Magical Location Clock app](https://mlc.jolanrensen.nl/help/group-data-api#h.oan9zw3gh0ca)
```json
{
  "0": {
    "name": "Kwijt 🤷",
    "people": [
      "Ronald",
      "Jolan"
    ],
    "peopleIds": [
      1544198659362550,
      1556728145307124
    ]
  },
  "164468523846839": {
    "name": "Joyce ❤️",
    "people": [],
    "peopleIds": [],
    "gps": {
      "latitude": 59.5391564418419,
      "longitude": 8.09191055862616
    }
  },
  "1644685267012957": {
    "name": "Joyce ❤️",
    "people": [],
    "peopleIds": [],
    "gps": {
      "latitude": 52.19508702169525,
      "longitude": 1.414078822875235
    }
  },
  "1659257255454626": {
    "name": "Camping 🏕️",
    "people": [
      "Pascale"
    ],
    "peopleIds": [
      1545419932562464
    ],
    "gps": {
      "latitude": 43.1857803898036,
      "longitude": 2.1369805421961723
    }
  }
}
```
Then the gradle plugin generates:
```kotlin
@DataSchema(isOpen = false)
interface Mlc1 {
    val name: String
    val people: kotlin.collections.List<kotlin.String>
    val peopleIds: kotlin.collections.List<kotlin.Long>
}

@DataSchema(isOpen = false)
interface Mlc3 {
    val latitude: Double
    val longitude: Double
}

@DataSchema(isOpen = false)
interface Mlc4

@DataSchema(isOpen = false)
interface Mlc2 {
    val gps: DataRow<Mlc3>
    val name: String
    val people: DataFrame<Mlc4>
    val peopleIds: DataFrame<Mlc4>
}

@DataSchema(isOpen = false)
interface Mlc5 : Mlc1 {
    val gps: DataRow<Mlc3>
}

@DataSchema
interface Mlc {
    @ColumnName("0")
    val `0`: DataRow<Mlc1>
    @ColumnName("164468523846839")
    val `164468523846839`: DataRow<Mlc2>
    @ColumnName("1644685267012957")
    val `1644685267012957`: DataRow<Mlc2>
    @ColumnName("1659257255454626")
    val `1659257255454626`: DataRow<Mlc5>
    public companion object {
[...]

PR #149, opened 2022-08-12
Json conversion improvements

Jolanrensen:
Fixes the empty array conversion described in https://github.com/Kotlin/dataframe/issues/143

I'll try to see if I can fix the other issues too, as well as check the generation using the plugin :)


Jolanrensen:
tackled in https://github.com/Kotlin/dataframe/pull/173

Issue #147, opened 2022-08-11
Additional feedback from dataframe operations


koperagen:
Some read/write operations may yield unexpected results, for example
1. reading CSV file with non-default delimiter could produce a dataframe with one column instead of 5. There is a %trackExecutions that prints generated schema, but it's either prints after each cell, or doesn't print at all
2. dataframe saves nested objects with just "toString", so you couldn't read it back. Might be helpful to print paths to all columns with non-serializable objects
3. ...?

One of the possible solutions is some sort of logging:
https://github.com/Kotlin/dataframe/issues/138#issuecomment-1210644392

I think we need more specific use cases

Kopilov:
```
else -> {
    this.setCellValue(any.toString())
}
``` [here](https://github.com/Kotlin/dataframe/blob/bdcee02e4a374423e363d28f615d58e022da5c51/dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt#L353) might be one of them. Or you are looking for some more specific?

The same case I have in saving to Arrow now.

Also I want to make saving to Arrow with some pre-defined schema and print warnings (or throw exceptions in strict mode) if some column does not exist in actual data or can not be converted to target type.

koperagen:
> Also I want to make saving to Arrow with some pre-defined schema and print warnings (or throw exceptions in strict mode) if some column does not exist in actual data or can not be converted to target type.
Sounds like AnyFrame.convertTo 

> [here](https://github.com/Kotlin/dataframe/blob/bdcee02e4a374423e363d28f615d58e022da5c51/dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt#L353) might be one of them. Or you are looking for something more specific?
this is a good example

koperagen:
> > Sounds like AnyFrame.convertTo
> 
> Does it sound good? Saving "as is" wold be default behavior but since Arrow supports explicit schemas, I want to use them also. Some another system might expect the data with some declared schema (and expose that), like inserting to existing SQL table.

I probably misunderstood a little bit. So you want to make some dataframe to match specific schema, like 
`val df: DataFrame<YourSchema> = df.yourFunction<YourSchema>(Mode.Warning)` or
`df.writeArrow("file.feather", arrowSchema, Mode.Warning)`?

Kopilov:
[...]

PR #162, opened 2022-08-29
Arrow writing

Kopilov:
Saving DataFrame in Arrow format, as is or per schema.

Warnings are muted by default, this may be changed after #147 resolving.


dataframe-arrow/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ArrowKtTest.kt
```diff
@@ -0,0 +1,237 @@
...
+
+    @Test
+    fun testWritingGeneral() {
+        fun assertEstimation(citiesDeserialized: DataFrame<*>) {
+            citiesDeserialized["name"] shouldBe citiesExampleFrame["name"]
+            citiesDeserialized["affiliation"] shouldBe citiesExampleFrame["affiliation"]
+            citiesDeserialized["is_capital"] shouldBe citiesExampleFrame["is_capital"]
+            citiesDeserialized["population"] shouldBe citiesExampleFrame["population"]
+            citiesDeserialized["area"] shouldBe citiesExampleFrame["area"]
+            citiesDeserialized["settled"].type() shouldBe typeOf<LocalDate>() // cities["settled"].type() refers to FlexibleTypeImpl(LocalDate..LocalDate?) and does not match typeOf<LocalDate>()
```
koperagen, line 111: Nice catch, i think it should be fixed in `readArrow` later as well

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/ArrowWriter.kt
```diff
@@ -0,0 +1,491 @@
...
+            check(actualField.isNullable)
+            infillWithNulls(vector, dataFrame.rowsCount())
+        } else {
+            infillVector(vector, convertedColumn)
+        }
+        return vector
+    }
+
+    private fun List<AnyCol>.toVectors(): List<FieldVector> = this.toArrowSchema(warningSubscriber).fields.mapIndexed { i, field ->
+        allocateVectorAndInfill(field, this[i], true, true)
```
koperagen, line 296: Maybe convert AnyCol to Field one by one, so you can get rid of this nullability in `this[i]`?

koperagen: like 
```
map { 
   val field = it.toArrowField() 
   allocateVectorAndInfill(field, it, true, true)
}
```

koperagen:
Wow, hi! That's a lot of work, thank you for the PR. I'll leave a couple of comments about the details, but right now i seem to lack general understanding, so i'll review it again a bit later. 

What do you think about making possible warnings a sealed class with overloaded toString for default behaviour? So that you can access actual exception in conversion or DataColumn and KType, etc.

koperagen:
Sorry, sorry :D I just kept this open while merging other PR. Can you re-open?

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/ArrowWriter.kt
Kopilov: Accepted, thanks

dataframe-arrow/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ArrowKtTest.kt
Kopilov: `FlexibleTypeImpl` comes not from Arrow but from `listOf(LocalDate.of())` in the sample generation.
What exactly do you want to fix?

Kopilov:
It seems that this can not be reopened. Duplicated in #169 by cherry-picking the branch manually.

Issue #142, opened 2022-08-08
[Feature] Swagger / OpenAPI support


Jolanrensen:
While the auto inferring of types is cool, it makes mistakes. This is logical since there are some things it just cannot know. For instance in json: What's the type of an empty list? What to do with undefined properties?

[OpenApi Swagger](https://swagger.io/specification/) could solve this. It exists for many APIs and essentially describes the structure of the JSON that will be expected. Generating Dataschemas from those should be a breeze.

Jolanrensen:
https://github.com/instructure/swagger-codegen-kotlin-retrofit-coroutines maybe something like this might help

Jolanrensen:
https://openapi-generator.tech/docs/generators/kotlin might actually be cool if this could be extended with dataframes as optional output type option. I'll try and see if I can create a template to generate DataSchema's

Jolanrensen:
might be easier to use our own json data inference, using openapi for help. A generator might be too advanced, just a library that can understand openapi might already suffice, checkout https://openapi.tools/, otherwise just do it ourselves.

PR #173, opened 2022-09-30
OpenAPI/Swagger JSON type schema support + many small fixes I came across

Jolanrensen:
As described by https://github.com/Kotlin/dataframe/issues/142

This PR contains a custom OpenAPI -> DF Marker conversion and implementation in the Gradle- and KSP plugin.
There are also a lot of tests, but I'm not yet 100% confident I've caught every edge case (there are many...).
Docs are updated too.
Let me know if there are any major things (or minor ones, all are welcome) that need changing or if you have some more testing ideas.


plugins/symbol-processor/src/test/kotlin/org/jetbrains/dataframe/ksp/DataFrameSymbolProcessorTest.kt
```diff
@@ -888,6 +1043,173 @@ class DataFrameSymbolProcessorTest {
         }
     }
 
+    private val petstoreYaml = File("../../core/src/test/resources/petstore.yaml")
+    private val petstoreYamlUrl = "https://raw.githubusercontent.com/OAI/OpenAPI-Specification/main/examples/v3.0/petstore.yaml"
```
koperagen, line 1047: is there a reason why this is downloaded directly from their github instead of using local copy + useHostedFile?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/openapi.kt
```diff
@@ -0,0 +1,1218 @@
...
+import org.jetbrains.kotlinx.dataframe.codeGen.ValidFieldName
+import org.jetbrains.kotlinx.dataframe.codeGen.plus
+import org.jetbrains.kotlinx.dataframe.schema.ColumnSchema
+import java.io.File
+import java.io.InputStream
+import java.net.URL
+import kotlin.reflect.KType
+import kotlin.reflect.typeOf
+
+public fun main() {
```
koperagen, line 38: also look like leftovers

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/openapi.kt
```diff
@@ -0,0 +1,1218 @@
...
+    }
+}
+
+/**
+ * Converts named OpenApi schemas to a list of [MyMarker]s.
+ * Will cause an exception for circular references, however they shouldn't occur in OpenApi specs.
+ */
+private fun Map<String, Schema<*>>.toMarkers(): List<MyMarker> {
+    // Convert the schemas to toMarker calls that can be repeated to resolve references.
+    val retrievableMarkers = mapValues { (typeName, value) ->
```
koperagen, line 334: Can you explain this part with references, resolve, retrieve and  ProduceAdditionalMarker? It seems i don't know enough about open api to understand it

koperagen (dismissed):

examples/idea-examples/movies/build.gradle.kts
```diff
@@ -22,3 +22,10 @@ tasks.withType<KotlinCompile> {
     kotlinOptions.jvmTarget = "1.8"
 }
 
+//dataframes {
```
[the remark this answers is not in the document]
Jolanrensen: nope XD thanks!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/guess.kt
```diff
@@ -17,17 +18,27 @@ import java.util.ServiceLoader
...
 
     // `DataFrame.Companion.read` methods uses this to sort list of all supported formats in ascending order (-1, 2, 10)
     // sorted list is used to test if any format can read given input
     public val testOrder: Int
 
-    public fun createDefaultReadMethod(pathRepresentation: String?): DefaultReadDfMethod
+    public fun createDefaultReadMethod(pathRepresentation: String? = null): DefaultReadDfMethod
+}
+
+public interface SupportedDataFrameFormat : SupportedFormat {
```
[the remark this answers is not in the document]
Jolanrensen: cool! didn't know that

plugins/symbol-processor/src/test/kotlin/org/jetbrains/dataframe/ksp/DataFrameSymbolProcessorTest.kt
Jolanrensen: I wanted to test whether that would work too, but local might be better

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/openapi.kt
Jolanrensen: This is actually something I came up with myself. OpenApi provides schemas for all of the types used. For each type, we want to generate a Marker (Which can be an interface, enum, etc.). However, the OpenApi schema is not ordered per se, so when we are reading the schema it might be that we have a reference to a (super)type (which are queried using `getRefMarker`) for which we have not yet created a Marker. In that case, I "pause" that one (by returning CannotFindRefMarker) and try to read another type schema first. Circular references cannot exist since it's encoded in JSON, so we never get stuck in an infinite loop. When all markers are "retrieved" (so turned from a RetrievableMarker to a MarkerResult.Success), we're done and have converted everything!

As for `produceAdditionalMarker`: In OpenAPI not all enums/objects have to be defined as a separate schema. Although recommended, you can still define an object anonymously directly as a type. For this, I have `produceAdditionalMarker` since during the conversion of a schema->Marker we get an additional new Marker.

Jolanrensen:
okay, I looked over additionalProperties... https://swagger.io/docs/specification/data-models/dictionaries/
Let's see how I can add that

Jolanrensen:
Since I changed so much I'm gonna provide a small overview of the changes I made. Might make it easier to review. I'll go through the changed files:

- Linting changes (sorry, but improves readability)
- Docs for OpenAPI and jsonoptions
- replaced all occurrences of "splitted" with "split" because English
- updated kotlinDatetime and made it api() for jupyter etc.
- ImportDataSchema gained jsonOptions with typeClashTactic and keyValuePaths
- starting with kdoc in some functions (more will come later)
- convertTo
  - `convertIf` in dsl: way more powerful than `convert(KType, KType)`, allows you to specify via a condition function whether you want to do a certain conversion and provides fromType and toSchema in ConverterScope. `convert` still has priority over `convertIf`
  - convertTo can now happen for any empty df (both 0 rows or 0 columns)
  - manual converters can now happen for any type of target column, not just value columns. user conversion happens before df conversion making some conversions that were previously impossible possible.
  - Value columns of datarows can become column groups, same as value columns with all nulls.
  - value columns of dataframes can become frame columns
  - absent columns can be created iff it's a nullable typed value column, DataRow<Something?> for a group column or frame column.
- containsNoData() helper function (created that once but don't use it anymore, might still be useful? opinion needed)
- enums can implement `DataSchemaEnum` to control how they are (en/de)coded from/to datasets instead of using just their name
- code generation can now generate enums and type aliases too
- `DefaultReadDfMethod` now provides actual `Marker` instead of just a name
- nullability helpers for FieldType
- bugfix for `MarkersExtractor` regarding `nullableProperties`
- `contains` in `BaseColumn` is now `operator fun`
- Bugfix for concat of DataFrames without columns (rows must still count up, else schema conversions will break)
- String+Number can become Comparable, this was always the case but order dependent. That now always gives the same result thanks to a bugfix in `commonParents()` and `commonType()`
- ISO_DATE_TIME support in parser from string
- `CodeGenerator.Companion.urlReader` is now split into `urlDfReader` and `urlCodeGenReader` where the former is used to generate a dataframe from the data and types from that, while the latter directly generates types (for openapi)
- Similar split for `SupportedFormat` into `SupportedDataFrameFormat` and `SupportedCodeGenerationFormat`
- `createColumn()`
  - bugfix for empty iterables always making column groups. Now uses guesstype or creates value column
  - bugfix for iterables with just nulls always making a frame column, now also uses guesstype or creates value column
  - can now create Column group with iterable of datarows too
- small fix in printing newlines of data schemas
-  `createEmptyColumn` and `createEmptyDataFrame` variants with `numberOfRows`
- `ColumnSchema`, aside from `type` now also has `contentType` for extra type info for Group- and Frame columns (instead of just getting `DataRow<*>` or `DataFrame<*>`. Very useful for new `convertIf` method. `intersectSchema` also tries to merge these `contentType`s if possible. `extractSchema` will make them `Any?`.
- `toSnakeCase()` helper function
- TODO more...

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt
```diff
@@ -164,20 +175,30 @@ internal fun commonParents(classes: Iterable<KClass<*>>): List<KClass<*>> =
...
-    }
+    }.sortedBy { it.simpleName } // make sure the order is stable to avoid bugs
 
-internal fun baseType(types: Set<KType>): KType {
+/**
+ * Returns the common type of the given [types].
+ *
+ * @param types the types to find the common type for
+ * @param listifyValues if true, then values and nulls will be wrapped in a list if they appear among other lists.
+ *   For example: `[Int, Any?, List<Int>]` will become `List<Int>` instead of `Any?`. If there is another collection
```
koperagen, line 198: is that so? 
It seems like a bug. Because type of null should be Nothing?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -9,27 +9,42 @@ class GroupByTests {
 
     @Test
     fun `groupBy values with nulls`() {
-        val df = dataFrameOf("a", "b")(
+        val df = dataFrameOf(
```
koperagen, line 12: i don't think this file should be reformatted in this commit / PR. It makes it harder to see actual changes. i mean, there are none, right?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
```diff
@@ -62,70 +64,82 @@ internal interface TypeRenderingStrategy {
...
-    override fun BaseField.renderFieldType(): Code =
+    override fun BaseField.renderFieldType(asAccessor: Boolean): Code =
         when (val fieldType = fieldType) {
-            is FieldType.ValueFieldType -> fieldType.typeFqName.shorten()
-            is FieldType.GroupFieldType -> "${DataRow::class.simpleName}<${fieldType.markerName}>"
+            is FieldType.ValueFieldType ->
+                fieldType.typeFqName.shorten()
+
+            is FieldType.GroupFieldType ->
+                if (asAccessor) "$DataRow<${fieldType.markerName}>"
```
koperagen, line 144: Similar conversion happens for FrameFieldType and List. You can generate List<fieldType.markerName> for usages inside data schema declaration and DataFrame<>... for usages in accessors. What do you think about changing this argument to an enum UseSite { DataSchemaDeclaration, GeneratedAccessor }? Or even splitting in two functions. Took me quite a while to understand what's this about

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/openapi.kt
```diff
@@ -0,0 +1,1464 @@
...
+    // convert DataRows to Maps if required by the schema
+    convert(
+        from = { it == typeOf<DataRow<*>>() },
+        to = { // any type of Map<String, _> or Map<String, _>?
+            (it.isSubtypeOf(typeOf<Map<*, *>>()) || it.isSubtypeOf(typeOf<Map<*, *>?>())) &&
+                it.arguments.getOrNull(0)?.type == typeOf<String>()
+        }
+    ) { (it as DataRow<*>).toMap() }
+
+    // convert DataFrame to DataFrame<Any?> if required by the schema
```
koperagen: What's the special meaning of DataFrame<Any?>? 
Previously, i think, you could just take any existing dataframe would be a valid value of type DataFrame<Any?>. But here's converting

koperagen: ok, it got commented out :D

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt
Jolanrensen: Definitely not `Nothing?` right? That only appears for thrown errors and return statements. `Nothing?` cannot have an instance. `Int, Any?, List<Int>` will become `Any?` as far as I know.

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
Jolanrensen: Unfortunately I have many of these reformats :/ Also for files I've touched but reverted in later commits. So there's no beginning trying to revert them all again I think. We could also view them as one of the "small fixes" part of the PR if that's okay with you :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/openapi.kt
Jolanrensen: This is not present anymore in later commits :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt
koperagen: Nothing appears for thrown errors, but `Nothing?` is a special type for null values 
image

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
koperagen: Sure, let's keep these changes. Just saying for the future

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt
Jolanrensen: Oh, I see now. I made an error in the KDoc. `[Int, Any?, List<Int>]` becomes `List<Any?>?`.
`[Int, Nothing?, List<Int>]` will become `List<Int>` (or at least, it should, but it doesn't atm... Bug :)  )

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
Jolanrensen: 2 functions instead of 1 boolean argument is always better indeed :) 
Where is this conversion in lists you talk about?

Jolanrensen: Okay, check out the latest commit, there I split the methods, renamed a bit and added explanation+examples :) should be a lot clearer what's happening now!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/ReplCodeGeneratorImpl.kt
```diff
@@ -68,7 +68,7 @@ internal class ReplCodeGeneratorImpl : ReplCodeGenerator {
                         // property scheme is valid for current data frame, but we should also check that all compatible open markers are implemented by it
                         val requiredBaseMarkers =
                             getRequiredMarkers(columnSchema, registeredMarkers.values)
-                        if (requiredBaseMarkers.all { currentMarker.implements(it) }) {
+                        if (requiredBaseMarkers.any() && requiredBaseMarkers.all { currentMarker.implements(it) }) {
```
koperagen, line 71: is this a fix for some bug? would be nice to have a test to see why this code was changed (or maybe there is? i didn't find it)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
koperagen: I mean
@DataSchema
interface Name {
    val firstName: String
}   

@DataSchema
interface Departments {
    val people: List<Name> // Name is DataSchema, so accessor field type is DataFrame. 
    // same as
    val people: DataFrame<Name>
}

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/ReplCodeGeneratorImpl.kt
Jolanrensen: yup. Accessors wouldn't be generated for my interface-nested DataSchemas without this fix. I think it was just a misinterpretation of `.all {}` again, that's why I didn't write an explicit test here, but the OpenApi Jupyter test breaks if this `.any()` check isn't there.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/guess.kt
```diff
@@ -16,18 +19,47 @@ import java.net.URL
...
+ * Note, this doesn't add functionality to [DataFrame.Companion.read], just [ImportDataSchema] and Gradle plugin.
+ *
+ * Return type will be a [CodeWithConverter] for which [CodeWithConverter.converter] needs to be called
+ * to get the generated code with the correct top-level interface name.
+ */
+public interface SupportedCodeGenerationFormat : SupportedFormat {
+
+    public fun readCodeForGeneration(
+        stream: InputStream,
+        generateHelperCompanionObject: Boolean = false,
```
koperagen, line 63: this in fact needed to support val ApiGuru = importDataSchema() notation in Jupyter, right? I think it's worth mentioning

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/KeyValueProperty.kt
```diff
@@ -0,0 +1,14 @@
+package org.jetbrains.kotlinx.dataframe.api
+
+import org.jetbrains.kotlinx.dataframe.annotations.ColumnName
+import org.jetbrains.kotlinx.dataframe.annotations.DataSchema
+
+/** A [DataSchema] interface / class can implement this if it represents a map-like data schema (so key: value). */
+public interface KeyValueProperty<T> {
+    // needs to be explicitly overridden in @DataSchema interface, otherwise extension functions won't generate (TODO)
```
koperagen, line 8: this interface can be annotated with @DataSchema itself. then extensions going to be a part of the library API and could be called on every marker that implements this interface

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convert.kt
```diff
@@ -76,41 +77,49 @@ internal fun AnyCol.convertToTypeImpl(to: KType): AnyCol {
...
-                    DataColumn.createValueColumn(name, values, to.withNullability(nullsFound))
-                }
-                else -> throw TypeConverterNotFoundException(from, to)
-            }
-            else -> {
+    if (from == to) return this
+
+    if (from.isSubtypeOf(to)) try {
+        return (this as DataColumnInternal<*>).changeType(to.withNullability(hasNulls()))
+    } catch (e: UnsupportedOperationException) { /* */
```
koperagen, line 88: do you remember why?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convertTo.kt
```diff
@@ -17,106 +25,209 @@ import org.jetbrains.kotlinx.dataframe.exceptions.ExcessiveColumnsException
...
+    @Suppress("UNCHECKED_CAST")
     override fun <A, B> convert(from: KType, to: KType, converter: (A) -> B) {
-        converters[from.withNullability(false) to to.withNullability(false)] = Converter(converter as (Any?) -> Any?, !from.isMarkedNullable)
+        converters[from.withNullability(false) to to.withNullability(false)] =
+            Converter({ converter(it as A) }, !from.isMarkedNullable)
     }
 
-    fun getConverter(from: KType, to: KType): Converter? {
-        return converters[from.withNullability(false) to to.withNullability(false)]
+    override fun convertIf(
```
koperagen, line 53: Do you have some representative example for this API? Honestly, i can't easily understand recursive list thing :(

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/openapi.kt
```diff
@@ -0,0 +1,2173 @@
...
+
+    if (file.extension.lowercase() != "json") {
+        return false
+    }
+
+    return isOpenApiStr(file.readText())
+}
+
+/** Parse and read OpenApi specification to [DataSchema] interfaces. */
+public fun readOpenApi(
```
koperagen, line 345: why not make DataFrame.Companion receiver?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/openapi.kt
```diff
@@ -0,0 +1,2173 @@
...
+         * DataFrame. This is to prevent `gather { all() }` from creating a lot of `null` values due to
+         * merging of ColumnGroups.
+         *
+         *
+         * @receiver the DataFrame to convert to an [AdditionalProperty] [DataFrame].
+         * @param convertTo optional [ConvertSchemaDsl] to specify extra conversions that need to occur.
+         * @return a [DataFrame] of [AdditionalProperty]s.
+         */
+        @Suppress("UNCHECKED_CAST")
+        public fun AnyFrame.convertToAdditionalProperties(
```
koperagen, line 531: maybe mention that this used in generated code?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/openapi.kt
```diff
@@ -0,0 +1,2173 @@
...
+import java.io.File
+import java.io.InputStream
+import java.net.URL
+import kotlin.String
+import kotlin.reflect.KType
+import kotlin.reflect.full.hasAnnotation
+import kotlin.reflect.full.isSubtypeOf
+import kotlin.reflect.jvm.jvmErasure
+import kotlin.reflect.typeOf
+
```
koperagen, line 75: wdyt about splitting this file? also, at some point we started to put formats in different modules

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt
Jolanrensen: All handled in latest commit :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
Jolanrensen: Hmm, I think this is handled on the schema generation side already. `List<Name>` is a `ValueColumn` type. Can't really seem to find where the distinction happens with wrapping `DataRow` around it or not

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/guess.kt
Jolanrensen: yup!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/KeyValueProperty.kt
Jolanrensen: sure, why not :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convert.kt
Jolanrensen: Yes, I believe this fixed a bug where converting from ColumnGroupImpl or FrameColumnImpl to something would crash because it doesn't have the `changeType` option. However, `convertTo` can now support conversions for any type of column, so the exception is caught here to give any user-given conversions a try first, otherwise, a TypeConverterNotFoundException is thrown.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convertTo.kt
Jolanrensen: For `convertIf`? Sure! It's not that difficult. `convert.with` can only support exact type matches, which can be a problem when you cannot define exact types for everything you want to catch. Let's say you want to write a converter that would convert all numbers to strings:
```kotlin
convert<Number>.with { it.toString() }
```
This wouldn't work, since `typeOf<Double>() != typeOf<Number>()`. That's where `convertIf` can come in handy. With this function you can manufacture any type of converter you like and you can use the full capabilities of Kotlin reflect you like:
```kotlin
convertIf({ fromType, toSchema -> 
    fromType.withNullability(false).isSubtypeOf(typeOf<Number>()) &&
        toSchema.type.withNullability(false) == typeOf<String>()
}) {
    it.toString()
}
```

For OpenApi I needed to provide a converter for `DataRow -> Any`, however, these converters only work on top-level types, so if a `DataRow->Any` conversion needs to happen in a `List<List<List<DataFrame<*>>>`, then we need my recursive function which makes `convertIf` accept any of these types and convert it accordingly.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/openapi.kt
Jolanrensen: Because you don't create a `DataFrame` ;)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataRowApi.kt
```diff
@@ -22,6 +23,24 @@ import kotlin.reflect.KType
 public fun AnyRow.isEmpty(): Boolean = owner.columns().all { it[index] == null }
 public fun AnyRow.isNotEmpty(): Boolean = !isEmpty()
 
+/** Returns `true` even if the columns contain empty iterables, empty dataframes etc. */
+public fun Any?.containsNoData(): Boolean =
```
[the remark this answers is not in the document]
Jolanrensen: Okay I agree :) will remove it

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/openapi.kt
Jolanrensen: sure

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/openapi.kt
Jolanrensen: Okay splitting off openapi works up 'till the part where I decide to read a file as json or as openapi. Need to think of something clever or universal to make that work elegantly :)

PR #183, opened 2022-11-04
add delimiter option for solving read error

LeeMH:
The example file(titanic.csv) is a csv file, but the delimiter is a semi-colon.
Added a delimiter option so that the sample code can run right out of the box.


Jolanrensen:
thanks! I agree it's a good idea to have this mentioned in te samples :)

PR #169, opened 2022-09-17
Arrow writing

Kopilov:
#162 duplicate

PR #187, opened 2022-11-22
Converting to date, Cell conversion exception

Kopilov:
When `AnyCol.convertTo` fails, we now can get more detailed information about the problem cell.

Also `convertToLocalDate` function added.

(This patch is a part of the next step of saving to Arrow per schema (#169) with encapsulating possible mismatches in a sealed class, but can be useful itself.)


koperagen:
Hi! Nice addition indeed, thank you :)

PR #190, opened 2022-11-28
CellConversionException in DataColumn<String?>.convertToDouble(locale)

Kopilov:
CellConversionException logic at parsing Double-s per locale


Jolanrensen:
return can be lifted out of if statement.
Can you provide some tests where the CellConversionException happens as well? :)

Kopilov:
> return can be lifted out of if statement.

Done

> Can you provide some tests where the CellConversionException happens as well? :)

Added to ```fun `convert to value class exceptions`()```

Jolanrensen:
Awesome! Looks good :)

Issue #139, opened 2022-08-05
readCSV using ImportDataSchema does not yield correct path


Jolanrensen:
Using `@file:ImportDataSchema()` or the gradle plugin doesn't seem to generate the correct path for `readCSV()`. It's probably because the relative path differs for the plugin compared to running the code from the IDE. The titanic example tries to circumvent this using `System.setProperty("user.dir", System.getProperty("user.dir") + "/examples/idea-examples/titanic")` but that doesn't seem to do anything on my machine (Arch Linux). 

setting `ksp { arg("dataframe.resolutionDir", ...) }` also doesn't seem to help.

Maybe only the absolute path should be saved as `defaultPath`?

Also, it's kinda strange that `ImportDataSchema` exclusively allows a relative path. It should definitely always allow an absolute path.

PR #191, opened 2022-11-28
Absolute path fixes in plugins

Jolanrensen:
`defaultPath` in `AbstractDefaultReadMethod` implementing IO methods will now be absolute path to solve difference between calls from gradle/the IDE. `@ImportDataSchema` and `dataframes { schemas { data = "" } }` now support absolute path too. Relative path will still be attempted first.

Fixes https://github.com/Kotlin/dataframe/issues/139

### DIFF
--- a/examples/idea-examples/titanic/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/titanic/ml/titanic.kt
+++ b/examples/idea-examples/titanic/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/titanic/ml/titanic.kt
@@ -2,8 +2,24 @@ package org.jetbrains.kotlinx.dataframe.examples.titanic.ml
 
 import org.jetbrains.kotlinx.dataframe.ColumnSelector
 import org.jetbrains.kotlinx.dataframe.DataFrame
-import org.jetbrains.kotlinx.dataframe.api.*
+import org.jetbrains.kotlinx.dataframe.api.by
 import org.jetbrains.kotlinx.dataframe.api.column
+import org.jetbrains.kotlinx.dataframe.api.convert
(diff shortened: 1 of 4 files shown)


examples/idea-examples/titanic/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/titanic/ml/titanic.kt
```diff
@@ -64,18 +79,23 @@ fun main() {
     }
 }
 
-fun <T> DataFrame<T>.toTrainTest(trainRatio: Double, yColumn: ColumnSelector<T, Number>): Pair<OnHeapDataset, OnHeapDataset> =
-    toOnHeapDataset(yColumn).split(trainRatio)
+fun <T> DataFrame<T>.toTrainTest(
```
koperagen, line 82: let's agree to create separate commits for this kind of changes? it will be easier to review without taking time to make sure semantics haven't changed, because now it's a bit painful

koperagen (approved):

examples/idea-examples/titanic/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/titanic/ml/titanic.kt
Jolanrensen: I want to, but even having a file opened, after I press run, then IntelliJ lints it XD Then it feels wrong to undo it again. It's good practice for code I write myself.
I'll go look for that setting next. I understand it's difficult to check it like this, so I'll try. 
However, the entire project could do with a good linting, cause in almost every file there are long lines, mixed named and positional arguments, missing spaces before brackets etc. etc. Will get to that one day :)

Issue #177, opened 2022-10-05
dataFrameOf doesn't recognize @DataSchema instances


Jolanrensen:
given
```kotlin
@DataSchema
data class Location(
    val name: String,
    val gps: Gps?,
)

@DataSchema
data class Gps(
    val latitude: Double,
    val longitude: Double,
)

val a: DataFrame<Location> = listOf(
    Location("Home", Gps(0.0, 0.0)),
    Location("Away", null),
).toDataFrame()
// ⌌----------------------------------------------------------⌍
// |  | name:String| gps:{latitude:Double?, longitude:Double?}|
// |--|------------|------------------------------------------|
// | 0|        Home|  { latitude:0.000000, longitude:0.0000...|
// | 1|        Away|                                       { }|
// ⌎----------------------------------------------------------⌏
// 
// name: String
// gps:
//     latitude: Double?
//     longitude: Double?

val b: AnyFrame = dataFrameOf("name", "gps")(
    "Home", Gps(0.0, 0.0),
    "Away", null,
)
// ⌌-------------------------------------------------⌍
// |  | name:String|                         gps:Gps?|
// |--|------------|---------------------------------|
// | 0|        Home| Gps(latitude=0.0, longitude=0.0)|
// | 1|        Away|                             null|
// ⌎-------------------------------------------------⌏
// 
// name: String
// gps: Gps?
```

One would expect `a` and `b` to give the same result. However, in `a` `gps` is recognized as a `DataRow` and in the second as a `Value`.

2x2xplz:
I'm not on the JetBrains team, so take this with a grain of salt, but I don't agree with your assumption that `a` and `b` should give the same result. Every row in `a` is specifically defined to be a `Location` (and can reference its schema), `b` never states the intended type.

The docs for `dataFrameOf` state: `Returns DataFrame with given column names and values`. It seems consistent with those docs that `b` would define the second column as **values** (of Gps), and not a `ColumnGroup`. Meanwhile `toDataFrame`'s docs mention `perform deep object graph traversal and convert nested objects into ColumnGroups and FrameColumns`, so not surprising to see independent nested cols for `latitude` and `longitude`.

Even if it had the capability to do so, wouldn't expect the program to search through all of your available `DataSchema`s to see which is the best fit... you have to specify the schema somewhere. And now that I say that, it would actually be potentially dangerous to have the program just assume that every pair of String + Gps is automatically a Location, without you specifying that.

[...]

Issue #113, opened 2022-05-21
Generate constructor for DataSchema


nikitinas:
Support `GenerateConstructor` annotation on `companion object` of schema interface to generate constructor for that interface and `append` overload for `DataFrame`:

```kotlin
@DataSchema
interface Schema {
    val a: Int
    val b: String

    @GenerateInterface
    companion object
}

// region Generated Code

operator fun Schema.Companion.invoke(a: Int, b: String): Schema =
    object: Schema {
        override val a = a
        override val b = b
    }

fun DataFrame<Schema>.append(vararg rows: Schema) = concat(rows.asIterable().toDataFrame())

// endregion
```

nikitinas:
Better design.

DataFrame API:

```kotlin

interface DataRowSchema

inline fun <reified T:DataRowSchema> dataFrameOf(vararg rows: T) = rows.asIterable().toDataFrame()

inline fun <reified T:DataRowSchema> DataFrame<T>.append(vararg rows: T) = concat(rows.asIterable().toDataFrame())
```

User code:

```kotlin
@DataSchema
interface Record: DataRowSchema {
    val a: Int
    val b: Int

    companion object
}

// region Generated code

operator fun Record.Companion.invoke(a: Int, b: Int): Record =
    object: Record {
        override val a = a
        override val b = b
    }

// endregion

// usage:

dataFrameOf(Record(1,2), Record(3,4))
  .append(Record(5,6))
  .add("sum") { a + b }
```

koperagen:
Can we somehow get rid of `DataRowSchema`? 
In FIR, it's possible to add supertypes to annotated class, so at least it can disappear from user code.

And here is another example to be aware of 
```Kotlin
// User code:
@DataSchema
interface AnotherRecord: DataRowSchema {
    val a: Int
    val b: Int

    companion object
}

@DataSchema
interface Record: DataRowSchema {
    val a: List<AnotherRecord> // same as DataFrame<AnotherRecord>
    val b: AnotherRecord // same as DataRow<AnotherRecord>

    companion object
}

// usage:
dataFrameOf(listOf(AnotherRecord(5,6)), AnotherRecord(5,6))
```

zaleslaw:
Working by default for Data Classes, we will not support it for interfaces

Jolanrensen:
Plus, for schema's `DataRowSchema` is already added as supertype, making `.append()` work :D

PR #197, opened 2022-11-29
Column creation of @DataSchema instances to produce ColumnGroups

Jolanrensen:
Based on https://github.com/Kotlin/dataframe/issues/177. See this issue for more details and motivation.

Might be related to `dataFrameOf()` overloads in https://github.com/Kotlin/dataframe/issues/113

With these changes:
```kotlin
dataFrameOf("name", "gps")(
    "Home", Gps(0.0, 0.0),
    "Away", null,
)
// ⌌----------------------------------------------------------⌍
// |  | name:String| gps:{latitude:Double?, longitude:Double?}|
// |--|------------|------------------------------------------|
// | 0|        Home|  { latitude:0.000000, longitude:0.0000...|
// | 1|        Away|                                       { }|
// ⌎----------------------------------------------------------⌏
// 
// name: String
// gps:
//     latitude: Double?
//     longitude: Double?
```
instead of 
```kotlin
// ⌌-------------------------------------------------⌍
// |  | name:String|                         gps:Gps?|
// |--|------------|---------------------------------|
// | 0|        Home| Gps(latitude=0.0, longitude=0.0)|
// | 1|        Away|                             null|
// ⌎-------------------------------------------------⌏
// 
// name: String
// gps: Gps?
```

Also changes the behavior of `columnOf(Gps(...))` to produce a `ColumnGroup` now. To be specific, if a new column is created from values where all values are either `null` or the same `@DataSchema` annotated type instance, a `ColumnGroup` will be formed instead of the previous `ValueColumn`.

Would this be acceptable behavior? Better than before? Worse?
@koperagen 
@nikitinas

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataColumn.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataColumn.kt
@@ -19,7 +19,7 @@ import org.jetbrains.kotlinx.dataframe.impl.columns.ColumnGroupImpl
 import org.jetbrains.kotlinx.dataframe.impl.columns.FrameColumnImpl
 import org.jetbrains.kotlinx.dataframe.impl.columns.ValueColumnImpl
 import org.jetbrains.kotlinx.dataframe.impl.columns.addPath
-import org.jetbrains.kotlinx.dataframe.impl.columns.guessColumnType
+import org.jetbrains.kotlinx.dataframe.impl.columns.createColumnGuessingType
 import org.jetbrains.kotlinx.dataframe.impl.columns.toColumnKind
 import org.jetbrains.kotlinx.dataframe.impl.getValuesType
 import org.jetbrains.kotlinx.dataframe.impl.splitByIndices
@@ -54,7 +54,7 @@ public interface DataColumn<out T> : BaseColumn<T> {
             values: List<T>,
             type: KType,
             infer: Infer = Infer.None,
-            defaultValue: T? = null
+            defaultValue: T? = null,
         ): ValueColumn<T> = ValueColumnImpl(values, name, getValuesType(values, type, infer), defaultValue)
 
         /**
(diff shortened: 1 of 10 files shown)


koperagen:
I reckon `convert` also gives special treatment to values of types `DataRow`, `DataFrame`
i.e. 
```
val df = dataFrameOf("abc")(123)
df.convert { abc }.with { DataRow.read("something.json") }
```

makes abc a ColumnGroup.

The same goes for 

`df.convert { abc }.with { DataFrame.read() }`

here abc is FrameColumn (DataColumn<DataFrame<>>).

Now, i believe, types which declarations are marked with `@DataSchema` are mapped by marker extractors in KSP and REPL. I.e. 
```
@DataSchema
class Name(
    val firstName: String
    val lastName: String
)

@DataSchema
class Person(
    val name: Name / DataRow<Name> 
    val names: List<Name> // DataFrame<Name>
)
```

both return types are identical from markers extractors perpective. It's easier to create values of Name / List<Name>.
Now the only operation that considers values of type Name / List<Name> as DataRow<Name> / DataFrame<Name> is `toDataFrame`.

So if `dataFrameOf` sees `Name` as `DataRow` and makes `ColumnGroup` from it, should `convert` do the same?

Jolanrensen:
I didn't completely understand it at first, but I agree. If `dataFrameOf()` should recognize `@DataSchema` instances and turn them into `DataRow` or `DataFrame` (for lists), then so should `convert`. If `convert` uses the `columnOf` calls this is already what happens using this PR, but I'll need to check that.

Btw. I don't think I've implemented the recognition of `List<T> -> DataFrame<T>` where `T` is annotated with `@DataSchema`. Will need to check that too.

nikitinas:
I think that `convert` should not unfold objects by default because this operation may need additional configuration of traversal depth, excludes etc. i suggest to unify `dataFrameOf` and `.toDataFrame` behaviour and configuration, but don't allow to unfold objects within `convert`. It can be performed by separate `unfold` operation if needed.

nikitinas:
See [my comment](https://github.com/Kotlin/dataframe/issues/177#issuecomment-1358968584) for #177

Issue #181, opened 2022-10-28
Bug: @DataSchema fails with generic types


Jolanrensen:
Say we have
```kotlin
interface KeyValue<T> {
    val key: String
    val value: T
}

@DataSchema
interface MySchema : KeyValue<Int>
```
You'd expect `MySchema` to work with `key: String`, `value: Int`, instead it fails cause it considers `value` as `value: T`.
You need an explicit `override val value: Int` inside `MySchema` for it to work correctly.

Jolanrensen:
Also, extensions won't be generated unless explicitly overriden

PR #200, opened 2022-12-05
Dataschema generic inheritance

Jolanrensen:
Fix for https://github.com/Kotlin/dataframe/issues/181

Embraces inherited properties in @DataSchemas now.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MarkersExtractor.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MarkersExtractor.kt
@@ -8,9 +8,9 @@ import org.jetbrains.kotlinx.dataframe.impl.schema.getPropertiesOrder
 import org.jetbrains.kotlinx.dataframe.schema.ColumnSchema
 import kotlin.reflect.KClass
 import kotlin.reflect.KType
-import kotlin.reflect.full.declaredMemberProperties
 import kotlin.reflect.full.findAnnotation
 import kotlin.reflect.full.hasAnnotation
+import kotlin.reflect.full.memberProperties
 import kotlin.reflect.full.superclasses
 import kotlin.reflect.full.withNullability
 import kotlin.reflect.jvm.jvmErasure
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MarkersExtractor.kt
```diff
@@ -54,7 +54,7 @@ internal object MarkersExtractor {
 
     private fun getFields(markerClass: KClass<*>, nullableProperties: Boolean): List<GeneratedField> {
         val order = getPropertiesOrder(markerClass)
-        return markerClass.declaredMemberProperties.sortedBy { order[it.name] ?: Int.MAX_VALUE }.mapIndexed { _, it ->
```
koperagen, line 57: I wonder if it was a deliberate choice to reduce amount of generated code in Jupyter. 
could you check what is generated after executing the second cell?
```
val df = dataFrameOf("a")(123)
```

second cell
```
val df1 = df.add("b") { 321 }
```

koperagen (changes requested):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MarkersExtractor.kt
koperagen: `Firstly, I need to make the first line a var to make isOpen = true` 
This sounds suprising. I don't think it was the case before, was it? 

`So, I suspect for Jupyter it gets the properties for extensions in yet another way.`
Oh, right. I think Jupyter works on DataFrame / ColumnSchema level and uses existing markers somehow.

Jolanrensen: It is the case since https://github.com/Kotlin/dataframe/pull/193. Or well, it was the case but not everywhere and not consistently (Even the tests were inconsistent about it). Since that PR it's always consistent.

koperagen (approved):
Let's move conversation here https://github.com/Kotlin/dataframe/pull/193

Issue #161, opened 2022-08-28
It fails to sort a grouped data frame


holgerbrandl:
When running
```kotlin
irisData.groupBy("Species").sortBy("Sepal.Length")
```
it fails with 
```
java.lang.IllegalStateException: Can not use ColumnGroup as sort column
```

However, sorting by group is a common usecase. In particular in combination with next() and prev()

koperagen:
Thank you for the report! It looks like a bug, because this seems to be working fine:
```
irisData.groupBy("Species")
        .sortBy { it["Petal.Length"] }
```
> In particular in combination with next() and prev()
Can you elaborate?

holgerbrandl:
I'm happy to help but I'm unsure about what you would need. It seems, you can reproduce the problem already?

koperagen:
Yes, i can reproduce the problem. I don't understand how and when one could use next() and prev() in sorting though.  So i'm curious about the use case and maybe an example?

holgerbrandl:
One use case that comes to mind are tool-state changes (ie. in manufacturing). Tools periodically change their status (idle, productive, maintenance, etc.). Later one, one may want to compute the duration for which each tool is in a specific state. To do s, we need to 
```
toolEvents.groupBy("tool_id").sortedBy("timestamp").add("state_duration"){ last timestamp - current).aggregate ... plot...
```

Unfortunately I do not have shareable example data at the moment.

koperagen:
I really like this example, thank you.

Jolanrensen:
fixed and merged!

PR #204, opened 2022-12-08
Sort grouped df

Jolanrensen:
Work in progress fix for https://github.com/Kotlin/dataframe/issues/161.
The first example is now fixed, but for the second example, I might need some test data to be sure it now works as expected.


Jolanrensen:
Thanks to @koperagen the other example from the issue is now also tested and appears to be working correctly

plugins/symbol-processor/src/main/kotlin/org/jetbrains/dataframe/ksp/GetQualifiedNameOrThrow.kt
```diff
@@ -1,7 +1,20 @@
 package org.jetbrains.dataframe.ksp
 
 import com.google.devtools.ksp.symbol.KSDeclaration
+import org.jetbrains.kotlinx.dataframe.impl.codeGen.quoteIfNeeded
 
-fun KSDeclaration.getQualifiedNameOrThrow(): String {
-    return (qualifiedName ?: error("@DataSchema declaration ${simpleName.asString()} at $location must have qualified name")).asString()
-}
+fun KSDeclaration.getQualifiedNameOrThrow(): String =
```
koperagen, line 6: isn't this commit merged already? is it intentional or something strange happened?

Jolanrensen: huh you're right. That's odd

Jolanrensen:
okay now I really messed it up

Issue #159, opened 2022-08-28
Add API to unfold object-columns into properties via reflection


holgerbrandl:
As with `unfold()` in krangl (see https://holgerbrandl.github.io/krangl/data_model/#to-type-or-not-to-type), it would be great if kotlin-df could provide a similar API. Clearly, this could be done manually using `df.add()`, but with many attributes this is very tedious. And in krangl we've seen that this can be done very efficiently via reflection.

Example (from krangl but yet missing in kdf):
```kotlin
data class City(val name:String, val code:Int)
data class Person(val name:String, val address:City)

val persons : List<Person> = listOf(
    Person("Max", City("Dresden", 12309)),
    Person("Anna", City("Berlin", 10115))
)

val personsDF: DataFrame = persons.asDataFrame() // <- Also sems missing API in kdf?

// unfold City attributes into different columns
personsDF.unfold<City>("address")  // <- Missing API in kdf
```

For impl see https://github.com/holgerbrandl/krangl/blob/49418b2ca6ee6ae165c034e56e4da77e4707f7ad/src/main/kotlin/krangl/Builder.kt#L115

koperagen:
`persons.asDataFrame()` -- i believe persons.toDataFrame() does the trick https://kotlin.github.io/dataframe/createdataframe.html#todataframe

holgerbrandl:
@koperagen Thanks for the pointer. 

I'd still think that the key aspect of this ticket is still valid: Once we have objects of type E in a column of a data-frame we can not conveniently expand E's properties into columns (without writing it out attribute by attribute using `add()`). That's what `unfold` was/is doing in krangl and I used it quite frequently.

koperagen:
> @koperagen Thanks for the pointer.
> 
> I'd still think that the key aspect of this ticket is still valid: Once we have objects of type E in a column of a data-frame we can not conveniently expand E's properties into columns (without writing it out attribute by attribute using `add()`). That's what `unfold` was/is doing in krangl and I used it quite frequently.

Indeed. We'll introduce this API. I also made an example of how it can be done now. 

```
class RepositoryInfo(val data: Any)

fun download(url: String) = RepositoryInfo("fancy response from the API")

val interestingRepos = dataFrameOf("name", "url")(
    "dataframe", "/dataframe",
    "kotlin", "/kotlin",
)

val initialData = interestingRepos
    .add("response") { download("url"()) }

[...]

PR #206, opened 2022-12-09
read -> unfold + Add documentation for unfold #159

koperagen:


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -38,7 +38,7 @@ public inline fun <reified T> Iterable<T>.toDataFrame(vararg props: KProperty<*>
         properties(roots = props, maxDepth = maxDepth)
     }
 
-public inline fun <reified T> DataColumn<T>.read(): AnyCol = when (kind()) {
+public inline fun <reified T> DataColumn<T>.unfold(): AnyCol = when (kind()) {
     ColumnKind.Group, ColumnKind.Frame -> this
     else -> when {
         isPrimitive() -> this
```
Jolanrensen, line 44: While the name `unfold()` does make sense for `isPrimitive()` and the `else` statement where the object is turned into a dataframe one property at a time, for the reading  of a file or url the name `read()` makes way more sense. Would it be a good idea to have both names? Or is this what you're planning to do?

Jolanrensen (changes requested):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
koperagen: I reckon people would use it for the same purpose, thus one function and one name.
It's not clear what function does from its name though. Even `read` is more like `read` + `replace`. So people would have to read docs / kdocs to find out anyway.

koperagen: speaking of, how would you make kdocs in this situation? i'm in dither over this because duplication and maintenance

Jolanrensen: hmm I'd definitely keep the old functions, annotate them with 
```kotlin
@Deprecated("read is now called unfold", ReplaceWith("unfold"), DeprecationLevel.ERROR)
```
And then simply KDoc the `read()` functions with `@see unfold`

Jolanrensen: But what does read/unfold do with a file or url? I'd say it still "read"s those right? I'd see "unfold" as only "unfolding" an object into a dataframe.

Jolanrensen (approved):

PR #207, opened 2022-12-11
Support custom filling of missing columns in `convertTo`

nikitinas:
Support `fill` operation in `convertTo` DSL to fill missing columns with custom row expressions.
Print column path in `TypeConversionException`.
Update docs for `convertTo`.


Jolanrensen (approved):
Very useful addition!

Issue #208, opened 2022-12-13
It's not clear when you need to update Kotlin kernel in your env


koperagen:
https://github.com/Kotlin/dataframe/commit/3db46ccccaa1291c0627307d64133317f545e6ae
This commit introduces breaking change in Jupyter integration, so dev version of dataframe starting with 1194 don't work with kernel versions < 0.11.0-180
Exception that's thrown when importing dataframe is not at all clear 
![image](https://user-images.githubusercontent.com/12936457/207312741-8c99ae3d-d5d9-48f8-993f-70ae075a154b.png)

So we need to find a way to tell people to update their kernel and to be aware of such changes ourselves

Jolanrensen:
Library version 0.11.0-198 will allow setting a minKernelVersion inside the Integration class. We should probably set that before the 0.9.0 release :)

PR #209, opened 2022-12-13
Min kernel version fix

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/208 by setting a minimal kernel version for the Kotlin Jupyter kernel in the library.
Users will see an error provided by the Kotlin Jupyter library if their kernel is too old


Jolanrensen:
image
LOL. Yes that makes sense haha. This method cannot be called on older versions of the kernel since it doesn't exist yet.

PR #214, opened 2022-12-28
KDocs start using Doc Preprocessor Gradle plugin

Jolanrensen:
The final version of this PR consists of:

 - the `ColumnExpression` alias and use of it across the API instead of a specific `Selector`.
 - General concept docs with reusable examples in the `:core` `documentation` package
 - Docs for update.kt
 - Docs for Nulls.kt

The rendered version of the docs can be found at `:core/generated-sources` and it would probably be useful to review those as well as the source files. I'm curious to hear about how you find the resulting docs and whether the structure I provided for the two API files (and the common docs) is repeatable for the rest of the API or if you have some additions, concerns etc.

Old message:
> Exploratory method for being able to reuse KDocs.
> After trying, it seems that compiler plugins won't do the trick, since the javadoc and sources.jar are generated from the actual sources before the plugin is run. 
> So, to have the sources.jar be modified, we need something to run from Gradle itself. 
> Cue [JCP](https://github.com/raydac/java-comment-preprocessor). I've used this in the [Kotlin Spark API](https://github.com/Kotlin/kotlin-spark-api) as well. 
> 
> The current implementation takes the sources from :core, runs the preprocessor on them, changes the Kotlin compile sources to the files modified by the preprocessor, and after compilation resets them back. With this method, sources.jar will contain the properly preprocessed files.
> One downside is that compile errors will thus point you to the preprocessor files.
> 
> Since we (probably) will only use JCP to generate KDocs I will experiment and see if there's a way to only change the sources when generating sources.jar and not during compilation.
> 
> I've also filed an [issue](https://github.com/raydac/java-comment-preprocessor/issues/44) for multiline string variables for JCP since defining reusable comments in the same file now requires them to be a single line... It is possible to define the comments in the Gradle build files, but this doesn't seem right to me.
> 
> Check add.kt for an example.
> 
> Edit: JCP output is now only used for `Jar` tasks :D


Jolanrensen:
The only problem atm is in usability. Due to the lack of multiline strings, we need to do:

```kotlin
import org.jetbrains.kotlinx.dataframe.exceptions.DuplicateColumnNamesException
import org.jetbrains.kotlinx.dataframe.exceptions.UnequalColumnSizesException

/** either
//#local ADD = evalFile("add.txt")
*/

/** or
//#local ADD0 = " * Original [DataFrame] is not modified."
//#local ADD1 = " *"
//#local ADD2 = " * @throws [DuplicateColumnNamesException] if columns in expected result have repeated names"
//#local ADD3 = " * @throws [UnequalColumnSizesException] if columns in expected result have different sizes"
//#local ADD4 = " * @return new [DataFrame] with added columns"
//
//#local ADD = ADD0 + "\n" + ADD1 + "\n" + ADD2 + "\n" + ADD3 + "\n" + ADD4
 */


/**
 * Creates new [DataFrame] with given columns added to the end of original [DataFrame.columns] list.
 *
/*$ADD$*/
 * @param columns columns to add
 */
public fun <T> DataFrame<T>.add(vararg columns: AnyBaseCol): DataFrame<T> = addAll(columns.asIterable())
```

Jolanrensen:
Using my self-built gradle plugin is probably a better solution https://github.com/Jolanrensen/kdocIncludeGradlePlugin

Jolanrensen:
Updated to use my own gradle plugin now via JitPack :)

Example can be found in `add.kt`:
image

Jolanrensen:
https://github.com/Jolanrensen/docProcessorGradlePlugin/tree/main was updated to allow other processors as well. Maybe `@sample` is also interesting for DataFrame :)

Jolanrensen:
`update.kt` now has all its methods documented and can be used as an example :)

I'll still look into including the generated docs in the PR so difs can be seen from github

Jolanrensen:
https://github.com/Jolanrensen/docProcessorGradlePlugin might need a thorough check from someone else before merging this branch in master. I'll see if I can clean it up so it is easier to understand.

zaleslaw:
What is the future of this PR?

Jolanrensen:
> What is the future of this PR?

Will rebase on master, fix every conflict and then ask for a review :)

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
Jolanrensen, line 1: This file I just started with, will probably work on docs after this PR is merged in a separate branch

Jolanrensen:
ready for review!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
```diff
@@ -1,36 +1,106 @@
...
+/**
+ * @include [CommonFillNullsFunctionDoc]
+ * @include [SelectingColumns.ColumnAccessors.WithExample] {@include [SetFillNullsOperationArg]}
+ * @include [Update.ColumnAccessorsParam]
+ */
+public fun <T, C> DataFrame<T>.fillNulls(vararg columns: ColumnReference<C>): Update<T, C?> =
+    fillNulls { columns.toColumns() }
+
+/**
+ * TODO this will be deprecated
```
zaleslaw, line 100: Make it deprecated or create an issue

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
```diff
@@ -1,36 +1,206 @@
...
-    fillNulls { cols.toColumns() }
+/**
+ * ## The Fill Nulls Operation
+ *
+ * Replaces `null` values with given value or expression.
+ * Specific case of [update][org.jetbrains.kotlinx.dataframe.api.update].
+ *
+ * Check out the [`fillNulls` Operation Usage][org.jetbrains.kotlinx.dataframe.api.FillNulls.Usage].
+ *
+ * For more information: [See `fillNulls` on the documentation website.](https://kotlin.github.io/dataframe/fill.html#fillnulls)
```
zaleslaw, line 129: Could we use the @see tag instead of markdown link

like here

@see <a href="https://arxiv.org/abs/1512.03385">
         *     Deep Residual Learning for Image Recognition (CVPR 2015)</a>

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
```diff
@@ -1,36 +1,206 @@
...
+ *
+ * Replaces `null` values with given value or expression.
+ * Specific case of [update][org.jetbrains.kotlinx.dataframe.api.update].
+ *
+ * Check out the [`fillNulls` Operation Usage][org.jetbrains.kotlinx.dataframe.api.FillNulls.Usage].
+ *
+ * For more information: [See `fillNulls` on the documentation website.](https://kotlin.github.io/dataframe/fill.html#fillnulls) 
+ * ## ‎
+ * The columns to update need to be selected. See [Selecting Columns][org.jetbrains.kotlinx.dataframe.documentation.SelectingColumns] for all the selecting options. 
+ * ## This Fill Nulls Overload
```
zaleslaw, line 132: Probably we need a line break before
## This Fill Nulls Overload

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
```diff
@@ -1,36 +1,206 @@
...
+ * Select columns using their [column names][String]
+ * ([String API][org.jetbrains.kotlinx.dataframe.documentation.AccessApi.StringApi]).
+ *
+ * For example:
+ *
+ * `df.`[fillNulls][org.jetbrains.kotlinx.dataframe.api.fillNulls]`("length", "age")`
+ *  
+ * @param columns The [Column names][org.jetbrains.kotlinx.dataframe.documentation.SelectingColumns.ColumnNames.WithExample] belonging to this [DataFrame][org.jetbrains.kotlinx.dataframe.DataFrame] to update.
```
zaleslaw, line 140: Could we use [columns] instead of columns without bracket for the parameters in the documentation?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
```diff
@@ -1,36 +1,206 @@
...
+ * Check out the [`fillNulls` Operation Usage][org.jetbrains.kotlinx.dataframe.api.FillNulls.Usage].
+ *
+ * For more information: [See `fillNulls` on the documentation website.](https://kotlin.github.io/dataframe/fill.html#fillnulls) 
+ * ## ‎
+ * The columns to update need to be selected. See [Selecting Columns][org.jetbrains.kotlinx.dataframe.documentation.SelectingColumns] for all the selecting options. 
+ * ## This Fill Nulls Overload
+ * Select columns using [KProperties][KProperty] ([KProperties API][org.jetbrains.kotlinx.dataframe.documentation.AccessApi.KPropertiesApi]).
+ *
+ * For example:
+ * ```kotlin
```
zaleslaw, line 160: in some places for inserted code we have 
```kotlin
```

but sometimes
just ``

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
```diff
@@ -1,36 +1,206 @@
...
+ * For more information: [See `fillNulls` on the documentation website.](https://kotlin.github.io/dataframe/fill.html#fillnulls) 
+ * ## ‎
+ * The columns to update need to be selected. See [Selecting Columns][org.jetbrains.kotlinx.dataframe.documentation.SelectingColumns] for all the selecting options. 
+ * ## This Fill Nulls Overload
+ * Select columns using [column accessors][org.jetbrains.kotlinx.dataframe.columns.ColumnReference]
+ * ([Column Accessors API][org.jetbrains.kotlinx.dataframe.documentation.AccessApi.ColumnAccessorsApi]).
+ *
+ * For example:
+ *
+ * `val length by `[column][org.jetbrains.kotlinx.dataframe.api.column]`<`[Double][Double]`>()`
```
zaleslaw, line 188: Maybe better to wrap all the rows with the ```kotlin?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
```diff
@@ -1,36 +1,206 @@
...
+ *
+ * `df.`[fillNulls][org.jetbrains.kotlinx.dataframe.api.fillNulls]`(length, age)`
+ *  
+ * @param columns The [Column references][org.jetbrains.kotlinx.dataframe.documentation.SelectingColumns.ColumnAccessors.WithExample] of this [DataFrame][org.jetbrains.kotlinx.dataframe.DataFrame] to update.
+ */
+public fun <T, C> DataFrame<T>.fillNulls(vararg columns: ColumnReference<C>): Update<T, C?> =
+    fillNulls { columns.toColumns() }
+
+/**
+ * TODO this will be deprecated
```
zaleslaw, line 200: Issue or fix?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
```diff
@@ -1,36 +1,106 @@
...
+ * ## `NaN`
+ * [Floats][Float] or [Doubles][Double] can be represented as [Float.NaN] or [Double.NaN], respectively,
+ * in cases where a mathematical operation is undefined, such as dividing by zero.
+ *
+ * You can also use [fillNaNs][fillNaNs] to replace `NaNs` in certain columns with a given value or expression
+ * or [dropNaNs][dropNaNs] to drop rows with `NaNs` in them.
+ *
+ * @see NA
+ */
+internal interface NaN
```
zaleslaw, line 22: If all these interfaces are related to the documentation, probably better to move into the documentation package or put near the Nulls but in the separate file with name NullsDocumentation.kt.

Why it's important. It seems that these interfaces are part of the architecture, but it doesn't reflect the reality

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
```diff
@@ -56,94 +126,350 @@ internal inline val Float?.isNA: Boolean get() = this == null || this.isNaN()
...
+/**
+ * @include [CommonFillNaNsFunctionDoc]
+ * @include [SelectingColumns.ColumnAccessors.WithExample] {@include [SetFillNaNsOperationArg]}
+ * @include [Update.ColumnAccessorsParam]
+ */
+public fun <T, C> DataFrame<T>.fillNaNs(vararg columns: ColumnReference<C>): Update<T, C> =
+    fillNaNs { columns.toColumns() }
+
+/**
+ * TODO this will be deprecated
```
zaleslaw, line 189: Oh, sorry, all these stuff will be deprecated in the https://github.com/Kotlin/dataframe/pull/320
Ignore my comments about deprecation under the Iterables

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/ExpressionsGivenRow.kt
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * - Providing a new value for every selected cell given the row of that cell ({@include [RowExpressionLink]}),
+ * for instance in [map][DataFrame.map], [add][DataFrame.add], and [insert][DataFrame.insert]
+ * (using [RowExpression][DfRowExpression]).
+ *
+ * - Providing a new value for every selected cell given the row of that cell and its previous value ({@include [RowValueExpressionLink]}),
+ * for instance in [update.with][Update.with], and [convert.notNull][Convert.notNull]
+ * (using [RowValueExpression][DfRowValueExpression]).
+ *
+ * Note:
```
zaleslaw, line 23: Probaly Note-> NOTE

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/SelectingColumns.kt
```diff
@@ -0,0 +1,127 @@
...
+import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.api.*
+import org.jetbrains.kotlinx.dataframe.columns.ColumnReference
+import org.jetbrains.kotlinx.dataframe.columns.ColumnSet
+import org.jetbrains.kotlinx.dataframe.columns.SingleColumn
+import org.jetbrains.kotlinx.dataframe.documentation.SelectingColumns.*
+import kotlin.reflect.KProperty
+
+/** {@comment
+ * In this file we provide documentation for high-level operations such as
```
zaleslaw, line 14: is it possible to have comments without any classes or methods or constants

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/SelectingRows.kt
```diff
@@ -0,0 +1,78 @@
...
+ * Selecting rows that satisfy a "Row Condition" ({@include [DocumentationUrls.DataRow.RowConditions]})
+ * can occur in the following two types of operations:
+ * - Selecting entire rows ({@include [RowConditionLink]}), for instance in [filter], [drop], [first], and [count]
+ * (using [RowFilter]).
+ * - Selecting parts of rows using a `where` operation after selecting columns ({@include [RowValueConditionLink]}),
+ * such as with [update], [gather], and [format]
+ * (using [RowValueFilter]).
+ *
+ * A Row Condition is similar to a {@include [RowExpressionsLink]} but expects a [Boolean] as result.
+ * {@comment TODO Is `where` present everywhere it should be?}
```
zaleslaw, line 17: sorry, but ticket or fix

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/samples/ApiLevels.kt
```diff
@@ -0,0 +1,146 @@
...
+            .filter { survived && home.endsWith("NY") && age in 10..20 }
+        // SampleEnd
+    }
+
+    fun extensionProperties1() {
+        val df = DataFrame.read("titanic.csv")
+    }
+}
+
+internal val ColumnsContainer<ApiLevels.TitanicPassenger>.age: DataColumn<Int> @JvmName("TitanicPassenger_age") get() = this["age"] as DataColumn<Int>
```
zaleslaw, line 131: Let's add empty rows between code rows or regroup it somehow

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/ExpressionsGivenRow.kt
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * - Providing a new value for every selected cell given the row of that cell ([Row Expression][org.jetbrains.kotlinx.dataframe.documentation.ExpressionsGivenRow.RowExpression.WithExample]),
+ * for instance in [map][DataFrame.map], [add][DataFrame.add], and [insert][DataFrame.insert]
+ * (using [RowExpression][DfRowExpression]).
+ *
+ * - Providing a new value for every selected cell given the row of that cell and its previous value ([Row Value Expression][org.jetbrains.kotlinx.dataframe.documentation.ExpressionsGivenRow.RowValueExpression.WithExample]),
+ * for instance in [update.with][Update.with], and [convert.notNull][Convert.notNull]
+ * (using [RowValueExpression][DfRowValueExpression]).
+ *
+ * Note:
```
zaleslaw, line 23: Note-> NOTE

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/ExpressionsGivenRow.kt
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * [update with][org.jetbrains.kotlinx.dataframe.api.Update.with]- and [add][org.jetbrains.kotlinx.dataframe.api.add]-like expressions use [AddDataRow][org.jetbrains.kotlinx.dataframe.api.AddDataRow] instead of [DataRow][org.jetbrains.kotlinx.dataframe.DataRow] as the DSL's receiver type.
```
zaleslaw, line 25: tab before -

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/ExpressionsGivenRow.kt
```diff
@@ -0,0 +1,93 @@
...
+ * This is an extension to [RowValueExpression][org.jetbrains.kotlinx.dataframe.RowValueExpression] and
+ * [RowExpression][org.jetbrains.kotlinx.dataframe.RowExpression] that provides access to
+ * the modified/generated value of the preceding row ([AddDataRow.newValue][org.jetbrains.kotlinx.dataframe.api.AddDataRow.newValue]).
+ *
+ * A Row Expression is similar to a [Row Condition][org.jetbrains.kotlinx.dataframe.documentation.SelectingRows] but that expects a [Boolean][Boolean] as result.
```
zaleslaw, line 30: [Row Expression]?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/ExpressionsGivenRow.kt
```diff
@@ -0,0 +1,93 @@
...
+ */
+internal interface ExpressionsGivenRow {
+
+    /**
+     * The key for an @arg that will define the operation name for the examples below.
+     * Make sure to [alias][your examples].
```
zaleslaw, line 36: what is the [your examples]?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/SelectingColumns.kt
```diff
@@ -0,0 +1,171 @@
...
+ * For example:
+ *
+ * `df.`operation`("length", "age")`
+ *
+ * - Select columns using [column accessors][org.jetbrains.kotlinx.dataframe.columns.ColumnReference]
+ * ([Column Accessors API][org.jetbrains.kotlinx.dataframe.documentation.AccessApi.ColumnAccessorsApi]).
+ *
+ * For example:
+ *
+ * `val length by `[column][org.jetbrains.kotlinx.dataframe.api.column]`<`[Double][Double]`>()`
```
zaleslaw, line 54: probably better to wrap with kotlin code?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/DocumentationUrls.kt
```diff
@@ -0,0 +1,66 @@
...
+    /** [See Access APIs on the documentation website.](https://kotlin.github.io/dataframe/apilevels.html) */
+    interface AccessApis {
+
+        /** [See String API on the documentation website.](https://kotlin.github.io/dataframe/stringapi.html) */
+        interface StringApi
+
+        /** [See Column Accessors API on the documentation website.](https://kotlin.github.io/dataframe/columnaccessorsapi.html) */
+        interface ColumnAccessorsApi
+
+        /** [See KProperties API on the documentation website.](https://kotlin.github.io/dataframe/kpropertiesapi.html) */
```
zaleslaw, line 60: What is about @See tag ?

zaleslaw (changes requested):
Not so many issues in reality, I will try to summarize them
- be accurate with inserted code and probably wrap with the kotlin code insertion
- convert some todos into issues or fix them
- answer some questions
- if possible, extract abstraction required for documentation building to the separate places

Some notes for the future:
- Did we have a chance to use constants instead of interfaces?
- Could we use @see tag instead of direct links with []() pattern?
- Could we reuse the same snipppets of code used in the documentation to insert them and checking compiling it once

Jolanrensen:
> Not so many issues in reality, I will try to summarize them
> 
> * be accurate with inserted code and probably wrap with the kotlin code insertion
> * convert some todos into issues or fix them
> * answer some questions
> * if possible, extract abstraction required for documentation building to the separate places
> 
> Some notes for the future:
> 
> * Did we have a chance to use constants instead of interfaces?
> * Could we use @see tag instead of direct links with  pattern?
> * Could we reuse the same snipppets of code used in the documentation to insert them and checking compiling it once

I cannot wrap code samples with triple backticks because then links inside won't be clickable anymore. If that was what you meant with `be accurate with inserted code and probably wrap with the kotlin code insertion`.

`if possible, extract abstraction required for documentation building to the separate places`
Do you suggest moving general docs from, say `update.kt`, to a separate file? say `updateDocs.kt`? I was debating myself whether to do that or not, because it would clutter the `internal` namespace more as well as moving the information about a file away from the file. I'd rather have them close together, I think. That said, the API is less clear if it's filled with docs.

`Did we have a chance to use constants instead of interfaces?`
Yes, constants are possible, but unlike interfaces they cannot be nested. Since the interfaces are never instantiated I don't think there's a speed loss. Isn't unreachable/unused code removed from the bytecode?

`Could we use @see tag instead of direct links with  pattern?`
yes, but currently `@see` tags only show up rendered at the bottom of the doc with no description. That's how I came to the `{@include [SomethingLink]}` pattern, where `interface SomethingLink` contains the proper name and destination for what you're pointing to.

`Could we reuse the same snipppets of code used in the documentation to insert them and checking compiling it once`
We cannot check compilation from the Doc processor. It just processes docs ;). But with `@sample` you can include the same snippets from the documentation (if they are in the same module, I believe).

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/DocumentationUrls.kt
Jolanrensen: Currently `@see` in Kotlin only takes 1 parameter and can just link to another element, not a website. Plus it can only be shown at the bottom and cannot have any extra information. Could make my own `@see` tag processor, but that wouldn't help much I think

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/SelectingColumns.kt
Jolanrensen: Then `[links][to code]` don't work ;)

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/ExpressionsGivenRow.kt
Jolanrensen: it's nothing, just instructions to the user to alias their examples, while also showing what I mean with aliasing.

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/ExpressionsGivenRow.kt
Jolanrensen: image

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/samples/ApiLevels.kt
```diff
@@ -0,0 +1,146 @@
...
+import org.jetbrains.kotlinx.dataframe.annotations.ColumnName
+import org.jetbrains.kotlinx.dataframe.api.add
+import org.jetbrains.kotlinx.dataframe.api.cast
+import org.jetbrains.kotlinx.dataframe.api.column
+import org.jetbrains.kotlinx.dataframe.api.dropNulls
+import org.jetbrains.kotlinx.dataframe.api.filter
+import org.jetbrains.kotlinx.dataframe.api.toListOf
+import org.jetbrains.kotlinx.dataframe.io.read
+
+internal interface ApiLevels {
```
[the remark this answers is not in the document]
Jolanrensen: Yes, it's used as `@sample` in `AccessApi.kt`. I'm not sure my plugin actually works cross-module to target the samples we use for Korro... need to check that.

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
Jolanrensen: Hmm, but then it can only be at the bottom and cannot easily be included in one "common docs" interface

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
Jolanrensen: It will be rendered as a header, there's no difference then, but for consistency I'll add it

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
Jolanrensen: Actually no XD My plugin will turn it into `@param [columns][columns]` since it aliases every link, which renders like "columns - columns" for some reason. I'll probably fix that in the plugin.

Jolanrensen: https://github.com/Jolanrensen/docProcessorGradlePlugin/issues/17

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
Jolanrensen: depends on whether I need to link to something from within the example.

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
Jolanrensen: Actually, this is your PR that will fix it. So depending on which gets merged first...

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/ExpressionsGivenRow.kt
Jolanrensen: yeah? so aggressive? haha

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
Jolanrensen: will move them indeed

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/SelectingColumns.kt
Jolanrensen: indeed no, removed it

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/samples/ApiLevels.kt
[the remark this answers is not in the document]
Jolanrensen: Yes! I managed to include the test module in the sources for the doc processor, we can now reuse the same samples file from the website :) I'll remove this file

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/DocumentationUrls.kt
zaleslaw: mmm, @see could refer on the external site, I suppose, but probably the issue that it added at the bottom is important

zaleslaw:
`Do you suggest moving general docs from, say update.kt, to a separate file? say updateDocs.kt? I was debating myself whether to do that or not, because it would clutter the internal namespace more as well as moving the information about a file away from the file. I'd rather have them close together, I think. That said, the API is less clear if it's filled with docs.`

Probably, I'll try to do it in my PR with docs and we could compare two approaches

PR #220, opened 2023-01-13
Describe how to control content cell limit from Jupyter Notebook

alllex:
This is a docs change that shortly describes how to control content cell limit from Jupyter Notebook.

Without this information is highly cumbersome to work with dataframes that have long text values.


Jolanrensen:
thanks!

PR #221, opened 2023-01-14
Add nestedRowsLimit parameter to DisplayConfiguration

pacher:
By default `DataFrame<T>.toHTML()` renders only first 20 rows of the dataset. We can override it by supplying a `DisplayConfiguration` object with desired `rowsLimit`.

However if dataset contains `FrameColumn` those nested datasets are rendered with hard-coded 5 rows limit and there is no way of changing it.

This PR adds `nestedRowsLimit` parameter to DisplayConfiguration used to limit rendered rows number of nested datasets in FrameColumn.
Also it allows to use `Int.MAX_VALUE` as a sentinel value to turn the limit off in order to render complete dataset.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
@@ -120,8 +120,8 @@ internal fun AnyFrame.toHtmlData(
     val scripts = mutableListOf<String>()
     val queue = LinkedList<Pair<AnyFrame, Int>>()
 
-    fun AnyFrame.columnToJs(col: AnyCol, rowsLimit: Int): ColumnDataForJs {
-        val values = if (rowsLimit == Int.MAX_VALUE) rows() else rows().take(rowsLimit)
+    fun AnyFrame.columnToJs(col: AnyCol, rowsLimit: Int?): ColumnDataForJs {
+        val values = if (rowsLimit != null) rows().take(rowsLimit) else rows()
         val scale = if (col.isNumber()) col.asNumbers().scale() else 1
         val format = if (scale > 0) {
             RendererDecimalFormat.fromPrecision(scale)
@@ -189,7 +189,7 @@ public fun <T> DataFrame<T>.toHTML(
     cellRenderer: CellRenderer = org.jetbrains.kotlinx.dataframe.jupyter.DefaultCellRenderer,
     getFooter: (DataFrame<T>) -> String = { "DataFrame [${it.size}]" },
 ): HtmlData {
-    val limit = configuration.rowsLimit
+    val limit = configuration.rowsLimit ?: Int.MAX_VALUE
 
     val footer = getFooter(this)
     val bodyFooter = buildString {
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
```diff
@@ -121,7 +121,7 @@ internal fun AnyFrame.toHtmlData(
     val queue = LinkedList<Pair<AnyFrame, Int>>()
 
     fun AnyFrame.columnToJs(col: AnyCol, rowsLimit: Int): ColumnDataForJs {
-        val values = rows().take(rowsLimit)
+        val values = if (rowsLimit == Int.MAX_VALUE) rows() else rows().take(rowsLimit)
```
Jolanrensen, line 124: `take()` already just returns the list if `n >= size`. So this change is not needed.

Jolanrensen (changes requested):

Jolanrensen:
Good idea! :D

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
pacher: `rows()` returns Iterable which is not a Collection. In this case inside `take()` there is a line 
```kotlin
val list = ArrayList<T>(n)
```
which throws `OutOfMemoryError: Requested array size exceeds VM limit` if n=Int.MAX_VALUE
So this change is necessary to allow using Int.MAX_VALUE as a sentinel value and turn the limit off if we want to render all the data without guessing the upper limit first.

pacher: I can use `df.count()` to set the limit to the number of rows for an outer dataset to render it fully, but I can't do this for nested dataframes. So I would like to be able to use Int.MAX_VALUE or any other value to turn the limit off and just render everything there is.

Jolanrensen: Ah I see, interesting find! I'd maybe add the check for if rowsLimit < 0, so people can then put in `-1` if they want no limit. I think that's more common.

Jolanrensen: Okay, you convinced me. 
I can only think of one edge case, namely if the list you're taking from is larger than Int.MAX_VALUE, but I think you've reached OOM errors long before that. Unfortunately, there's no "Infinity" option for integers, so this is the best option.

Maybe we'll convert the rows iterable into a `Collection` with `size` to avoid this error in the future.

I will also pass on this case (`Iterable.take()`) to the Kotlin team and see what they think about it.

Jolanrensen (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
Jolanrensen: Actually, we could also just change it to `rows().toList().take(rowsLimit)` since it creates a list anyways.

pacher: Done. I agree that nullable is even more intuitive and hopefully idiomatic  :)

PR #226, opened 2023-01-19
Parse String column to nullable Double with correct locale

Kopilov:
We have already made system locale considering on parsing String to Double but it worked correctly only when converting nullable to nullable or not nullable to not nullable. This affected writing Arrow per schema in particular.


Kopilov:
@zaleslaw please notice that this PR contains test (based on real use case) that does not work originally and patch that fixes it

zaleslaw:
Yeah, thanks for the comments, I am going to make a review next 2 weeks

zaleslaw:
So, I need to run TC and it will be merged

PR #236, opened 2023-01-23
Remove unused useDarkColorScheme parameter from DisplayConfiguration

pacher:
`useDarkColorScheme` parameter was added to `DisplayConfiguration` in cffe4041ebe680195b097811c2a231ca967362f2
but the code using it was removed in 3db46ccccaa1291c0627307d64133317f545e6ae

This PR is cleaning up leftover parameter which is no longer used and could only confuse future readers


Jolanrensen:
@ileasile Can we indeed safely delete this?

koperagen:
Thanks for pointing that out, i removed it in different PR with a rework of this API

Issue #224, opened 2023-01-17
Redundant numeric suffix


koperagen:
In the following example columns under "second" column group should be named "left", "right", without numeric suffix

**Code to reproduce problem:**
```kotlin
val df = dataFrameOf("first", "second")("22-65", "22-66")
df.split("first", "second").by("-").inward("left", "right")
```

![image](https://user-images.githubusercontent.com/12936457/212975600-fcc30df0-ef3b-4c7c-b9f3-44954ea8684c.png)

PR #240, opened 2023-01-25
Fix new columns names generation during split operation (#224)

ermolenkodev:
* Introduce a new scheme for the generation of columns names during split, which takes hierarchy into account

* Add tests


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/split.kt
```diff
@@ -46,22 +50,52 @@ internal fun <T, C, R> splitImpl(
...
+
+internal fun generateUnusedName(
+    df: DataFrame<*>,
+    preferredName: String?,
+    insertPath: ColumnPath,
+    columnsToBeInserted: List<ColumnToInsert>
+): String {
+    // check if column with this name already exists in the df in the same position in the hierarchy,
+    // or we already have a column with this name in the list of columns to be inserted to the same position in the hierarchy
+    fun isUsed(name: String) =
```
Jolanrensen, line 80: Not very fond of inner functions, but you do reuse arguments from the outer function and it makes things more readable, so I think I'll let it slide :)

Jolanrensen (approved):

PR #279, opened 2023-02-22
Rename Column typealias to AnyColumnReference

Jolanrensen:
Since all type aliases with a `*` are called `AnySomething`, it felt odd to have `ColumnReference<*>` be called `Column`. This insinuates that it is a Column object instance, which is not the case. `AnyColumnReference` (or a variant of that) seems more in line with the other aliases.


koperagen:
I'm not sure if it was intentional or indeed an inconsistency
For example, groupBy operation syntax in documentation also uses `column`, `columns` https://kotlin.github.io/dataframe/groupby.html
What do you think?

Jolanrensen:
So does [update](https://kotlin.github.io/dataframe/update.html), but those are still of type `ColumnReference<C>`. It's odd that `groupBy` has the same type of argument, just unbound (`ColumnReference<*>`), but it's suddenly called `Column`. So, I'd suggest making those `ColumnReference<*>` or `AnyColumnReference` for clarity and consistency :)

koperagen (approved):

Issue #280, opened 2023-02-22
Update `where` is ignored if followed by `asFrame`


Jolanrensen:
To reproduce, make a dataframe with a columnGroup containing a nullable String column "A". Then:

```kotlin
df.update { myGroupedColumn }
    .where { it.a == null } // assumingly filtering to only get rows where this holds
    .asFrame {
        update { a }.with { "None" }
    }
```
Running this, makes all values of "a" in the columnGroup turn to "None" instead of honoring the where statement.

Jolanrensen:
Could be solved by something like this:
```kotlin
public infix fun <T, C, R> Update<T, DataRow<C>>.asFrame(expression: DataFrameExpression<C, DataFrame<R>>): DataFrame<T> =
    if (df.isEmpty()) df
    else df.replace(columns)
        .with {
            val processedCol = it
                .asColumnGroup()
                .let { expression(it, it) }
                .asColumnGroup(it.name())

            it.updateImpl(df, filter) { row, _, _ ->
                processedCol[row.index] as DataRow<C>?
            }
        }
```

I'll make a PR with tests later :)

zaleslaw:
I will help with the review

Jolanrensen:
Actually, I see my fix is wrong. Currently processedCol still takes the entire col, which is not expected if you, for instance, want to calculate the mean of some values of the filtered column.
Filter needs to be applied before `expression` is executed

PR #283, opened 2023-02-27
[Fix] Update.asFrame now takes filter into account.

Jolanrensen:
Fix for https://github.com/Kotlin/dataframe/issues/280.

Similar to `Update.perCol {}`, in the body of `asFrame`, users will have access to the entire selected group column (treated as a DF), but now the result of the operation will only be applied to rows that satisfy a given `where/at` clause of the `update` call chain.

See the new test for an example.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
@@ -4,15 +4,20 @@ import org.jetbrains.kotlinx.dataframe.AnyFrame
 import org.jetbrains.kotlinx.dataframe.AnyRow
 import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.DataFrameExpression
+import org.jetbrains.kotlinx.dataframe.DataRow
 import org.jetbrains.kotlinx.dataframe.RowValueFilter
 import org.jetbrains.kotlinx.dataframe.Selector
 import org.jetbrains.kotlinx.dataframe.api.AddDataRow
 import org.jetbrains.kotlinx.dataframe.api.Update
+import org.jetbrains.kotlinx.dataframe.api.asColumnGroup
 import org.jetbrains.kotlinx.dataframe.api.cast
 import org.jetbrains.kotlinx.dataframe.api.indices
 import org.jetbrains.kotlinx.dataframe.api.isEmpty
 import org.jetbrains.kotlinx.dataframe.api.name
 import org.jetbrains.kotlinx.dataframe.api.replace
+import org.jetbrains.kotlinx.dataframe.api.rows
+import org.jetbrains.kotlinx.dataframe.api.toColumn
 import org.jetbrains.kotlinx.dataframe.api.toDataFrame
 import org.jetbrains.kotlinx.dataframe.api.with
 import org.jetbrains.kotlinx.dataframe.columns.ColumnGroup
@@ -20,6 +25,7 @@ import org.jetbrains.kotlinx.dataframe.columns.FrameColumn
 import org.jetbrains.kotlinx.dataframe.columns.size
 import org.jetbrains.kotlinx.dataframe.impl.columns.AddDataRowImpl
 import org.jetbrains.kotlinx.dataframe.impl.createDataCollector
+import org.jetbrains.kotlinx.dataframe.index
 import org.jetbrains.kotlinx.dataframe.type
 import kotlin.reflect.full.isSubclassOf
 import kotlin.reflect.full.withNullability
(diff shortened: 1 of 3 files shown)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/update.kt
```diff
@@ -14,6 +15,47 @@ class UpdateTests {
         df.update { col }.with { 2 } shouldBe df
     }
 
+    @DataSchema
+    interface AAndB {
```
zaleslaw, line 19: I need to say, that for me as a reviewer naming in this test is unclear, especially data class Update.

Could we rename it somehow to be clear from the reading or test repairing point of view

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
```diff
@@ -40,10 +46,45 @@ internal fun <T, C> Update<T, C>.updateWithValuePerColumnImpl(selector: Selector
...
+/**
+ * Implementation for Update As Frame:
+ * Replaces selected column groups with the result of the expression only where the filter is true.
+ */
+internal fun <T, C, R> Update<T, DataRow<C>>.asFrameImpl(expression: DataFrameExpression<C, DataFrame<R>>): DataFrame<T> =
+    if (df.isEmpty()) df
+    else df.replace(columns).with {
+        val src = it.asColumnGroup()
+        val updatedColumn = expression(src, src).asColumnGroup(src.name())
+        if (filter == null) {
```
zaleslaw, line 58: What's happened if we will have some other operations here

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
```diff
@@ -40,10 +46,45 @@ internal fun <T, C> Update<T, C>.updateWithValuePerColumnImpl(selector: Selector
...
     }
 
+/**
+ * Implementation for Update As Frame:
+ * Replaces selected column groups with the result of the expression only where the filter is true.
+ */
+internal fun <T, C, R> Update<T, DataRow<C>>.asFrameImpl(expression: DataFrameExpression<C, DataFrame<R>>): DataFrame<T> =
+    if (df.isEmpty()) df
+    else df.replace(columns).with {
+        val src = it.asColumnGroup()
```
zaleslaw, line 56: Could we rename it somehow related to the Column Group, it's not too easy to understand the algorithm in the last section?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
```diff
@@ -40,10 +46,45 @@ internal fun <T, C> Update<T, C>.updateWithValuePerColumnImpl(selector: Selector
...
+                if (index == -1) it else updatedColumn[index]
+            }.toColumn(src.name)
+        }
+    }
+
 internal fun <T, C> DataColumn<C>.updateImpl(
     df: DataFrame<T>,
     filter: RowValueFilter<T, C>?,
-    expression: (AddDataRow<T>, DataColumn<C>, C) -> C?
+    expression: (AddDataRow<T>, DataColumn<C>, C) -> C?,
```
zaleslaw, line 87: Why do we need here ,?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/update.kt
```diff
@@ -14,6 +15,47 @@ class UpdateTests {
...
+        override val a: Int,
+        override val b: String,
+        val c: Boolean,
+    ) : AAndB
+
+    @Test
+    fun `update asFrame`() {
+        val df = listOf(
+            Update(1, "a", true),
+            Update(2, "b", false),
```
zaleslaw, line 35: Sorry, probably I miss the idea, but where are the null data mentioned in the issue?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
```diff
@@ -40,10 +46,45 @@ internal fun <T, C> Update<T, C>.updateWithValuePerColumnImpl(selector: Selector
...
+
+                    if (filter.invoke(row, currentValue)) {
+                        this += rowIndex
+                        collector.add(currentValue)
+                    }
+                }
+            }
+
+            // Then we only replace the original rows with the updated rows that are inside the filter
+            src.rows().map {
```
zaleslaw, line 77: could we extract it to the separate function with the name related to this comment (to be sure that it mutates the data in the dataframe from the function signature, for example)

zaleslaw (changes requested):
Could you please answer my questions?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/update.kt
Jolanrensen: That's irrelevant. The issue is that .where {} is not taken into account no matter what's inside it or which data is used

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/update.kt
Jolanrensen: Agreed, I'll change the names to something better :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
Jolanrensen: You mean like `asColumnGroup` instead of `asFrame`? That's more a design question. `ColumnGroup` is a specialized version of `DataFrame` (DF with name), so it would behave similarly. Let's discuss that later, for now I just fixed the behavior of the function already present.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
Jolanrensen: probably it was an auto-linting thing, but I am a fan of trailing commas as it is easier to check git diff in the long run.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
Jolanrensen: sure

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
zaleslaw: let me think, I don't remember all the scope of operations

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
zaleslaw: I mean the src variable

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
Jolanrensen: I do, I just kdoc'ed the entirety of update.kt XD

Jolanrensen:
@zaleslaw Greatly simplified the function :) didn't know I could make it this small. Should also make it easier to understand what's going on

zaleslaw (approved):

PR #285, opened 2023-02-28
fixed 1password openapi example

Jolanrensen:


examples/idea-examples/json/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/openapi/openApiApiGuru.kt
```diff
@@ -3,7 +3,7 @@
     name = "ApiGuruOpenApiKsp",
 )
 @file:ImportDataSchema(
-    path = "https://api.apis.guru/v2/specs/1password.local/connect/1.3.0/openapi.json",
+    path = "https://api.apis.guru/v2/specs/1password.local/connect/1.5.7/openapi.json",
```
koperagen, line 6: This example uses URL deliberately? It's a shame that our build depends on an external resource that can be deleted :c

Jolanrensen: yeah... We could also grab this json and put it inside the project. I did that for Apis.guru's OpenApi spec, but that was because theirs was/is broken.

PR #290, opened 2023-03-03
Add support for swing tables integration in IDEA plugin

ermolenkodev:
* Update kotlin-jupyter-api to 0.11.0-311

* Add common interface for classes convertable to AnyFrame

* Add functions for dynamic loading of dataframe rows in IDEA plugin

* Temporary ignore vine notebook test

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
@@ -35,7 +35,7 @@ import kotlin.reflect.KProperty
 import kotlin.reflect.full.isSubtypeOf
 
 /** Users will get an error if their Kotlin Jupyter kernel is older than this version. */
-private const val MIN_KERNEL_VERSION = "0.11.0.198"
+private const val MIN_KERNEL_VERSION = "0.11.0.311"
 
 internal val newDataSchemas = mutableListOf<KClass<*>>()
 
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
@@ -6,29 +6,8 @@ import org.jetbrains.kotlinx.dataframe.AnyCol
 import org.jetbrains.kotlinx.dataframe.AnyFrame
 import org.jetbrains.kotlinx.dataframe.AnyRow
 import org.jetbrains.kotlinx.dataframe.annotations.DataSchema
-import org.jetbrains.kotlinx.dataframe.api.Convert
-import org.jetbrains.kotlinx.dataframe.api.FormattedFrame
-import org.jetbrains.kotlinx.dataframe.api.Gather
-import org.jetbrains.kotlinx.dataframe.api.GroupBy
-import org.jetbrains.kotlinx.dataframe.api.Merge
-import org.jetbrains.kotlinx.dataframe.api.Pivot
-import org.jetbrains.kotlinx.dataframe.api.PivotGroupBy
-import org.jetbrains.kotlinx.dataframe.api.ReducedGroupBy
-import org.jetbrains.kotlinx.dataframe.api.ReducedPivot
-import org.jetbrains.kotlinx.dataframe.api.ReducedPivotGroupBy
-import org.jetbrains.kotlinx.dataframe.api.Split
-import org.jetbrains.kotlinx.dataframe.api.SplitWithTransform
-import org.jetbrains.kotlinx.dataframe.api.Update
-import org.jetbrains.kotlinx.dataframe.api.asColumnGroup
-import org.jetbrains.kotlinx.dataframe.api.asDataFrame
-import org.jetbrains.kotlinx.dataframe.api.columnsCount
-import org.jetbrains.kotlinx.dataframe.api.dataFrameOf
-import org.jetbrains.kotlinx.dataframe.api.frames
-import org.jetbrains.kotlinx.dataframe.api.into
-import org.jetbrains.kotlinx.dataframe.api.isColumnGroup
+import org.jetbrains.kotlinx.dataframe.api.*
 import org.jetbrains.kotlinx.dataframe.api.name
-import org.jetbrains.kotlinx.dataframe.api.toDataFrame
-import org.jetbrains.kotlinx.dataframe.api.values
 import org.jetbrains.kotlinx.dataframe.codeGen.CodeWithConverter
 import org.jetbrains.kotlinx.dataframe.columns.ColumnGroup
 import org.jetbrains.kotlinx.dataframe.columns.ColumnReference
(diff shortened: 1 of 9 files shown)


Jolanrensen:
Is there still a way for users to switch between Swing and HTML tables? Since Swing afaik doesn't support hierarchical tables yet.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataframeConvertable.kt
```diff
@@ -0,0 +1,11 @@
+package org.jetbrains.kotlinx.dataframe.api
+
+import org.jetbrains.kotlinx.dataframe.DataFrame
+
+/**
+ * Common interface for all objects that can be converted to [DataFrame].
+ * Useful for rendering in Jupyter integration.
+ */
+public interface DataframeConvertable {
```
koperagen, line 9: Since it's useful mostly for Jupyter Integration, let's make in an internal interface? I imagine having "df.update("").convertToDataFrame()" in completion might be somewhat confusing

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
```diff
@@ -156,6 +137,7 @@ internal class Integration(
         import("org.jetbrains.kotlinx.dataframe.columns.*")
         import("org.jetbrains.kotlinx.dataframe.jupyter.ImportDataSchema")
         import("org.jetbrains.kotlinx.dataframe.jupyter.importDataSchema")
+        import("org.jetbrains.kotlinx.dataframe.jupyter.getRowsSubsetForRendering")
```
koperagen, line 140: Why should this be imported by default? You want people to use this as an API or there is another reason?

ermolenkodev:
> Is there still a way for users to switch between Swing and HTML tables? Since Swing afaik doesn't support hierarchical tables yet.

Sure, but it is handled on IDEA side. Swing tables UI will be off by default until a hierarchical view support is added.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
ermolenkodev: The Swing tables component uses this function to query subsets of the dataframe. This way, pagination is implemented. It can be imported on every query, but It makes sense to import once because it's a frequent operation.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataframeConvertable.kt
ermolenkodev: Yes, you are right. It's unacceptable to pollute high-level API. But the internal interface will be hard to maintain because, in the future, it will require not to forget to implement DataframeConvertable on every new implementation of core interfaces such as DataColumn, GroupBy, etc. And it will be hard to understand why rendering is not working as expected.
What do you think if we eliminate the idea of a common interface for conversion and just make a conversion function?
````
public fun convertToDataFrame(dataframeLike: Any): DataFrame<*> =
    when (dataframeLike) {
        is Pivot<*> -> dataframeLike.frames().toDataFrame()
        is ReducedPivot<*> -> dataframeLike.values().toDataFrame()
        is PivotGroupBy<*> -> dataframeLike .frames()
        is ReducedPivotGroupBy<*> -> dataframeLike.values()
        is SplitWithTransform<*, *, *> -> dataframeLike.into()
        is Merge<*, *, *> -> dataframeLike.into("merged")
        is Gather<*, *, *, *> -> dataframeLike.into("key", "value")
        is Update<*, *> -> dataframeLike.df
        is Convert<*, *> -> dataframeLike.df
        is FormattedFrame<*> -> dataframeLike.df
        is AnyCol -> dataFrameOf(dataframeLike)
        is AnyRow -> dataframeLike.toDataFrame()
        is GroupBy<*, *> -> dataframeLike.toDataFrame()
        is AnyFrame -> dataframeLike
        else -> throw IllegalArgumentException("Unsupported type: ${dataframeLike::class}")
    }
`
At least it will fail soon at runtime with a clear exception.

Comment (koperagen) [reply core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataframeConvertable.kt:9]: Sure, sounds good

File: core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt (hunk 2 above)
Comment (koperagen) [reply core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt:140]: Hm, i don't fully understand how this integration works. Where is this function called? I don't see usages in dataframe repository, and at the same time i don't understand how it could be possible to call dataframe function from the plugin

Comment (koperagen) [reply core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt:140]: I mean, it looks like this function will be called from generated code, but i somehow missed where this code is generated

Comment (ermolenkodev) [reply core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt:140]: In this case, the IDEA plugin is just a web client of the Jupyter kernel. The plugin sends Kotlin code to the kernel to evaluate and asynchronously receives a response.
something like this 
``
DISPLAY(getRowsSubsetForRendering(Out[x], 0, 20), "")
``

Comment (koperagen) [reply core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt:138]: Ok, i understand now. Could you add this explanation as a comment?  Also, I've been thinking about limiting the scope of this function somehow. Apparently we need it as a part of public API, but now it's also a global function. Would be good to make it not global

File: core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/JupyterHtmlRenderer.kt [hunk 3]
```diff
@@ -1,26 +1,46 @@
...
+    }
+
+    val jsonEncodedDf = json {
+        obj(
+            "nrow" to df.size.nrow,
+            "ncol" to df.size.ncol,
+            "columns" to df.columnNames(),
+            "kotlin_dataframe" to encodeFrame(df.rows().take(limit).toDataFrame())
+        )
+    }.toJsonString()
```
ileasile, line 42: Don't serialize it here, there is no need in it

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/JupyterHtmlRenderer.kt
```diff
@@ -30,5 +50,18 @@ internal inline fun <reified T : Any> JupyterHtmlRenderer.render(
...
+internal fun Notebook.renderAsIFrameAsNeeded(data: HtmlData, jsonEncodedDf: String): MimeTypedResult {
+    val textHtml = if (jupyterClientType == JupyterClientType.KOTLIN_NOTEBOOK) {
+        data.generateIframePlaneText(currentColorScheme)
+    } else {
+        data.toString(currentColorScheme)
+    }
+
+    return mimeResult(
+        "text/html" to textHtml,
+        "application/json" to jsonEncodedDf
```
ileasile, line 65: Use another MIME type here

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/JupyterHtmlRenderer.kt
```diff
@@ -30,5 +50,18 @@ internal inline fun <reified T : Any> JupyterHtmlRenderer.render(
...
+}
+
+internal fun Notebook.renderAsIFrameAsNeeded(data: HtmlData, jsonEncodedDf: String): MimeTypedResult {
+    val textHtml = if (jupyterClientType == JupyterClientType.KOTLIN_NOTEBOOK) {
+        data.generateIframePlaneText(currentColorScheme)
+    } else {
+        data.toString(currentColorScheme)
+    }
+
+    return mimeResult(
```
ileasile, line 63: Use MimeTypedResultEx, it accepts any json objects

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
```diff
@@ -6,29 +6,8 @@ import org.jetbrains.kotlinx.dataframe.AnyCol
...
-import org.jetbrains.kotlinx.dataframe.api.frames
-import org.jetbrains.kotlinx.dataframe.api.into
-import org.jetbrains.kotlinx.dataframe.api.isColumnGroup
+import org.jetbrains.kotlinx.dataframe.api.*
 import org.jetbrains.kotlinx.dataframe.api.name
-import org.jetbrains.kotlinx.dataframe.api.toDataFrame
-import org.jetbrains.kotlinx.dataframe.api.values
 import org.jetbrains.kotlinx.dataframe.codeGen.CodeWithConverter
 import org.jetbrains.kotlinx.dataframe.columns.ColumnGroup
 import org.jetbrains.kotlinx.dataframe.columns.ColumnReference
```
ileasile, line 13: Also update MIN_KERNEL_VERSION in this file

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * A class with utility methods for Kotlin Notebook Plugin integration.
+ * Kotlin Notebook Plugin is acts as a client of Kotlin Jupyter kernel and use this functionality
+ * for dynamic pagination when rendering dataframes.
+ * The plugin sends Kotlin following code to the kernel to evaluate
+ * DISPLAY(KotlinNotebooksPluginUtils.getRowsSubsetForRendering(Out[x], 0, 20), "")
+ */
+public class KotlinNotebookPluginUtils {
+    public companion object {
```
ileasile, line 14:
```suggestion
public object KotlinNotebookPluginUtils {
```

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/JupyterHtmlRenderer.kt
ermolenkodev: There are a few issues:

- in the data frame there, klaxon json is used everywhere, so it needs to be converted to kotlinx.serialization to pass to kernel side functions without serialization. It certainly can be done, but I don't want to mess up with some edge cases)
- on the UI side, there are a few existing interfaces where string serialized data is expected, so it would require the serializing part of the response immediately after deserializing

ileasile: Ok, no problem then. If it won't reduce the number of changes, it doesn't worth it

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
```diff
@@ -56,7 +35,7 @@ import kotlin.reflect.KProperty
 import kotlin.reflect.full.isSubtypeOf
 
 /** Users will get an error if their Kotlin Jupyter kernel is older than this version. */
-private const val MIN_KERNEL_VERSION = "0.11.0.198"
+private const val MIN_KERNEL_VERSION = "0.11.0.311"
```
koperagen, line 38: We only have to update min kernel version if compatibility is broken, i.e. new version of the dataframe won't work with old kernels. Is it broken now? I see that render produces a different result than before. Also, did you check how everything works in Jupyter notebooks?

koperagen (approved):

koperagen:
Thank you!

PR #300, opened 2023-03-14
Improve experience with HTML rendering of dataframes

koperagen:
In this PR i'll to address issues that exist with our HTML rendering
1. It wasn't possible to render DataFrame so that it can be rendered in the browser (initHtml is an internal function, and it's not clear what it does)
2. We lack documentation for custom-rendering-in-Jupyter use-case
3. It exposes HtmlData class that is not a part of our API
4. ...?

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
@@ -1,5 +1,6 @@
 package org.jetbrains.kotlinx.dataframe.io
 
+import org.intellij.lang.annotations.Language
 import org.jetbrains.kotlinx.dataframe.AnyCol
 import org.jetbrains.kotlinx.dataframe.AnyFrame
 import org.jetbrains.kotlinx.dataframe.AnyRow
@@ -21,7 +22,8 @@ import org.jetbrains.kotlinx.dataframe.jupyter.RenderedContent
 import org.jetbrains.kotlinx.dataframe.name
 import org.jetbrains.kotlinx.dataframe.nrow
 import org.jetbrains.kotlinx.dataframe.size
-import org.jetbrains.kotlinx.jupyter.api.HtmlData
+import java.awt.Desktop
+import java.io.File
 import java.io.InputStreamReader
 import java.net.URL
 import java.util.LinkedList
@@ -116,7 +118,7 @@ internal fun nextTableId() = sessionId + (tableInSessionId++)
 internal fun AnyFrame.toHtmlData(
     configuration: DisplayConfiguration = DisplayConfiguration.DEFAULT,
     cellRenderer: CellRenderer,
-): HtmlData {
+): DataFrameHtmlData {
     val scripts = mutableListOf<String>()
     val queue = LinkedList<Pair<AnyFrame, Int>>()
 
@@ -165,40 +167,31 @@ internal fun AnyFrame.toHtmlData(
     }
     val body = getResourceText("/table.html", "ID" to rootId)
     val script = scripts.joinToString("\n") + "\n" + getResourceText("/renderTable.js", "___ID___" to rootId)
-    return HtmlData("", body, script)
+    return DataFrameHtmlData("", body, script)
 }
 
-internal fun HtmlData.print() = println(this)
+internal fun DataFrameHtmlData.print() = println(this)
 
-internal fun initHtml(
-    includeJs: Boolean = true,
-    includeCss: Boolean = true,
-    useDarkColorScheme: Boolean = false,
-): HtmlData =
-    HtmlData(
-        style = if (includeCss) getResources("/table.css") else "",
-        script = if (includeJs) getResourceText("/init.js") else "",
-        body = "",
-    )
-
-@Deprecated("Clarify difference with .toHTML()", ReplaceWith("this.toStandaloneHTML()", "org.jetbrains.kotlinx.dataframe.io.toStandaloneHTML"))
-public fun <T> DataFrame<T>.html(): String = toHTML(extraHtml = initHtml()).toString()
+@Deprecated("Clarify difference with .toHTML()", ReplaceWith("this.toStandaloneHTML().toString()", "org.jetbrains.kotlinx.dataframe.io.toStandaloneHTML"))
+public fun <T> DataFrame<T>.html(): String = toStandaloneHTML().toString()
 
 /**
- * @return HTML String that be saved as an *.html file and displayed in the browser
+ * @return DataFrameHtmlData with table script and css definitions. Can be saved as an *.html file and displayed in the browser
  */
 public fun <T> DataFrame<T>.toStandaloneHTML(
     configuration: DisplayConfiguration = DisplayConfiguration.DEFAULT,
     cellRenderer: CellRenderer = org.jetbrains.kotlinx.dataframe.jupyter.DefaultCellRenderer,
     getFooter: (DataFrame<T>) -> String = { "DataFrame [${it.size}]" },
-): String = toHTML(configuration, extraHtml = initHtml(), cellRenderer, getFooter).toString()
+): DataFrameHtmlData = toHTML(configuration, cellRenderer, getFooter).withTableDefinitions()
 
+/**
+ * @return DataFrameHtmlData without additional definitions. Can be rendered in Jupyter kernel environments
+ */
 public fun <T> DataFrame<T>.toHTML(
     configuration: DisplayConfiguration = DisplayConfiguration.DEFAULT,
-    extraHtml: HtmlData? = null,
     cellRenderer: CellRenderer = org.jetbrains.kotlinx.dataframe.jupyter.DefaultCellRenderer,
     getFooter: (DataFrame<T>) -> String = { "DataFrame [${it.size}]" },
-): HtmlData {
+): DataFrameHtmlData {
     val limit = configuration.rowsLimit ?: Int.MAX_VALUE
 
     val footer = getFooter(this)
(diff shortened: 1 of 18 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
```diff
@@ -181,8 +181,18 @@ internal fun initHtml(
...
 
+@Deprecated("Clarify difference with .toHTML()", ReplaceWith("this.toStandaloneHTML()", "org.jetbrains.kotlinx.dataframe.io.toStandaloneHTML"))
 public fun <T> DataFrame<T>.html(): String = toHTML(extraHtml = initHtml()).toString()
 
+/**
+ * @return HTML String that be saved as an *.html file and displayed in the browser
+ */
+public fun <T> DataFrame<T>.toStandaloneHTML(
+    configuration: DisplayConfiguration = DisplayConfiguration.DEFAULT,
+    cellRenderer: CellRenderer = org.jetbrains.kotlinx.dataframe.jupyter.DefaultCellRenderer,
```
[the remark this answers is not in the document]
koperagen: Thing is, this is our class :) I was confused too. The idea was to move Integration to another module, so we cannot accidentally use something that is not available outside notebooks, so i think this one is ok

pacher:
Can we please, please just make that `extraHtml` default to `initHtml()` instead of `null`. Please?

Before 8e596e6da39e56bf7dfdd6c6913313a6c114a3fd it was just a boolean named `includeInit` and everything was fine.
Now it is called `extraHtml` and defaults to `null` which is causing many problems for no apparent reason.
- `toHtml()` by default does not include javascript and css because `extraHtml` is null, so the output is not very usable. And I can not just do `toHtml(extraHtml = initHtml())` because `initHtml()` is `internal`. So I had to copy-paste and adapt initHtml where I load resources from the dataframe dependency (!). And now I depend on implementation detail and have to monitor every commit in dataframe just in case something has changed.
- there is a `format` operator which produces `FormattedFrame` and the only thing one can do with it is to render it to html using `FormattedFrame.toHTML()`. But as you might have already guessed, it is calling `DataFrame.toHtml()` without `extraHtml = initHtml()` and the output is unusable due to the lack of javascript and css. Clearly it was overlooked after change in 8e596e6da39e56bf7dfdd6c6913313a6c114a3fd 

And the thing is, I can't think of a use case where `extraHtml` should be null, let alone by default. The only place I have found in the dataframe codebase where it is used with default `null` is in some tests where it does not really matter, just saves a few bytes of memory and a few cpu cycles.
I do believe that if such a use case exists, it is justifiable to explicitly pass `null` as a parameter, also in those tests.

P.S. How can I make code suggestions? Should I make a PR against this PR branch or something?

koperagen:
Hi! Sorry for the inconvenience, it was indeed overlooked. With this PR  `toStandaloneHTML` is `toHTML` with default to `initHtml`. I think it's what you want, no? I want to keep `toHTML` with null becuase at least on Kotlin Notebooks our script for tables is provided by the environment, so no need to include it. I.e.
Use `df.toStandaloneHTML()` to save this HTML as a page
Use `DISPLAY(HTML(df.toHTML()))` in notebooks to render HTML with custom DisplayConfiguration or formatting. Or am i wrong and this output doesn't work in notebooks?

pacher:
@koperagen Thanks for quick reply.
I did not use notebooks yet, so can't say what works there. So I did not know about `DISPLAY(HTML(df.toHTML()))`, unfortunately for me it looks like a valid use case.

Yes, I've seen that `toStandaloneHTML()` includes `initHtml` and that's why my thumbs up.
I would still use `toHtml` because it returns `HtmlData` where it is easy to add my own bits of javascript and css. Which brings us to 
>     3. It exposes HtmlData class that is not a part of our API

I've started to use `HtmlData` when it **still was** a part of dataframe public api before it was removed in 3db46cc 
Now it still works, but I need to declare `org.jetbrains.kotlinx:kotlin-jupyter-api:0.11.0-198` dependency and again not to forget to watch every commit in dataframe for a version change. Would be nice to have it as `api` level dependency of dataframe, but I guess it's not going to happen.

Back to the topic, can we maybe than rename parameter `extraHtml` to `initHtml` and make `initHtml()` function public with some other more descriptive name? Like `loadInitHtmlFromResources` or something..

koperagen:
Oh, okay, thank you for your perspective. I wasn't sure if we create our own version of HtmlData, but now i see that we should. For external use, it looks like `plus` overload and `toString` is enough, right?
Unfortunately can't make `org.jetbrains.kotlinx:kotlin-jupyter-api` an api dependency, it's only a `compileOnly` because Jupyter kernel provides these classes.

pacher:
> For external use, it looks like `plus` overload and `toString` is enough, right?

Right, that's the only things dataframe's HtmlData had and the only ones I use.

> Oh, okay, thank you for your perspective. I wasn't sure if we should create our own public version of HtmlData, but now i see that we should. 

Now I feel bad if public version will be created only because of me. Maybe what I do is too much special and could/should be done otherwise. I just do a little tweaking, for example I want to align everything right, not just numbers, so I have
```kotlin
private fun HtmlData.alignAllRight() = this + HtmlData("""
    table.dataframe th, td {
        text-align: right;
    }
""".trimIndent(), "", "")
```
but that is bad example, because it could be done with `format`
Another example is that I want all column groups be expanded by default, so I have
```kotlin
private fun HtmlData.addExpandAllColumnsFunction() = this + HtmlData("", "", """
    function expandAllColumns() {
        const tableRootId = document.getElementsByClassName('dataframe')[0].df.rootId;
        const dframe = document.getElementById("df_" + tableRootId).df;
        dframe.cols.forEach((col) => {if (col.children.length > 0) col.expanded = true});
        for (frameId in dframe.childFrames) dframe.childFrames[frameId].cols.forEach((col) => {if (col.children.length > 0) col.expanded = true});
        DataFrame.renderTable(tableRootId);
    }                    
""".trimIndent())

private fun HtmlData.expandAllColumns() = addExpandAllColumnsFunction() + HtmlData("", "", """
    expandAllColumns();
""".trimIndent())
```
and similar to expand all nested frames.
This kind of stuff. (I know that I depend on javascript in resources, that's on me)
There are probably too many of such possible little tweaks to try and have configuration options in `toHtml()`.

At the end I display it in JavaFx WebView, so I can even interact with it if needed and call javascript functions from kotlin.

koperagen:
Hi @pacher i published this draft as a dev version `10.0-dev-1456`, check it out if you want c: 
I think it is a good idea to have such API for customizing resulting HTML, because we indeed cannot cover all possible tweaks. I've also added a couple of short cut functions that might prove useful when working with HtmlData: `writeHTML` and `openInBrowser`

pacher:
@koperagen 
Awesome! Thanks a lot.

Couple of remarks:
1.  From what I can see javadocs got swapped for `toHtml` and `toStandaloneHTML` in FormattedFrame. Based on implementation `toHtml` should be without definitions and `toStandaloneHTML` should be with definitions.
2. If we are targeting modern kotlin and JDK, maybe we should add an overload with `java.nio.file.Path` to `writeHTML(destination: File)`?
3. Maybe you can have a look and close PR #236. Accept or reject it does not matter. It's not important, but would make merge conflicts or just hang there forever.

koperagen:
> @koperagen Awesome! Thanks a lot.
> 
> Couple of remarks:
> 
> 1. From what I can see javadocs got swapped for `toHtml` and `toStandaloneHTML` in FormattedFrame. Based on implementation `toHtml` should be without definitions and `toStandaloneHTML` should be with definitions.
> 2. If we are targeting modern kotlin and JDK, maybe we should add an overload with `java.nio.file.Path` to `writeHTML(destination: File)`?
> 3. Maybe you can have a look and close PR [Remove unused useDarkColorScheme parameter from DisplayConfiguration #236](https://github.com/Kotlin/dataframe/pull/236). Accept or reject it does not matter. It's not important, but would make merge conflicts or just hang there forever.

Thanks! I checked that `Path` has some advantages over `File` even in such a simple case, so i'll add it.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
```diff
@@ -95,7 +94,14 @@ internal class Integration(
                 applyRowsLimit = false
             )
 
-            render<HtmlData> { notebook.renderHtmlAsIFrameIfNeeded(it) }
+            render<DataFrameHtmlData> {
+                if (notebook.jupyterClientType == JupyterClientType.KOTLIN_NOTEBOOK) {
+                    (it.withTableDefinitions().toJupyterHtmlData()).toIFrame(notebook.currentColorScheme)
```
Jolanrensen, line 99: What's up with these brackets? ;P

docs/StardustDocs/topics/toHTML.md
```diff
@@ -0,0 +1,57 @@
...
+Rendering of hierarchical tables in HTML is supported by JS and CSS definitions
+that can be found in project resources.
+
+Depending on your environment there can be different ways to use result of `toHTML` functions
+
+## IntelliJ IDEA
+
+### Working with result
+
+Following function produces HTML that includes JS and CSS definitions. It can be displayed in the browser and has parameters for customization.
```
Jolanrensen, line 15: The following

docs/StardustDocs/topics/toHTML.md
```diff
@@ -0,0 +1,57 @@
...
+df.toStandaloneHTML(DisplayConfiguration(rowsLimit = null)).openInBrowser()
+df.toStandaloneHTML(DisplayConfiguration(rowsLimit = null)).writeHTML(File("/path/to/file"))
+df.toStandaloneHTML(DisplayConfiguration(rowsLimit = null)).writeHTML(Path("/path/to/file"))
+```
+
+<!---END-->
+
+### Composing multiple tables
+
+`toHTML` and `toStandaloneHTML` return composable `DataFrameHtmlData`. You can use it to include additional scripts, elements, styles on final page or just merge together multiple tables
```
Jolanrensen, line 29: period at the end :)

docs/StardustDocs/topics/toHTML.md
```diff
@@ -0,0 +1,57 @@
...
+listOf(df1, df2, df3).fold(DataFrameHtmlData.tableDefinitions()) { acc, df -> acc + df.toHTML() }
+```
+
+<!---END-->
+
+## Jupyter Notebooks
+
+### Configuring display for individual output
+
+`toHTML` is useful if you want to configure display for single cell, not whole notebook as described in section for [Jupyter Notebooks](jupyterRendering.md)
```
Jolanrensen, line 47: grammar ;P
configure how a single cell is displayed. How to configure this for the entire notebook, check out this section..

Jolanrensen (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
koperagen: Refactoring leftovers c:

docs/StardustDocs/topics/toHTML.md
koperagen: Yes, thanks for pointing that out. I rephrased this sentence

Issue #254, opened 2023-02-07
`rowColExpression` for `perRowCol` wrong in docs


Jolanrensen:
https://kotlin.github.io/dataframe/convert.html,
 https://kotlin.github.io/dataframe/update.html
`rowColExpression` should be `(DataRow, DataColumn) -> NewValue`
Plus, the examples of `perRowCol` are not present in `convert`.

PR #301, opened 2023-03-15
fixed rowColExpression type in docs

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/254

### DIFF
--- a/docs/StardustDocs/topics/convert.md
+++ b/docs/StardustDocs/topics/convert.md
@@ -8,7 +8,7 @@ convert { columnsSelector }
     .with { rowExpression } | .perRowCol { rowColExpression } | .withValue(value)  | to<Type>() | to { colExpression }
 
 rowExpression = DataRow.(OldValue) -> NewValue
-rowColExpression = DataRow.(DataColumn) -> NewValue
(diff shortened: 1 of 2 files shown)


docs/StardustDocs/topics/convert.md
```diff
@@ -8,7 +8,7 @@ convert { columnsSelector }
     .with { rowExpression } | .perRowCol { rowColExpression } | .withValue(value)  | to<Type>() | to { colExpression }
 
 rowExpression = DataRow.(OldValue) -> NewValue
-rowColExpression = DataRow.(DataColumn) -> NewValue
+rowColExpression = (DataRow, DataColumn) -> NewValue
```
koperagen, line 11: Maybe we should fix the type instead?

Jolanrensen: no I think the type is right. Usually we have the context as `this` and then the specific thing you mention as `it`. Here there are two specific things. They cannot both be `it`, so they should be manually named `row` and `col` by the user of the lambda :)

koperagen: Right, two parameters here make more sense

koperagen (approved):

PR #303, opened 2023-03-16
[Bug fix] "SyntaxError: Invalid hexadecimal escape sequence" for rendering DataFrames with "\" in content

Jolanrensen:
Fixed bug where reading attached file using `DataFrame.read("bug.csv")` and rendering in notebooks would result in no rendering at all. This was caused by a "\x" not being escaped and considered an "Invalid hexadecimal escape sequence". This fix replaces "\" characters with proper hex codes.

[bug.csv](https://github.com/Kotlin/dataframe/files/10991059/bug.csv)


koperagen (approved):

koperagen:
Hm, something is wrong with the build

PR #307, opened 2023-03-16
Added 3 APIs for Flatten function

### LINKED ISSUE #292: `flatten` lacks overloads
`flatten` only has a `ColumnSelector` DSL and lacks the other 3 abilities to select columns (column names, column accessors, KProperties)

zaleslaw:
Fixed #292

### DIFF
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/flatten.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/flatten.kt
@@ -3,7 +3,6 @@ package org.jetbrains.kotlinx.dataframe.api
 import io.kotest.matchers.shouldBe
 import org.jetbrains.kotlinx.dataframe.DataRow
 import org.jetbrains.kotlinx.dataframe.annotations.DataSchema
-import org.jetbrains.kotlinx.dataframe.samples.api.TestBase
 import org.junit.Test
 
 class FlattenTests {
@@ -38,7 +37,7 @@ class FlattenTests {
         val castedGroupedDF = grouped.cast<Grouped>()
 
         // KProperties API
-        castedGroupedDF.flatten (Grouped::d) shouldBe df
+        castedGroupedDF.flatten(Grouped::d) shouldBe df
 
         // Extension properties API
         castedGroupedDF.flatten { d } shouldBe df
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/flatten.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/flatten.kt
@@ -1,6 +1,9 @@
 package org.jetbrains.kotlinx.dataframe.api
 
 import io.kotest.matchers.shouldBe
+import org.jetbrains.kotlinx.dataframe.DataRow
+import org.jetbrains.kotlinx.dataframe.annotations.DataSchema
+import org.jetbrains.kotlinx.dataframe.samples.api.TestBase
 import org.junit.Test
 
 class FlattenTests {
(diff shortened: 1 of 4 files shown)


zaleslaw:
@koperagen @Jolanrensen I have a question: why do we miss KProperties as an Access API in the documentation (I could not find tabs with that kind of access). Is it related that it requires a plugin and it is somehow "difficult" for a user?

I also suggest updating the documentation in PRs together with the API changes as a part of triad (code, tests, docs)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/flatten.kt
```diff
@@ -1,6 +1,9 @@
 package org.jetbrains.kotlinx.dataframe.api
 
 import io.kotest.matchers.shouldBe
+import org.jetbrains.kotlinx.dataframe.DataRow
+import org.jetbrains.kotlinx.dataframe.annotations.DataSchema
+import org.jetbrains.kotlinx.dataframe.samples.api.TestBase
```
Jolanrensen, line 6: Unused import

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/flatten.kt
```diff
@@ -13,6 +16,41 @@ class FlattenTests {
...
+    fun `flatten access APIs`() {
+        val df = dataFrameOf("a", "b", "c")(1, 2, 3)
+        val grouped = df.group("a", "b").into("d")
+
+        // String API
+        grouped.flatten("d") shouldBe df
+        val castedGroupedDF = grouped.cast<Grouped>()
+
+        // KProperties API
+        castedGroupedDF.flatten (Grouped::d) shouldBe df
```
Jolanrensen, line 41: random space

zaleslaw:
Not, probably I wrote unclear. We have 4 apis but on many pages in operations we have 3 tabs only without Kproperties

docs/StardustDocs/topics/flatten.md
```diff
@@ -11,13 +11,36 @@ flatten  [ { columns } ]
 Columns after flattening will keep their original names. Potential column name clashes are resolved by adding minimal possible name prefix from ancestor columns.
 
 <!---FUN flatten-->
```
Jolanrensen, line 13: you renamed this function to `flatten_properties` in Modify.kt

We use Korro to check that samples in stardust are actually compilable. So make sure that all examples in markdown are surrounded like
`<!---FUN someFunctionName-->` 
your example, which is between // SampleStart and // SampleEnd comments in samples/api/Modify.kt inside a function with the same name.
`<!---END-->`

Jolanrensen (changes requested):

koperagen:
> Not, probably I wrote unclear. We have 4 apis but on many pages in operations we have 3 tabs only without Kproperties

I think it was done deliberately, but might be discussed. There's not much value in using this API outside its rather narrow use case, like described here https://kotlin.github.io/dataframe/kpropertiesapi.html. So using only 3 APIs in samples was probably an attempt to find a balance between providing this feature for advanced use cases and not confusing people as to what to use.

docs/StardustDocs/topics/flatten.md
zaleslaw: Template looks like: name of function underscore acess api

zaleslaw:
@Jolanrensen fixed the review

docs/StardustDocs/topics/flatten.md
Jolanrensen: Oh now I see! Didn't know this was possible. Cool :)

Jolanrensen (approved):

Issue #262, opened 2023-02-14
[Bug] Disable logging of exceptions from Swagger Parser / Jackson


Jolanrensen:
The `io.swagger.parser.v3:swagger-parser` dependency uses `org.slf4j:slf4j-simple` which needs its logs to be disabled. 
The exception thrown in the build logs is purely visual and doesn't affect the code, but it's annoying nonetheless.

zaleslaw:
I will try to solve this problem somehow, give me a chance

Jolanrensen:
Need to check if this error message is gone after the bootstrap version is bumped to 0.10.0. It should be fixed then

PR #311, opened 2023-03-20
Set logging to QUIET for generateDataFrameXXX tasks

Jolanrensen:
Set logging to QUIET for generateDataFrameXXX tasks in an attempt to shut up swagger parser errors.

Note: This does also disable logs for the WARNING level and lower in this task. Currently I haven't figured out a way yet to only disable logs for a certain package.

Potential fix for https://github.com/Kotlin/dataframe/issues/262


Jolanrensen:
@koperagen We do need to be careful publishing logs with level WARNING from building data schemas now. They won't show up either

PR #314, opened 2023-03-20
Overload toDataFrame for basic types to avoid surprising results

koperagen:
Without overloads Iterable<String>.toDataFrame produces unexpected result

```
val string = listOf("aaa", "aa", null)
string.toDataFrame()
```
=>
  length
0    3
1    2
2 null

I think it will be more reasonable to create a dataframe with a single column for all basic types (numbers, strings, char, boolean), and do reflection scan for everything else

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
@@ -7,6 +7,7 @@ import org.jetbrains.kotlinx.dataframe.ColumnsSelector
 import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.DataFrame
 import org.jetbrains.kotlinx.dataframe.DataRow
+import org.jetbrains.kotlinx.dataframe.annotations.DataSchema
 import org.jetbrains.kotlinx.dataframe.columns.ColumnPath
 import org.jetbrains.kotlinx.dataframe.impl.ColumnNameGenerator
 import org.jetbrains.kotlinx.dataframe.impl.api.createDataFrameImpl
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -187,6 +188,83 @@ public abstract class CreateDataFrameDsl<T> : TraversePropertiesDsl {
...
+    OfSingleValueColumn<U>::value from { it }
+}.cast()
+
+@JvmName("toDataFrameULong")
+public inline fun <reified U : ULong?> Iterable<U>.toDataFrame(): DataFrame<OfSingleValueColumn<U>> = toDataFrame {
+    OfSingleValueColumn<U>::value from { it }
+}.cast()
+
+@DataSchema
+public interface OfSingleValueColumn<T> {
```
Jolanrensen, line 264: Bit of a strange name. I see you were going for "DataFrame of single value column", but hmm... I think "ValueColumn" would be also fine

Jolanrensen (commented):
I agree we need to take care of primitives. A list of strings gives me a dataframe of string lengths without the new overloads haha.
However, do all of these need to be overloads for `toDataFrame`? Or can these just be moved to `createDataFrameImpl`?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
koperagen: But we already have ValueColumn :c I wanted to avoid confusion with any "columns" that we have in API

Jolanrensen:
> > I agree we need to take care of primitives. A list of strings gives me a dataframe of string lengths without the new overloads haha. However, do all of these need to be overloads for `toDataFrame`? Or can these just be moved to `createDataFrameImpl`?
> 
> I don't think so. At east we'll lose the ability to return this typed dataframe

Of course! You're right :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
Jolanrensen: I see. Maybe `ValueProperty<T>`? Since we already have `KeyValueProperty<T>` as default `@DataSchema` interface.

koperagen: sounds good

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -187,6 +188,83 @@ public abstract class CreateDataFrameDsl<T> : TraversePropertiesDsl {
...
+    ValueProperty<U>::value from { it }
+}.cast()
+
+@JvmName("toDataFrameULong")
+public inline fun <reified U : ULong?> Iterable<U>.toDataFrame(): DataFrame<ValueProperty<U>> = toDataFrame {
+    ValueProperty<U>::value from { it }
+}.cast()
+
+@DataSchema
+public interface ValueProperty<T> {
```
Jolanrensen, line 264: Could maybe be moved to a separate file, similar to KeyValueProperty :)

Jolanrensen (approved):
The rest looks good! 
Btw, maybe we should also add overloads for (primitive) arrays toDataFrame, since they aren't iterables

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
koperagen: I'd prefer it here, doesn't have much value on its own i think

koperagen:
> The rest looks good! Btw, maybe we should also add overloads for (primitive) arrays toDataFrame, since they aren't iterables

Not sure, so let's return to it later c: I'll merge this

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
Jolanrensen: thought the same about KeyValueProperty, but who knows, we cannot know everything a user might want :)  And a value-only DataSchema seems to me exactly the sort of thing people might want to use (or create themselves)

PR #316, opened 2023-03-21
Made the API formatted in the same way

zaleslaw:


zaleslaw:
@Jolanrensen @koperagen please have a look, it's not a final version and formatting, but I need your opinion

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
```diff
@@ -166,7 +166,14 @@ public fun DataColumn<String?>.convertToDouble(locale: Locale? = null): DataColu
         try {
             return mapIndexed { row, value ->
                 currentRow = row
-                value?.let { parser(value.trim()) ?: throw TypeConversionException(value, typeOf<String>(), typeOf<Double>(), path) }
+                value?.let {
+                    parser(value.trim()) ?: throw TypeConversionException(
+                        value,
+                        typeOf<String>(),
+                        typeOf<Double>(),
+                        path
```
Jolanrensen, line 174: trailing comma

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
```diff
@@ -201,8 +208,14 @@ public fun <T : Any> DataColumn<T?>.convertToBoolean(): DataColumn<Boolean?> = c
 
 // region convert URL
 
-public fun <T, R : URL?> Convert<T, R>.toIFrame(border: Boolean = false, width: Int? = null, height: Int? = null): DataFrame<T> = to { it.map { IFRAME(it.toString(), border, width, height) } }
-public fun <T, R : URL?> Convert<T, R>.toImg(width: Int? = null, height: Int? = null): DataFrame<T> = to { it.map { IMG(it.toString(), width, height) } }
+public fun <T, R : URL?> Convert<T, R>.toIFrame(
+    border: Boolean = false,
+    width: Int? = null,
+    height: Int? = null
```
Jolanrensen, line 214: trailing comma

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/remove.kt
```diff
@@ -10,15 +10,44 @@ import kotlin.reflect.KProperty
...
 public fun <T> DataFrame<T>.remove(vararg columns: KProperty<*>): DataFrame<T> = remove { columns.toColumns() }
+
 public fun <T> DataFrame<T>.remove(vararg columns: String): DataFrame<T> = remove { columns.toColumns() }
+
 public fun <T> DataFrame<T>.remove(vararg columns: AnyColumnReference): DataFrame<T> = remove { columns.toColumns() }
-public fun <T> DataFrame<T>.remove(columns: Iterable<AnyColumnReference>): DataFrame<T> = remove { columns.toColumnSet() }
+
+@Deprecated(
+    message = "It will be removed in the 0.10.0 release",
+    level = DeprecationLevel.WARNING,
```
Jolanrensen, line 26: Could be level ERROR

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/remove.kt
```diff
@@ -10,15 +10,44 @@ import kotlin.reflect.KProperty
...
+    replaceWith = ReplaceWith("remove(columns)")
```
Jolanrensen, line 27: and `remove(columns)` is what we're going to remove. We still need to decide what to replace it with but it will probably be `remove { columns.toColumnSet() }`

Jolanrensen (commented):
Good start :) 
General things I noticed: Trailing comma's. Nikita was worried about the git-blame, but with large refactorings like this it will probably mess up git history anyways. Trailing commas might make future git changes better readable.

Many times I see 
```
fun someShortFun(something: Something): T = something

fun someLongerFunction(something: Something): SomethingLong =
    blahblah
```
Would it be better if we convert all functions to 2 lines instead of just the ones getting past the break-off line?

There should probably be a newline between each function.

I don't care whether there should be newlines in between // region starts and // endregion, but it should be consistent everywhere :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/remove.kt
zaleslaw: It doesn't work with the ERROR level, because it depends on the function in implementation with the same level of deprecation and it will be not compiled

zaleslaw: Oh, do you suggest change the implementation? In this way it should work, I suppose. I started to think about binary compatibility and don't think about replacement of the implementation.

Issue #317, opened 2023-03-22
ColumnSelectionDsl does not support KProperties API


Jolanrensen:
https://kotlin.github.io/dataframe/columnselectors.html doesn't support KPoperties except in some select places:

```kotlin
df.select { it[Person::name] } // works
df.select { Person::name } // cannot work
df.select { Person::name() } // this also not actually...
df.select { (Person::name)() } // this could

// by column path
df.select { it[Person::name][Name::firstName] } // works
df.select { Person::name[Name::firstName] }

// with a new name
df.select { Person::name named "Full Name" }

// converted
df.select { Person::name[Name::firstName].map { it.lowercase() } }

// column arithmetics
df.select { 2021 - Person::age }

// two columns
df.select { Person::name and Person::age } // works

// range of columns
df.select { Person::name..Person::age }

// all children of ColumnGroup
df.select { Person::name.all() }

// dfs traversal of all children columns
df.select { Person::name.allDfs() }
```
If we claim to have all these APIs, then this should also be supported

PR #318, opened 2023-03-22
Column Selection DSL improvements

Jolanrensen:
Based on https://github.com/Kotlin/dataframe/issues/317
I went over the Column Selection DSL and added support for KProperties everywhere.
Additionally, I saw there were some missing APIs for Strings as well, like
`select { "myGroup".allAfter("someCol") }`, `select { "myCol".groups { true } }` etc.

Finally, I renamed `allUntil` to `allUpTo`, as "until" means excluding the last one, while "up to" does not. And I renamed `allSince` to `allFrom`.

Docs are updated too.


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Access.kt
```diff
@@ -706,6 +706,41 @@ class Access : TestBase() {
...
+        df.select { Person::name[Name::firstName].map { it.lowercase() } }
+
+        // column arithmetics
+        df.select { 2021 - (Person::age)() }
+
+        // two columns
+        df.select { Person::name and Person::age }
+
+        // range of columns
+        df.select { Person::name..Person::age }
```
zaleslaw, line 734: Very strange API:)

Jolanrensen: not weirder than "name".."age" imo

zaleslaw: Of course not, I mean totally the range of columns based on name, I could imagine ranging the ordered things or alphabetically, but not the range of column names in the structure, is this order guaranteed by compiler or somehow?

Jolanrensen: Nope, that's impossible afaik. However, it's still very useful to select multiple columns quickly. But, as with all non-typesafe operations (resulting in `Any?` types), it works best in Jupyter or with the upcoming plugin, which generate accessors on the fly.

Jolanrensen:
If we don't merge this PR, it will be joint with https://github.com/Kotlin/dataframe/pull/331, since that builds on top of this PR adding docs and other missing functions

PR #331, opened 2023-03-30
Columns Selection DSL KDocs and missing API overloads

Jolanrensen:
Adds docs surrounding the Column Selection DSLs

Depends on https://github.com/Kotlin/dataframe/pull/318


core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
```diff
@@ -0,0 +1,515 @@
...
+            val secondName by it.map<_, String?> { null }.asValueColumn()
+            val thirdName by it.map<_, String?> { null }.asValueColumn()
+
+            dataFrameOf(firstName, secondName, thirdName)
+                .cast<FirstNames>(verify = true)
+                .asColumnGroup(firstNames)
+        }
+
+        dfGroup.select { colGroup("name") } shouldBe dfGroup.select { name }
+        dfGroup.select { colGroup<String>("name") } shouldBe dfGroup.select { name }
```
koperagen, line 166: what is colGroup of String?

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
```diff
@@ -0,0 +1,515 @@
...
+        }
+
+        df.select {
+            "name".cols("firstName", "lastName")
+        } shouldBe df.select {
+            Person::name.cols("firstName", "lastName")
+        }
+
+        df.select {
+            "name"["firstName", "lastName"]
```
koperagen, line 337: it still looks more like "pathOf" for me than slicing

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Access.kt
```diff
@@ -630,6 +630,7 @@ class Access : TestBase() {
         df.select { age and name }
         df.fillNaNs { dfsOf<Double>() }.withZero()
         df.remove { cols { it.hasNulls() } }
+        df.group { cols { it.data != name } }.into { "nameless" }
```
koperagen, line 633: i don't understand this line

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
```diff
@@ -110,6 +124,11 @@ public fun frameColumn(path: ColumnPath): ColumnAccessor<AnyFrame> = column(path
 @JvmName("frameColumnTyped")
 public fun <T> frameColumn(path: ColumnPath): ColumnAccessor<DataFrame<T>> = column(path)
 
+@JvmName("frameColumnDataFrameKProperty")
+public fun <T> frameColumn(property: KProperty<DataFrame<T>>): ColumnAccessor<DataFrame<T>> = column(property)
+
+public fun <T> frameColumn(property: KProperty<T>): ColumnAccessor<DataFrame<T>> = column(property.name)
```
koperagen, line 130: again, it's probably `List<T>`

koperagen: oh, ok, i'm probably starting to review the same changes again :D

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/ColumnExpression.kt
```diff
@@ -0,0 +1,28 @@
...
+ * [The Add DSL][AddDsl.expr], [The Columns Selection DSL][ColumnsSelectionDsl.expr], and
+ * [The Create DataFrame DSL][CreateDataFrameDsl.expr].
+ *
+ * The `expr {}` call functions like a mapping statement iterating over the object it's called on.
+ */
+internal interface ColumnExpression {
+
+    /**
+     * ## Column Expression
+     * Create a temporary new column by defining an expression to fill up each row.
```
koperagen, line 20: Not sure if the word temporary fits here. In `df.select { expr { } }` it will become "permanent. I'd call it an adapter from column-based DSL to row-based (or the other way around)

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/ColumnExpression.kt
```diff
@@ -0,0 +1,28 @@
+package org.jetbrains.kotlinx.dataframe.documentation
+
+import org.jetbrains.kotlinx.dataframe.api.*
+
+/**
+ * ## Column Expression
+ * In many DSLs, the lambda `expr {}` can be used to
```
koperagen, line 7: Sometimes it's "expr("name") { }`. And look at my comment below

koperagen (changes requested):
Nice c: Let's see how this works out.

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
```diff
@@ -110,6 +124,11 @@ public fun frameColumn(path: ColumnPath): ColumnAccessor<AnyFrame> = column(path
 @JvmName("frameColumnTyped")
 public fun <T> frameColumn(path: ColumnPath): ColumnAccessor<DataFrame<T>> = column(path)
 
+@JvmName("frameColumnDataFrameKProperty")
+public fun <T> frameColumn(property: KProperty<DataFrame<T>>): ColumnAccessor<DataFrame<T>> = column(property)
+
+public fun <T> frameColumn(property: KProperty<T>): ColumnAccessor<DataFrame<T>> = column(property.name)
```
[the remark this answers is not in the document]
Jolanrensen: yes good one!

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
```diff
@@ -11,77 +10,730 @@ import org.jetbrains.kotlinx.dataframe.DataColumn
...
+     * @return A [ColumnSet][org.jetbrains.kotlinx.dataframe.columns.ColumnSet] containing all columns from [this] to [endInclusive].
+     * @throws [IllegalArgumentException] if the columns have different parents.
+     *
+     */
+    public operator fun KProperty<*>.rangeTo(endInclusive: String): ColumnSet<*> =
+        toColumnAccessor().rangeTo(endInclusive.toColumnAccessor())
+
+    /**
+     * ## Subset of Columns
+     * Returns a [ColumnSet][org.jetbrains.kotlinx.dataframe.columns.ColumnSet] containing all columns from [this] to [endInclusive].
```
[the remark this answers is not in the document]
Jolanrensen: I do want to keep it there at the top, people don't usually read `@return` unless they want more information. I can rephrase it a bit though.

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
```diff
@@ -100,33 +752,3093 @@ public interface ColumnsSelectionDsl<out T> : ColumnSelectionDsl<T>, SingleColum
...
+     * #### For example:
+     *
+     * `df.`[select][org.jetbrains.kotlinx.dataframe.api.ColumnsSelectionDsl.select]` { `[col][org.jetbrains.kotlinx.dataframe.api.ColumnsSelectionDsl.col]`("columnGroup"["columnName"]) }`
+     *
+     * `df.`[select][org.jetbrains.kotlinx.dataframe.api.ColumnsSelectionDsl.select]` { myColGroup.`[col][org.jetbrains.kotlinx.dataframe.api.ColumnsSelectionDsl.col]`("columnGroup"["columnName"]) }`
+     *
+     * @return A [ColumnAccessor][org.jetbrains.kotlinx.dataframe.columns.ColumnAccessor] for the column with the given argument.
+     * @see [column]
+     * @see [colGroup]
+     * @see [frameCol]
```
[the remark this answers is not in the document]
Jolanrensen: colGroup and frameCol are of the same group of functions, and column is what is called. Unfortunately kdocs doesn't support extra info after `@see`, they just get summed up in a list.
image
doesn't look that weird, but I could remove them if you want to

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
```diff
@@ -100,33 +752,3093 @@ public interface ColumnsSelectionDsl<out T> : ColumnSelectionDsl<T>, SingleColum
...
+     * #### For example:
+     * TODO: [Issue #286](https://github.com/Kotlin/dataframe/issues/286)
+     *
+     * #### Examples for this overload:
+     *
+     * `// although these can be shortened to just the `[colsOf][colsOf]` call`
+     *
+     * `df.`[select][select]` { `[colsOf][colsOf]`<`[String][String]`>().`[cols][cols]` { "e" `[in][String.contains]` it.`[name][ColumnPath.name]`() } }`
+     *
+     * `df.`[select][select]` { `[colsOf][colsOf]`<`[String][String]`>()`[`[`][cols]`{ it.`[any][ColumnWithPath.any]` { it == "Alice" } }`[`]`][cols]` }`
```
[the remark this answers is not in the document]
Jolanrensen: image
yes, it renders like this

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
```diff
@@ -440,16 +4166,102 @@ public interface ColumnsSelectionDsl<out T> : ColumnSelectionDsl<T>, SingleColum
...
+     *
+     * `df.`[select][org.jetbrains.kotlinx.dataframe.DataFrame.select]` { myColumnGroup.`[colsOf][org.jetbrains.kotlinx.dataframe.api.colsOf]`<`[Int][Int]`>() }`
+     *
+     * `df.`[select][org.jetbrains.kotlinx.dataframe.DataFrame.select]` { "myColumnGroup"<Type>().`[colsOf][org.jetbrains.kotlinx.dataframe.api.colsOf]`<`[Int][Int]`> { it.`[size][org.jetbrains.kotlinx.dataframe.DataColumn.size]` > 10 } }`
+     *
+     * `df.`[select][org.jetbrains.kotlinx.dataframe.DataFrame.select]` { (Type::myColumnGroup)().`[colsOf][org.jetbrains.kotlinx.dataframe.api.colsOf]`<`[Double][Double]`>() }`
+     * ## ‎
+     * Finally, [colsOf][org.jetbrains.kotlinx.dataframe.api.colsOf] can also take a [KType] argument instead of a reified type.
+     * This is useful when the type is not known at compile time or when the API function cannot be inlined.
+     * (TODO: [Issue: #325, context receiver support](https://github.com/Kotlin/dataframe/issues/325))
```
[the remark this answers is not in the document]
Jolanrensen: Does show we're working towards it and why we need the KType argument overload, right?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/ColumnExpression.kt
```diff
@@ -0,0 +1,28 @@
...
+ * create a temporary new column by defining an expression to fill up each row.
+ *
+ * These DSLs include (but are not limited to):
+ * [The Add DSL][AddDsl.expr], [The Columns Selection DSL][ColumnsSelectionDsl.expr], and
+ * [The Create DataFrame DSL][CreateDataFrameDsl.expr].
+ *
+ * The `expr {}` call functions like a mapping statement iterating over the object it's called on.
```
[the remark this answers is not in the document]
Jolanrensen: I'll rephrase

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
```diff
@@ -0,0 +1,515 @@
...
+            }
+        } shouldBe df.select { name.firstName }
+    }
+
+    @Test
+    fun col() {
+        df.select { col("age") } shouldBe df.select { age }
+        df.select { col<Int>("age") } shouldBe df.select { age }
+        df.select { col(pathOf("age")) } shouldBe df.select { age }
+        df.select { col<Int>(pathOf("age")) } shouldBe df.select { age }
```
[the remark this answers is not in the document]
Jolanrensen: yes, it definitely works like that. This is for completeness, since we also have `colGroup(path)` and `frameCol(path)`

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
Jolanrensen: creates a column accessor for a column group named "name"

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
Jolanrensen: pathOf would be `"name"["firstName"]["lastName"]`. I actually think this is kinda neat. It allows you to select multiple columns way down in a path: `"some"["path"]["to"]["multiple", "columns"]`

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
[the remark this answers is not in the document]
koperagen: oh, i see now. Yes, without additional information it's not clear that they're functions from the same "group" :c

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Access.kt
Jolanrensen: It used to be:
```kotlin
df.group { cols { it != name } }.into { "nameless" }
```
because `cols` took a `(AnyCol) -> Boolean`. However, I converted the argument to a `ColumnFilter<*>` which is like `(ColumnWithPath<*>) -> Boolean`. This also in preparation for being able to do `cols {}.recursively()` (Since `dfs {}` also works with columns with path).
However, since the column is now wrapped into a ColumnWithPath ,you need to call `.data.` to get the actual column and be able to compare it with some accessor. If you have a better idea how to solve this, let me know :)

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/ColumnExpression.kt
Jolanrensen: good point

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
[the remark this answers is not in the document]
koperagen: Yes, but this information is not for an API reference IMO.

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/ColumnExpression.kt
Jolanrensen: yes, but not everywhere. This is the common docs part, so if a version of `expr` supports a name, I'll mention it at the specific version.

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
koperagen: I mean type parameter

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
koperagen: I think it takes a bit more effort to "parse" this expression and recognize the difference and to understand that the result is multiple columns.
Here the difference is more obvious
`"some"["path"]["to"].cols("multiple", "columns")`

Jolanrensen: Yeah for sure :) But again for completeness, should we remove it?
`pathOf("path", "to")["a", "b"]` doesn't look that wrong

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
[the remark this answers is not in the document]
Jolanrensen: Right, I'll put it in a `{@comment }` block :) So it's there for us but not for the API docs

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Access.kt
koperagen: Can you mention it in that issue? And, how that `it.hasNulls()` above works?

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
Jolanrensen: `ColumnAccessor<DataRow<String>>`

Jolanrensen: doesn't make much sense now I think of it haha

Jolanrensen: `colGroup<Name>("name")` is better :)

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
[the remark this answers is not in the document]
koperagen: What if we make it "deprecated"? Instead of `df.select { col(pathOf("age")) }` use `df.select { pathOf("age") }`

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
koperagen: Ok, but type parameter in ColumnGroup and DataRow is a data schema, i.e. "MyDataSchema". So i think this example is misleading - String is not a DataSchema

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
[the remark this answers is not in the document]
Jolanrensen: hmm, could do that... But then we'd need to do that for all identity functions. I'd rather just mention in the docs it doesn't do anything and leave it for completeness. Makes it also easy to refactor for users if they can just swap out `colGroup` for `col`.

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
koperagen: Maybe? I don't know, let's check how it looks in completion. How discoverable is this API, how many options there are

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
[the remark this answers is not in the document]
Jolanrensen: I will add a note in the docs for it though, cause I see it's missing

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
Jolanrensen: I agree, Name is

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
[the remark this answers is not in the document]
Jolanrensen: the typed version is still useful though :)

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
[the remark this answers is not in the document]
koperagen: do you think it worth adding examples for an overload with type argument here? kdoc mentions "typed" and "untyped". This is new. Maybe rephrase it to "with and without type parameter"?

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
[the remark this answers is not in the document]
koperagen: maybe it means we lack a typed version of pathOf? c: There should be "cast" though

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Access.kt
Jolanrensen: ColumnWithPath still inherits from DataColumn, so the only thing changing is exact equality

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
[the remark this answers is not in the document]
Jolanrensen: yeah you can also cast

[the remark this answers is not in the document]
Jolanrensen: but I'd rather have one overload too many and know we've considered all cases than not enough because some obscure overload prevents it from working correctly

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Access.kt
koperagen: Ok, then let's continue there later. Not a show stopper

Jolanrensen: yeah, we'll get back to it in the dfs renaming issue :)

Issue #333, opened 2023-03-31
Rename dfs function


Jolanrensen:
When I first read 
```kotlin
df.select { dfsOf<Type> { it.size() > 10 } }
```
I thought it meant "Select dataframes of given type where their size is larger than 10".
Asking a friend of mine, they recognized "dfs" as "depth first search" but then they assumed it would return the deepest nested column, since a search is done when a result is found.

Plus, since this function traverses all column groups, it could just as well have been named bfs (breadth first search) and give the same result.

Clearly, this function needs renaming to avoid further confusion.

The functions that need to be renamed include `dfs {}`, `dfsOf<> {}` and `allDfs(includeGroups: Boolean)`. Plus, in the internal parts of the library, the name DFS is also sometimes used to just mean a "recursive flattening" of columns, which is definitely not a "search".

Candidates for a rename include:
```kotlin
df.select { traversingGroups { it.size() > 10 } }
df.select { traversingGroupsColsOf<Type> { it.size() > 10 } } // ??
df.select { allTraversingGroups(includeGroups = false) } // ?
```
or
```kotlin
df.select { cols(traversingGroups = true) { it.size() > 10 } }
df.select { colsOf<Type>(traversingGroups = true) { it.size() > 10 } }
df.select { all(traversingGroups = false, includeGroups = true) } // ?
```
or
```kotlin
df.select { allLevels { it.size() > 10 } }
df.select { allLevelsOf<Type> { it.size() > 10 } } // ?
df.select { allLevels(includeGroups = false) } // ?
```

Other ideas are welcome :)

pacher:
Plus many. I thought the same thing, dfs = dataframes so probably something to do with frame columns. And I was always thinking "if only there would be a way to get all the columns recursively from all the nested groups". I was doing some workarounds and planning to try to implement it at some point and make a PR.
Until I finally saw "depth first search" mentioned once somewhere in the docs at the bottom of the page.
But now that I know it, it is hard to beet the name `dfs` in conciseness. And I use it quite a lot, so every letter counts. Having `traversingGroups` everywhere would look awful and be a big bummer for me.

What would help is a dedicated page in the docs visible in the index tree or a more prominent section with the words "traverse" or "recursive". I would certainly remember that such functionality exists when reading the docs for the first time. But at that point `dfs` was mentioned only once in some table with other functions and I missed it.

Jolanrensen:
[...]

PR #361, opened 2023-04-20
Rework of dfs functions to recursively

Jolanrensen:
https://github.com/Kotlin/dataframe/issues/333


Jolanrensen:
`df.select { first { ... }.recursively() }` now works, but I 'm not sure why. Needs more tests and maybe a clearer way to indicate a `ColumnSet` passes on its `recursive` behavior to its parent or not.
For `first {}` which uses `singleImp()`, `singleImpl()` was modified to pass on its recursive behavior. This might however be needed in more places.

Jolanrensen:
`df.select { it[0, 3].recursively() }` 
what does this even mean? Take index 0 and 3 of every column group maybe. But this would then break when it hits a non-columnGroup. 
Allowing `.recursively()` on any columnSet might open a can of worms we don't want. I think I'm gonna have another attempt at changing the return type of functions that allow a .recursively call  in favor of this generic but messy approach.

Jolanrensen:
This will be closed in favor of https://github.com/Kotlin/dataframe/pull/363

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
```diff
@@ -3868,24 +4085,18 @@ public interface ColumnsSelectionDsl<out T> : ColumnSelectionDsl<T>, SingleColum
...
-    public fun SingleColumn<AnyRow>.cols(range: IntRange): ColumnSet<*> =
-        transform { it.flatMap { it.children().subList(range.first, range.last + 1) } }
+    public fun SingleColumn<*>.cols(range: IntRange): ColumnSet<*> =
+        colsInternal(range)
 
     /**
      *
      */
-    public operator fun SingleColumn<AnyRow>.get(range: IntRange): ColumnSet<*> = cols(range)
+    public operator fun SingleColumn<*>.get(range: IntRange): ColumnSet<*> = cols(range)
```
koperagen, line 4099: Why? I think SingleColumn<AnyRow> is a ColumnGroup and this operations make sense, but not for SingleColumn<*>

Jolanrensen: Well yes, but then it breaks when you 
```kotlin
df.select { it["name"][0..5] }
```
since it doesn't know the type of the single column you select that way.

Also, I'm gonna close this PR in favor of dfs-rename3, just waiting for Tolya to answer.

PR #363, opened 2023-04-22
Rework of dfs functions to recursively (changing return type of recurse-able functions)

Jolanrensen:
Same as https://github.com/Kotlin/dataframe/pull/361 but using a different tactic.


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
```diff
@@ -3,181 +3,159 @@ package org.jetbrains.kotlinx.dataframe.api
...
-            name.colsOf<String>().first { col ->
-                col.any { it == "Alice" }
-            }
-        }
+        listOf(
+            df.select { name },
+            df.select { first() },
+            df.select { all().first() },
+            df.select { first { it.name().startsWith("n") } },
+        ).shouldAllBeEqual()
```
zaleslaw, line 19: Should they be equal between each other? Or to another snippet of API

I mean, that for now, the test structure is hidden a little bit and naming could lead to the misunderstanding

But the goal of refactoring is clear

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/replace.kt
```diff
@@ -37,7 +37,7 @@ public fun <T, C> DataFrame<T>.replace(columns: Iterable<ColumnReference<C>>): R
 
 public fun <T> DataFrame<T>.replaceAll(
     vararg valuePairs: Pair<Any?, Any?>,
-    columns: ColumnsSelector<T, *> = { allDfs() },
+    columns: ColumnsSelector<T, *> = { all().recursively(includeGroups = false) },
```
koperagen, line 40: Why has the default value for includeGroups changed?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/tree/TreeNode.kt
```diff
@@ -48,19 +48,34 @@ internal class TreeNode<T>(
         return addChild(childName, createData())
     }
 
-    fun dfs(enterCondition: (TreeNode<T>) -> Boolean = { true }, yieldCondition: (TreeNode<T>) -> Boolean = { true }): List<TreeNode<T>> {
+    @Deprecated("Use allChildren instead", ReplaceWith("allChildren(enterCondition, yieldCondition)"))
+    fun dfs(
```
koperagen, line 52: Why deprecate it? It's an internal class

docs/StardustDocs/topics/ColumnSelectors.md
```diff
@@ -251,18 +250,18 @@ df.select { take(2) and col(3) }
 
 ```kotlin
 // first/last n columns in column set
-df.select { allDfs().take(3) }
-df.select { allDfs().takeLast(3) }
+df.select { allRec(includeGroups = false).take(3) }
```
koperagen, line 253: hm, i see various different options in this PR. Sometimes it's all().recursively(), sometimes all().rec(), and this. I saw that you did it intentionally, but why?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/replace.kt
koperagen: Is it correct that it's because of unification of overloads?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataColumn.kt
```diff
@@ -1,26 +1,8 @@
...
-import org.jetbrains.kotlinx.dataframe.columns.ColumnWithPath
-import org.jetbrains.kotlinx.dataframe.columns.FrameColumn
-import org.jetbrains.kotlinx.dataframe.columns.ValueColumn
-import org.jetbrains.kotlinx.dataframe.impl.columns.ColumnGroupImpl
-import org.jetbrains.kotlinx.dataframe.impl.columns.FrameColumnImpl
-import org.jetbrains.kotlinx.dataframe.impl.columns.ValueColumnImpl
-import org.jetbrains.kotlinx.dataframe.impl.columns.addPath
-import org.jetbrains.kotlinx.dataframe.impl.columns.guessColumnType
-import org.jetbrains.kotlinx.dataframe.impl.columns.toColumnKind
+import org.jetbrains.kotlinx.dataframe.api.*
```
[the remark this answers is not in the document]
Jolanrensen: No, we don't but this will be fixed with a linter update. Currently, the linter breaks for "unused imports" if we only use an import in KDocs. Adding this star import fixes that.

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
```diff
@@ -4490,7 +5094,7 @@ public fun <C> ColumnSet<*>.colsOf(type: KType, filter: (DataColumn<C>) -> Boole
  * @param [filter] an optional filter function that takes a column of type [C] and returns `true` if the column should be included.
  * @return A [ColumnSet][org.jetbrains.kotlinx.dataframe.columns.ColumnSet] containing the columns of given type that were included by [filter].
  */
-public inline fun <reified C> ColumnSet<*>.colsOf(noinline filter: (DataColumn<C>) -> Boolean = { true }): ColumnSet<C> =
+public inline fun <reified C> SingleColumn<*>.colsOf(noinline filter: (DataColumn<C>) -> Boolean = { true }): TransformableColumnSet<C> =
     colsOf(typeOf<C>(), filter)
```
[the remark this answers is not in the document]
Jolanrensen: No we can't, unfortunately, unless we get context receivers. All those extension functions need to be inside the ColumnsSelectionDsl interface

docs/StardustDocs/topics/ColumnSelectors.md
Jolanrensen: Yes, it does exactly the same, similar to allDfs and all().dfs. First I decided to use all().rec(ursively) everywhere, but indeed for completion I decided to have allRec(ursively) present too. I think we can just keep all options.

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
```diff
@@ -1700,7 +1642,11 @@ public interface ColumnsSelectionDsl<out T> : ColumnSelectionDsl<T>, SingleColum
 
         /**
          * ## Cols
-         * Creates a subset of columns ([ColumnSet][org.jetbrains.kotlinx.dataframe.columns.ColumnSet]) from a parent [ColumnSet][org.jetbrains.kotlinx.dataframe.columns.ColumnSet], -[ColumnGroup][org.jetbrains.kotlinx.dataframe.columns.ColumnGroup], or -[DataFrame][org.jetbrains.kotlinx.dataframe.DataFrame].
+         * Creates a subset of columns ([ColumnSet][org.jetbrains.kotlinx.dataframe.columns.ColumnSet]) from the current [ColumnSet][org.jetbrains.kotlinx.dataframe.columns.ColumnSet].
+         *
+         * If the current [ColumnSet][org.jetbrains.kotlinx.dataframe.columns.ColumnSet] is a [SingleColumn][org.jetbrains.kotlinx.dataframe.columns.SingleColumn] (and thus consists of only one column (or [column group][org.jetbrains.kotlinx.dataframe.columns.ColumnGroup])),
```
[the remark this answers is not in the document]
Jolanrensen: It's only too large because the doc processor enters the FQ name. I don't think it's a concern

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
```diff
@@ -4085,68 +4443,127 @@ public interface ColumnsSelectionDsl<out T> : ColumnSelectionDsl<T>, SingleColum
...
+    public fun <C> ColumnSet<C>.filter(predicate: ColumnFilter<C>): TransformableColumnSet<C> =
+        colsInternal(predicate as ColumnFilter<*>) as TransformableColumnSet<C>
+
+    public fun SingleColumn<*>.nameContains(text: CharSequence): TransformableColumnSet<*> =
+        cols { it.name.contains(text) }
+
+    public fun <C> ColumnSet<C>.nameContains(text: CharSequence): TransformableColumnSet<C> =
+        cols { it.name.contains(text) }
+
+    public fun SingleColumn<*>.nameContains(regex: Regex): TransformableColumnSet<*> = cols { it.name.contains(regex) }
```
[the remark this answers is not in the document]
Jolanrensen: Haven't reached that far down yet with kdocs/refactoring. Will fix everything in this file eventually. This pr just adds the recursively functionality

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/TypeConversions.kt
```diff
@@ -95,6 +95,14 @@ public fun <T> DataColumn<T>.castToNullable(): DataColumn<T?> = cast()
 
 public fun <T> ColumnReference<T>.castToNullable(): ColumnReference<T?> = cast()
 
+public fun AnyCol.setNullable(nullable: Boolean): AnyCol {
```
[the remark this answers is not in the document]
Jolanrensen: Sure, but that should be part of a different pr probably

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/Recursively.kt
```diff
@@ -0,0 +1,75 @@
...
+        this@recursivelyImpl.transformResolveSingle(
+            context = context,
+            transformer = RecursivelyTransformer(
+                includeGroups = includeGroups,
+                includeTopLevel = includeTopLevel,
+            ),
+        )
+}
+
+private class RecursivelyTransformer(
```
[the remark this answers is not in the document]
Jolanrensen: good idea

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/Utils.kt
```diff
@@ -146,25 +167,31 @@ internal fun <C> ColumnsContainer<*>.getColumn(path: ColumnPath, policy: Unresol
...
-        node?.dfs()?.forEach { it.data = null }
+        node?.allChildren()?.forEach { it.data = null }
         while (node != null) {
             node.data = null
             node = node.parent
         }
     }
-    val dfs = fullTree.topDfs { it.data != null }
-    return dfs.map { it.data!!.addPath(it.pathFromRoot()) }
+    val subtrees = fullTree.topmostChildren { it.data != null }
```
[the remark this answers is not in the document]
Jolanrensen: what would you suggest?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/Utils.kt
```diff
@@ -93,34 +78,70 @@ internal fun <T> DataColumn<T>.assertIsComparable(): DataColumn<T> {
     return this
 }
 
-internal fun <A, B> SingleColumn<A>.transformSingle(converter: (ColumnWithPath<A>) -> List<ColumnWithPath<B>>): ColumnSet<B> =
-    object : ColumnSet<B> {
-        override fun resolve(context: ColumnResolutionContext): List<ColumnWithPath<B>> =
-            this@transformSingle.resolveSingle(context)?.let { converter(it) } ?: emptyList()
-    }
+internal fun <A, B> SingleColumn<A>.transformSingle(
```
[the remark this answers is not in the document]
Jolanrensen: This was already in the project before my PR, so not sure if I should explain it here or maybe just add a quick doc above it?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/Utils.kt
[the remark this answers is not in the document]
Jolanrensen: As my doc explains it takes the tree and returns first the children, closest to the root of the tree, where the predicate holds. It doesn't matter whether the children of those nodes also hold, those are ignored. So since we now have `allChildren { something == true }`, I thought `topmostChildren { something == true }`. It can also be viewed as subtree-roots where the predicate holds.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/tree/Utils.kt
```diff
@@ -29,9 +29,22 @@ internal fun <T> TreeNode<T>.getOrPut(path: ColumnPath, createData: (ColumnPath)
     return node
 }
 
-internal fun <T> TreeNode<T>.topDfs(yieldCondition: (TreeNode<T>) -> Boolean): List<TreeNode<T>> = dfs(enterCondition = { !yieldCondition(it) }, yieldCondition = yieldCondition)
+/**
+ * Traverses all children in the tree in depth-first order and returns the top-most nodes that satisfy
+ * [yieldCondition]. This means that if a node satisfies [yieldCondition], its children are not traversed, regardless of
+ * whether they satisfy [yieldCondition] or not.
+ */
+internal fun <T> TreeNode<T>.topmostChildren(yieldCondition: (TreeNode<T>) -> Boolean): List<TreeNode<T>> =
```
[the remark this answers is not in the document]
Jolanrensen: that wouldn't reflect how the function works

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
```diff
@@ -74,8 +74,8 @@ internal fun tableJs(columns: List<ColumnDataForJs>, id: Int, rootId: Int, nrow:
     var index = 0
     val data = buildString {
         append("[")
-        fun dfs(col: ColumnDataForJs): Int {
-            val children = col.nested.map { dfs(it) }
+        fun depthFirstSearch(col: ColumnDataForJs): Int {
```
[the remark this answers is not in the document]
Jolanrensen: indeed, the name should reflect the goal, not how it is done

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnsSelectionDsl.kt
Jolanrensen: Yes, they should all be equal to each other. I think the same is clear. It cannot mean anything else, right?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/replace.kt
Jolanrensen: Yes, by default groups were included in dfs {}, but excluded in allDfs() if I recall correctly. I had to choose a sensible default. But if you use colsOf<Int>().rec(), then includeGroups irrelevant anyways, since Int is not a Group.

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/tree/TreeNode.kt
Jolanrensen: probably yes, but since it's used a lot around the code base and people (Anatoly) might be used to it, I thought it would be helpful to have a redirect.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/Utils.kt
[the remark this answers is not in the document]
Jolanrensen: adding docs

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/replace.kt
koperagen: Hmm, this is interesting and a bit confusing. I think this parameter only exists for allDfs. For normal dfs it would be `dfs { !it.isColumnGroup() }`. Now it's also present for `rec`, `recursively`.

Jolanrensen: `.rec()` can not only be called on `all()`, you can also call `first { }.rec(includeGroups = false)`, so I think keeping it in rec/recursively too would be best.

koperagen: I'm not sure first + includeGroups will work the way i'd expect. In other cases you can write `colsOf { it.name.startsWith("hi") && !it.isColumnGroup() }.rec()` for the same result. So, parameter doesn't bring anything new. But sometimes it won't make sense: `first { it.type == typeOf<Int> }.rec()`

koperagen:
Another thing that i noticed now is that it's hard to tell what rec does for `first { it.type == typeOf<Int> }.rec()`. Will it give the same result as `dfs { it.type == typeOf<Int>  }.first()`?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/replace.kt
Jolanrensen: `first { predicate }.rec(includeGroups = false)` returns a `SingleColumn` which will resolve to `null` if there is no non-group column adhering to the predicate, as you'd expect. Recursively modifies what goes into `first`, filtering out groups if needed.
But you're right, it doesn't always add something useful. Let me go over all the uses and see if I can remove it from `recursively` while keeping it for `allRec`

Jolanrensen:
> Another thing that i noticed now is that it's hard to tell what rec does for `first { it.type == typeOf<Int> }.rec()`. Will it give the same result as `dfs { it.type == typeOf<Int> }.first()`?

Yes, it works the same. You can read it like a sentence: "Select the first column of type Int recursively", which sort of makes sense.
Recursively modifies the call before such that it runs on all columns recursively instead of just on the top-level.

docs/StardustDocs/topics/addDf.md
```diff
@@ -10,7 +10,7 @@ Returns [`DataFrame`](DataFrame.md) with union of columns from several given [`D
 df.add(df1, df2)
 ```
 
-<dataFrame src="org.jetbrains.kotlinx.dataframe.samples.api.Modify.addDfs.html"/>
+<dataFrame src="org.jetbrains.kotlinx.dataframe.samples.api.Modify.addDataFrames.html"/>
```
koperagen, line 13: Hm, i don't see this file in the snippets. You forgot to commit it or it wasn't generated?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/replace.kt
koperagen: Maybe to come up with another name? I think the primary purpose of this function is shorthand to `dfs { !it.isColumnGroup() }` + for newcomers it might not be obvious that (or why) column groups should be included / excluded, so includeGroups = false by default is important, i think

docs/StardustDocs/topics/addDf.md
Jolanrensen: lol, we need an auto-commit for these too

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/replace.kt
Jolanrensen: I don't think we have a function that can give all valueColumns right? If we did we could have `valueCols {}.recursively()` or `valueColsRecursively()` but that might get a bit long.

Jolanrensen: Plus it's not just ColumnGroups that get included right? also FrameColumns and ValueColumns. So, just all of them. I think that's a good default

Jolanrensen: @koperagen Actually I do want to add `valueCols {}` and `frameCols {}` since we already have `groups {}` as well. Probably in another PR though. Then you can call `valueCols().rec()` if you want. Sounds okay?

koperagen: Could be a nice addition, not 100% related to this problem though. For this function something like "allLeafs" might fit better. I think it was named like that before and renamed later

Jolanrensen: We could also take it in another direction eventually, similar to `colsOf<Type>()` have `colsOfKind(vararg ColumnKind).rec()`, that might be clearer

PR #367, opened 2023-04-27
Fixed formatting

### LINKED ISSUE #315: Make method declaration for four APIs more consistent
- [x] Be sure that all methods are ordered by the same order (all, ColumnSelector, String, Reference, KProperties)
- [x] Be sure that there are no missed APIs
- [x] Reformat the code
- [x] Add spaces correctly between functions
- [x] add some docs if new method was added or removed

zaleslaw:
Fixes #315

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/tsv.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/tsv.kt
@@ -28,7 +28,7 @@ public class TSV : SupportedDataFrameFormat {
     }
 }
 
-private val tabChar = '\t'
+private const val tabChar = '\t'
 
 public fun DataFrame.Companion.readTSV(
     fileOrUrl: String,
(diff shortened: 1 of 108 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/tsv.kt
```diff
@@ -28,7 +28,7 @@ public class TSV : SupportedDataFrameFormat {
     }
 }
 
-private val tabChar = '\t'
+private const val tabChar = '\t'
 
 public fun DataFrame.Companion.readTSV(
```
Jolanrensen, line 33: While you're at it, could you check this file and probably csv.kt for named arguments? I see many usages of readDelim with 10+ params and no names, which will break stuff quickly upon refactoring or the addition of extra params in the future

zaleslaw: Looks like we need to fix and discuss the case with the named arguments separately.

Because in many cases it just doubles the code without additional sense.

Jolanrensen: I think it's always sensible, especially when there are multiple arguments with the same type involved.

PR #370, opened 2023-05-03
Korro outputs

koperagen:
Now running `korro` task will not only update code snippets, but also copy-paste content of the file with name corresponding to FQ name of the korro FUN directive. So, it means that we can append arbitrary content to the docs by simply writing it into the file during test execution. korro will keep docs in sync with it.


docs/StardustDocs/topics/ColumnSelectors.md
```diff
@@ -127,42 +94,39 @@ df.select { name.allDfs() }
...
-df.select { it[Person::name] }
-df.select { (Person::name)() }
-df.select { col(Person::name) }
+df.select { it["name"] }
 
 // by column path
-df.select { it[Person::name][Name::firstName] }
-df.select { Person::name[Name::firstName] }
+df.select { it["name"]["firstName"] }
+df.select { "name"["firstName"] }
```
zaleslaw, line 105: is it really works this way, call [] on String? I don't know about it

docs/StardustDocs/topics/ColumnSelectors.md
```diff
@@ -127,42 +94,39 @@ df.select { name.allDfs() }
...
-df.select { Person::name named "Full Name" }
+df.select { "name" named "Full Name" }
 
 // converted
-df.select { Person::name[Name::firstName].map { it.lowercase() } }
+df.select { "name"["firstName"]<String>().map { it.uppercase() } }
 
 // column arithmetics
-df.select { 2021 - (Person::age)() }
+df.select { 2021 - "age"<Int>() }
```
zaleslaw, line 114: honestly need to say, that Person::age was better than "age"<Int>

docs/StardustDocs/topics/ColumnSelectors.md
```diff
@@ -127,42 +94,39 @@ df.select { name.allDfs() }
...
+df.select { "name".."age" }
 
 // all children of ColumnGroup
-df.select { Person::name.all() }
+df.select { "name".all() }
 
 // depth-first-search traversal of all children columns
-df.select { Person::name.allDfs() }
+df.select { "name".allDfs() }
 ```
```
zaleslaw, line 127: looks like you and @Jolanrensen will have a clash here

docs/StardustDocs/topics/toHTML.md
```diff
@@ -38,6 +38,7 @@ val df3 = df.sortByDesc { age }
 listOf(df1, df2, df3).fold(DataFrameHtmlData.tableDefinitions()) { acc, df -> acc + df.toHTML() }
 ```
 
+<dataFrame src="org.jetbrains.kotlinx.dataframe.samples.api.Render.composeTables.html"/>
```
zaleslaw, line 41: Why do we need this line here?

docs/StardustDocs/topics/ColumnSelectors.md
koperagen: It is! It's an overloaded "get" operator.

docs/StardustDocs/topics/ColumnSelectors.md
koperagen: The PR is a bit unclear in this part. In fact, it's a very bad "diff". In this file after update of korro KProperties API tab was removed (we discussed it with Jolan). And "String API" tab was already there

docs/StardustDocs/topics/toHTML.md
koperagen: It's how we embed the table component into the documentation now

docs/StardustDocs/topics/ColumnSelectors.md
Jolanrensen: There has also been some shifting around, that's where the clashes come from

Jolanrensen:
Could you run `assemble` again to regenerate the docs before each commit?

koperagen:
> Could you run `assemble` again to regenerate the docs before each commit?

I updated the PR with updated Render.kt in generated sources. Is it what you asked for?

gradle.properties
```diff
@@ -2,6 +2,6 @@ projectName=dataframe
 version=0.11.0
 jupyterApiTCRepo=
 kotlin.jupyter.add.scanner=false
-org.gradle.jvmargs=-Xmx2G
+org.gradle.jvmargs=-Xmx4G
```
Jolanrensen, line 5: back to 4 hey?

Jolanrensen (approved):

gradle.properties
koperagen: Yes, korro needs it, otherwise fails with OOM

Jolanrensen: nice

PR #372, opened 2023-05-10
ColumnsSelection DSL continued

Jolanrensen:
WIP


zaleslaw:
It's interesting, could we split the model/plugin changes from the documentation changes?

Jolanrensen:
Looks like I can split up the Selection DSL file into separate files/interfaces by making the SelectionsDsl inherit other interfaces, which is probably a good idea, since the file has turned into 5k+ lines. I do need to take care of proper KDoc linking due to Kotlin's limitations for referring to extension functions.

Jolanrensen:
Splitting off is done! Now let's figure out how/if I can contribute it in parts (probably in separate PRs) to be reviewed and added to the master branch. Might be difficult since the change in structure affects every function, but hey, let's try :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/recursively.kt
```diff
@@ -0,0 +1,305 @@
...
+    @Deprecated(
+        message = "dfs is deprecated, use recursively instead.",
+        replaceWith = ReplaceWith("this.cols(predicate).recursively()"),
+        level = DeprecationLevel.WARNING,
+    )
+    public fun String.dfs(predicate: (ColumnWithPath<*>) -> Boolean): ColumnSet<*> =
+        columnGroup(this).dfs(predicate)
+
+    @Deprecated(
+        message = "dfs is deprecated, use recursively instead.",
```
zaleslaw, line 132: It gives me some idea: looks like we have no strict rules for removing deprecated stuff.
Could we somehow comment it with "it will be removed 0.13, for example).
Also, please find the file with String constants, containing deprecated messages and refactor all these String objects with the constants

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/frameCols.kt
```diff
@@ -0,0 +1,119 @@
...
+     * @arg [CommonFrameColsDocs.ExampleArg]
+     *
+     * `df.`[select][DataFrame.select]` { `[frameCols][SingleColumn.frameCols]`() }`
+     *
+     * `df.`[select][DataFrame.select]` { myColGroup.`[frameCols][SingleColumn.frameCols]` { it.`[name][ColumnReference.name]`.`[startsWith][String.startsWith]`("my") } }`
+     */
+    public fun SingleColumn<DataRow<*>>.frameCols(filter: Predicate<FrameColumn<*>> = { true }): TransformableColumnSet<DataFrame<*>> =
+        ensureIsColGroup().frameColumnsInternal(filter)
+
+    /** TODO */
```
zaleslaw, line 77: TODO what?))

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/filter.kt
```diff
@@ -36,3 +48,124 @@ public fun <T> DataFrame<T>.filterBy(column: ColumnReference<Boolean>): DataFram
 public fun <T> DataFrame<T>.filterBy(column: KProperty<Boolean>): DataFrame<T> = filterBy { column.toColumnAccessor() }
 
 // endregion
+
+internal fun <T, C> ColumnsSelector<T, C>.filter(predicate: (ColumnWithPath<C>) -> Boolean): ColumnsSelector<T, C> =
+    { this@filter(it, it).asColumnSet().filter(predicate) }
+
+// region ColumnsSelectionDsl
+
+// TODO remove confusing overloads
```
zaleslaw, line 57: convert to issue or fix

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/except.kt
```diff
@@ -0,0 +1,256 @@
...
+import org.jetbrains.kotlinx.dataframe.impl.columns.changePath
+import org.jetbrains.kotlinx.dataframe.impl.columns.createColumnSet
+import org.jetbrains.kotlinx.dataframe.impl.columns.transformSingle
+import kotlin.reflect.KProperty
+
+// region ColumnsSelectionDsl
+public interface ExceptColumnsSelectionDsl<out T> {
+    // region except
+
+    /** TODO tbd */
```
zaleslaw, line 24: convert to issue or fix

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/except.kt
```diff
@@ -0,0 +1,256 @@
...
+    @Suppress("UNCHECKED_CAST")
+    public fun <C> ColumnSet<C>.colsExcept(predicate: ColumnFilter<C>): TransformableColumnSet<C> =
+        colsInternal { !predicate(it as ColumnWithPath<C>) } as TransformableColumnSet<C>
+
+    /** TODO tbd */
+    public fun SingleColumn<DataRow<*>>.colsExcept(predicate: ColumnFilter<*>): TransformableColumnSet<*> =
+        colsInternal { !predicate(it) }
+
+    // TODO Same as select and cols but then inverted
```
zaleslaw, line 33: convert to issue or fix

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/colsOfKind.kt
```diff
@@ -0,0 +1,86 @@
...
+import org.jetbrains.kotlinx.dataframe.impl.headPlusArray
+import kotlin.reflect.KProperty
+
+// region ColumnsSelectionDsl
+
+// TODO tbd
+public interface ColsOfKindColumnsSelectionDsl {
+
+    /** TODO tbd */
+    public fun ColumnSet<*>.colsOfKind(
```
zaleslaw, line 20: is Kind and Child the same abstraction? I suggest descendant instead of Child/Kind/Parent terminology

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/colGroups.kt
```diff
@@ -0,0 +1,141 @@
...
+
+    @Deprecated("Use colGroups instead", ReplaceWith("this.colGroups(filter)"))
+    public fun ColumnSet<*>.groups(filter: Predicate<ColumnGroup<*>> = { true }): TransformableColumnSet<AnyRow> =
+        columnGroupsInternal(filter)
+
+    @Deprecated("Use colGroups instead", ReplaceWith("this.colGroups(filter)"))
+    public fun SingleColumn<DataRow<*>>.groups(filter: Predicate<ColumnGroup<*>> = { true }): TransformableColumnSet<AnyRow> =
+        ensureIsColGroup().columnGroupsInternal(filter)
+
+    @Deprecated("Use colGroups instead", ReplaceWith("this.colGroups(filter)"))
```
zaleslaw, line 28: Move Strings to the Strings constants

Jolanrensen:
> I suggest the next group of changes to improve readability for future generations:
> 
> * renaming Kind/Child/Parent to the more neutral stuff like ancestor/descendant
> * converting all TODOs to the issues or removing
> * split some huge files like cols to smaller if possible
> * extracting same String rows in the Deprecation messages to the separate file

Thanks for the review! I'll go over the individual comments.

- I think the name `child` is better in the DSL than `descendant` as it's much shorter and more readable, especially since some of the function names are already getting large, a.k.a. `myGroupCol.takeLastChildrenWhile {}`. Also, for tree structures, names like parent, child, etc. are very common. It could however still be renamed to something else, since we're dealing with groups and not with trees (from a user's perspective), so maybe `contents` or `entries`? It would then become `takeLastEntriesWhile {}`. wdyt?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/except.kt
Jolanrensen: outside the scope of the current review cycle. We'll get there

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/except.kt
Jolanrensen: not yet ready to be reviewed

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/filter.kt
Jolanrensen: not ready to be reviewed

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/recursively.kt
```diff
@@ -0,0 +1,305 @@
...
+import org.jetbrains.kotlinx.dataframe.impl.columns.recursivelyImpl
+import org.jetbrains.kotlinx.dataframe.impl.columns.transform
+import org.jetbrains.kotlinx.dataframe.impl.columns.tree.flattenRecursively
+import kotlin.reflect.KProperty
+import kotlin.reflect.KType
+import kotlin.reflect.typeOf
+
+// region ColumnsSelectionDsl
+// TODO make path modification optional
+// TODO rename to atAllDepths()?
```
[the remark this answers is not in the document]
Jolanrensen: not ready to be reviewed yet

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/colsOfKind.kt
Jolanrensen: Kind != Child, we're not talking Dutch or German, haha. Kind is the type of column, so either Frame, Value, or ColumnGroup.

Jolanrensen: Also, not part of the current review cycle yet

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Access.kt
```diff
@@ -897,12 +897,12 @@ class Access : TestBase() {
...
         // find the last column inside a column group satisfying the condition
         df.select {
-            colGroup("name").last { it.name().endsWith("Name") }
+            colGroup("name").lastChild { it.name().endsWith("Name") }
         }
 
         // find the single column inside a column group satisfying the condition
         df.select {
-            Person::name.single { it.name().startsWith("first") }
+            Person::name.singleChild { it.name().startsWith("first") }
```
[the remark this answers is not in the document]
Jolanrensen: I think `descendant` is too long for a DSL. But `singleCol` might be a good alternative?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/ConfiguredAggregateColumn.kt
```diff
@@ -1,27 +1,37 @@
...
+    private val newPath: ColumnPath? = null,
 ) : ColumnSet<C> {
 
-    private fun ColumnWithPath<C>.toDescriptor(keepName: Boolean) = when (val col = this) {
-        is AggregateColumnDescriptor<C> -> {
-            val path = if (keepName) newPath?.plus(col.newPath ?: col.column.shortPath()) ?: col.newPath
-            else newPath ?: col.newPath
-            AggregateColumnDescriptor(col.column, default ?: col.default, path)
+    private fun ColumnWithPath<C>.toDescriptor(keepName: Boolean): AggregateColumnDescriptor<C> =
+        when (val col = this) {
```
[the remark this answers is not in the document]
Jolanrensen: I think Tolya wrote it like this to be able to rename `this` to `col`, perform the check and keep the function as an expression. Potentially it could be extended if something else should be done for other inheritors of ColumnWithPath.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/columns/ColumnsResolver.kt
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Implemented by [SingleColumn] and [ColumnSet].
+ *
+ * @param C common type of resolved columns
+ * @see [SingleColumn]
+ * @see [ColumnSet]
+ * @see [TransformableColumnSet]
+ * @see [TransformableSingleColumn]
+ */
+public interface ColumnsResolver<out C> {
```
koperagen, line 22: If you intend for it to be implemented by two classes, might consider making it sealed interface

Jolanrensen: image
I would, but that would impact the flexibility of the type. The original `ColumnSet` also wasn't sealed. I think the `ColumnsResolver` interface is simple enough to be reused in other parts of the API eventually.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cols.kt
```diff
@@ -0,0 +1,967 @@
...
+    private interface ColumnPathColsRangeDocs
+
+    /** @include [ColumnPathColsRangeDocs] */
+    public fun ColumnPath.cols(range: IntRange): ColumnSet<*> = columnGroup(this).cols(range)
+
+    /**
+     * ## Columns by Index Range from List of Columns
+     * Helper function to create a [ColumnSet] from a list of columns by specifying a range of indices.
+     *
+     * {@comment TODO find out usages of this function and add examples}
```
Jolanrensen, line 930: @koperagen any idea what this was for? It's present but never used. The only place I see the `List<DataColumn<C>>` return type is inside the `DataFrame` interface... any ideas?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/col.kt
```diff
@@ -0,0 +1,585 @@
...
+import org.jetbrains.kotlinx.dataframe.impl.columns.getAt
+import org.jetbrains.kotlinx.dataframe.impl.columns.singleImpl
+import org.jetbrains.kotlinx.dataframe.impl.columns.transformSingle
+import org.jetbrains.kotlinx.dataframe.util.COL_REPLACE
+import org.jetbrains.kotlinx.dataframe.util.IDENTITY_FUNCTION
+import kotlin.reflect.KProperty
+
+// region ColumnsSelectionDsl
+
+public interface ColColumnsSelectionDsl {
```
zaleslaw, line 22: Funny name

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/col.kt
```diff
@@ -0,0 +1,585 @@
...
+     * @return A [ColumnAccessor] for the column with the given argument if possible, else a [SingleColumn].
+     * @throws [IllegalStateException\] if the column with the given argument does not exist.
+     *
+     * @see [column\]
+     * @see [ColumnsSelectionDsl.colGroup\]
+     * @see [ColumnsSelectionDsl.frameCol\]
+     * @see [ColumnsSelectionDsl.valueCol\]
+     * {@arg [CommonColDocs.Note]}
+     */
+    private interface CommonColDocs {
```
zaleslaw, line 63: Is it correct to say, that this is a util documentation interface. If it is, probably better to move outside this file

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCol.kt
```diff
@@ -0,0 +1,592 @@
...
+     * @throws [IllegalStateException\] if the column with the given argument does not exist.
+     * @throws [IllegalArgumentException\] if the column with the given argument is not a value column.
+     *
+     * @see [valueColumn\]
+     * @see [ColumnsSelectionDsl.colGroup\]
+     * @see [ColumnsSelectionDsl.frameCol\]
+     * @see [ColumnsSelectionDsl.col\]
+     * {@arg [CommonValueColDocs.Note]}
+     */
+    private interface CommonValueColDocs {
```
zaleslaw, line 66: again about this kind of interfaces

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCol.kt
```diff
@@ -0,0 +1,592 @@
...
+    public fun <C> valueCol(valueCol: ColumnAccessor<C>): ColumnAccessor<C> = valueCol.ensureIsValueColumn()
+
+    /**
+     * @include [ValueColReferenceDocs] {@arg [CommonValueColDocs.ReceiverArg] myColumnGroup.}
+     */
+    public fun <C> SingleColumn<DataRow<*>>.valueCol(valueCol: ColumnAccessor<C>): SingleColumn<C> =
+        ensureIsColGroup()
+            .transformSingle {
+                it.getChild(valueCol)
+                    ?.cast<C>()
```
zaleslaw, line 119: Sometimes I really worry these paths of ???? because each of them multiplies the paths, do you think could it be handled otherwise?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/frameCol.kt
```diff
@@ -0,0 +1,653 @@
...
+
+    /**
+     * @include [FrameColIndexDocs] {@arg [CommonFrameColDocs.ReceiverArg] myColumnGroup.}
+     * @include [CommonFrameColDocs.FrameColumnTypeParam]
+     */
+    public fun <C> SingleColumn<DataRow<*>>.frameCol(index: Int): SingleColumn<DataFrame<C>> =
+        ensureIsColGroup()
+            .allColumnsInternal()
+            .getAt(index)
+            .cast<DataFrame<C>>()
```
zaleslaw, line 565: I started to think, could this cast not be applied? throw an exception?

zaleslaw (commented):
Just some minor comments to structure of each file with mixing DSL and doc interfaces in one large file.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/col.kt
Jolanrensen: I don't think it's worth it to clutter the (internal) naming space with documentation that only applies to this file. New inheritance support could help a tiny bit with that, though.
Although, if it's in a separate file, updating even a single letter of documentation will create a git diff of 4 files (2 files in the project and 2 generated ones).

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/col.kt
Jolanrensen: just following my structure :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCol.kt
Jolanrensen: Not necessarily, in this specific case, only `getChild()` can return a nullable value for which the `IllegalStateException` will be thrown.
I continue the `?.` chain downwards for readability since something like 
```kotlin
(it.getChild(valueCol)?.cast<C>() ?: throw IllegalStateException("ValueColumn '${valueCol.path()}' not found in column group '${it.path}'"))
                    .also { it.data.ensureIsValueColumn() }
                    .let(::listOf)
```
becomes ugly and unreadable quickly. Other notation suggestions are welcome tho :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/frameCol.kt
Jolanrensen: The cast can never fail since it casts `SingleColumn<*>` to `SingleColumn<DataFrame<C>>`. This type is erased at runtime.

The call below "`.ensureIsFrameColumn()`" performs the check to make sure it is actually a frame column at runtime. It also provides a helpful exception to the user if it isn't a frame column.

In theory, these two functions could be merged if that makes it clearer

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/col.kt
Jolanrensen: plus, this documentation describes the functions inside this interface. Usually, kdoc is always right next to the function, so it should probably still be close to the function

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCol.kt
Jolanrensen: With AI help:
```kotlin
ensureIsColGroup().transformSingle {
    val child = it.getChild(valueCol)/*?.cast<IfNeeded>()*/
        ?: throw IllegalStateException("ValueColumn '${valueCol.path()}' not found in column group '${it.path}'")
    child.data.ensureIsValueColumn()
    listOf(child)
}.singleImpl()
```

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
```diff
@@ -66,6 +69,48 @@ public inline fun <T, reified C> column(
 
 // endregion
 
+// region valueColumn
+
+public fun valueColumn(): ColumnDelegate<Any?> = column()
```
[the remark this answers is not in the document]
Jolanrensen: `val columnA by valueColumn<String>()` is indeed the same as
`val columnB by column<String>()` 
but
`val columnC by valueColumn()` is different than
`val columnD by column()` 
in the sense that columnD can be either a value-, group-, or frame column, while  columnC is explicitly a ValueColumn.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
```diff
@@ -164,6 +209,97 @@ public fun <T> ColumnGroupReference.frameColumn(property: KProperty<List<T>>): C
...
+ * `df.`[select][DataFrame.select]` { "myColumnGroup".`[asColumnGroup][String.asColumnGroup]`<GroupType>().`[valueCols][ColumnsSelectionDsl.valueCols]`() }`
+ */
+private interface StringAsColumnGroupDocs
+
+/** @include [StringAsColumnGroupDocs] */
+@JvmName("asColumnGroupTyped")
+public fun <C> String.asColumnGroup(): ColumnAccessor<DataRow<C>> = columnGroup<C>(this)
+
+/** @include [StringAsColumnGroupDocs] */
+public fun String.asColumnGroup(): ColumnAccessor<DataRow<*>> = columnGroup<Any?>(this)
```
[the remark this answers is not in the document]
Jolanrensen: valid concern. But should we leave it out if we have the same for KProperty, ColumnPath etc.?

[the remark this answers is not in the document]
koperagen: In my opinion function named `asColumnGroup` should be available for `DataColumn` or maybe `BaseColumn`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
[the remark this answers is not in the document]
Jolanrensen: I also like the other syntax more, but I find it weird if it's missing when we explicitly state our columns can have 3 kinds: ValueColumn, FrameColumn, and ColumnGroup, but the constructors only exist for the latter two and a generic one.

I think it wouldn't hurt to keep it. @zaleslaw, @ermolenkodev  What do you think?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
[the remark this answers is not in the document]
Jolanrensen: it already is. So then remove all these asColumnGroup functions?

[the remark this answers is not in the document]
Jolanrensen: I'll keep just SingleColumn, I think that makes the most sense.

Jolanrensen:
Small preview of the DSL grammar in rendered KDoc :)

image

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCols.kt
```diff
@@ -0,0 +1,376 @@
...
+     *
+     *  [columnGroup][org.jetbrains.kotlinx.dataframe.documentation.UsageTemplateColumnsSelectionDsl.UsageTemplate.ColumnGroupDef]
+     *
+     *  &nbsp;&nbsp;&nbsp;&nbsp;.[**valueCols**][org.jetbrains.kotlinx.dataframe.api.ColumnsSelectionDsl.valueCols]` [` **`{ `**[condition][org.jetbrains.kotlinx.dataframe.documentation.UsageTemplateColumnsSelectionDsl.UsageTemplate.ConditionDef]**` }`** `]` `[ `.[**atAnyDepth**][org.jetbrains.kotlinx.dataframe.api.ColumnsSelectionDsl.atAnyDepth]`()` ` ]`
+     *
+     *
+     *
+     *
+     *
+     *
```
zaleslaw, line 83: Why do we have here so many empty rows?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/all.kt
```diff
@@ -0,0 +1,302 @@
...
+            df.select { cols().allUpTo(pathOf("age")) },
+        ).shouldAllBeEqual()
+    }
+
+    @Test
+    fun `all subset at lower level after and from`() {
+        listOf(
+            df.select { age },
+            df.select { colsOf<Int?>().allBefore(weight) }, // TODO
+            df.select { allBefore(weight).colsOf<Int?>() },
```
zaleslaw, line 150: Do we have allExcept(<columnName>)?

zaleslaw (approved):
I only have two large questions:
- could we forbid mixing APIs in range operation
- how to document it later in our site documentation? Give an example for each API's combination? Or we will go to the site documentation, operation page generation?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCols.kt
Jolanrensen: It's because they are accumulated from the template arguments that are split by newlines for readability. Might need to find a solution for that in the doc processor. They don't show up in KDoc renders in the IDE which is what I'm optimizing all KDocs for.

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/columnRange.kt
```diff
@@ -0,0 +1,66 @@
...
+            df.select { Person::age..weight },
+            df.select { Person::age..pathOf("weight") },
+
+            df.select { age.."weight" },
+            df.select { age..Person::weight },
+            df.select { age..weight },
+            df.select { age..pathOf("weight") },
+
+            df.select { pathOf("age").."weight" },
+            df.select { pathOf("age")..Person::weight },
```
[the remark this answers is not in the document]
Jolanrensen: It's not forbidden. The overloads just didn't exist before. I think it makes a lot of sense to be able to mix APIs, especially with how DF currently works without compiler plugin: Say you have a dataframe where you just added a new column:
```kotlin
df.add("newCol") { 0 }
  .select { existingCol.."newCol" }
```
there's no accessor for newCol yet, so you need to reference it in another way. Sure, a user can add `newCol` by a column accessor, however, if they choose the String API is good enough for their new column, they should not be forced to use a different API suddenly or worse, have to downgrade to `"existingCol".."newCol"` just because we didn't provide the overloads.

Jolanrensen:
> I only have two large questions:
> 
> * could we forbid mixing APIs in range operation
> * how to document it later in our site documentation? Give an example for each API's combination? Or we will go to the site documentation, operation page generation?

I think it's best to allow mixing APIs in this case. The same will happen for the `and` operator. More explanation in my response above.
For the documentation site, I'd provide a handful of examples and use DSL Grammar for the full picture, like here:
image

koperagen:
> Question: should we keep ColumnSet.all(Before/ etc)(column)?

I'd say yes, i can imagine someone trying to write
```
df.select { colsOf<Int>().allAfter { nameContains("city").single() } }
```
One can filter `allAfter` result like below, but i don't see why not keep previous variant as well
```
df.select { allAfter { nameContains("city").single() }.colsOf<Int>() }
```

Jolanrensen:
> > Question: should we keep ColumnSet.all(Before/ etc)(column)?
> 
> I'd say yes, i can imagine someone trying to write
> 
> ```
> df.select { colsOf<Int>().allAfter { nameContains("city").single() } }
> ```
> 
> One can filter `allAfter` result like below, but i don't see why not keep previous variant as well
> 
> ```
> df.select { allAfter { nameContains("city").single() }.colsOf<Int>() }
> ```

@koperagen 
One reason to not do this is because it opens up a ColumnSelector on a ColumnSet, which breaks the design of the DSL a bit. A ColumnSet is a simple collection of columns and can contain multiple columns with the same name (as it can be a collection of columns from any depth). Opening a ColumnSelectionDsl on this is only possible if the ColumnSet is converted into an anonymous column group (which is how it's currently implemented) which will throw an exception if the ColumnSet contains multiple columns of the same name.
Alternatively we could make the `ColumnSet.allUpTo {}` behave like `first {}` and just give a ColumnFilter. That would make
```kotlin
df.select { colsOf<Int>().allAfter { "city" in it.name } }
```
still possible, but
```kotlin
df.select { colsOf<Int>().allAfter { city } } // getting `city` accessor from outer scope
```
not anymore.

I'm asking this now because it also influences the design of `except {}` and `select {}`.

koperagen:
> > > Question: should we keep ColumnSet.all(Before/ etc)(column)?
> > 
> > 
> > I'd say yes, i can imagine someone trying to write
> > ```
> > df.select { colsOf<Int>().allAfter { nameContains("city").single() } }
> > ```
> > 
> > 
> > One can filter `allAfter` result like below, but i don't see why not keep previous variant as well
> > ```
> > df.select { allAfter { nameContains("city").single() }.colsOf<Int>() }
> > ```
> 
> @koperagen One reason to not do this is because it opens up a ColumnSelector on a ColumnSet, which breaks the design of the DSL a bit. A ColumnSet is a simple collection of columns and can contain multiple columns with the same name (as it can be a collection of columns from any depth). Opening a ColumnSelectionDsl on this is only possible if the ColumnSet is converted into an anonymous column group (which is how it's currently implemented) which will throw an exception if the ColumnSet contains multiple columns of the same name. Alternatively, we could make the `ColumnSet.allUpTo {}` behave like `first {}` and just give a ColumnFilter. That would make
> 
> ```kotlin
> df.select { colsOf<Int>().allAfter { "city" in it.name } }
> ```
> 
> still possible, but
> 
> ```kotlin
> df.select { colsOf<Int>().allAfter { city } } // getting `city` accessor from outer scope
> ```
> 
> not anymore. (Also, `ColumnSet.allUpTo(String)` might also be ambiguous, since a column with the given name might occur more than once in the set, uh list... ColumnSet (I hate that name))
> 
> I'm asking this now because it also influences the design of `except {}` and `select {}`.

Ok, from this perspective it's indeed not very convenient to provide API with DSL argument for ColumnSet receivers. Is it how API was before you started to work on it or is it some new design decision to be made?

Jolanrensen:
> Ok, from this perspective it's indeed not very convenient to provide API with DSL argument for ColumnSet receivers. Is it how API was before you started to work on it or is it some new design decision to be made?

It's new; previously all(After/Before/etc.) was only defined on SingleColumns and the Plain DSL. So we could a) remove all functions on ColumnSets for this, b) implement them without `{}`, c), implement them with `{ ColumnFilter }`, d) implement them with `{ ColumnSelectionDsl }`. I think b) or c) are our best options. What do you think?

`except`, on the other hand, does still use a selector but it "leaks" the outside scope inside the selector to make it work:
```kotlin
df.select { // this: ColumnsSelectionDsl<T> 
  all() // ColumnSet 
    .except { // this: ColumnsSelectionDsl<T>
        age 
    } 
}
```
But this has many issues, such as:
```kotlin
df.select { // this: ColumnsSelectionDsl<T> 
  myColGroup.all() // ColumnSet based on columns inside myColGroup
    .except { // this: ColumnsSelectionDsl<T>
        age // still operates on outside T scope, not on MyColGroup scope
    } 
}
```
Not sure how to handle those yet..

Jolanrensen:
@koperagen @zaleslaw Whenever you have time, the functions in **bold** can be reviewed (`ColsAtAnyDepth` up until `colsOf`). Thanks in advance!

core/ExposedTest.ipynb
```diff
@@ -0,0 +1,304 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 36,
+   "outputs": [],
+   "source": [
+    "USE { dependencies(\"org.xerial:sqlite-jdbc:3.30.1\") }"
+   ],
+   "metadata": {
```
zaleslaw, line 10: Could you please move such notebooks like this or spark to "research branch or as a separate PR" when you finish or probably to your own repository with dependency to our KDF? It will be quite interesting!

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/colsAtAnyDepth.kt
```diff
@@ -0,0 +1,184 @@
...
+        asSingleColumn()
+            .ensureIsColumnGroup()
+            .asColumnSet()
+            .dfsInternal { it.isSubtypeOf(typeOf<String?>()) }
+    }.sortedBy { it.name }
+
+    @Test
+    fun `first, last, and single`() {
+        listOf(
+            dfGroup.select { name.firstName.firstName },
```
zaleslaw, line 69: could we change a subfield from firstName.firstName to firstName.newFirstName at least?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/colsAtAnyDepth.kt
```diff
@@ -0,0 +1,184 @@
...
+
+    fun List<ColumnWithPath<*>>.print() {
+        forEach {
+            if (it.isValueColumn()) println("${it.name}: ${it.type()}")
+            else it.print()
+        }
+        println()
+    }
+
+    infix fun List<ColumnWithPath<*>>.shouldBe(other: List<ColumnWithPath<*>>) {
```
zaleslaw, line 34: Please add some minor comments to these wonderful DSL helper should* function or probably move them to helper file and collect all the helper testing stuff there, it will be easy to looking on that, create new helper testing functions.

Because I see distributed parts of KDF / SelectionDSL testing framework

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/drop.kt
```diff
@@ -0,0 +1,47 @@
...
+            df.select { pathOf("name").dropLastCols(1) },
+        ).shouldAllBeEqual()
+    }
+
+    @Test
+    fun `dropWhile and dropLastWhile`() {
+        listOf(
+            df.select { name.lastName },
+            df.select { name.dropColsWhile { it.name == "firstName" } },
+            df.select { "name".dropColsWhile { it.name == "firstName" } },
```
zaleslaw, line 34: I'm starting to think that "While" postfix for working with Predicates is not so good here and in another functions.
What do you think? Is dropLastColsFilter or dropLastColsIf would be better? Looking at `while` I'm starting to think about cycles (it has predicates too, I agree).

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
```diff
@@ -0,0 +1,39 @@
...
+            df.select { all() },
+            df.select { all().distinct() },
+            df.select { (all() and age and city).distinct() },
+        ).shouldAllBeEqual()
+
+        val cols = dfGroup.getColumnsWithPaths {
+            cols(
+                name.firstName,
+                name.firstName,
+                name.firstName.firstName,
```
zaleslaw, line 24: could we do the same tests but with the paths as Strings?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/none.kt
```diff
@@ -0,0 +1,17 @@
...
+import org.jetbrains.kotlinx.dataframe.ncol
+import org.jetbrains.kotlinx.dataframe.nrow
+import org.junit.Test
+
+class NoneTests : ColumnsSelectionDslTests() {
+
+    @Test
+    fun none() {
+        df.select { none() }.let {
+            it.nrow shouldBe 0
```
zaleslaw, line 13: Could we add a test for df.select { none() and firstName } to have 1 column selected to keep the algebra:)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/simplify.kt
```diff
@@ -0,0 +1,37 @@
...
+import org.jetbrains.kotlinx.dataframe.samples.api.secondName
+import org.jetbrains.kotlinx.dataframe.samples.api.lastName
+import org.jetbrains.kotlinx.dataframe.samples.api.name
+import org.junit.Test
+
+class SimplifyTests : ColumnsSelectionDslTests() {
+
+    @Test
+    fun simplify() {
+        df.select { cols(name.firstName, name.lastName, age).simplify() } shouldBe
```
zaleslaw, line 25: Honestly, only learning these tests I started to understand what this function does. Could we add more complex cases here? Especially with repeated subcolumns, for example?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/simplify.kt
```diff
@@ -0,0 +1,84 @@
...
+     * `df.`[select][DataFrame.select]` { `[colsAtAnyDepth][ColumnsSelectionDsl.colsAtAnyDepth]` { "e" `[in][String.contains]` it.`[name][DataColumn.name]` }.`[simplify][ColumnSet.simplify]`() }`
+     *
+     * @return A [ColumnSet][ColumnSet]`<`[C][C]`>` containing only the columns that are not inside any column group in [this].
+     */
+    @Suppress("UNCHECKED_CAST")
+    public fun <C> ColumnSet<C>.simplify(): ColumnSet<C> = simplifyInternal() as ColumnSet<C>
+
+    // region deprecated
+
+    @Deprecated(TOP_MESSAGE, ReplaceWith("simplify()"), DeprecationLevel.WARNING)
```
zaleslaw, line 66: Need to say, that roots and top lead better to understanding what is doing this function? And it does not simplify my life, because semantic closer to (cut hierarchy or leave only high-level groups and leaves).

Also, for me, it's unclear, how it should work with 3 and more levels of hierarchy? Or could I apply on different level of our ColumnTree?

zaleslaw: Probably you could cover it with more interesting tests with 3 and more levels of hierarchy or change the functionality (like you add ignoreCase flag earlier, but with simplify(level=2))

zaleslaw (commented):
Great tests and docs, mostly it's clear for me as a user, excluding a couple of functions

core/ExposedTest.ipynb
Jolanrensen: oops these should not have been added. But yeah that sounds like a good idea :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/colsAtAnyDepth.kt
Jolanrensen: agreed!

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/drop.kt
Jolanrensen: just following standard names: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/drop-while.html

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
Jolanrensen: sure!

Thanks, this actually uncovered an issue. I wasn't able to type:
```kotlin
cols(
    "name"["firstName"], // ColumnPath
    "name"["firstName"], // ColumnPath
    "name"["firstName"]["firstName"], // ColumnSet?? huh
).distinct()
```
because I missed a `ColumnPath[String]: ColumnPath` overload. It worked fine like 
`df.select { "name"["firstName"]["firstName"] }` cause `select` doesn't mind whether it gets a `ColumnSet` or a `SingleColumn`, but `vararg` functions like `cols()` do require every argument to be the same type :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/simplify.kt
Jolanrensen: You cannot call it at a lower level of the column tree. It can only be called on a `ColumnSet` (so a result of `colsOf {}`, `colsAtAnyDepth {}`, `(a and b)`, etc.).

I didn't like the name `top` because that was totally non-descriptive. `roots` was better, but with that name one might expect `(a.b.c and a.b).roots()` to return the actual "root", which would be `a`, but it actually returns `a.b`...
I'm definitely open for other suggestions tho! Simplify works, but I also think it's too broad of a name for what it does.

Jolanrensen: Adding a level does not really make sense to me in the context of this function. All this function does is remove columns that already appear in groups (in groups in groups...) in the column set. It's already a niche function that's used in the internals of the library a lot and just happens to be in the public api.
I cannot think of a use-case that adding a level would solve.

Jolanrensen:
Some small improvements that need to be made before merging this PR:
 - [ ]  Documentation could be a bit simpler in some places (like in `all()`, the user doesn't immediately need to know what a `ColumnsResolver` is.
 - [ ]  Usage needs a bit more explanation and better titles (also maybe be named `Grammar`?)
 - [ ] `ColumnFilter` could have better docs, especially for `colsAtAnyDepth {}`
 - [ ]  Selection dsl return type might be unclear, because just returning a kproperty doesn't work etc. Clarify in Selection DSL docs
 - [ ]  each individual CS dsl interface doesn't have docs yet
 - [ ]  `colsOf`: clarify that this includes subtypes
 - [ ] Clear up relativity with `ColumnSet.except`

core/build.gradle.kts
```diff
@@ -205,6 +205,7 @@ val processKDocsMain by creatingProcessDocTask(
         ARG_DOC_PROCESSOR,
         COMMENT_DOC_PROCESSOR,
         SAMPLE_DOC_PROCESSOR,
+//        REMOVE_ESCAPE_CHARS_PROCESSOR,
```
zaleslaw, line 208: ?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/colsAtAnyDepth.kt
```diff
@@ -28,33 +29,38 @@ import kotlin.reflect.KType
 
 // region ColumnsSelectionDsl
 
+/**
+ * ## Cols At Any Depth {@include [ColumnsSelectionDslLink]}
+ *
+ * See [Grammar] for all functions in this interface.
+ */
 public interface ColsAtAnyDepthColumnsSelectionDsl {
```
zaleslaw, line 37: I'd like this Java-style naming

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/colsAtAnyDepth.kt
```diff
@@ -69,11 +75,12 @@ public interface ColsAtAnyDepthColumnsSelectionDsl {
...
      * ## Cols At Any Depth
      *
-     * Returns the columns of this [ColumnsResolver] at any depth if they satisfy the optional given predicate.
+     * Returns all columns in [this\] at any depth (so also inside [Column Groups][ColumnGroup]) if they satisfy the
+     * optional given predicate.
      *
-     * This function is especially powerful if followed by another filter function such as
-     * [colsOf][ColumnsSelectionDsl.colsOf], [single][ColumnsSelectionDsl.single], and [valueCols][ColumnsSelectionDsl.valueCols].
-     * ### Check out: [Usage]
+     * This function is especially powerful if followed by another [ColumnSet] filter function like
```
zaleslaw, line 81: ` This function is especially powerful if followed by another ` - it sounds a little bit emotional for technical documentation

core/build.gradle.kts
Jolanrensen: will be in the next release of the doc preprocessor, I must not forget to enable it then :)

zaleslaw (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/update.kt
```diff
@@ -89,11 +92,9 @@ public data class Update<T, C>(
      * {@include [Indent]}
      * `| .`[**asFrame**][Update.asFrame]**` { `**[dataFrameExpression][ExpressionsGivenDataFrame.DataFrameExpression.WithExample]**` }`**
      *
-     * {@include [LineBreak]}
-     * {@include [DslGrammarLink]}
      * {@setArg [UpdateOperationArg] [**update**][update]}{@comment The default name of the `update` operation function name.}
      */
-    public interface Usage
+    public interface Grammar
```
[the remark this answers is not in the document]
Jolanrensen: hmm, I don't know, as you can also copy docs from other functions or classes, which are not solely "documentation interfaces" like these and those won't have a special name. Plus, this name is also what people will see when they read the KDocs:
image
Now it just says "public interface Grammar", which looks nice, plus I can also refer to it without alias, like `See [Grammar] for...`.

What we could do, maybe in the future, is to have two main interfaces somewhere: `PublicDocInterface`, which all documentation interfaces that are user-facing implement, and `InternalDocInterFace` or `TempDocInterface` which would indicate which doc interfaces would be removed after the docs are processed. That might clean up all intermediate doc interfaces neatly. (The doc processor cannot do this (yet) btw, just an idea).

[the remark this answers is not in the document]
Jolanrensen: Actually, that's where `@Annotations` would make more sense I think, like:

```kotlin
/**
 * Some Doc
 */
 @ExcludeFromSources
 internal interface Doc
```

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/last.kt
```diff
@@ -127,13 +129,14 @@ public interface LastColumnsSelectionDsl {
 
     /**
      * ## Last (Col)
-     * Returns the last column in this [ColumnSet] or [ColumnGroup]
-     * that adheres to the given [condition\].
+     * Returns the last column from [this\] that adheres to the optional given [condition\].
      * If no column adheres to the given [condition\], [NoSuchElementException] is thrown.
      *
+     * This function only looks at columns at the top-level.
```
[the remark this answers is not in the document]
Jolanrensen: replaced with "operates"

PR #377, opened 2023-05-11
prevent CI/CD from publishing artifacts where version = build number

koperagen:
this feature was used to publish versions such as "0.8.0-RC-1", etc. For now it will be impossible, but we can re-introduce it later.
We discussed and decided to always use version from the repository as a prefix to any build number. It should make it less prone to mistakes (i.e. impossible to publish version "123")


Jolanrensen:
Could we at least still support RC? It's a good safety check before releasing a full version

PR #378, opened 2023-05-16
Adds the parent name in the flatten operation

### LINKED ISSUE #229: Incorrect column names when using computing aggregates
Consider the following example
```
 val aggregate = irisData.groupBy("Species")
          .aggregate {
              mean() into "mean"
              std() into "sd"
          }

aggregate.flatten().print()
```

The result is 
```
      Species Sepal.Length Sepal.Width Petal.Length Petal.Width Sepal.Length1 Sepal.Width1 Petal.Length1 Petal.Width1
 0     setosa     5.006000    3.428000     1.462000    0.246000      0.352490     0.379064      0.173664     0.105386
 1 versicolor     5.936000    2.770000     4.260000    1.326000      0.516171     0.313798      0.469911     0.197753
 2  virginica     6.588000    2.974000     5.552000    2.026000      0.635880     0.322497      0.551895     0.274650

      Species Sepal.Length Sepal.Width Petal.Length Petal.Width
 0     setosa     5.006000    3.428000     1.462000    0.246000
 1 versicolor     5.936000    2.770000     4.260000    1.326000
 2  virginica     6.588000    2.974000     5.552000    2.026000

```

Observed: column names are not indicating the type of aggregate

Expected: Sepal.Length.mean, Sepal.Length.sd, and so on on

zaleslaw:
Fixes #229

### DIFF
--- a/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/flatten.kt
+++ b/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/flatten.kt
@@ -13,7 +13,8 @@ import org.jetbrains.kotlinx.dataframe.impl.ColumnNameGenerator
 import org.jetbrains.kotlinx.dataframe.impl.columns.toColumnSet
 
 internal fun <T, C> DataFrame<T>.flattenImpl(
-    columns: ColumnsSelector<T, C>
+    columns: ColumnsSelector<T, C>,
+    keepParentNameForColumns: Boolean = false
 ): DataFrame<T> {
     val rootColumns = getColumnsWithPaths { columns.toColumnSet().filter { it.isColumnGroup() }.top() }
     val rootPrefixes = rootColumns.map { it.path }.toSet()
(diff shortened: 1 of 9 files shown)


Jolanrensen:
Don't forget to regenerate docs, any build/assemble call will do so.

Maybe it's a good idea to make this behavior optional? I can see many cases where flattening should result in all child columns with their actual name, not including the parent name.

zaleslaw:
Could you please share the case, I will add it to the tests and could make ot optional

Jolanrensen:
Just take any Dataframe containing some column groups where the children (value- or frame-columns) names are unique and flatten it. This should produce a new data frame with all value- and frame-columns regardless of whether they were nested in a column group or not

zaleslaw:
In SQL, it's done by default, you know all this stuff: "MAX (name)" and so on.
Ok, I will add at Friday some examples and make it optional

Jolanrensen:
@zaleslaw "keepParentNameForColumns" is a bit long isn't it? Also, "forColumns" is a bit redundant IMO. I'd consider something like "renameByPath".

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/flatten.kt
```diff
@@ -29,7 +30,8 @@ internal fun <T, C> DataFrame<T>.flattenImpl(
         .into {
             val targetPath = getRootPrefix(it.path).dropLast(1)
             val nameGen = nameGenerators[targetPath]!!
-            val name = nameGen.addUnique(it.name())
+            val preferredName = if (keepParentNameForColumns) "${it.name()}.${it.parentName}" else it.name()
```
Jolanrensen, line 33: Looks like this is the wrong way around, should be parentName.name instead.

PR #385, opened 2023-05-28
Added RELEASE_CHECK_LIST.md

### LINKED ISSUE #246: Add Release Check List
To be sure that anybody from the maintainers could release a new version of the library and reduce the release bus-factor

The Release Check list should contain the full list of all required steps before and during the release, except the private credentials to the inner resources.

zaleslaw:
Fixes #246

### DIFF
--- a/RELEASE_CHECK_LIST.md
+++ b/RELEASE_CHECK_LIST.md
@@ -1,0 +1,25 @@
+**Release activities check-list for releases:**
+
+0. Run code inspections (fix typos, Kotlin issues, fix code formatting, linter)
+1. Write missed KDocs for new APIs
+2. Remove methods deprecated in the previous release 
+3. Update tutorials according last code changes
+4. Update README.MD according last code changes
+   - update artifact version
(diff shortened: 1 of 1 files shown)


koperagen:
Documentation is being published from the GitHub actions. And along with checking Maven Central, we need to check Gradle Plugin portal availability. It mirrors dataframe dependency used in the plugin, it usually takes around 12 hours.
Also, the library version needs to be changed in the https://github.com/Kotlin/dataframe/blame/master/docs/StardustDocs/v.list and maybe https://github.com/Kotlin/dataframe/blob/master/.github/workflows/main.yml for "stable" releases.

koperagen (changes requested):
Check my previous comment for requested changes. Maybe some of these actions can be done by a script. Quite a big list we have here

koperagen:
Another thing is updating a version in the jupyter descriptor, renovate bot sends PRs here https://github.com/Kotlin/kotlin-jupyter-libraries/blob/master/dataframe.json

Issue #383, opened 2023-05-28
Library doesn't follow official naming conventions


dunca:
The ["Choose good names" section](https://kotlinlang.org/docs/coding-conventions.html#choose-good-names) of the Kotlin Coding Conventions documentation, mentions that acronyms longer than two characters shouldn't be capitalized.
By capitalizing CSV in class, function, etc. names, the library doesn't follow standard Kotlin naming conventions.

Actual examples from the codebase (list is not exhaustive):
_org.jetbrains.kotlinx.dataframe.io.CSV
org.jetbrains.kotlinx.dataframe.io.CSVType
org.jetbrains.kotlinx.dataframe.io.DataFrame.Companion.readCSV
org.jetbrains.kotlinx.dataframe.io.DataFrame.Companion.writeCSV_

Expected:
_org.jetbrains.kotlinx.dataframe.io.Csv
org.jetbrains.kotlinx.dataframe.io.CsvType
org.jetbrains.kotlinx.dataframe.io.DataFrame.Companion.readCsv
org.jetbrains.kotlinx.dataframe.io.DataFrame.Companion.writeCsv_

zaleslaw:
@dunca Yes, sometimes naming violates a little bit naming conventions for acronyms, but from the other side we could have our own approach to name IO sources; personally, I prefer to use CSV instead of Csv and HTML instead of Html, I think it looks prettier and clearer

dunca:
@zaleslaw, is there even a point in having language-official naming conventions, if they're not followed by official libraries?

zaleslaw:
Probably, you are right, especially regarding what is pretty or not. Of course , it's subjective, as anming conventions is agreed average of the leading subjectives.

Thank you for lighting the topic

zaleslaw:
Also, @dunca if you tasted the library, do you have something to share yet, it will be great to read, regarding interfaces/docs, how is easy to start. 

It will be great to read, you could it via issues or discussion thread if it's huge

Jolanrensen:
I agree, it's important for the library to be written as expected. Renamings like these will probably take a lot of time, however. 
We're looking into a good linter/formatter that can also detect stuff like this, however, with the size of the library we get flooded with hundreds of issues when we enable such a thing. It's probably best to go one file at a time and gradually reformat and rename stuff.

Jolanrensen:
It's been a while since I've come across an incorrectly camel cased name. Feel free to reopen this issue if we missed one :)

Issue #302, opened 2023-03-15
`renameToCamelCase` for columns with names starting with capital letter should lowercase it


belovrv:

zaleslaw:
@belovrv what kind of CamelCase do you want?
- CamelCase?
- camelCase? (style lowerCamelCase)

For now, we have no documentation for this function, please give some examples of usage or share some insights

zaleslaw:
@Jolanrensen how do you estimate this code?

For me, it looks like a strange 2-3 level repeating function s with recursive call.

If it's a common pattern to apply changes in the dataframe, could you advise how to interpret correctly the code?

```

public fun <T> DataFrame<T>.renameToCamelCase(): DataFrame<T> {
    return rename {
        dfs { it.isColumnGroup() && it.name() matches DELIMITED_STRING_REGEX }
    }.toCamelCase()
        .rename {
            dfs { !it.isColumnGroup() && it.name() matches DELIMITED_STRING_REGEX }
        }.toCamelCase()
        .update {
            dfsOf<AnyFrame>()
        }.with { it.renameToCamelCase() }
}
```

Jolanrensen:
@zaleslaw 
It's strange actually, the first two statements seem to be useless but I cannot merge them into 1.
 
It says:
First, recursively rename all column groups to camel case, 
then recursively rename all other columns to camel case, 
then finally, take all frame columns and call `renameToCamelCase()` on all dataframes inside.

It might look clearer after my rename:
```kotlin
public fun <T> DataFrame<T>.renameToCamelCase(): DataFrame<T> =
    rename {
        cols { it.isColumnGroup() && it.name() matches DELIMITED_STRING_REGEX }.recursively()
    }.toCamelCase()

    .rename {
        cols { !it.isColumnGroup() && it.name() matches DELIMITED_STRING_REGEX }.recursively()
    }.toCamelCase()

    .update {
        colsOf<AnyFrame>().recursively()
    }.with { it.renameToCamelCase() }
```

PR #386, opened 2023-05-28
Added some name repairing strategies

### LINKED ISSUE #342: Disambiguate duplicated column names when reading data
E.g. when using `DataFrame.readExcel()`, it fails to read sheets with duplicated column names. It's bad practice to have this type of duplication in data, but that's how data ends up on one's desk quite frequently.

kdf should follow the approach implemented in other tabular-data-APIs, to disambiguate (or repair) duplications, e.g by correcting duplicatedcolumn name
* foo (first appearance)
* foo_1 (second appearance)
* foo_2 (third appearance)

Such functionality is also referred to as name-repair strategy, e.g. see https://readr.tidyverse.org/reference/read_delim.html (name_repair)

The function should be applied/provided to/by all `DataFrame.read*` methods for API consistency.

Optionally, the user could be given more control over the repair strategy.

zaleslaw:
Fixed #342


zaleslaw:
Need to say, that this is a solution for one IO only to solve user needs, but it's a good idea to make it generic and move later to API module to be available for other IO, because it's not a guarantee that other IO can handle properly the empty names, for example

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/NameRepairStrategy.kt
```diff
@@ -0,0 +1,16 @@
+package org.jetbrains.kotlinx.dataframe.io
+
+/**
+ * This strategy defines how the repeatable name column will be handled
+ * during the creation new dataframe from the IO sources.
+ */
+public enum class NameRepairStrategy {
```
Jolanrensen, line 7: It's unclear that it targets duplicate names just by looking at the enum class name.  Maybe call it `DuplicateNameStrategy`?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/NameRepairStrategy.kt
```diff
@@ -0,0 +1,16 @@
...
+    /** No actions, keep as is. */
+    NO,
```
Jolanrensen, line 9: `DO_NOTHING`? Also, I wonder when this option might be used, since DataFrames cannot contain duplicate column names.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/NameRepairStrategy.kt
```diff
@@ -0,0 +1,16 @@
...
+
+    /** Check the uniqueness of the column names without any actions. */
```
Jolanrensen, line 11: I assume this would then throw an exception?

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -60,96 +61,114 @@ private const val readExcel = "readExcel"
...
+ * @param nameRepairStrategy handling of column names.
+ * The default behavior is to ensure column names are [NameRepairStrategy.CHECK_UNIQUE]
  */
 public fun DataFrame.Companion.readExcel(
     url: URL,
     sheetName: String? = null,
     skipRows: Int = 0,
     columns: String? = null,
     rowsCount: Int? = null,
+    nameRepairStrategy: NameRepairStrategy = NameRepairStrategy.CHECK_UNIQUE,
```
Jolanrensen, line 73: again probably `duplicateNameStrategy`

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -60,96 +61,114 @@ private const val readExcel = "readExcel"
...
+    return wb.use { readExcel(wb, sheetName, skipRows, columns, rowsCount, nameRepairStrategy) }
 }
 
 /**
  * @param sheetName sheet to read. By default, first sheet in the document
  * @param columns comma separated list of Excel column letters and column ranges (e.g. “A:E” or “A,C,E:F”)
  * @param skipRows number of rows before header
  * @param rowsCount number of rows to read.
+ * @param nameRepairStrategy handling of column names.
+ * The default behavior is to ensure column names are [NameRepairStrategy.CHECK_UNIQUE]
```
Jolanrensen, line 85: column names are what?

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -195,6 +218,29 @@ public fun DataFrame.Companion.readExcel(
     return dataFrameOf(columns)
 }
 
+/**
+ * This is a universal function for name repairing
```
Jolanrensen, line 222: looks like a good place for a TODO :)

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -195,6 +218,29 @@ public fun DataFrame.Companion.readExcel(
...
+ * and should be moved to the API module later,
+ * when the functionality will be enabled for all IO sources.
+ */
+private fun repairNameIfRequired(nameFromCell: String, columnNameCounters: MutableMap<String, Int>, nameRepairStrategy: NameRepairStrategy): String {
+    return when(nameRepairStrategy) {
```
Jolanrensen, line 227: linting

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -195,6 +218,29 @@ public fun DataFrame.Companion.readExcel(
...
+        NameRepairStrategy.NO -> nameFromCell
+        NameRepairStrategy.CHECK_UNIQUE -> if (columnNameCounters.contains(nameFromCell)) throw DuplicateColumnNamesException(columnNameCounters.keys.toList()) else nameFromCell
+        NameRepairStrategy.MAKE_UNIQUE -> if (nameFromCell.isEmpty()) { // probably it's never empty because of filling empty column names earlier
+            val emptyName = "Unknown column"
+            if(columnNameCounters.contains(emptyName)) "${emptyName}_${columnNameCounters[emptyName]}"
```
Jolanrensen, line 232: linting

Jolanrensen: Also not sure whether we should use a `_`. In Kotlin they are only used for ENUM_NAMES and constants, whereas column names will be used as accessors. I think we can drop the `_` and just put the number after it. wdyt?

Jolanrensen (changes requested):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/NameRepairStrategy.kt
zaleslaw: In many dataframe frameworks it's named as NameRepairing, better to keep naming to reduce learning curve

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
zaleslaw: I thought about it, it became uglier and badly readable, but let it be.
From other side, column names are not a part of Kotlin, but became part when we generate schemas

zaleslaw:
Thanks for the review, good points @Jolanrensen

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/NameRepairStrategy.kt
Jolanrensen: I think "repairing" insinuates a name is "broken", e.g. unsupported characters etc. This is definitely different IMO. Can you give an example from another library?

zaleslaw: https://readr.tidyverse.org/reference/read_delim.html#arguments
go to the name_repair

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
Jolanrensen: Yeah, but we mainly push in the column accessor / DataSchema API, due to its type safety etc., so shouldn't we mainly focus on that? Especially also with the [camelCase issue](https://github.com/Kotlin/dataframe/issues/302) we had lately, and with  [naming convention concerns](https://github.com/Kotlin/dataframe/issues/383). I think we should play it safe and not become looking like python ;)

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
Jolanrensen: There's no space after `when`, same with the `if`s a bit below. No clue why the linter is not picking those up.

zaleslaw: ok, will check twice, thank you

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/NameRepairStrategy.kt
Jolanrensen: I see! Well, if in the future we also add "universal", which fixes name syntax too, I'm fine with the name. Then it won't be just about uniqueness

PR #400, opened 2023-06-12
Split the Gradle and Kotlin build snippets for Android and Server-side

### LINKED ISSUE #374: Improve Gradle snippets for Server-side and Android developers in Readme.md
- [x] split Server and Android parts
- [x] check that all stuff works for both sides
- [x] add support for Gradle and Gradle.kts
- [x] push the changes to the Documentation

zaleslaw:
Fixed #374


zaleslaw:
Hi, I've built projects with build.gradle for server side and Android; snippets should work correctly

README.md
```diff
@@ -5,12 +5,12 @@
 [![Maven Central](https://img.shields.io/maven-central/v/org.jetbrains.kotlinx/dataframe?color=blue&label=Maven%20Central)](https://search.maven.org/artifact/org.jetbrains.kotlinx/dataframe)
 [![GitHub License](https://img.shields.io/badge/license-Apache%20License%202.0-blue.svg?style=flat)](http://www.apache.org/licenses/LICENSE-2.0)
 
-Kotlin Dataframe aims to reconcile Kotlin static typing with dynamic nature of data by utilizing both the full power of Kotlin language and opportunities provided by intermittent code execution in Jupyter notebooks and REPL.   
+Kotlin Dataframe aims to reconcile Kotlin static typing with the dynamic nature of data by utilizing both the full power of Kotlin language and opportunities provided by intermittent code execution in Jupyter notebooks and REPL.
```
Jolanrensen, line 8: ...reconcile Kotlin's static typing...power of the Kotlin language and the opportunities...

README.md
```diff
@@ -23,23 +23,94 @@ Explore [**documentation**](https://kotlin.github.io/dataframe/overview.html) fo
...
 ## Setup
 
-### Gradle
-```kotlin
+### Gradle for Server-side
+```groovy
+// build.gradle
+
 plugins {
     // Optional Gradle plugin for enhanced type safety and schema generation
```
Jolanrensen, line 31: This hint should also be in the other two examples

README.md
```diff
@@ -23,23 +23,94 @@ Explore [**documentation**](https://kotlin.github.io/dataframe/overview.html) fo
 
 ## Setup
 
-### Gradle
-```kotlin
+### Gradle for Server-side
```
Jolanrensen, line 26: I would call it "Gradle for JVM". It's not like DataFrame is just for servers. Many people use it on their desktop.

README.md
```diff
@@ -70,10 +141,11 @@ android {
         }
     }
 }
-tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> { 
+tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
```
Jolanrensen, line 144: this part is missing from the groovy example.
It broke some KSP KotlinCompile task if I left it out (the last time I tried it). Maybe it's not needed anymore, but then it should be removed here.

README.md
```diff
@@ -97,7 +169,7 @@ or specific version:
 
 ## Kotlin, Kotlin Jupyter, OpenAPI, Arrow and JDK versions
 
-This table shows the mapping between main library components versions and minimum supported Java versions.
+This table shows the mapping between main library component versions and minimum supported Java versions.
 
 | Kotlin DataFrame Version | Minimum Java Version | Kotlin Version | Kotlin Jupyter Version | OpenAPI version | Apache Arrow version |
```
Jolanrensen, line 174: make sure to rebase on the master branch, so these versions are correct again :)

README.md
```diff
@@ -114,8 +186,11 @@ val flightNumber by columnOf(10045.0, Double.NaN, 10065.0, Double.NaN, 10085.0)
 val recentDelays by columnOf("23,47", null, "24, 43, 87", "13", "67, 32")
 val airline by columnOf("KLM(!)", "{Air France} (12)", "(British Airways. )", "12. Air France", "'Swiss Air'")
 
-// create dataframe
+// create the dataframe
```
Jolanrensen, line 189: I'd say "a" here. But it's a comment, which are better when they're compact, so maybe it could be left as is.

README.md
```diff
@@ -114,8 +186,11 @@ val flightNumber by columnOf(10045.0, Double.NaN, 10065.0, Double.NaN, 10085.0)
...
 val df = dataFrameOf(fromTo, flightNumber, recentDelays, airline)
+
+// print the dataframe
```
Jolanrensen, line 192: Here "the" is correct :) since you're referring to something that's already there. In the previous line you created something new

README.md
```diff
@@ -155,7 +230,7 @@ val clean = df
 clean
     // group by the flight origin renamed into "from"
     .groupBy { origin named "from" }.aggregate {
-        // we are in the context of single data group
+        // we are in the context of a single data group
```
Jolanrensen, line 233: good :)

Jolanrensen (changes requested):

Jolanrensen:
I think the examples didn't change from https://github.com/Jolanrensen/DataframeTestApplication right? Otherwise, we should also update that.

README.md
zaleslaw: Could it be for complex examples with Schema description? I have tested on the example from the readme with dataframe creation without Schema description.

zaleslaw: Looks like I need a double check here for all possible variants (4 cases), but I need more complex example.
What example did you use to get the broken KSP task?

Jolanrensen: I just used the project at my branch linked below. It used the DataFrame plugin and so KSP to generate @DataSchema accessors. It might be that it broke because the Gradle JVM version was not 8 on my machine and then I got a weird error related to KSP. Adding these lines fixed that. If you cannot reproduce it, it might be fine.

zaleslaw: Ok, will try with your project, need to double check

Jolanrensen (approved):

Jolanrensen:
@zaleslaw not sure why TC is complaining, but it might be better for us to create branches inside this project, not your own fork, since it tries to authenticate for your fork currently

README.md
zaleslaw: It's very interesting that this task is really required on Kotlin DSL, but could be ignored (with different JDKs) on the Gradle.  I tested a lot of different combinations. Let's wait something from users here.

Issue #284, opened 2023-02-27
Jupyter compile-time DF type not recognized


Jolanrensen:
> @ileasile:  Noticed interesting thing: if the property has compile-time type DataFrame  then it's typed and property accessors are generated. But if it has compile-time type Any, it is not typed in any way and remains Any. Most likely could be fixed on the integration side
![image_720](https://user-images.githubusercontent.com/17594275/221631408-bda29f38-c26d-479e-a99e-6ac403f57497.png)
it's about onVariable {}  and updateVariable{}  callbacks - or FieldHandlers in general
We pass compile-time type (KType) to org.jetbrains.kotlinx.jupyter.api.FieldHandler#acceptsType. But we could also provide another handler with runtime type (KClass) if you need. Feel free to ask

PR #401, opened 2023-06-12
Fix: Jupyter compile-time DF type not recognized

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/284.

Basically, an `updateVariable` case was added for the `Any` type, which then checks (in the correct "reversed" order) the instances of the types to update the variable (including a new cast-operation).

Tests appear to be working fine, as well as actual notebooks, but feel free to try it for yourself.


zaleslaw (approved):
It looks that better to have some String constants separately with giving some explanation about types and so on

Jolanrensen:
> It looks that better to have some String constants separately with giving some explanation about types and so on

Agreed, I thought of using KotlinPoet or something to generate the name from the class name, but I couldn't yet figure out how to generate a `<*>` after `ColumnGroup`. Will try again

koperagen (approved):

Issue #253, opened 2023-02-07
`withValue(value)` consistency


Jolanrensen:
Some functions in the API have the option `withValue` such as `update { something }.withValue(value)` while other functions, such as `convert {}` say they have it in the docs, yet don't.

We need to either add `withValue` everywhere a `with` occurs or just remove `withValue` from the entire API since `with { value }` does the same thing.

Jolanrensen:
Let's remove `withValue` from update (neat deprecation with ReplaceWith ofc)
and remove it everywhere in the docs

koperagen:
Hm, i have concerns about deprecations of this method because we can break examples in notebooks (especially in datalore) and there is at least one screenshot of some code in docs

Jolanrensen:
Well, we can change those right? Good to know but it doesn't seem like a showstopper :)

Jolanrensen:
Aah wait, `Update.withValue` does have some use, namely keeping the type the same :) `with {}` allows the type to change

Jolanrensen:
Btw `Update.asNullable` is fairly useless. All it does is cast `Update<T, C>` -> `Update<T, C?>`.
This is then only useful for `Update.withValue()` because it expects an exact type match. But we can just as easily allow withValue to change nullability or just make users use `with {}` since that allows all type/nullability changes.

koperagen:
> Well, we can change those right? Good to know but it doesn't seem like a showstopper :)

Sure, just something to keep in mind 

Regarding nullability, if `with` indeed allows nullability change, then `asNullable` looks like a leftover. Maybe it wasn't always possible. I vaguely remember there were design discussions about it, but i don't remember the final decision. 
Also, apart from `withValue` we also have `withZero`, `withNull`.

Jolanrensen:
Yes, I'm indifferent to keep or remove `withZero` and `withNull`. Could be a handy addition since it's less typing than `with { 0.0 }` especially if autocompleted. I see more use in those than `withValue(value)`

nikitinas:
Originally `withValue` was added to support possible performance optimizations, for example when user fills a long column with a single constant value (#291 )

zaleslaw:
@Jolanrensen please revisit this

PR #403, opened 2023-06-13
withValue() -> with { value }

Jolanrensen:
Discussed in https://github.com/Kotlin/dataframe/issues/253#issuecomment-1587226245

Since we have no plans atm to implement single-value columns, I think it might be best to remove `withValue()` from the `Update` function (and the docs of `Convert`).

This PR replaces usage of it across the library and examples and provides a helpful deprecation annotation.

### DIFF
--- a/examples/notebooks/titanic/Titanic.ipynb
+++ b/examples/notebooks/titanic/Titanic.ipynb
@@ -1111,7 +1111,7 @@
     "val df1 = df\n",
     "    // imputing\n",
     "    .fillNulls { sibsp and parch and age and fare }.perCol { mean() }\n",
-    "    .fillNulls { sex }.withValue(\"female\")\n",
+    "    .fillNulls { sex }.with { \"female\" }\n",
     "    .fillNulls { embarked }.with { \"S\" }\n",
     "    .convert { sibsp and parch and age and fare }.toDouble()\n",
     "\n",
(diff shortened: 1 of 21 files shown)


zaleslaw (approved):

examples/notebooks/titanic/Titanic.ipynb
```diff
@@ -1111,7 +1111,7 @@
     "val df1 = df\n",
     "    // imputing\n",
     "    .fillNulls { sibsp and parch and age and fare }.perCol { mean() }\n",
-    "    .fillNulls { sex }.withValue(\"female\")\n",
+    "    .fillNulls { sex }.with { \"female\" }\n",
```
koperagen, line 1114: Is it updated on datalore?

koperagen (approved):

examples/notebooks/titanic/Titanic.ipynb
Jolanrensen: no, datalore runs the release version of DF, updating it now would break it

koperagen: Hm, probably not in this case, `with` is available in 0.10.1

Jolanrensen: oh of course XD, you're right

Jolanrensen: done!

Issue #109, opened 2022-05-13
Documentation for `NA` term


koperagen:
This term is used in valueCounts and statistical operations: std, mean, varianceAndMean
https://kotlin.github.io/dataframe/valuecounts.html
In documentation, there is no explanation of what it means. What's the difference from NaN?
In the following function, skipNA used when the value is NaN.
```
public fun Sequence<Float>.mean(skipNA: Boolean = skipNA_default): Double {
    var count = 0
    var sum: Double = 0.toDouble()
    for (element in this) {
        if (element.isNaN()) {
            if (skipNA) continue
            else return Double.NaN
        }
        sum += element
        count++
    }
    return if (count > 0) sum / count else Double.NaN
}
```

Jolanrensen:
I mentioned it in the KDocs in https://github.com/Kotlin/dataframe/pull/214, but indeed, we need it on the documentation website too.

PR #411, opened 2023-06-20
NA and NaN docs

Jolanrensen:
https://github.com/Kotlin/dataframe/issues/109

I added docs and references to it from other files + kdocs.

I don't know where to put this file in the tree, however.... It's not an operation, it's closest to a data abstraction maybe, but not entirely. I made it a separate tab for now.


koperagen:
You want to add only interfaces with kdocs for now, right? And reference it from related operations later. Just making sure. Otherwise looks good.

Jolanrensen:
> You want to add only interfaces with kdocs for now, right? And reference it from related operations later. Just making sure. Otherwise looks good.

The NA and NaN interfaces are already referenced to from fillNA, dropNA etc. in the code :) People can click on the reference to read more about the concepts. I just updated them with a URL.

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/DocumentationUrls.kt
```diff
@@ -19,6 +19,17 @@ internal interface DocumentationUrls {
         interface RowConditions
     }
 
+    /** [See `NaN` and `NA` on the documentation website.](https://kotlin.github.io/dataframe/nanAndNa.html) */
```
zaleslaw, line 22: Looks like a new pattern to refer on the Documentation site. Is it correct that we did not do it earlier.
It seems good idea, but probably let's discuss it widely.

Because it looks like adding yet one detail to the hierarchy of documents produced by your plugin

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/DocumentationUrls.kt
```diff
@@ -19,6 +19,17 @@ internal interface DocumentationUrls {
...
+    interface NanAndNa {
+
+        /** [See `NaN` on the documentation website.](https://kotlin.github.io/dataframe/nanAndNa.html#nan) */
+        interface NaN
+
+        /** [See `NA` on the documentation website.](https://kotlin.github.io/dataframe/nanAndNa.html#na) */
+
```
zaleslaw, line 29: remove blank line

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/DocumentationUrls.kt
```diff
@@ -19,6 +19,17 @@ internal interface DocumentationUrls {
...
     }
 
+    /** [{@include [Text]}{@arg [NameArg] `NaN` and `NA`}]({@include [Url]}/nanAndNa.html) */
+    interface NanAndNa {
+
+        /** [{@include [Text]}{@arg [NameArg] `NaN`}]({@include [Url]}/nanAndNa.html#nan) */
+        interface NaN
+
+        /** [{@include [Text]}{@arg [NameArg] `NA`}]({@include [Url]}/nanAndNa.html#na) */
+
```
zaleslaw, line 29: blank line

docs/StardustDocs/d.tree
```diff
@@ -35,6 +35,7 @@
         <toc-element topic="DataColumn.md"/>
         <toc-element topic="DataRow.md"/>
     </toc-element>
+    <toc-element topic="nanAndNa.md"/>
```
zaleslaw, line 38: probably better to move under operations (for first time)

docs/StardustDocs/topics/nanAndNa.md
```diff
@@ -0,0 +1,25 @@
...
+
+## NaN
+
+`Float` or `Double` values can be represented as `NaN`,
+in cases where a mathematical operation is undefined, such as for dividing by zero. The
+result of such an operation can only be described as "**N**ot **a** **N**umber".
+
+This is different from `null`, which means that a value is missing and, in Kotlin, can only occur
+for `Float?` and `Double?` types.
+
```
zaleslaw, line 14: blank line?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/DocumentationUrls.kt
Jolanrensen: No, it was used before in Update.kt and Nulls.kt, you can see the other URLs in the file ;)

docs/StardustDocs/d.tree
Jolanrensen: But it's not an operation. I'm not comfortable having it there, seems weird. Probably we can find a better place after restructuring the docs/flattening them.

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/DocumentationUrls.kt
zaleslaw: In this way, it looks that we should include somehow the reusing the info in our documentation writing pipeline.
What I mean: imagine, we have a new function or uncovered function, it's great to have an instruction or an example: where you should go and put your sentences. Here, it looks like the main Documentation could be a final judge for our doubts.

Issue #405, opened 2023-06-14
Cannot rename `ColumnGroup` and its children at the same time


Jolanrensen:
As the title says, renaming of columns inside column groups cannot be done if the column group itself is renamed in the same call.

For instance:
```kotlin
val myDf = dataFrameOf("a", "b", "c")(1, 2, 3).group("a", "b", "c").into("letters")
//⌌---------------------⌍
//|  |           letters|
//|--|------------------|
//| 0| { a:1, b:2, c:3 }|
//⌎---------------------⌏

myDf.rename {
    all().recursively()
    // or using the old notation
    allDfs(includeGroups = true)
}.into { it.name + "_renamed" }

//⌌---------------------⌍
//|  |   letters_renamed|
//|--|------------------|
//| 0| { a:1, b:2, c:3 }|
//⌎---------------------⌏
```

It needs to be done in two separate operations, like:
```kotlin
myDf
    .rename {
        cols { !it.isColumnGroup() }.recursively()
    }.into { it.name + "_renamed" }
    .rename {
        cols { it.isColumnGroup() }.recursively()
    }.into { it.name + "_renamed" }

//⌌--------------------------------------------⌍
//|  |                          letters_renamed|
//|--|-----------------------------------------|
//| 0| { a_renamed:1, b_renamed:2, c_renamed...|
//⌎--------------------------------------------⌏
```

From a user-perspective, this makes little sense and we should probably find a way to fix this.

Jolanrensen:
This actually also breaks `renameToCamelCase()` which breaks running a test like:
```kotlin
val doublyNestedColumnGroup = dataFrameOf("test_name")(
    dataFrameOf("another_name")(
        dataFrameOf("third_name")(1).first()
    ).first()
)

val df = doublyNestedColumnGroup.renameToCamelCase()
df.columnNames() shouldBe listOf("testName")
df["testName"].asColumnGroup().columnNames() shouldBe listOf("anotherName") // still named "another_name"
df["testName"]["anotherName"].asColumnGroup().columnNames() shouldBe listOf("thirdName")
```

So an all-encompassing solution is needed.

Jolanrensen:
The problem runs deeper... It also applies to 
```kotlin
df.move { group and group.child }.into { ... }
```
and
```kotlin
df.replace { group and group.child }.with { ... }
```
If the name changes of a group, its children won't be updated since their path now differs from their new path.

PR #419, opened 2023-06-28
Fixed rename behavior

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/405

Column groups and their children can now be renamed in the same call using a rewritten method.
`renameToCamelCase` was also updated with this new method (no longer does it need 2 or 3 runs over the dataframe) and I actually moved the the fix for https://github.com/Kotlin/dataframe/issues/302 to `toCamelCase()`, since it saves a run over the DF and makes the calls more consistent.

### DIFF
--- a/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/rename.kt
+++ b/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/rename.kt
@@ -2,14 +2,16 @@ package org.jetbrains.kotlinx.dataframe.impl.api
 
 import org.jetbrains.kotlinx.dataframe.DataFrame
 import org.jetbrains.kotlinx.dataframe.api.RenameClause
+import org.jetbrains.kotlinx.dataframe.api.asColumnGroup
 import org.jetbrains.kotlinx.dataframe.api.cast
 import org.jetbrains.kotlinx.dataframe.api.getColumnsWithPaths
-import org.jetbrains.kotlinx.dataframe.api.insert
-import org.jetbrains.kotlinx.dataframe.api.under
+import org.jetbrains.kotlinx.dataframe.api.toDataFrame
+import org.jetbrains.kotlinx.dataframe.columns.ColumnGroup
 import org.jetbrains.kotlinx.dataframe.columns.ColumnKind
 import org.jetbrains.kotlinx.dataframe.columns.ColumnWithPath
 import org.jetbrains.kotlinx.dataframe.impl.columns.tree.allChildrenNotNull
 import org.jetbrains.kotlinx.dataframe.impl.columns.tree.collectTree
+import org.jetbrains.kotlinx.dataframe.impl.columns.tree.map
 import org.jetbrains.kotlinx.dataframe.kind
 
 internal fun <T, C> RenameClause<T, C>.renameImpl(newNames: Array<out String>): DataFrame<T> {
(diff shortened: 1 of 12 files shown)


core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/rename.kt
```diff
@@ -0,0 +1,46 @@
...
+}
+
+internal fun <T, C> RenameClause<T, C>.renameImpl(transform: (ColumnWithPath<C>) -> String): DataFrame<T> {
+    val selectedColumns = df.getColumnsWithPaths(columns)
+    val tree = df.getColumnsWithPaths { all().rec() }.collectTree()
+
+    // perform rename in nodes
+    tree.allChildrenNotNull().forEach { node ->
+        val column = selectedColumns.find { it.data == node.data } ?: return@forEach
+        val newName = transform(column)
```
zaleslaw, line 27: some unclear naming here: is it transform column to name?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/rename.kt
```diff
@@ -0,0 +1,46 @@
...
+        node.name = newName
+    }
+
+    // build up a new DataFrame using the modified names
+    var newDf = DataFrame.empty(df.rowsCount()).cast<T>()
```
zaleslaw, line 32: Probably dfWithModifiedNames is better naming

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/rename.kt
```diff
@@ -36,6 +101,20 @@ class RenameTests {
...
 
+    @Test
+    fun `doubly nested row`() {
+        val doublyNestedColumnGroup = dataFrameOf("test_name")(
+            dataFrameOf("another_name")(
+                dataFrameOf("third_name")(1).first()
+            ).first()
+        )
+
+        val df = doublyNestedColumnGroup.renameToCamelCase()//.alsoDebug()
```
zaleslaw, line 112: remove comment?

zaleslaw (approved):
I left minor comments for generated docs instead of the real sources

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/rename.kt
```diff
@@ -0,0 +1,46 @@
...
+    return renameImpl { newNames[i++] }
+}
+
+internal fun <T, C> RenameClause<T, C>.renameImpl(transform: (ColumnWithPath<C>) -> String): DataFrame<T> {
+    val selectedColumns = df.getColumnsWithPaths(columns)
+    val tree = df.getColumnsWithPaths { all().rec() }.collectTree()
+
+    // perform rename in nodes
+    tree.allChildrenNotNull().forEach { node ->
+        val column = selectedColumns.find { it.data == node.data } ?: return@forEach
```
koperagen, line 26: What's data here? From collectTree definition it seems its a DataColumn, and in this case equals will iterate over elements of the column. Quite a heavy operation

koperagen: is it possible that there's two identical columns and second one will never be renamed?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/rename.kt
```diff
@@ -0,0 +1,46 @@
...
+        val newName = transform(column)
+        node.name = newName
+    }
+
+    // build up a new DataFrame using the modified names
+    var newDf = DataFrame.empty(df.rowsCount()).cast<T>()
+    tree.allChildrenNotNull().forEach { node ->
```
koperagen, line 33: This tree structure is used in other places to create dataframes? Is there an existing code to re-create a df from it? Approach with insert under looks suspicious because it's a "mutation" of an immutable data structure in a loop

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/rename.kt
Jolanrensen: `data` is a column with no path anymore (`collectTree()` removes it and ColumnWithPath.data yields the original column). So to get that path back I need to find it in `selectedColumns` (I'll change this to a Map actually). I would like to use `node.pathFromRoot()` but unfortunately, the tree might already have had some node renamed, so this would generate wrong results.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/rename.kt
Jolanrensen: There doesn't appear to be such a thing. Usually dataframes are built from the ground up with columns recursively. I'll try to create a sort-of mapping function for TreeNode with which I can create a new dataframe with columns I create on the fly

PR #427, opened 2023-07-20
fix compilation for markers with empty bodies in Jupyter

koperagen:


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
```diff
@@ -483,7 +476,7 @@ internal class CodeGeneratorImpl(typeRendering: TypeRenderingStrategy = FullyQua
                 append("\n}")
             }
         } else {
-            ""
+            " { }"
```
zaleslaw, line 479: is here an empy symbol before brackets?

zaleslaw (approved):
Nice PR, it was tiny and helped me to understand some things during the process useful for another task!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
koperagen: yes, it's intentional so that generated code is `interface Name { }`

Issue #428, opened 2023-07-22
Column `hasNulls` property is actually isNullable


Kopilov:
Hello

If I get some dataframe with some column and I want to check if it contains null values, I use `hasNulls` property.
```
val nullable = DataColumn.create("nullable", listOf("a" as String?, "b" as String?, "c" as String?))
println(nullable.hasNulls)
```
expected: false
actual: true

From another side, column can be marked as not nullable but contain null values (this behavior might be a critical issue itself)
```
    val notNullable = DataColumn.create(
        "notNullable",
        listOf("a" as String?, "b" as String?, "c" as String?, null as String?),
        typeOf<String>().withNullability(false)
    )
    println(notNullable.hasNulls)
```
expected: true or IllegalArgumentException
actual: false

koperagen:
Hi! It's indeed seems confusing, but at the same time expected.
`DataColumn` tries to be an efficient constructor, that's why it relies on supplied reified type parameter by default and doesn't check anything. It has `infer` parameter to match nullability / type to data though. 
Maybe it'll make sense to change its default behaviour to `Infer.Nulls` and maybe introduce another function with explicit naming, like `createWithoutTypeInference`

Kopilov:
`DataColumn.create` behavior can be justified but what about `hasNulls`? The issue is about `hasNulls` primarily, without regard for data origin.

In real life, I want to save nullable (but without null values) column as not nullable in Arrow and get unexpected NullValueError

Jolanrensen:
I agree about your change for `hasNulls()`, it should only return `true` if there are actually nulls in the column. But, do we also have a `canHaveNulls()`? Because that sounds like useful behavior.

koperagen:
I still see it as a `DataColumn.create` problem, maybe i miss something.
Its `DataColumn.create` responsibility to match actual values with type and nullability, and `hasNulls` only looks at type supplied to constructor. So, wouldn't `DataColumn.create("col", listOf("a" as String?, "b" as String?, "c" as String?), Infer.Throw)` solve this problem?

Jolanrensen:
[...]

PR #429, opened 2023-07-22
Expected `hasNulls` behavior for #428 

Kopilov:
Expected `hasNulls` behavior for #428

### DIFF
--- a/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataColumn.kt
+++ b/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataColumn.kt
@@ -97,7 +97,7 @@ public interface DataColumn<out T> : BaseColumn<T> {
         public fun empty(name: String = ""): AnyCol = createValueColumn(name, emptyList<Unit>(), typeOf<Unit>())
     }
 
-    public fun hasNulls(): Boolean = type().isMarkedNullable
+    public fun hasNulls(): Boolean = values().any { it == null }
 
     override fun distinct(): DataColumn<T>
 
--- a/core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/contains.kt
+++ b/core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/contains.kt
@@ -1,7 +1,11 @@
 package org.jetbrains.kotlinx.dataframe.api
 
 import io.kotest.matchers.shouldBe
+import org.jetbrains.kotlinx.dataframe.DataColumn
+import org.jetbrains.kotlinx.dataframe.hasNulls
 import org.junit.Test
+import kotlin.reflect.full.withNullability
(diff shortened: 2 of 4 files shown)


Jolanrensen:
Since we're not sure about the performance impact this would have and based on the philosophy of DF I propose the following solution:

1. Deprecate the original `hasNulls` getter in favor of a new `isNullable` getter.
2. In the release afterwards introduce the `hasNulls` implementation of this PR

That way, parts of the library that need performance can use `isNullable` and `hasNulls` can be used for a more thorough check. @koperagen wdyt?

koperagen:
What i have in mind: in most places where `DataColumn` is created inside the library, `hasNulls` is already known and used as a nullability flag for column type. For example,
https://github.com/Kotlin/dataframe/blob/9713fa265588d7decce47308cb8d275ca0ba72bb/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/join.kt#L175
https://github.com/Kotlin/dataframe/blob/d281ab9c542b4cb4beb3cc2b2a0108cc4bdd1831/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/csv.kt#L294
https://github.com/Kotlin/dataframe/blob/704200cb86e7bdc07b800a7cfef48de408bd5fe8/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnImpl.kt#L29
I want to use this information when available. Now this information is stored as a nullability flag in the type.

It works everywhere except `arrow`, there's `NullabilityOptions.Widening` that makes column without null values (hasNulls = false) to have nullable type (isNullable = true). Here https://github.com/Kotlin/dataframe/issues/78 i said it's needed for generating schema from samples, but maybe there are others. @Kopilov interesting what you think about it) 

Assuming it's the only such place, i suggest to see if it's feasible to change something in the arrow module without hurting existing use cases. If not, then maybe make `hasNulls` an optional constructor parameter for `DataColumn` and pass this parameter whenever possible

Kopilov:
> It works everywhere except `arrow`, there's `NullabilityOptions.Widening` that makes column without null values (hasNulls = false) to have nullable type (isNullable = true). 

Yes but `NullabilityOptions.Infer` (Use only actual data, set `DataColumn.hasNulls` to true if and only if there are null values in the column)  is still default behavior. Using NullabilityOptions.Widening is just as available as using explicit type in #428 expected/actual example.

> Assuming it's the only such place, i suggest to see if it's feasible to change something in the arrow module without hurting existing use cases. 

In arrow reading I do not see any problems (if you do, please, open the issue with unexpected behavior example). In arrow writing we can use custom thorough check instead of [hasNulls](https://github.com/Kotlin/dataframe/blob/de6e9c2ad2732cabd25385d7d8517af6197a7e85/dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/ArrowWriterImpl.kt#L158C56-L158C56) to make saving nullable without nulls as not nullable to work correctly.

PR #430, opened 2023-07-24
Deprecation messages

Jolanrensen:
As a tip from @zaleslaw in https://github.com/Kotlin/dataframe/pull/372, we currently don't have any central indication as to when deprecated functions will be removed and whatnot. We did already have a `deprecationMessages.kt` file, containing some const messages. 
I updated the file with instructions and added versions to the remove-messages.

I also went through the rest of the core module as well to get the deprecations in the same central place. The columns selection DSL will also deprecate a lot of functions, so it's good to have this present ahead of that. (I won't include that in this PR as that's too large)

### DIFF
--- a/RELEASE_CHECK_LIST.md
+++ b/RELEASE_CHECK_LIST.md
@@ -29,3 +29,8 @@
 17. Create Release from the release tag on GitHub 
 18. Update a KDF version in the [Kotlin Jupyter Descriptor](https://github.com/Kotlin/kotlin-jupyter-libraries/blob/master/dataframe.json). Now the Renovate bot doing this 
 19. Update DataFrame version in gradle.properties file for next release cycle (i.e. 0.10.0 -> 0.11.0)
+20. Update deprecated functions in [deprecationMessages.kt](/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt)
+    such that 
+    - `Level.WARNING` messages are changed to `Level.ERROR`
+    - `Level.ERROR` messages and their functions are removed.
+    - Update regions in the file accordingly.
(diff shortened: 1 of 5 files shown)


RELEASE_CHECK_LIST.md
```diff
@@ -29,3 +29,8 @@
 17. Create Release from the release tag on GitHub 
 18. Update a KDF version in the [Kotlin Jupyter Descriptor](https://github.com/Kotlin/kotlin-jupyter-libraries/blob/master/dataframe.json). Now the Renovate bot doing this 
 19. Update DataFrame version in gradle.properties file for next release cycle (i.e. 0.10.0 -> 0.11.0)
+20. Update deprecated functions in [deprecationMessages.kt](/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt)
+    such that 
+    - `Level.WARNING` messages are changed to `Level.ERROR`
```
zaleslaw, line 34: Great, love this!

zaleslaw: Looks like a post-release section with 19 paragraph

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt
```diff
@@ -1,13 +1,35 @@
...
-internal const val ITERABLE_COLUMNS_DEPRECATION_MESSAGE = "Replaced with `toColumnSet()` operation."
+internal const val ITERABLE_COLUMNS_DEPRECATION_MESSAGE = "Replaced with `toColumnSet()` operation. Removed in 0.11.0."
 
-internal const val DIFF_DEPRECATION_MESSAGE = "Replaced to explicitly indicate nullable return value; added a new non-null overload."
+// endregion
 
-internal const val DIFF_REPLACE_MESSAGE = "this.diffOrNull(expression)"
+// region WARNING in 0.11.0, ERROR in 0.12.0
 
+internal const val DIFF_DEPRECATION_MESSAGE = "Replaced to explicitly indicate nullable return value; added a new non-null overload. Will be removed in 0.12.0."
```
zaleslaw, line 21: We could go deeper and refactor the version to the constant

zaleslaw (approved):
Great refactoring and process improvement

PR #431, opened 2023-07-26
Temporary documentation sitemap solution

Jolanrensen:
While Writerside cannot [yet](https://youtrack.jetbrains.com/issue/WRS-2357/Generate-a-sitemap-for-external-users) produce a sitemap.xml (for indexability of the documentation website in search consoles), we can add such a file manually. This PR adds a temporary static sitemap file to the documentation website.


koperagen:
Looks good. Can you interactively rebase your branch to squash 3 commits with the same name together? You can also use amend when commiting, makes history a bit easier to read imo. 
Also, we'll have to update sitemap manually for now, until writerside comes up with their solution?

Jolanrensen:
> Looks good. Can you interactively rebase your branch to squash 3 commits with the same name together? You can also use amend when commiting, makes history a bit easier to read imo. Also, we'll have to update sitemap manually for now, until writerside comes up with their solution?

yep that's right. But unless we add or remove pages we won't have to change it.

PR #434, opened 2023-08-01
Predicate join operation 

koperagen:
Main purpose: joining data on operations other than equals


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -10,7 +10,9 @@ public fun <A, B> DataFrame<A>.join(
     other: DataFrame<B>,
     type: JoinType = JoinType.Inner,
     selector: JoinColumnsSelector<A, B>? = null
-): DataFrame<A> = joinImpl(other, type, true, selector)
+): DataFrame<A> {
+    return joinImpl(other, type, addNewColumns = type.addNewColumns, selector)
```
Jolanrensen, line 14: probably provide name for each argument in the function, it looks odd like this :). Also, I'm fine with it being an expression function if it's just a one line thing

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -116,10 +118,20 @@ public enum class JoinType {
     Left, // all data from left data frame, nulls for mismatches in right data frame
     Right, // all data from right data frame, nulls for mismatches in left data frame
     Inner, // only matched data from right and left data frame
+    Filter, // only matched data from left data frame
     Full, // all data from left and from right data frame, nulls for any mismatches
     Exclude // mismatched rows from left data frame
 }
 
+internal val JoinType.addNewColumns: Boolean
+    get() {
```
Jolanrensen, line 127: also here is an expression body probably cleaner, especially since you use the same name as the getter :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/predicateJoin.kt
```diff
@@ -0,0 +1,50 @@
...
+}
+
+public typealias JoinExpression<A, B> = Selector<JoinedDataRow<A, B>, Boolean>
+
+public fun <A, B> DataFrame<A>.predicateJoin(
+    right: DataFrame<B>,
+    type: JoinType = JoinType.Inner,
+    joinExpression: JoinExpression<A, B>
+): DataFrame<A> {
+    return predicateJoinImpl(right, type, addNewColumns = type.addNewColumns, joinExpression)
```
Jolanrensen, line 19: same as in join, I'd make it an expression function and provide named arguments for all

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/predicateJoin.kt
```diff
@@ -0,0 +1,50 @@
...
+
+public fun <A, B> DataFrame<A>.fullPredicateJoin(
+    right: DataFrame<B>,
+    joinExpression: JoinExpression<A, B>
+): DataFrame<A> = predicateJoin(right, JoinType.Full, joinExpression)
+
+public fun <A, B> DataFrame<A>.filterPredicateJoin(
+    right: DataFrame<B>,
+    joinExpression: JoinExpression<A, B>
+): DataFrame<A> = predicateJoinImpl(right, JoinType.Inner, addNewColumns = false, joinExpression)
```
Jolanrensen, line 45: argument names :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/predicateJoin.kt
```diff
@@ -0,0 +1,50 @@
...
+public interface JoinedDataRow<out A, out B> : DataRow<A> {
+    public val right: DataRow<B>
+}
+
+public typealias JoinExpression<A, B> = Selector<JoinedDataRow<A, B>, Boolean>
+
+public fun <A, B> DataFrame<A>.predicateJoin(
+    right: DataFrame<B>,
+    type: JoinType = JoinType.Inner,
+    joinExpression: JoinExpression<A, B>
```
Jolanrensen, line 17: trailing commas everywhere please :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/predicateJoin.kt
```diff
@@ -0,0 +1,96 @@
...
+internal class JoinedDataRowImpl<A, B>(
+    leftOwner: DataFrame<A>,
+    val index: Int,
+    rightOwner: DataFrame<B>,
+    index1: Int
+) : JoinedDataRow<A, B>, DataRowImpl<A>(index, leftOwner) {
+    override val right: DataRow<B> = DataRowImpl(index1, rightOwner)
+}
+
+internal fun <A, B> DataFrame<A>.predicateJoinImpl(
```
Jolanrensen, line 27: comments would be nice XD, bit hard to check it like this

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/testSets/person/PredicateJoinTests.kt
```diff
@@ -0,0 +1,129 @@
...
+import org.jetbrains.kotlinx.dataframe.api.leftPredicateJoin
+import org.jetbrains.kotlinx.dataframe.api.predicateJoin
+import org.jetbrains.kotlinx.dataframe.api.print
+import org.jetbrains.kotlinx.dataframe.api.remove
+import org.jetbrains.kotlinx.dataframe.api.rightPredicateJoin
+import org.jetbrains.kotlinx.dataframe.api.select
+import org.junit.Test
+
+class PredicateJoinTests : BaseJoinTest() {
+    @Test
```
Jolanrensen, line 27: newline

docs/StardustDocs/topics/operations.md
```diff
@@ -78,6 +78,7 @@ Most multiplex operations end with `into` or `with` function. The following nami
 * [inferType](inferType.md) — infer column type from column values
 * [insert](insert.md) — insert column
 * [join](join.md) — join [`DataFrames`](DataFrame.md) by key columns
+  * [predicateJoin](predicateJoin.md) — join [`DataFrames`](DataFrame.md) by an expression that evaluates joined [DataRows](DataRow.md) to Boolean
```
Jolanrensen, line 81: is this indent correct?

docs/StardustDocs/topics/predicateJoin.md
```diff
@@ -0,0 +1,283 @@
+[//]: # (title: predicateJoin)
+
+<!---IMPORT org.jetbrains.kotlinx.dataframe.samples.api.Join-->
+
+Joins two [`DataFrames`](DataFrame.md) by a join expression. 
+
+```kotlin
```
Jolanrensen, line 7: code block contains non-kotlin code, so the highlighting goes wrong

docs/StardustDocs/topics/predicateJoin.md
```diff
@@ -0,0 +1,283 @@
...
+joinExpression: JoinedDataRow.(JoinedDataRow) -> Boolean
+
+interface JoinedDataRow: LeftDataRow {
+    
+    val right: RightDataRow
+    
+}
+```
+
+Unlike [join](join.md) similar to one from relational databases, where relations between tables are based on IDs and usually equality between values, predicate join allows any operations that returns Boolean value. That gives you the ability to join data based on various logical operations and heuristics. This can also be useful when dealing with data from disparate sources that are connected, but lack common identifiers or structured relations typical of SQL tables with indices.
```
Jolanrensen, line 19: please fold the line so I don't have to scroll horizontally to read it, haha

docs/StardustDocs/topics/join.md
```diff
@@ -94,6 +94,7 @@ df.join(other)
 
 Supported join types:
 * `Inner` (default) — only matched rows from left and right [`DataFrames`](DataFrame.md)
+* `Filter` — only matched rows from left [`DataFrame`](DataFrame.md)
```
Jolanrensen, line 97: probably add predicateJoin below to the joinSpecials as well

Jolanrensen:
the name `predicateJoin` might actually be a bit long and cumbersome, especially if expanded like `innerPredicateJoin`. Maybe it could be shortened like `joinWhere` or even better, be combined with `join`:
```kotlin
campaigns.innerJoin(visits).where {
    right.date in startDate..endDate
}
```
or in the join dsl:
```kotlin
campaigns.innerJoin(visits) {
    where { right.date in startDate..endDate }
}
```

docs/StardustDocs/topics/predicateJoin.md
```diff
@@ -0,0 +1,283 @@
...
+    right[date] in startDate()..endDate()
+}
+```
+
+</tab>
+<tab title="Strings">
+
+```kotlin
+campaigns.predicateJoin(visits) {
+    right[{ "date"<LocalDate>() }] in "startDate"<LocalDate>().."endDate"<LocalDate>()
```
Jolanrensen, line 69: `right.col<LocalDate>("date")` is surely easier to type XD But it's also cool to show this notation. Didn't occur to me this was also possible :)

Jolanrensen, line 69: is this the only notation possible to get the typed column/cell from `right`? Looks cool, but overcomplicated.

koperagen: Another notation is `right.getValue<LocalDate>("date")`. I was trying to make it shorter because it didn't fit in single line, but now samples are split in 5 different cells, so no longer an issue. I'll return `right.getValue<LocalDate>("date")`

koperagen:
> the name `predicateJoin` might actually be a bit long and cumbersome, especially if expanded like `innerPredicateJoin`. Maybe it could be shortened like `joinWhere` or even better, be combined with `join`:
> 
> ```kotlin
> campaigns.innerJoin(visits).where {
>     right.date in startDate..endDate
> }
> ```
> 
> or in the join dsl:
> 
> ```kotlin
> campaigns.innerJoin(visits) {
>     where { right.date in startDate..endDate }
>     // or
>     withPredicate {}
> }
> ```

I agree that `predicateInnerJoin` is too long. I'd prefer to have original join renamed to joinBy and have this to be named join. Another option i can suggest is `joinWith`. Combining this operation with column join is a stretch in my opinion. Columnar join has default join statement (columns with the same names from both dataframes). So, those 2 samples will have very different meaning

```
 campaigns.innerJoin(visits).where {
     right.date in startDate..endDate
 }
```

```
 campaigns.innerJoin(visits)
```

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
koperagen: wouldn't it look odd too? plus, it's not even me who wrote this initially :D 
```Kotlin
joinImpl(
  other = other, 
  type = type, 
  addNewColumns = type.addNewColumns,
  selector = selector
)
```

Jolanrensen: no not at all XD I write functions always like that. IMO positional arguments are a nightmare for refactoring and I think originally they were only allowed áfter the positional ones so mixing them looks extra odd to me.

docs/StardustDocs/topics/operations.md
Jolanrensen: Aah, I agree then :)

docs/StardustDocs/topics/predicateJoin.md
koperagen: I think these blocks use pseudocode intentionally everywhere we have them. So should be ok to use here as well. As for highlighting, with `kotlin` at least valid parts are highlighted. Like interface :)

Jolanrensen: you could also split it up in multiple blocks if you want pseudocode :)
Or do like I usually do:
```kotlin
// some pseudo code () -> Something
fun kotlinCode() {}
```

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
koperagen: done

docs/StardustDocs/topics/predicateJoin.md
koperagen: i don't think that will work, you can see that even inside kotlin code there are symbols like [ ] for optional parts or | for enumeration, etc

docs/StardustDocs/topics/predicateJoin.md
koperagen: i rewrote it completely in the end

docs/StardustDocs/topics/joinWith.md
```diff
@@ -0,0 +1,333 @@
...
+joinExpression: JoinedDataRow.(JoinedDataRow) -> Boolean
+
+interface JoinedDataRow: LeftDataRow {
+    
+    val right: RightDataRow
+    
+}
+```
+
+This function is a [join](join.md) variant that lets you match data using any expression that returns Boolean,
```
Jolanrensen, line 19: a Boolean. It gives the opportunity...

docs/StardustDocs/topics/joinWith.md
```diff
@@ -0,0 +1,333 @@
...
+interface JoinedDataRow: LeftDataRow {
+    
+    val right: RightDataRow
+    
+}
+```
+
+This function is a [join](join.md) variant that lets you match data using any expression that returns Boolean, 
+which also gives opportunity to perform operations that require values from both matching rows.
+Can be helpful if data you want to join wasn't designed relational and requires heuristics to tell if rows are matching,
```
Jolanrensen, line 21: This function can be helpful if the data...

docs/StardustDocs/topics/predicateJoin.md
Jolanrensen: much better :)

koperagen:
Thank you!

PR #435, opened 2023-08-01
Docs update & DynamicDataFrameBuilder

koperagen:
DynamicDataFrameBuilder here is a composition of name generator and mutable list. Useful for Kandy because they create dataframes from columns and need to know column names.

Also changed split documentation and made couple of small changes in other places, so i'd recommend to see commits one by one after https://github.com/Kotlin/dataframe/pull/434 is merged (this contains commits from that PR)

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -154,6 +154,22 @@ tasks.withType<KorroTask> {
     dependsOn(copySamplesOutputs)
 }
 
+// This task installs the pre-commit hook to the local machine the first time the project is built
+// The pre-commit hook contains the command to run processKDocsMain before each commit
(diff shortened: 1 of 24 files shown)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
```diff
@@ -0,0 +1,16 @@
+package org.jetbrains.kotlinx.dataframe.api
+
+import io.kotest.matchers.shouldBe
+import org.junit.Test
+
+class ConstructorsTests {
```
Jolanrensen, line 6: Could you maybe also add a test with duplicate column names? Since DynamicDataFrameBuilder seems to be able to handle those

PR #440, opened 2023-08-08
Updating docs processor

Jolanrensen:
Updating docs preprocessor to 0.2.3 and updating deprecated @tags
`@arg -> @setArg` and `@includeArg -> @getArg` as updated by the docs preprocessor.
Also, inheritance is now supported, bringing the operability of `[references]` in the Gradle plugin closer to how IntelliJ links up references.

I created a pre-commit hook that gets installed locally the first time you assemble the project. This hook runs the doc preprocessor each time you commit something.
One downside is that each commit now takes about 1m30s, but it does come with generated docs safety.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -157,6 +157,8 @@ tasks.withType<KorroTask> {
 // This task installs the pre-commit hook to the local machine the first time the project is built
 // The pre-commit hook contains the command to run processKDocsMain before each commit
 val installGitPreCommitHook by tasks.creating(Copy::class) {
+    doNotTrackState(/* reasonNotToTrackState = */ "Fails on TeamCity otherwise.")
+
     from(File(rootProject.rootDir, "gradle/scripts/pre-commit"))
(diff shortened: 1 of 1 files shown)


koperagen:
Something with the build :c As for pre-commit hooks, let's see how it works

core/build.gradle.kts
```diff
@@ -157,6 +157,8 @@ tasks.withType<KorroTask> {
 // This task installs the pre-commit hook to the local machine the first time the project is built
 // The pre-commit hook contains the command to run processKDocsMain before each commit
 val installGitPreCommitHook by tasks.creating(Copy::class) {
+    doNotTrackState(/* reasonNotToTrackState = */ "Fails on TeamCity otherwise.")
```
koperagen, line 160: curious

koperagen (approved):

core/build.gradle.kts
Jolanrensen: I think it's because TC has a protected/non-existent .git folder

PR #441, opened 2023-08-11
Docs update & DynamicDataFrameBuilder

koperagen:
DynamicDataFrameBuilder here is a composition of name generator and mutable list. Useful for Kandy because they create dataframes from columns and need to know column names.

Also changed split documentation and made couple of small changes in other places, so i'd recommend to see commits one by one after https://github.com/Kotlin/dataframe/pull/434 is merged (this contains commits from that PR)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
```diff
@@ -0,0 +1,16 @@
+package org.jetbrains.kotlinx.dataframe.api
+
+import io.kotest.matchers.shouldBe
+import org.junit.Test
+
+class ConstructorsTests {
```
Jolanrensen, line 6: Could you maybe also add a test with duplicate column names? Since DynamicDataFrameBuilder seems to be able to handle those

docs/StardustDocs/topics/createDataFrame.md
```diff
@@ -213,3 +215,28 @@ val df = students.toDataFrame {
 ```
 
 <!---END-->
+
+### DynamicDataFrameBuilder
+
+Previously mentioned dataframe constructors throw an exception when column names are duplicated. 
+When implementing a custom operation involving multiple dataframes, computed columns or parsing some third-party data,
```
Jolanrensen, line 222: difficult sentence mixing two summations. I'd rephrase it like:

"When implementing a custom operation involving multiple dataframes, or computed columns or when parsing some third-party data..."

Jolanrensen (approved):

docs/StardustDocs/topics/createDataFrame.md
koperagen: Indeed, thank you c:

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
koperagen: Right, makes sense since it's a primary use case :o

PR #449, opened 2023-09-07
Added message for incompatible libraries in Jupyter notebooks

Jolanrensen:
We faced an incompatibility issue by one of our users using Krangl in the same notebook. it broke because Krangl uses Apache commons csv v1.6 while DF uses 1.7: https://kotlinlang.slack.com/archives/C05333T208Y/p1693933250176339

I added a small generic message in the docs regarding this.


Jolanrensen:
Let's wait and see if @ileasile can fix these compatibility issues on the notebook side first.

Issue #124, opened 2022-07-05
Support result sets


john-kramer-upwave:
It would be great if dataframe supported java.sql.ResultSet.

Something like:

`DataFrame.fromResultSet(resultSet)`

This would give you an equivalent data frame as if you exported the result set to a CSV and imported it via DataFrame.readCSV().

Thanks!

zaleslaw:
Closed in #451

PR #451, opened 2023-09-08
Added JDBC-integration

### LINKED ISSUE #212: JDBC Support
Any objections for supporting creating a dataframe directly from a result set? as opposed to have to migrate the results to CSV or JSON or using Arrow. I see the SupportedDataFrameFormat interface only allows for a file or an InputStream. Are there any impediments that would make a result set difficult to support ?

zaleslaw:
**Enhanced SQL Database Integration via JDBC API**

This PR marks a significant advancement in our project by introducing seamless integration with SQL databases through the JDBC API, addressing issues #212 and #124.

**Key Features:**

This enhancement empowers our framework to read data from SQL databases with remarkable flexibility, offering the following reading regimes:

- **Table-based Reading:** Retrieve data from specific tables by name.
- **Database Exploration:** Read all non-system tables within the database.
- **ResultSet Access:** Directly access and read data from ResultSet objects.
- **Complex Query Results:** Fetch results from intricate SQL queries, including joins between two or more tables.

Each reading regime allows for either utilizing a provided database connection or establishing a connection internally if the database configuration details are available. Furthermore, every method within these regimes offers two variants: with and without result limits.

**Schema Information:**

In addition to data retrieval, this enhancement provides the capability to obtain schema information for the results. You can retrieve the schema of individual tables, ResultSet objects, SQL queries, or even acquire schemas for all non-system tables within the database.

**Supported Databases:**

This implementation currently supports the following databases:

- MySQL
- MariaDB
- SQLite
- PostgreSQL
- H2

Please note that while PostgreSQL is supported, certain data types may not be fully accommodated yet.

To gain a comprehensive understanding of this new functionality, I strongly recommend consulting the [documentation](https://github.com/Kotlin/dataframe/blob/8419bb05fa2a98ee888d54e21759e4b379ec5b6d/docs/StardustDocs/topics/readSqlDatabases.md).

Note: The plugin support is a work in progress and not yet complete; kindly disregard this aspect for now. Your feedback and contributions are highly valued as we continue refining this feature.


zaleslaw:
@koperagen @Jolanrensen @ermolenkodev please have a look to be familiar with the code/changes.

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/h2Test.kt
```diff
@@ -0,0 +1,428 @@
...
+                    345.67, 3.45, 5.67, 4.71,
+                    '2023-07-22', '20:45:00', '03:30:00', '2023-07-21 23:45:15',
+                    '2023-07-20 23:45:15', NULL,
+                    'Option3', '{"key": "yet_another_value"}', '345e6789-e89b-12d3-a456-426655440002'
+                )
+            """.trimIndent()
+        ).executeUpdate()
+
+        val df = DataFrame.readSqlTable(connection, "TestTable").cast<TestTableData>()
+        assertEquals(3, df.rowsCount())
```
Jolanrensen, line 207: We usually use `shouldBe` from https://kotest.io/docs/assertions/assertions.html across the library

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/h2Test.kt
```diff
@@ -0,0 +1,428 @@
...
+    @Test
+    fun `read from non-existing table`() {
+        shouldThrow<JdbcSQLSyntaxErrorException> {
+            DataFrame.readSqlTable(connection, "WrongTableName").cast<Customer>()
+        }
+    }
+
+    @Test
+    fun `read from sql query`() {
+        val sqlQuery = """
```
Jolanrensen, line 306: you can @Language("SQL") these for proper highlighting :)

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/imdbTest.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    DriverManager.getConnection(URL, props).use { connection ->
+        val df = DataFrame.readSqlTable(connection, "actors", 100).cast<ActorKDF>()
+        df.print()
+    }
+}
+
+@Test
+fun `read sql query`() {
+    val sqlQuery = "select name, year, rank,\n" +
```
Jolanrensen, line 49: multiline string with @Language("SQL") may be easier

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/MariadbTest.kt
```diff
@@ -0,0 +1,322 @@
...
+
+        val df2 = DataFrame.readSqlTable(connection, "table2").cast<Table1MariaDb>()
+        df2.print()
+        assertEquals(3, df2.rowsCount())
+    }
+
+    @Test
+    fun `read from sql query`() {
+        val sqlQuery = """
+SELECT
```
Jolanrensen, line 292: weird indent

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/imdbTest.kt
```diff
@@ -0,0 +1,71 @@
...
+import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.annotations.DataSchema
+import org.jetbrains.kotlinx.dataframe.api.cast
+import org.jetbrains.kotlinx.dataframe.api.print
+import org.junit.Test
+import java.sql.DriverManager
+import java.util.*
+
+
+private const val URL = "jdbc:mariadb://localhost:3306/imdb"
```
Jolanrensen, line 12: Yeah like you said, we probably need to run these only before a release, maybe on TC? In other cases we should probably disable the tests as it will throw unnecessary errors when building the library.

plugins/dataframe-gradle-plugin/src/integrationTest/kotlin/org/jetbrains/dataframe/gradle/SchemaGeneratorPluginIntegrationTest.kt
```diff
@@ -372,6 +375,127 @@ class SchemaGeneratorPluginIntegrationTest : AbstractDataFramePluginIntegrationT
         result.task(":build")?.outcome shouldBe TaskOutcome.SUCCESS
     }
 
+    @Test
+    // TODO: test is broken
```
Jolanrensen, line 379: Sounds like a dataframe version mismatch

Jolanrensen:
Awesome job! I'll leave some small comments here and there for now and play around with it a bit :)

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -0,0 +1,635 @@
...
+}
+
+/**
+ * Reads the data from a ResultSet and converts it into a DataFrame.
+ *
+ * @param [resultSet] the ResultSet containing the data to read.
+ * @param [dbType] the type of database that the ResultSet belongs to.
+ * @return the DataFrame generated from the ResultSet data.
+ */
+public fun DataFrame.Companion.readResultSet(resultSet: ResultSet, dbType: DbType): AnyFrame {
```
Jolanrensen, line 197: Also extension function on ResultSet might be beneficial :)

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -0,0 +1,635 @@
...
+}
+
+/**
+ * Reads all non-system tables from a database and returns them as a list of data frames
+ * using the provided database configuration.
+ *
+ * @param [dbConfig] the database configuration to connect to the database, including URL, user, and password.
+ * @return a list of [AnyFrame] objects representing the non-system tables from the database.
+ */
+public fun DataFrame.Companion.readAllTables(dbConfig: DatabaseConfiguration): List<AnyFrame> {
```
Jolanrensen, line 248: readAllSqlTables maybe? This is defined on the entire DataFrame.Companion scope

Jolanrensen: Also, this doesn't return the names of the tables right? may be better to return a map

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
zaleslaw: It's a good idea, but I want to postpone all the extension functions for the SQL objects for the 2nd wave

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -0,0 +1,635 @@
...
+}
+
+/**
+ * Reads data from an SQL table and converts it into a DataFrame.
+ *
+ * @param [connection] the database connection to read tables from.
+ * @param [tableName] the name of the table to read data from.
+ * @return the DataFrame containing the data from the SQL table.
+ */
+public fun DataFrame.Companion.readSqlTable(connection: Connection, tableName: String): AnyFrame {
```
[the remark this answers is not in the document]
zaleslaw: Created an issue #458

PaulWoitaschek:
Thanks for the initiative!

We are using Trino:
https://trino.io/docs/current/client/jdbc.html

Would it be possible to make the DbType not sealed?
Then in the `DatabaseConfiguration` we could pass in a custom implementation of the `DbType`, which would superceed the automatic selection.

zaleslaw:
@PaulWoitaschek thanks for the feedback, I will think about making nor sealed

PaulWoitaschek:
@zaleslaw 

For inspiration you can also check how exposed did it:
https://github.com/JetBrains/Exposed/blob/30a9ce0f7d6a98630147bcd8da648d974c6b63c0/documentation-website/Writerside/topics/Frequently-Asked-Questions.md?plain=1#L58

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/MariaDb.kt
```diff
@@ -0,0 +1,83 @@
...
+            "SMALLINT" -> ColumnSchema.Value(typeOf<Int>())
+            "MEDIUMINT"-> ColumnSchema.Value(typeOf<Int>())
+            "MEDIUMINT UNSIGNED" -> ColumnSchema.Value(typeOf<Long>())
+            "INTEGER", "INT" -> ColumnSchema.Value(typeOf<Int>())
+            "INTEGER UNSIGNED", "INT UNSIGNED" -> ColumnSchema.Value(typeOf<Long>())
+            "BIGINT" -> ColumnSchema.Value(typeOf<Long>())
+            "FLOAT" -> ColumnSchema.Value(typeOf<Float>())
+            "DOUBLE" -> ColumnSchema.Value(typeOf<Double>())
+            "DECIMAL" -> ColumnSchema.Value(typeOf<Double>())
+            "DATE" -> ColumnSchema.Value(typeOf<String>())
```
PaulWoitaschek, line 63: I don't fully comprehend the logic but wold this in the end lead to all date types being shown as a `String` instead of the equivalent kotlinx.time classes? (like for instance done in the csv auto inference)

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -0,0 +1,635 @@
...
+public data class DatabaseConfiguration(val url:String, val user: String = "", val password: String = "")
+
+/**
+ * Reads data from an SQL table and converts it into a DataFrame.
+ *
+ * @param [dbConfig] the configuration for the database, including URL, user, and password.
+ * @param [tableName] the name of the table to read data from.
+ * @return the DataFrame containing the data from the SQL table.
+ */
+public fun DataFrame.Companion.readSqlTable(dbConfig: DatabaseConfiguration, tableName: String): AnyFrame {
```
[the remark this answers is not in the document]
koperagen: Data class is easier to reuse imo. Like, i tried to experiment with SQLite db: read multiple tables, made multiple `readSqlTable` calls.  With data class i created a variable. With DSL it would be more tiresome, especially in notebook where there're no refactorings (?) like extract

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
koperagen: Another suggestion for the future: better notebooks support. We can generate a type safe wrapper for this map. From user perspective it'll be:
```
val tables = DataFrame.readAllTables(...)
```
next cell (each table has schema):
```
tables.employee
tables.tracks
...
```

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -0,0 +1,37 @@
...
+public object Sqlite : DbType("sqlite") {
+    override fun convertDataFromResultSet(rs: ResultSet, tableColumn: TableColumnMetadata): Any? {
+        val name = tableColumn.name
+        return when (tableColumn.sqlType) {
+            "INTEGER", "INTEGER AUTO_INCREMENT" -> rs.getInt(name)
+            "TEXT" -> rs.getString(name)
+            "REAL" -> rs.getDouble(name)
+            "NUMERIC" -> rs.getDouble(name)
+            "BLOB" -> rs.getBytes(name)
+            else -> throw IllegalArgumentException("Unsupported SQLite type: ${tableColumn.sqlType}")
```
koperagen, line 23: I tried to read https://www.sqlitetutorial.net/sqlite-sample-database/
It failed with this error
`Unsupported SQLite type: NVARCHAR`.
There's also some unsupported type there that `ResultSet` sees as `DATE`
I think it's because SQLite doesn't truly has types and probably their names can vary greatly


I have several notes here:
1. I don't fully understand why interpret each column type from database metadata tables. I assume it has something to do with schema generation? JDBC interprets values and has information about how to read them: 
```
val type = rs.metaData.getColumnType(it)
when (type) {
java.sql.Types.FLOAT -> rs.getFloat()
...
}
```

2. Can we provide user with a way to support certain column types themselves? Like, if i tried to read that db without access to dataframe repository, i would have no way to fix the problem. Only to report to GitHub :( 

3. Probably worth to add column name to exception message

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -0,0 +1,635 @@
...
+public data class TableMetadata(val name: String, val schemaName: String?, val catalogue: String?)
+
+/**
+ * Represents the configuration for a database connection.
+ *
+ * @property [url] the URL of the database.
+ * @property [user] the username used for authentication (optional, default is empty string).
+ * @property [password] the password used for authentication (optional, default is empty string).
+ */
+public data class DatabaseConfiguration(val url:String, val user: String = "", val password: String = "")
```
koperagen, line 55: I tried this API without reading the docs (on purpose :) ), it wasn't clear for me what is URL. Only after some experiments i found what i should write something like this
"jdbc:sqlite:/home/path/chinook.db"

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -0,0 +1,635 @@
...
+    return readSqlTable(connection, tableName, DEFAULT_LIMIT)
+}
+
+/**
+ * Reads data from an SQL table and converts it into a DataFrame.
+ *
+ * @param [connection] the database connection to read tables from.
+ * @param [tableName] the name of the table to read data from.
+ * @param [limit] the maximum number of rows to retrieve from the table.
+ * @return the DataFrame containing the data from the SQL table.
```
koperagen, line 101: Let's add
```
@see DriverManager.getConnection
```

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -0,0 +1,635 @@
...
+}
+
+/**
+ * Converts the result of an SQL query to the DataFrame.
+ *
+ * @param [dbConfig] the database configuration to connect to the database, including URL, user, and password.
+ * @param [sqlQuery] the SQL query to execute.
+ * @return the DataFrame containing the result of the SQL query.
+ */
+public fun DataFrame.Companion.readSqlQuery(dbConfig: DatabaseConfiguration, sqlQuery: String): AnyFrame {
```
koperagen, line 130: Are there other options for sqlQuery type? Something like PreparedStatement, maybe? I'm afraid that with String it's tempting to use interpolation, and interpolation is prone to injections

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -0,0 +1,635 @@
...
+ * @param [connection] the database connection to read tables from.
+ * @param [limit] the maximum number of rows to read from each table.
+ * @return a list of [AnyFrame] objects representing the non-system tables from the database.
+ */
+public fun DataFrame.Companion.readAllTables(connection: Connection, limit: Int): List<AnyFrame> {
+    val metaData = connection.metaData
+    val url = connection.metaData.url
+    val dbType = extractDBTypeFromURL(url)
+
+    val tableTypes = arrayOf("TABLE")
```
koperagen, line 287: Looks like a static property, let's move it to top level?

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
zaleslaw: 1. The problem with java.sql.Types that many types in the SQL databases are mapped to Types.OTHER (Postgres, MariaDB, MSSQL) and better to control it manually, if you test it on your synthetic tests. If in the future the shorter schema will pass all my tests, I will replace on mapping with codes

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
zaleslaw: Thanks, will improve and probably add one require about possible String formats

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
zaleslaw: From other side, this give freedom to write what user wants, it's not a client library for Java Enterprise with UI elements and library user has its own responsibility to protect his database

I'm trying to avoid to expose low-level java.sql objects in the API, but probably it's not a possible.
Also I thought about parsing sqlQuery/tableName Strings with regular expression to be a valid table name (based on metadata) or valid SELECT - like query (without ; and DELETE/UPDATE sql keywords)

But it's a good idea to provide additional function with Statement or PreparedStatement

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/ImportDataSchema.kt
```diff
@@ -45,6 +49,12 @@ public annotation class CsvOptions(
     public val delimiter: Char,
 )
 
+public annotation class JdbcOptions(
+    public val user: String = "", // TODO: I'm not sure about the default parameters
```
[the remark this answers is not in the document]
Jolanrensen: Then this is the best option

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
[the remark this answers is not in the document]
Jolanrensen: Easier to reuse yes, but not easier to type when calling the function. Maybe we could both support a DatabaseConfiguration and just arguments in the function call?

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/h2Test.kt
zaleslaw: I am not a big fan of that approach, but it should be consistent, thanks

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
zaleslaw: I explored API after proposed changes and found that for meta information better to add separate methods or use methods returning a data schema

When I changed it to map, it overcomplicated a path to the dataframes

But I also thought, that it's great to have a field name for dataframes and one global space of all created dataframes to search between them by name

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/H2.kt
```diff
@@ -0,0 +1,83 @@
...
+            "ENUM" -> rs.getString(name)
+            "JSON" -> rs.getString(name)
+            "UUID" -> rs.getString(name)
+            else -> throw IllegalArgumentException("Unsupported H2 type: ${tableColumn.sqlType}")
+        }
+    }
+
+    override fun toColumnSchema(tableColumnMetadata: TableColumnMetadata): ColumnSchema {
+        return when (tableColumnMetadata.sqlType) {
+            "CHARACTER", "CHAR" -> ColumnSchema.Value(typeOf<String>())
```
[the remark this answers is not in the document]
zaleslaw: You mean all the column values with NULL or justa few NULLs in the column?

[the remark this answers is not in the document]
zaleslaw: You mean all the column values with NULL or justa few NULLs in the column?

plugins/dataframe-gradle-plugin/src/integrationTest/kotlin/org/jetbrains/dataframe/gradle/SchemaGeneratorPluginIntegrationTest.kt
zaleslaw: Sorry, what do you mean?

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -0,0 +1,635 @@
...
+ * @param [rs] the result set
+ * @return a mutable map of column names to [TableColumnMetadata] objects,
+ * where each TableColumnMetadata object contains information such as the column type,
+ * JDBC type, size, and name.
+ */
+private fun getTableColumnsMetadata(rs: ResultSet): MutableMap<String, TableColumnMetadata> {
+    val metaData: ResultSetMetaData = rs.metaData
+    val numberOfColumns: Int = metaData.columnCount
+
+    val tableColumns = mutableMapOf<String, TableColumnMetadata>()
```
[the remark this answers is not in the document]
zaleslaw: If we keep in mind, that usually result mapped to the data class/schema which could not contain repeated fields I will throw an exception for repeated names, it's a user responsibility to make synonyms for repeated columns (the same in the pandas/spark)

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/H2.kt
```diff
@@ -0,0 +1,83 @@
...
+import java.sql.ResultSet
+import kotlin.reflect.typeOf
+
+/**
+ * Represents the H2 database type.
+ *
+ * This class provides methods to convert data from a ResultSet to the appropriate type for H2,
+ * and to generate the corresponding column schema.
+ *
+ * NOTE: All date and timestamp related types are converted to String to avoid java.sql.* types.
```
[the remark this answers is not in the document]
zaleslaw: Created an issue https://github.com/Kotlin/dataframe/issues/461

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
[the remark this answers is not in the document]
zaleslaw: Again, for now, a lot of functions, we tracked all the ideas here or I created a separate issues for that.

I want to give users ability to taste it with the status experimental to handle more cases

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
Jolanrensen: One global space of all created dataframes would break the functional aspect of Kotlin DataFrame. They are immutable and any modification makes a new instance.
We could have a DataFrame Collection of some kind, but that would not be much different than a Map, or be able to name DataFrames.
Naming a DataFrame is not that far-fetched, since essentially, a DataFrame and ColumnGroup are the same and ColumnGroups can be named too. That said, currently variable names are used to refer to a DataFrame. There would be a mismatch when people would write: `val dataFrameName = dataFrameOf(...) named "otherDataFrameName"`.

Maybe in the meantime, you could return a DataFrame containing a "name: ValueColumn<String>" and "df: FrameColumn" column, which, especially in notebooks, would be a clear way for users to access multiple dataframes using a single one.

plugins/dataframe-gradle-plugin/src/integrationTest/kotlin/org/jetbrains/dataframe/gradle/SchemaGeneratorPluginIntegrationTest.kt
Jolanrensen: I think the gradle plugin uses the bootstrap version of dataframe, which does not yet include your added functions (like `readSqlTable` etc.). I remember from working on OpenAPI and adding support to the Gradle plugin that I continuously had to publishToMavenLocal because the Gradle plugin cannot use the source files directly.

Jolanrensen: @koperagen am I correct?

PR #452, opened 2023-09-08
Public `empty(DataFrameSchema)` API

koperagen:
It's needed in Kandy to create GroupBy with empty groups that can still be aggregated. You can see the difference in the test i added

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataFrame.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataFrame.kt
@@ -16,7 +16,9 @@ import org.jetbrains.kotlinx.dataframe.impl.DataFrameSize
 import org.jetbrains.kotlinx.dataframe.impl.getColumnsImpl
 import org.jetbrains.kotlinx.dataframe.impl.headPlusArray
 import org.jetbrains.kotlinx.dataframe.impl.headPlusIterable
+import org.jetbrains.kotlinx.dataframe.impl.schema.createEmptyDataFrame
 import org.jetbrains.kotlinx.dataframe.impl.schema.createEmptyDataFrameOf
+import org.jetbrains.kotlinx.dataframe.schema.DataFrameSchema
 import kotlin.reflect.KType
 
 /**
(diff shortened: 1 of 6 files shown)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/testSets/person/DataFrameTests.kt
```diff
@@ -2383,4 +2383,26 @@ class DataFrameTests : BaseTest() {
...
+
+    @Test
+    fun `api for creating GroupBy with empty groups which can be aggregated using statistics`() {
+        val df1 = dataFrameOf("a", "b")(1, "c")
+        val df2 = DataFrame.empty()
+        val groupBy = dataFrameOf(columnOf("group1", "group2") named "group", columnOf(df1, df2)).asGroupBy()
+
+        shouldThrow<IllegalStateException> {
+            groupBy.aggregate {
+                sum("a")
```
zaleslaw, line 2395: is it illegal because..?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataFrame.kt
```diff
@@ -33,6 +35,8 @@ public interface DataFrame<out T> : Aggregatable<T>, ColumnsContainer<T> {
         public fun empty(nrow: Int = 0): AnyFrame = if (nrow == 0) Empty else DataFrameImpl<Unit>(emptyList(), nrow)
 
         public inline fun <reified T> emptyOf(): DataFrame<T> = createEmptyDataFrameOf(T::class).cast()
+
+        public fun empty(schema: DataFrameSchema): AnyFrame = schema.createEmptyDataFrame()
```
zaleslaw, line 39: Could you please provide minor KDocs for new method and group of similar methods (like empty)?

zaleslaw (approved):

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/testSets/person/DataFrameTests.kt
koperagen: Added another assertion. It's illegal because column a is not found :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataFrame.kt
koperagen: Sure, provided one for `emptyOf` and `empty`

Issue #402, opened 2023-06-12
Permission denied when opening xlsx on linux


holgerbrandl:
How to reproduce use:
1) Use system with linux OS (observed on ubunu)
2) Use DataFrame.readXls to read any excel file
3) switch to **another** user account
4) rerun the the same tests

_Expected_: It should read the file

_Observed_: It creates a directory called `/tmp/poifiles` but with just write permission for the first users. So the second attempt from the other account will fail as shown in the screenshots All later users will struggle with this problem until /tmp is cleaned up manually.

![image](https://github.com/Kotlin/dataframe/assets/200952/d5425753-215f-4f08-8e7b-59c3457319ff)

Jolanrensen:
Luckily it's an easy fix :) https://stackoverflow.com/questions/5297911/changing-location-of-temp-files-created-using-apache-poi

PR #472, opened 2023-10-16
[Fix] Linux multiple user permission reading xlsx files

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/402


dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -56,6 +59,20 @@ internal class DefaultReadExcelMethod(path: String?) : AbstractDefaultReadMethod
...
 private const val readExcel = "readExcel"
 
+/**
+ * To prevent [Issue #402](https://github.com/Kotlin/dataframe/issues/402):
+ *
+ * Creates new temp directory instead of the default `/tmp/poifiles` which would
+ * cause permission issues for multiple users.
+ */
+private fun setWorkbookTempDirectory() {
+    val tempDir = Files.createTempDirectory("dataframe-excel").toFile()
```
zaleslaw, line 69: Please move directory name to constants

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -72,6 +89,7 @@ public fun DataFrame.Companion.readExcel(
     rowsCount: Int? = null,
     nameRepairStrategy: NameRepairStrategy = NameRepairStrategy.CHECK_UNIQUE,
 ): AnyFrame {
+    setWorkbookTempDirectory()
```
zaleslaw, line 92: Should we handle some exception here, if we could not create a temp directory?

zaleslaw (changes requested):
Please answer the questions

zaleslaw (approved):

PR #480, opened 2023-10-20
Add static rendering for GitHub

### LINKED ISSUE #476: Static HTML output for dataframes in the Notebook for rendering on GitHub
Now DataFrame tables are not rendered in the notebooks when these notebooks are placed on GitHub. We could change it with the [approach](https://github.com/Kotlin/kandy/pull/155/files#diff-3aaadf3300c1cde75007e0241f7d11cfc13450d09ce493a7eb1afbd20c6cfeddR54-R61) used in Kandy library.

ileasile:
Fixes #476

I tried to make the code effective and clean enough. However, there might be problems with hierarchical dataframes. Please check if it's the case and adjust the code, it shouldn't be hard. Don't forget that you can't add any JavaScript though.

An example notebook rendered with the library from this PR is here: https://gist.github.com/ileasile/a0b52c0df65bb401078cd845387dd6f0

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
@@ -18,6 +18,7 @@ import org.jetbrains.kotlinx.dataframe.impl.renderType
 import org.jetbrains.kotlinx.dataframe.impl.scale
 import org.jetbrains.kotlinx.dataframe.impl.truncate
 import org.jetbrains.kotlinx.dataframe.jupyter.CellRenderer
+import org.jetbrains.kotlinx.dataframe.jupyter.DefaultCellRenderer
 import org.jetbrains.kotlinx.dataframe.jupyter.RenderedContent
 import org.jetbrains.kotlinx.dataframe.name
 import org.jetbrains.kotlinx.dataframe.nrow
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
```diff
@@ -172,6 +172,68 @@ internal fun AnyFrame.toHtmlData(
...
+        columnsToRender.forEach { col ->
+            emitTag("td") {
+                append(cellRenderer.content(row[col.path()], configuration).truncatedContent)
+            }
+        }
+    }
+
+    fun StringBuilder.emitBody() = emitTag("tbody") {
+        val rowsCountToRender = minOf(rowsCount(), configuration.rowsLimit ?: Int.MAX_VALUE)
+        for (rowIndex in 0..<rowsCountToRender) {
```
Jolanrensen, line 219: wow first time I see ..< in action, cool

Jolanrensen (approved):
Can be merged if we do https://github.com/Kotlin/dataframe/issues/503 afterwards

Jolanrensen:
Actually, adding static rendering could throw OOM errors where they previously wouldn't happen. How should we handle those?
image

Jolanrensen:
Thanks for the performance fixes :)

Jolanrensen (approved):

Jolanrensen:
I'll also cherry-pick it into 0.12.1

Issue #483, opened 2023-10-24
Error: Cannot get a NUMERIC value from a STRING cell


jmrsnt:
There is an error when reading an xlsx file that has cells that have formulas that return string. The file [formula_cell.xlsx](https://github.com/Kotlin/dataframe/files/13127219/formula_cell.xlsx) was used as an example.

```kotlin
import org.jetbrains.kotlinx.dataframe.DataFrame
import org.jetbrains.kotlinx.dataframe.io.*
import org.jetbrains.kotlinx.dataframe.api.*

class MyApp {
    fun run() {
        val path = this.javaClass.getResource("formula_cell.xlsx")!!
        val df = DataFrame.readExcel(path)

        df.print()
    }
}

fun main(args: Array<String>) {
    MyApp().run()
}
```

This produces the error below:

```log

Exception in thread "main" java.lang.IllegalStateException: Cannot get a NUMERIC value from a STRING cell
	at org.apache.poi.xssf.usermodel.XSSFCell.typeMismatch(XSSFCell.java:946)
	at org.apache.poi.xssf.usermodel.XSSFCell.getNumericCellValue(XSSFCell.java:260)
	at org.jetbrains.kotlinx.dataframe.io.XlsxKt.cellValue(xlsx.kt:293)
	at org.jetbrains.kotlinx.dataframe.io.XlsxKt.readExcel(xlsx.kt:242)
	at org.jetbrains.kotlinx.dataframe.io.XlsxKt.readExcel(xlsx.kt:182)
	at org.jetbrains.kotlinx.dataframe.io.XlsxKt.readExcel(xlsx.kt:101)
	at org.jetbrains.kotlinx.dataframe.io.XlsxKt.readExcel$default(xlsx.kt:91)
	at MyApp.run(Main.kt:8)
	at MainKt.main(Main.kt:15)
```

This occurs because when reading the cell value, if it is a formula, the value is being returned as stringCellValue, at `xlsx.kt:293`

https://github.com/Kotlin/dataframe/blob/c8ae4f21bc4b71e2ee8322dba6a004651aaade4e/dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt#L281-L298

Jolanrensen:
fix merged, thanks!

PR #484, opened 2023-10-24
Fix reading Excel file that has cells with formulas that return string

jmrsnt:
This PR is to fix the issue #483


Jolanrensen:
Thanks for the fix! Might I just request a recursive call instead of listing all cellType->valueTypes twice? Might also make it easier to extend in the future if we get more types. Something like:
```kotlin
private fun Cell?.cellValue(sheetName: String): Any? {
    if (this == null) return null
    fun getValueFromType(type: CellType?): Any? = when (type) {
        CellType._NONE -> error("Cell $address of sheet $sheetName has a CellType that should only be used internally. This is a bug, please report https://github.com/Kotlin/dataframe/issues")
        CellType.NUMERIC -> {
            val number = numericCellValue
            when {
                DateUtil.isCellDateFormatted(this) -> DateUtil.getLocalDateTime(number).toKotlinLocalDateTime()
                else -> number
            }
        }

        CellType.STRING -> stringCellValue
        CellType.FORMULA -> getValueFromType(cachedFormulaResultType)
        CellType.BLANK -> stringCellValue
        CellType.BOOLEAN -> booleanCellValue
        CellType.ERROR -> errorCellValue
        null -> null
    }

    return getValueFromType(cellType)
}
```

(and renaming "greather" to "greater" in the test :) )

jmrsnt:
Yes, it makes sense. Thank you for observing, I am sending the requested changes.

Jolanrensen (approved):

Issue #489, opened 2023-11-06
Catch Error as wel as Exception when parsing OpenAPI


Jolanrensen:
When using `DataFrame.read()` with some JSON, we get this error. It doesn't occur when using `DataFrame.readJson()`.
```
USE_FAST_DOUBLE_PARSER
java.lang.NoSuchFieldError: USE_FAST_DOUBLE_PARSER
	at com.fasterxml.jackson.dataformat.yaml.YAMLParser._parseNumericValue(YAMLParser.java:1055)
	at com.fasterxml.jackson.core.base.ParserBase.getNumberType(ParserBase.java:660)
	at com.fasterxml.jackson.databind.deser.std.BaseNodeDeserializer._fromFloat(JsonNodeDeserializer.java:650)
	at com.fasterxml.jackson.databind.deser.std.BaseNodeDeserializer._deserializeContainerNoRecursion(JsonNodeDeserializer.java:479)
	at com.fasterxml.jackson.databind.deser.std.JsonNodeDeserializer.deserialize(JsonNodeDeserializer.java:88)
	at com.fasterxml.jackson.databind.deser.std.JsonNodeDeserializer.deserialize(JsonNodeDeserializer.java:20)
	at com.fasterxml.jackson.databind.deser.DefaultDeserializationContext.readRootValue(DefaultDeserializationContext.java:323)
	at com.fasterxml.jackson.databind.ObjectMapper._readTreeAndClose(ObjectMapper.java:4716)
	at com.fasterxml.jackson.databind.ObjectMapper.readTree(ObjectMapper.java:3076)
	at io.swagger.v3.parser.util.DeserializationUtils.readYamlTree(DeserializationUtils.java:247)
	at io.swagger.v3.parser.util.DeserializationUtils.deserializeIntoTree(DeserializationUtils.java:145)
	at io.swagger.v3.parser.OpenAPIV3Parser.readContents(OpenAPIV3Parser.java:168)
	at io.swagger.v3.parser.OpenAPIV3Parser.readContents(OpenAPIV3Parser.java:107)
	at io.swagger.parser.OpenAPIParser.readContents(OpenAPIParser.java:29)
	at org.jetbrains.kotlinx.dataframe.io.IsOpenApiKt.isOpenApiStr(isOpenApi.kt:9)
	at org.jetbrains.kotlinx.dataframe.io.IsOpenApiKt.isOpenApi(isOpenApi.kt:37)
[...]

PR #490, opened 2023-11-06
Catching Error too in `isOpenApiStr`

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/489

### DIFF
--- a/dataframe-openapi/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/isOpenApi.kt
+++ b/dataframe-openapi/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/isOpenApi.kt
@@ -8,7 +8,7 @@ import java.net.URL
 public fun isOpenApiStr(text: String): Boolean = try {
     val parsed = OpenAPIParser().readContents(text, null, null)
     parsed.openAPI?.components?.schemas != null
-} catch (_: Exception) {
+} catch (_: Throwable) {
     false
 }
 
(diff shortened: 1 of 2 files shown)


dataframe-openapi/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/isOpenApi.kt
```diff
@@ -8,7 +8,7 @@ import java.net.URL
 public fun isOpenApiStr(text: String): Boolean = try {
     val parsed = OpenAPIParser().readContents(text, null, null)
     parsed.openAPI?.components?.schemas != null
-} catch (_: Exception) {
+} catch (_: Throwable) {
```
zaleslaw, line 11: I'm thinking, could we give more info here, just logging the reason of Error or Exception, now it silently happened

zaleslaw: Or, if logger is not setup, create a TODO with "add logger"

zaleslaw (changes requested):
Silent Throwable

dataframe-openapi/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/isOpenApi.kt
Jolanrensen: hmm, we already have the problem of the openAPI parser logging exceptions each time we read anything with DataFrame. I don't want to needlessly log even more stuff in this case. That said, adding a logger at level INFO or DEBUG might still be beneficial indeed.

zaleslaw: So sad about openAPI parser logging during reading, Debug level will be a good option, so, probably a good idea to add todo or comment, to this issue or to the openAPI parser logging exception issue (if it is exists), to explain why we doing it silently, because for me it looks like anti-pattern

zaleslaw (approved):

zaleslaw (approved):

Jolanrensen:
Also cherry picked into https://github.com/Kotlin/dataframe/tree/0.12.1

PR #505, opened 2023-11-15
Type mapping rewriting for JDBC integration

zaleslaw:


koperagen (changes requested):
Let's discuss column types nullability before merging

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -25,10 +12,11 @@ import org.jetbrains.kotlinx.dataframe.io.db.DbType
 import org.jetbrains.kotlinx.dataframe.io.db.extractDBTypeFromUrl
 import org.jetbrains.kotlinx.dataframe.schema.ColumnSchema
 import org.jetbrains.kotlinx.dataframe.schema.DataFrameSchema
+import java.sql.*
```
Jolanrensen, line 15: We don't use *-imports in the library

zaleslaw: you are right, again new idea without settings did this (will check PR before the finalization)

PR #506, opened 2023-11-17
remove klaxon usage outside json.kt

koperagen:
Excluding klaxon transivite dependency which is fine if you don't use JSON reading can lead to an unexpected class not found exception now


Jolanrensen (approved):
Lol, indeed, we don't need klaxon for this single function

Issue #503, opened 2023-11-14
Regenerate notebooks with static output


Jolanrensen:
Based on https://github.com/Kotlin/dataframe/pull/480
the example notebooks would have to be regenerated

PR #511, opened 2023-11-21
Static notebook fixes and examples

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/503 and https://github.com/Kotlin/dataframe/issues/476.

This PR adds static HTML DataFrames to .ipynb files, both when rendering DataFrames in Swing and HTML such that they can be displayed on Github. For example: https://github.com/Kotlin/dataframe/blob/static-examples2/examples/notebooks/github/github.ipynb

At the moment, I excluded the [`json` example](https://github.com/Kotlin/dataframe/blob/static-examples2/examples/notebooks/json/KeyValueAndOpenApi.ipynb) because enabling static output makes the file 170MB+ and thus too large for GitHub. This probably needs https://github.com/Kotlin/dataframe/issues/512 to be fixed beforehand. 

Non-JS collapsible DataFrames also should be explored later: https://github.com/Kotlin/dataframe/issues/513.


koperagen (approved):
Nice!

Jolanrensen:
Now I'll just have to redo everything when https://github.com/Kotlin/dataframe/issues/513 is done

Jolanrensen:
And cherry picked into 0.12.1

PR #514, opened 2023-11-23
use range indexing to get rows subset

koperagen:
I noticed that previous implementation took 200 ms to execute on 10m rows dataframe because it iterates over all rows. Indexing should improve it

### DIFF
--- a/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
+++ b/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
@@ -24,7 +24,6 @@ import org.jetbrains.kotlinx.dataframe.api.SplitWithTransform
 import org.jetbrains.kotlinx.dataframe.api.Update
 import org.jetbrains.kotlinx.dataframe.api.at
 import org.jetbrains.kotlinx.dataframe.api.dataFrameOf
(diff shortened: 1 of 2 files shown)


core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
```diff
@@ -60,7 +59,7 @@ public object KotlinNotebookPluginUtils {
      * It's used for example for dynamic pagination in Kotlin Notebook Plugin.
      */
     public fun getRowsSubsetForRendering(df: AnyFrame, startIdx: Int, endIdx: Int): DisableRowsLimitWrapper =
-        DisableRowsLimitWrapper(df.filter { it.index() in startIdx until endIdx })
+        DisableRowsLimitWrapper(df[startIdx until endIdx])
```
Jolanrensen, line 62: cool place to try the new `..<` operator :D

Jolanrensen (approved):
funny how such a simple change can cause such a great impact :)

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
koperagen: new shiny! i'll use it indeed

PR #519, opened 2023-11-27
Better static tables

Jolanrensen:
Based upon https://github.com/Kotlin/dataframe/issues/513, this fix adds better static tables with uncollapsed column groups as well as interactive frame columns using `<details>` (without JS and optional CSS).
This allows ipynb notebook files to be viewed by simple renderers that don't allow iFrames or JS to be run (such as GitHub). Note that Github strips out the CSS and `<details>` tags and displays the nested DataFrames in full, so I made sure the limits for nested dataframes were reasonable.
This means: We adhere to the given `rowsLimit` and `nestedRowsLimit` (20 and 5 by default respectively) for the top-level and we halve the `nestedRowsLimit` (rounding up) for each deeper nested DF.

I also updated all notebooks except for the json one (it's too large for GH now)


Jolanrensen:
Okay, outside of Github, this static rendering works great, however, Github cannot render anything even smelling interactive. I tried CSS tricks, `<details>`, but it doesn't help. CSS just doesn't do anything, as well as JS. 
So, this is the result we get... What looks better in your opinion?

Before:
image
After:
image

Before:
image
After:
image

Before:
image
After (too large to fit, but is scrollable):
image

koperagen:
Ok, i think i understand what's going on in that matrix creation. Looks good! Few suggestions: move implementation to staticHtml.kt, there's now a lot of specific things not related to regular HTML. Second, have you checked escaping? `toHTML` does escape symbols, maybe `cellFormatter` that is used will do it, not sure

Jolanrensen:
@koperagen Thanks! I'll move it to another file.
As for escaping, yes, that's done by this line:
```append(cellRenderer.content(cellValue, configuration).truncatedContent)```
The `CellRenderer` calls `renderValueForHtml()` which uses the `formatter` :).

Issue #530, opened 2023-12-06
Document and test DataColumnArithmetics.kt


Jolanrensen:
DataFrame supports many DataColumn arithmetics like `/`, `+`, `*`, etc. but they are undocumented (both in KDocs and on the [website](https://kotlin.github.io/dataframe/columnarithmetics.html)).

This should be done and made sure it's discoverable.

~Also, `!` is missing for boolean columns.~

Jolanrensen:
AI likes to think it's possible to write `colA + colB`: pairwise operations. At the moment we only support `colA + value`, but it may be worth considering the other notation as well.

There are some downsides though; there would need to be quite a few overloads for this, combining all number types etc.

koperagen:
Please refer to this discussion when implementing https://github.com/Kotlin/dataframe/pull/1617

PR #531, opened 2023-12-06
Added ! to DataColumnArithmetics

Jolanrensen:
Very small addition which adds `!booleanColumn` notation across the library similar to the other arithmetics we support.

 As referenced by https://github.com/Kotlin/dataframe/issues/530.


zaleslaw (approved):
New tricky pattern to handle Nullability in Generics:)

Issue #526, opened 2023-12-04
It fails to rename nullable inline value column 


holgerbrandl:
**Expected**: It should run the following example
```kotlin

import org.jetbrains.kotlinx.dataframe.api.*

@JvmInline
value class Speed(val kmh: Number){
}

class PathSegment(
    val id: String, val speedLimit: Speed? = null,
)


fun main() {
    val segments = listOf(PathSegment("foo", Speed(2.3)), PathSegment("bar",))

    val df = segments.toDataFrame()
    df.rename("speedLimit" to "speed_limit").print()
}
```

**Observed**: It fails with 
```
Exception in thread "main" java.lang.NullPointerException
	at org.kalasim.scratch.Speed.hashCode-impl(InlineClassDf.kt)
	at org.kalasim.scratch.Speed.hashCode(InlineClassDf.kt)
	at org.jetbrains.kotlinx.dataframe.impl.UtilsKt.rollingHash(Utils.kt:99)
	at org.jetbrains.kotlinx.dataframe.impl.columns.UtilsKt.getHashCode(Utils.kt:39)
	at org.jetbrains.kotlinx.dataframe.impl.columns.DataColumnImpl$hashCode$2.invoke(DataColumnImpl.kt:36)
	at org.jetbrains.kotlinx.dataframe.impl.columns.DataColumnImpl$hashCode$2.invoke(DataColumnImpl.kt:36)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
	at org.jetbrains.kotlinx.dataframe.impl.columns.DataColumnImpl.getHashCode(DataColumnImpl.kt:36)
	at org.jetbrains.kotlinx.dataframe.impl.columns.DataColumnImpl.hashCode(DataColumnImpl.kt:38)
	at java.base/java.util.HashMap.hash(HashMap.java:340)
	at java.base/java.util.HashMap.put(HashMap.java:608)
	at org.jetbrains.kotlinx.dataframe.impl.api.RenameKt.renameImpl(rename.kt:64)
	at org.jetbrains.kotlinx.dataframe.impl.api.RenameKt.renameImpl(rename.kt:19)
	at org.jetbrains.kotlinx.dataframe.api.RenameKt.into(rename.kt:57)
	at org.jetbrains.kotlinx.dataframe.api.RenameKt.rename(rename.kt:22)
	at org.kalasim.scratch.InlineClassDfKt.main(InlineClassDf.kt:18)
[...]

PR #532, opened 2023-12-07
Fix performance problem in `rename` implementation

nikitinas:
Currently `DataColumn` object is used as a key for `associateBy` in `rename` implementation. It leads to computation of rolling hash function and scanning through all column data. This problem can be seen in stack trace for https://github.com/Kotlin/dataframe/issues/526

Instead, a column path should be used as column id.


Jolanrensen:
Thanks! Could you rebase on the master so that the TeamCity tests pass (includes https://github.com/Kotlin/dataframe/pull/535) and `assemble` the project to make sure the doc processor is run and the generated-sources are updated? Then I can merge it :)

nikitinas:
Done

Jolanrensen (approved):

PR #535, opened 2023-12-12
Update simple-git version to 2.0.3

Jolanrensen:
This was already updated but reverted somehow. Fixes unexpected character bug on windows machines.


Jolanrensen:
Cherry picked into 0.12.1 too

PR #542, opened 2023-12-15
Fix toDataFrame for value classes

koperagen:
Problem can be seen here
https://github.com/Kotlin/dataframe/issues/526

### DIFF
--- a/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
+++ b/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
@@ -163,7 +163,7 @@ internal fun convertToDataFrame(
             }
         }
         property.javaField?.isAccessible = true
-//        property.isAccessible = true
+        property.isAccessible = true
 
         var nullable = false
         var hasExceptions = false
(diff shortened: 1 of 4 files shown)


core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
```diff
@@ -141,7 +144,26 @@ internal fun convertToDataFrame(
         val property = it
         if (excludes.contains(property)) return@mapNotNull null
 
+        class ValueClassConverter(val unbox: Method, val box: Method)
+
+        val valueClassConverter = (it.returnType.classifier as? KClass<*>)?.let { kClass ->
+            if (!kClass.isValue) return@let null else {
```
nikitinas, line 150: Why do we need explicit return here? This seems readable enough:
```
if (!kClass.isValue) null else {
```

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -224,4 +225,63 @@ class CreateDataFrameTests {
...
+    internal value class Speed(val kmh: Number?){
+    }
+
+    internal class PathSegment(
+        val id: String, val speedLimit: Speed? = null,
+    )
+
+    @Test
+    fun valueClassNullableField() {
+        val segments = listOf(PathSegment("foo", Speed(2.3)), PathSegment("bar",))
```
nikitinas, line 239: Let's remove trailing comma

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
koperagen: right, thank you

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
koperagen: updated formatting here

Issue #437, opened 2023-08-02
Migrate netflix and titanic examples to Kandy


koperagen:

zaleslaw:
@koperagen could you please clarify the issue? Add links to the examples, motivation (why to do it), and what places should be changed (or added from scratch)

koperagen:
Sure. These 2 notebooks use lets-plot for building plots, but now we also have Kandy. I think it will be nice to replace usages of lets-plot API with kandy api because it's intended to provide nicer Kotlin DSL
https://github.com/Kotlin/dataframe/tree/master/examples/notebooks/titanic
https://github.com/Kotlin/dataframe/tree/master/examples/notebooks/netflix

zaleslaw:
@koperagen please, have a look, seems, like it's not actual anymore and I could close it

Jolanrensen:
Actually, it was fixed in https://github.com/Kotlin/dataframe/pull/548

PR #548, opened 2023-12-21
migrate Titanic.ipynb to Kandy

koperagen:
https://github.com/Kotlin/dataframe/issues/437
I realized that there's no CorrPlot a bit too late, but found that corr + gather can prepare data for plotting very well too, yielding almost the same results. Also adjusted the look.


zaleslaw:
I have an idea to post code changes here, if they are not so large, because Github doesn't support good diffs for Notebooks

Jolanrensen:
Please run it with the latest master so that the tables render correctly for github :)

Jolanrensen (changes requested):
I'll review like `cell:line: comment` :)

 - 15:6: Probably replace with `JetBrainsColors.darkGrey` right?
 - 19:1: Very long line, maybe split it across multiple 
 - 21:3: Also long, but better than the last one
 - 31:: Hmm, the color of the small circles is very hard to see, could these be scaled up?
 
Like I said in the previous comment, please run the notebook from the master branch published to mavenLocal such that the tables render on GH :)
Other than that, very nice! Kandy is so much easier to read than Let's Plot haha

Jolanrensen:
I'll merge this for now and check it for 0.13.1-RC1

PR #550, opened 2024-01-07
Reading Arrow NullVector

Kopilov:
Apache Arrow files might contain `NullVector` values (as result of saving null-infilled column in other libraries and languages without static types and target schema). From this PR they will be correctly read by Kotlin DataFrame instead of crushing. Also we can make saving to `NullVector`s, should we?

Among others, Arrow itself is upgraded to last stable version (14.0.2) and #428 problem is fixed for Arrow writing by replacing original `hasNulls` function with custom explicit checking.

### DIFF
--- a/dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
+++ b/dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
@@ -223,6 +223,15 @@ private inline fun <reified T> List<T?>.withTypeNullable(
     return this to typeOf<T>().withNullability(nullable)
 }
 
+@JvmName("withTypeNullableNothingList")
+private fun List<Nothing?>.withTypeNullable(
+    expectedNulls: Boolean,
+    nullabilityOptions: NullabilityOptions,
+): Pair<List<Nothing?>, KType> {
+    val nullable = nullabilityOptions.applyNullability(this, expectedNulls)
+    return this to nothingType(nullable)
+}
+
 private fun readField(root: VectorSchemaRoot, field: Field, nullability: NullabilityOptions): AnyBaseCol {
     try {
         val range = 0 until root.rowCount
@@ -256,7 +265,7 @@ private fun readField(root: VectorSchemaRoot, field: Field, nullability: Nullabi
             is TimeStampMilliVector -> vector.values(range).withTypeNullable(field.isNullable, nullability)
             is TimeStampSecVector -> vector.values(range).withTypeNullable(field.isNullable, nullability)
             is StructVector -> vector.values(range).withTypeNullable(field.isNullable, nullability)
-            is NullVector -> vector.values(range) to nothingType(field.isNullable)
+            is NullVector -> vector.values(range).withTypeNullable(field.isNullable, nullability)
             else -> {
                 throw NotImplementedError("reading from ${vector.javaClass.canonicalName} is not implemented")
             }


dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/ArrowWriterImpl.kt
```diff
@@ -249,7 +249,7 @@ internal class ArrowWriterImpl(
         strictType: Boolean,
         strictNullable: Boolean,
     ): FieldVector {
-        val containNulls = (column == null || column.hasNulls())
+        val containNulls = (column == null || column.values().any { it == null})
```
Jolanrensen, line 252: Please revert this line as it's less efficient. The `hasNulls()`/`isMarkedNullable` is set upon column creation and since columns are immutable we can safely assume all values are not null if `hasNulls() == false`. Maybe we could only run this check if `strictNullable == true`, wdyt @koperagen?

Kopilov: OK, I have removed that commit on rebasing to new master. But we still can specify nullable type for not-null values and vice versa (look first message in #428, that test is reproduceable just now).

> Maybe we could only run this check if `strictNullable == true`

Sounds good, I can make this in separate PR. Or we should fix #428 totally in any another way (such as keeping "nullable" for types and "hasNulls" for values independently). You were agree that `hasNulls()` should only return true if there are actually nulls in the column.

Jolanrensen: We're indeed still working on figuring out a good way to fix https://github.com/Kotlin/dataframe/issues/428 so that the cases you gave are not possible anymore. That way `hasNulls() == isMarkedNullable` at all times. The difficult part is finding a good balance between performance and safety. We don't want to have to scan an entire list of data if a user can ensure it has or has no nulls...

gradle/libs.versions.toml
```diff
@@ -30,7 +30,7 @@ sl4j = "2.0.7"
 junit = "4.13.2"
 kotestAsserions = "4.6.3"
 jsoup = "1.14.3"
-arrow = "11.0.0"
+arrow = "14.0.2"
```
Jolanrensen, line 33: I'm sorry for the late response. We want to merge NullVector reading support for version 0.13.0-RC1 soon, but  in the meantime arrow was updated to 15.0.0. (https://github.com/Kotlin/dataframe/pull/590)
Could you update it here too?

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
```diff
@@ -172,6 +173,10 @@ private fun StructVector.values(range: IntRange): List<Map<String, Any?>?> = ran
     getObject(it)
 }
 
+private fun NullVector.values(range: IntRange): List<Any?> = range.map {
```
Jolanrensen, line 176: And if I'm correct, this List will always only contain `null`s, right? Then the return type should probably be `List<Nothing?>`.
That's in line with the type of `val a = listOf(null)`.

Jolanrensen (changes requested):

gradle/libs.versions.toml
Kopilov: Done

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
Kopilov: I tried to apply
```patch
-private fun NullVector.values(range: IntRange): List<Any?> = range.map {
-    getObject(it)
+private fun NullVector.values(range: IntRange): List<Nothing?> = range.map {
+    assert(getObject(it) == null)
+    null
 }
 ```
 
This provides build error
```
 > Task :dataframe-arrow:compileKotlin FAILED
e: file:///home/akopilov@dellin.local/workspace/kotlin/dataframe/dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt:254:51 Cannot use 'Nothing' as reified type parameter
```

**Comment by Kopilov** [comment]
> Merge branch 'master' into NullVector

Should we apply `git rebase` instead (to avoid spahetty-like history)?

File: dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt (hunk 3 above)
Comment (Jolanrensen) [reply dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt:176]: Ah yes of course, `typeOf<Nothing?>()` also doesn't work.
But there are workarounds for that, like here https://github.com/Kotlin/dataframe/blob/341365f4b986a47e8955f369582620a147370d02/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt#L464

So you can make a workaround that still has the `applyNullability` (with checked `NullabilityOptions`) done to the list:

```kotlin
private fun nothingType(nullable: Boolean): KType =
    if (nullable) {
        typeOf<List<Nothing?>>()
    } else {
        typeOf<List<Nothing>>()
    }.arguments.first().type!!

@JvmName("withTypeNullableNothingList")
private fun List<Nothing?>.withTypeNullable(
    expectedNulls: Boolean,
    nullabilityOptions: NullabilityOptions,
): Pair<List<Nothing?>, KType> {
    val nullable = nullabilityOptions.applyNullability(this, expectedNulls)
    return this to nothingType(nullable)
}
```

At the end it's just a small piece of help to the user: for instance, if they add all values from their `Nothing?` column and, say, a `String` column, the result should be a `List<String?>`. If the first were `Any?` then the result would be `List<Any?>` which is not helpful.

Kopilov: Done (I hope that I understood correctly what do you mean)

Jolanrensen:
Thanks for the help! I'll run the CI and merge :)

Jolanrensen:
@Kopilov Looks like the test `org.jetbrains.kotlinx.dataframe.io.ArrowKtTest.testReadingAllTypesAsEstimatedNotNullableWithNulls` now fails: 
```
org.junit.ComparisonFailure: expected:<kotlin.Nothing?> but was:<kotlin.Nothing> expected:<kotlin.Nothing[?]> but was:<kotlin.Nothing[]>
  at org.jetbrains.kotlinx.dataframe.io.ExampleEstimatesAssertionsKt.assertEstimations(exampleEstimatesAssertions.kt:163)
  at org.jetbrains.kotlinx.dataframe.io.ArrowKtTest.testReadingAllTypesAsEstimatedNotNullableWithNulls(ArrowKtTest.kt:221)
```
This probably means we need my entire solution with `NullabilityOptions` after all... The tests are fine if I use this:

```kotlin
...

@JvmName("withTypeNullableNothingList")
private fun List<Nothing?>.withTypeNullable(
    expectedNulls: Boolean,
    nullabilityOptions: NullabilityOptions,
): Pair<List<Nothing?>, KType> {
    val nullable = nullabilityOptions.applyNullability(this, expectedNulls)
    return this to nothingType(nullable)
}
```

and then
```kotlin
is NullVector -> vector.values(range).withTypeNullable(field.isNullable, nullability)
```

Kopilov:
@Jolanrensen applied, thanks

Jolanrensen (approved):

PR #552, opened 2024-01-09
make sure to adjust nullability in order to fix platform types

koperagen:
Platform types can lead to unexpected NPE and the way types rendered now: "URI!" - ! sign appears. Code with such type cannot be compiled Given type URI!, type.withNullability will produce either URI or URI?. Both are OK for us.


Jolanrensen:
> Code with such type cannot be compiled Given type URI!, type.withNullability will produce either URI or URI?.

When will it produce either the nullable one or the non-nullable one?

koperagen:
I just wanted to say that calling `type.withNullability` fixes platform type :) In our case this function is called when `Infer.Nulls` is used, in which case actual nullability depends on data.

zaleslaw (approved):
Cool, this PR gives me an ability to dig into Infering of nullability

PR #561, opened 2024-01-16
Added examples for jdbc integration

zaleslaw:


examples/idea-examples/imdb/build.gradle.kts
```diff
@@ -0,0 +1,59 @@
...
+    kotlin("jvm")
+    id("org.jetbrains.kotlinx.dataframe")
+}
+
+repositories {
+    mavenLocal() // in case of local dataframe development
+    mavenCentral()
+}
+
+kotlin.sourceSets.getByName("main").kotlin.srcDir("build/generated/ksp/main/kotlin/")
```
koperagen, line 14: Does it resolve & compile without this line?

examples/idea-examples/imdb/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/jdbc/Example_1_Define_schema_manually.kt
```diff
@@ -0,0 +1,40 @@
...
+    val df = DataFrame.readSqlTable(dbConfig, TABLE_NAME_MOVIES, 10000).cast<Movies>()
+
+    // print the dataframe
+    df.print()
+
+    // print the dataframe metadata and statistics
+    df.describe().print()
+
+    // print names of top-10 rated films
+    df.sortByDesc { it[Movies::rank] }
```
koperagen, line 36: Strange that you use KProperties API here and not extension properties

examples/idea-examples/imdb/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/jdbc/Example_1_Define_schema_manually.kt
```diff
@@ -0,0 +1,40 @@
...
+    val year: String
+    val rank: String
+}
+
+fun main() {
+    // define the database configuration
+    val dbConfig = DatabaseConfiguration(URL, USER_NAME, PASSWORD)
+
+    // read the table
+    val df = DataFrame.readSqlTable(dbConfig, TABLE_NAME_MOVIES, 10000).cast<Movies>()
```
koperagen, line 27: I wonder.. `cast<Movies>()` isn't very safe, although we use it some times.  Let's use `cast<Movies>(verify=true)` in examples just to be sure and provide clear error and not some sudden NPE / class cast exceptions

koperagen (changes requested):
A lot of code :slightly_smiling_face: It looks interesting, i want to read it carefully. For now left comment on simple things that draw my attention

examples/idea-examples/imdb/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/jdbc/Example_1_Define_schema_manually.kt
zaleslaw: It does not work correctly in the one project and it was a reason, but now it works correctly, moved to extension

examples/idea-examples/imdb/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/jdbc/Example_1_Define_schema_manually.kt
zaleslaw: probably better to make verify=true - default option, if it's so important, what do you think? Is there another way to cast raw result to the schema?

examples/idea-examples/imdb/build.gradle.kts
zaleslaw: Better to create an issue for all examples (it's the same in all)

zaleslaw:
I decided to move it for a while into a separate repository

PR #571, opened 2024-01-25
Add option to prevent Gradle plugin from adding dependency on KSP

koperagen:
We add plugin automatically for better onboarding experience, but it makes project depend on some specific KSP version. KSP complains when people use it with different version of compiler:
`ksp-1.9.0-1.0.12 is too old for kotlin-1.9.21. Please upgrade ksp or downgrade kotlin-gradle-plugin to 1.9.0.`

Still, it's ok to use preprocessor built for 1.9.0 with KSP 1.9.21. That's why we need an option for people to decide version of KSP they need in their project.
People will be able to use our preprocessor built for older versions of KSP API most of the time without problems thanks to backward compatibility
In dataframe project itself we always need to update Kotlin version in two steps. With this change (and one more PR later) it won't be necessary anymore

### DIFF
--- a/plugins/dataframe-gradle-plugin/src/main/kotlin/org/jetbrains/dataframe/gradle/toCamelCaseByDelimiters.kt
+++ b/plugins/dataframe-gradle-plugin/src/main/kotlin/org/jetbrains/dataframe/gradle/toCamelCaseByDelimiters.kt
@@ -1,9 +1,11 @@
 package org.jetbrains.dataframe.gradle
 
+import java.util.Locale
+
 fun String.toCamelCaseByDelimiters(delimiters: Regex): String {
-    return split(delimiters).joinToCamelCaseString().decapitalize()
+    return split(delimiters).joinToCamelCaseString().replaceFirstChar { it.lowercase(Locale.getDefault()) }
 }
 
 fun List<String>.joinToCamelCaseString(): String {
-    return joinToString(separator = "") { it.capitalize() }
+    return joinToString(separator = "") { s -> s.replaceFirstChar { if (it.isLowerCase()) it.titlecase(Locale.getDefault()) else it.toString() } }
 }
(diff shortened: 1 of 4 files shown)


koperagen:
Let's keep this internal for a while. I'll come up with good explanation for setup page later. For now i'd like to merge it without docs and update our own project to use it to simplify publishing process.

Jolanrensen (approved):

plugins/dataframe-gradle-plugin/src/main/kotlin/org/jetbrains/dataframe/gradle/toCamelCaseByDelimiters.kt
```diff
@@ -1,9 +1,11 @@
...
+import java.util.Locale
+
 fun String.toCamelCaseByDelimiters(delimiters: Regex): String {
-    return split(delimiters).joinToCamelCaseString().decapitalize()
+    return split(delimiters).joinToCamelCaseString().replaceFirstChar { it.lowercase(Locale.getDefault()) }
 }
 
 fun List<String>.joinToCamelCaseString(): String {
-    return joinToString(separator = "") { it.capitalize() }
+    return joinToString(separator = "") { s -> s.replaceFirstChar { if (it.isLowerCase()) it.titlecase(Locale.getDefault()) else it.toString() } }
```
zaleslaw, line 10: Sorry, changes in the behaviour is not clear for me here. Is it related to the KSP update, isn't it?

zaleslaw (approved):

plugins/dataframe-gradle-plugin/src/main/kotlin/org/jetbrains/dataframe/gradle/toCamelCaseByDelimiters.kt
koperagen: `capitalize` was deprecated some time ago. I come across the warning and decided to fix it

Jolanrensen:
One problem I found: 
If `kotlin.dataframe.add.ksp=false`, `KspPluginApplier` isn't run which skips the line that makes ksp depend on `symbol-processor-all`.

PR #572, opened 2024-01-25
make `forEach` inline to enable composable and suspend calls in lambda

koperagen:


core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/forEach.kt
```diff
@@ -7,21 +7,21 @@ import org.jetbrains.kotlinx.dataframe.columns.values
 
 // region DataColumn
 
-public fun <T> DataColumn<T>.forEach(action: (T) -> Unit): Unit = values.forEach(action)
+public inline fun <T> DataColumn<T>.forEach(action: (T) -> Unit): Unit = values().forEach(action)
```
koperagen, line 10: values was an internal property

Jolanrensen (approved):
sounds reasonable :) might be more functions that need this actually

Issue #573, opened 2024-01-26
Change serialization format for rendering in IntelliJ IDEA


ermolenkodev:
The task is to change the serialization format to enable dynamic loading of nested frames in the Kotlin notebooks plugin in IntelliJ IDEA.

PR #574, opened 2024-01-26
Fix for #573: Change serialization format for rendering in IntelliJ IDEA 

ermolenkodev:
Fixes https://github.com/Kotlin/dataframe/issues/573


core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
```diff
@@ -859,42 +867,103 @@ private val valueTypes =
...
-        when {
-            col is ColumnGroup<*> -> encodeRow(col, index)
-            col is FrameColumn<*> -> encodeFrame(col[index])
-            col.isList() -> {
-                col[index]?.let { array(it as List<*>) } ?: array()
-            }
+        when (col) {
+            is ColumnGroup<*> -> encodeRow(col, index)
+            is FrameColumn<*> -> encodeFrame(col[index])
+            else -> encodePrimitiveData(col, index)
```
koperagen, line 873: let's call it `encodeValue`?  Name primitive made me think that some cases were removed (List, else)

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
```diff
@@ -941,6 +1031,26 @@ public fun AnyFrame.toJson(prettyPrint: Boolean = false, canonical: Boolean = fa
     }.toJsonString(prettyPrint, canonical)
 }
 
+public fun AnyFrame.toJsonWithMetadata(
+    rowLimit: Int,
+    nestedRowLimit: Int? = null
+): JsonObject {
```
koperagen, line 1037: Why do we need to expose this class in public API?

koperagen:
Can i ask you to move internal and private functions from this file to new writeJson.kt / readJson.kt files under impl.io package? This file became too big to navigate :c

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
```diff
@@ -859,42 +867,103 @@ private val valueTypes =
...
+    val values = frame.columns().map { col ->
+        when (col) {
+            is ColumnGroup<*> -> obj(
+                DATA to encodeRowWithMetadata(col, index, rowLimit),
+                METADATA to obj(KIND to ColumnKind.Group.toString())
+            )
+
+            is FrameColumn<*> -> {
+                val data = if (rowLimit == null) encodeFrameWithMetadata(col[index])
+                else encodeFrameWithMetadata(col[index].rows().take(rowLimit).toDataFrame(), rowLimit)
```
koperagen, line 907: I believe rows().take(rowLimit).toDataFrame() can be simplified to `take(rowLimit)`. Can you check if result is the same?

koperagen:
You changed handling of value and array columns. Can you tell me more about changes? This part of serialization was not tested by unit tests thoroughly before. I worry if things might break
(no test coverage)
image

docs/serialization_format.md
```diff
@@ -0,0 +1,49 @@
+## Serialization format for the Kotlin notebooks plugin
+This document is an informal specification of the serialization format used for rendering Kotlin dataframes in the Kotlin notebooks plugin of IntelliJ IDEA.
```
zaleslaw, line 2: Cool, many thanks for this document!

zaleslaw (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/JupyterHtmlRenderer.kt
```diff
@@ -60,21 +56,18 @@ internal inline fun <reified T : Any> JupyterHtmlRenderer.render(
...
-        val jsonEncodedDf = json {
-            obj(
-                "nrow" to df.size.nrow,
-                "ncol" to df.size.ncol,
-                "columns" to df.columnNames(),
-                "kotlin_dataframe" to encodeFrame(df.rows().take(limit).toDataFrame()),
-            )
-        }.toJsonString()
-        notebook.renderAsIFrameAsNeeded(html, staticHtml, jsonEncodedDf)
+        val jsonEncodedDf = df.toJsonWithMetadata(limit, reifiedDisplayConfiguration.rowsLimit)
```
koperagen, line 59: Let's discuss this. I'd say default row limit that we used to have for HTML table might be too small in some situations. For example, imagine groupBy with 10 rows and 1000 rows in each group. I'd expect that it will still work very fast. That's why i'd prefer not to have any pagination in this situation at all.

ermolenkodev:
> Can i ask you to move internal and private functions from this file to new writeJson.kt / readJson.kt files under impl.io package? This file became too big to navigate :c

done

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
ermolenkodev: I checked, but it doesn't work without an explicit cast as `take(rowLimit)` returns List<DataRow>

ermolenkodev:
> You changed handling of value and array columns. Can you tell me more about changes? This part of serialization was not tested by unit tests thoroughly before. I worry if things might break (no test coverage) image

I have extracted the code that filters autogenerated values and array columns into functions to be reused in new methods. I have double-checked the logic, and it seems it hasn't been altered. However, I can't guarantee this without running tests. If this is critical, we should add tests and merge them before proceeding with this pull request.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/JupyterHtmlRenderer.kt
ermolenkodev: It is problematic. When loading a dataset with numerous rows within nested frames, such as ten frames with 1000 rows each. This can result in a large JSON file when serialized. It has a multiplicative effect when the number of cells increases.  It is even quite easy making it impossible to open in a Kotlin Notebook due to the default limit on JSON size during parsing. This limit is shared with the Python notebook plugin.

In my experience, pagination improves responsiveness and does not make it slower. You have the option to manually adjust the page size to view the entire nested table or utilize the “Go to Row” action. 

Another possible option is to include a function in KotlinNotebookPluginUtils that can modify or disable pagination for specific use cases.

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
```diff
@@ -0,0 +1,212 @@
...
+internal val AnyFrame.isPossibleToFindUnnamedColumns: Boolean
+    get() = columns().size != 1
+
+internal fun AnyFrame.extractArrayColumn(): DataColumn<*>? {
+    val allColumns = columns()
+
+    return columns().filter { it.name.startsWith(arrayColumnName) }
+        .takeIf { isPossibleToFindUnnamedColumns }
+        ?.maxByOrNull { it.name }?.let { arrayCol ->
+            if (arrayCol.kind() == org.jetbrains.kotlinx.dataframe.columns.ColumnKind.Group) null
```
Jolanrensen, line 159: fq name again

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/JupyterHtmlRenderer.kt
koperagen: I see. Ok, i can agree with that. No need to change anything then

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
koperagen: `else encodeFrameWithMetadata(col[index].take(rowLimit), rowLimit)`
Like this

koperagen: and there's similar place in `internal fun KlaxonJson.encodeDataFrameWithMetadata`

koperagen: `frame.rows().take(rowLimit).toDataFrame()` -> `frame.take(rowLimit)`

koperagen (approved):
Approved, just one tiny fix related to take before merge please

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
ermolenkodev: Oh, now I got it! Thanks!

Issue #576, opened 2024-01-28
Add support for reading Parquet files


fb64:

zaleslaw:
Thanks for the issue, could you please provide a use-case for this task? Do you have data in this format to explore?

Jolanrensen:
Can confirm, Parquet support would be nice, it's [used](https://spark.apache.org/docs/latest/sql-data-sources-parquet.html) a lot with Spark as well.

kopczynski-9livesdata:
I second this. Parquet format is massively popular when working with Spark and having this in Kotlin Dataframe would be great.

PR #577, opened 2024-01-28
Add support for reading parquet file thanks to arrow-dataset #576

fb64:
Fixes https://github.com/Kotlin/dataframe/issues/576
- Using [arrow-dataset](https://arrow.apache.org/docs/java/dataset.html) to read parquet file
- Adding  test parquet file generated from a fork of [arrow_example](https://github.com/fb64/arrow_example/tree/write-parquet-file) that allows to write parquet file
- Adding arrow-dataset dependency
- Updating arrow version (from 11 to 15)


koperagen:
Hi and thanks for the PR. I have nothing to add to the code. But i get this exception trying to run the test on Linux with both Java 11 and 17. The issue seems to be on Arrow side. Do you know about any requirements for it to work?

```
java.lang.UnsatisfiedLinkError: /tmp/jnilib-15573607865820834233.tmp: /tmp/jnilib-15573607865820834233.tmp: undefined symbol: _ZTIN6google8protobuf7MessageE
	at java.base/jdk.internal.loader.NativeLibraries.load(Native Method)
	at java.base/jdk.internal.loader.NativeLibraries$NativeLibraryImpl.open(NativeLibraries.java:388)
	at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:232)
	at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:174)
	at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2389)
	at java.base/java.lang.Runtime.load0(Runtime.java:755)
	at java.base/java.lang.System.load(System.java:1953)
	at org.apache.arrow.dataset.jni.JniLoader.load(JniLoader.java:92)
	at org.apache.arrow.dataset.jni.JniLoader.loadRemaining(JniLoader.java:75)
	at org.apache.arrow.dataset.jni.JniLoader.ensureLoaded(JniLoader.java:61)
	at org.apache.arrow.dataset.jni.NativeMemoryPool.createListenable(NativeMemoryPool.java:44)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingImplKt.readArrowDataset(arrowReadingImpl.kt:327)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingKt.readParquet(arrowReading.kt:197)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingKt.readParquet$default(arrowReading.kt:194)
	at org.jetbrains.kotlinx.dataframe.io.ArrowKtTest.testReadParquet(ArrowKtTest.kt:590)
```

fb64:
Hi @koperagen it seems to be a JNI issue, I just checked and it works well both on my MacBook Pro (M1) and on a PC with  Windows 10 (intel core i7).  What is the processor architecture on your computer ? Normally `arrow-dataset` dependency provides the required native library but maybe it not fits with your hardware 🤔 
Did you launch tests with gradle ? Only the JVM arg `--add-opens java.base/java.nio=ALL-UNNAMED` is needed (as is configured in gradle tasks.test ...)
I'll try to reproduce on linux with docker ....
image

koperagen:
Yes, i do run them with Gradle. Processor is Intel core i7. I tried to run the test on TeamCity, but there it fails on Linux as well :( 
Upon inspecting that .so file content i found that this protobuf symbol is indeed undefined which means it's expected to be loaded from another library
0000000000000000         *UND*	0000000000000000              _ZTIN6google8protobuf7MessageE
`objdump /tmp/jnilib-11657767653473718381.tmp -x`

But the library doesn't have a dependency on any protobuf library, so i assume it could be a linkage error on Arrow side.. maybe? 
```
ldd /tmp/jnilib-11657767653473718381.tmp
	linux-vdso.so.1 (0x00007ffd3f689000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007fe7a31f0000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fe7a31eb000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fe7a31e6000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fe79f800000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fe7a30ff000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fe7a30dd000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fe79f400000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fe7a3219000)
```

fb64:
Effectively, I also reproduced the issue with docker, downgrading arrow dependency to the version `14.0.2` seems to fix the error. I'll update the PR and check try to dig on arrow side

koperagen:
Can confirm, 14.0.2 works. I tried it, have some requests
1. Can you clarify what are expected url values? 
Because following code throws an exception 
```val df = DataFrame.readParquet(URL("https://d37ci6vzurychx.cloudfront.net/trip-data/yellow_tripdata_2023-07.parquet"))```

```
Exception in thread "main" java.lang.RuntimeException: Unrecognized filesystem type in URI: https://d37ci6vzurychx.cloudfront.net/trip-data/yellow_tripdata_2023-07.parquet
	at org.apache.arrow.dataset.file.JniWrapper.makeFileSystemDatasetFactory(Native Method)
	at org.apache.arrow.dataset.file.FileSystemDatasetFactory.createNative(FileSystemDatasetFactory.java:40)
	at org.apache.arrow.dataset.file.FileSystemDatasetFactory.<init>(FileSystemDatasetFactory.java:31)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingImplKt.readArrowDataset(arrowReadingImpl.kt:328)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingKt.readParquet(arrowReading.kt:197)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingKt.readParquet$default(arrowReading.kt:194)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingImplKt.main(arrowReadingImpl.kt:348)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingImplKt.main(arrowReadingImpl.kt)
```
Looks like only URL that point to files are valid ones? Can we make this parameter a `File` then? 
```DataFrame.readParquet(URL("file:/home/nikita/Downloads/yellow_tripdata_2023-07.parquet"))```

2. I can't read https://d37ci6vzurychx.cloudfront.net/trip-data/yellow_tripdata_2023-07.parquet and get an exception. Does it work for you? 
```
/arrow/java/dataset/src/main/cpp/jni_util.cc:79: Failed to update reservation while freeing bytes: JNIEnv was not attached to current thread
/tmp/jnilib-17033017462678975899.tmp(+0x1623b38)[0x7fe9e1e23b38]
/tmp/jnilib-17033017462678975899.tmp(_ZN5arrow4util8ArrowLogD1Ev+0xed)[0x7fe9e1e23eed]
... etc
```

**Comment by fb64** [comment]
Actually URI parsing is done natively by arrow and it supports only few file systems and unfortunately http(s) is not supported yet : 
- local file:  `file:/filename.parquet` 
- AWS S3: `s3:/filename.parquet`
- Google Cloud Storage `gs:/filename.parquet` or `gcs:/filename.parquet`
- Hadoop : `hdhf:/filename.parquet`

**Comment by koperagen** [comment]
> unfortunately http(s) is not supported yet
I actually tried to read local copy of that file and it failed with `JNIEnv was not attached to current thread`. Want to see if it's a platform specific bug or something else 

Thanks for clarification about URI. Let's change that parameter type to `java.net.URI` and add a note about filesystems then?

**Comment by fb64** [comment]
I reached the same issue, another problem with JNI (and thread)...
Changing the creation of `NativeMemoryPool` with `NativeMemoryPool.getDefault()` [here](https://github.com/Kotlin/dataframe/pull/577/commits/0dd7498f7f7804a90c569c4b29845415b11e1e1f#diff-f9d6e7bffc8c69ddbbaf8e3ee4978759b79f73fb15fd32f3585c83017e1ab6bdR327) seems to fix the error.
By the way is not recommended to use it in production: https://arrow.apache.org/docs/java/dataset.html#native-memory-management
I also will update the URI part ...

**Comment by zaleslaw** [comment]
Hi, thanks to the PR, I could not understand will it cover any Parquet files or only Parquet files keeping the something in the Arrow format? I will collect a few parquet files and return to you

**Comment by fb64** [comment]
> Hi, thanks to the PR, sorry, I could not understand will it cover any Parquet files or only Parquet files keeping the something in the Arrow format? I will collect a few parquet files and return to you

I confirm that it should cover every parquet files. We facing to a [JNI error](https://github.com/Kotlin/dataframe/pull/577#issuecomment-1919445431) with some parquet files (not all). I created an  issue on arrow repository:  https://github.com/apache/arrow/issues/39919

**Comment by zaleslaw** [comment]
@fb64 we made a decision to not merge it immediately before three things happened:

1. This PR will be tested on our own Parquet files
2. We compare this approach with alternatives (without Apache Arrow, for example)
3. We decided to keep it in Arrow module or create a separate module for it (it could change our 2-level structure of modules and dependencies in the project.

Thanks again for your help and collaboration!

**Comment by fb64** [comment]
> @fb64 we made a decision to not merge it immediately before three things happened:
> 
> 1. This PR will be tested on our own Parquet files
> 2. We compare this approach with alternatives (without Apache Arrow, for example)
> 3. We decided to keep it in Arrow module or create a separate module for it (it could change our 2-level structure of modules and dependencies in the project.
> 
> Thanks again for your help and collaboration!

No problem ! 
From my experience the other alternative is to use the [Java Parquet library ](https://github.com/apache/parquet-mr/tree/master) which relies on Hadoop which can be difficult to run on windows because of certain native libraries (but this point has maybe been improved). On the other hand arrow-dataset seems to be still under development and not totally operational but it seems prometheus and could bring both parquet and orc format reading/writing feature easily.
Let's keep in touch

**Comment by fb64** [comment]
Related Arrow issue for `JNIEnv was not attached to current thread` error : https://github.com/apache/arrow/issues/20379

**Comment by fb64** [comment]
for information I just updated this PR with Arrow 16.0.0 that includes fixes for the 2 issues discovered previously : 
- `java.lang.UnsatisfiedLinkError`  --> https://github.com/apache/arrow/issues/39919
- `JNIEnv was not attached to current thread` --> https://github.com/apache/arrow/issues/20379

**Comment by laurentperez** [comment]
Hi

I'm interested in df support, I can confirm that cloning from https://github.com/fb64/dataframe/tree/arrow-read-parquet and building with `./gradlew build publishToMavenLocal -xtest -xintegrationTest` works. Very nice.

Please note that Arrow required me to set the following args in my build.gradle

`jvmArgs '--add-opens', 'java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED'`

However, when reading a.parquet with 38 millions of lines, it will OOM with `Java heap space: failed reallocation of scalar replaced objects` with the followig code snippet.

It works with smaller files though, around 5M rows per file work.

```
Caused by: java.lang.OutOfMemoryError: Java heap space: failed reallocation of scalar replaced objects
	at java.base/java.lang.Long.valueOf(Long.java:1206)
	at org.apache.arrow.vector.BigIntVector.getObject(BigIntVector.java:131)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingImplKt.values(arrowReadingImpl.kt:111)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingImplKt.readField(arrowReadingImpl.kt:301)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingImplKt.readArrowImpl(arrowReadingImpl.kt:389)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingKt.readArrow(arrowReading.kt:181)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingImplKt.readArrowDataset(arrowReadingImpl.kt:415)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingKt.readParquet(arrowReading.kt:195)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingKt.readParquet$default(arrowReading.kt:192)
```

```
import org.jetbrains.kotlinx.dataframe.io.readParquet
val frame = DataFrame.readParquet(            URL("file:/home/lperez/foo.parquet)        )
        val rowsCount = frame.rowsCount()
        val columnTypes = frame.columnTypes()
        println("columnTypes: $columnTypes")
        println("rowsCount: $rowsCount")
        println(frame)
```

Parquet size :

```
zsh 10147 [127]  (git)-[main]-% parquet-tools inspect foo.parquet

############ file meta data ############
created_by: Polars
num_columns: 7
num_rows: 38837326
num_row_groups: 148
format_version: 1.0
serialized_size: 100764
```
I've set Xmx to 4096M on my machine, should I be able to run 38M of lines or is it too much ?

thanks

**Comment by fb64** [comment]
Hello Laurent, I'm pleased to see your interest in this PR. 
Kotlin Dataframe keeps all data in memory, which may lead to out-of-memory issues with large datasets. 
What is the size of your parquet file and is it compressed ? Could you specify the column types in your dataset? 
For datasets too large for memory, I  can recommend to use DuckDB with Kotlin Dataframe for preprocessing before loading into a Dataframe using Arrow. 
I recently published an article about this : https://dev.to/fb64/kotlin-dataframe-arrow-54kh
Hope it'll help you.

**Comment by laurentperez** [comment]
Hi. My parquet is 126MB on disk, uses ZSTD compression, columns types are byte array, double, float (1), and has 38M lines.  
Theses lines are one month of measures from IoT devices in some industry.

In fact.. I'm already using DuckDB ;) Some of my parquets are generated either by python polars, or duck.  

Our company backends use both kotlin and python so I was looking at your PR with great interest.

I've used KDF before, my use case is to render base64 PNG plot renderings (using [Kandy](https://kotlin.github.io/kandy/quick-start-guide.html)) of the dataframes, I know it works, however, only on "small" datasets : 

I was not aware KDF would load the entire thing in memory.  
I _wrongly_ assumed it'd read using chunks or some sort of cursor or iterable. Perhaps if `readParquet`() delegated to an underlying `(RootAllocator(),512) # from your article`, or some sort of implementation based on https://kotlin.github.io/dataframe/read.html#read-apache-arrow-formats, then it'd fit.

This is not a problem _per se_ : I'm _already_ using duck, and your article is a great piece of information, I'll use it :) :+1: 

I was wondering how one would read such a massive dataset directly thru `readParquet` without knowledge of Arrow (or duck, for that matter) but I guess this is beyond the scope of this PR.

I hope your branch gets merged soon in 1.4.x ;)

(1)

############ file meta data ############
created_by: Polars
num_columns: 7
num_rows: 38837326
num_row_groups: 148
format_version: 1.0
serialized_size: 100764


############ Columns ############
name
id
value2
value3
timestamp
value4
processed_timestamp

############ Column(name) ############
name: name
path: name
max_definition_level: 1
max_repetition_level: 0
physical_type: BYTE_ARRAY
logical_type: String
converted_type (legacy): UTF8
compression: ZSTD (space_saved: -16%)

############ Column(id) ############
name: id
path: id
max_definition_level: 1
max_repetition_level: 0
physical_type: INT64
logical_type: None
converted_type (legacy): NONE
compression: ZSTD (space_saved: 86%)

############ Column(value2) ############
name: value2
path: value2
max_definition_level: 1
max_repetition_level: 0
physical_type: INT64
logical_type: None
converted_type (legacy): NONE
compression: ZSTD (space_saved: 99%)

############ Column(value3) ############
name: value3
path: value3
max_definition_level: 1
max_repetition_level: 0
physical_type: INT64
logical_type: None
converted_type (legacy): NONE
compression: ZSTD (space_saved: 65%)

############ Column(timestamp) ############
name: timestamp
path: timestamp
max_definition_level: 1
max_repetition_level: 0
physical_type: BYTE_ARRAY
logical_type: String
converted_type (legacy): UTF8
compression: ZSTD (space_saved: 84%)

############ Column(value4) ############
name: value4
path: value4
max_definition_level: 1
max_repetition_level: 0
physical_type: DOUBLE
logical_type: None
converted_type (legacy): NONE
compression: ZSTD (space_saved: 82%)

############ Column(processed_timestamp) ############
name: processed_timestamp
path: processed_timestamp
max_definition_level: 1
max_repetition_level: 0
physical_type: BYTE_ARRAY
logical_type: String
converted_type (legacy): UTF8
compression: ZSTD (space_saved: 81%)

**Comment by fb64** [comment]
Polars (and also Pandas) store DataFrames in memory. However, Polars uses Arrow Format for representation in RAM, which can be more efficient for handling large datasets compared to KDF.
Polars has also a lazy mode that allows performing some operations without loading all rows (streaming mode) but only some operations are supported. 
Maybe you can try to check how much memory takes your parquet file into a Polars DataFrame  and compare it to KDF by using : 
````python
import polars as pl
df_pl = pl.read_parquet('<your_parquet_file>')
print(f'Size: {df_pl.estimated_size()} bytes')
````
I could be useful to add a such method in Kotlin DataFrame

**Comment by laurentperez** [comment]
Polars estimated_size returned : 3650708644 == estimated 3.65GB

I made a mistake configuring my gradle build, it was not honoring my heap settings under tests, hence the OOM.

I made it work and it fits in memory using 16GB of heap which roughly translates using `htop` to 18-20GB of Resident Set Size (RSS is real physical RAM used by the JVM, Heap plus the other regions), I had to opt-in for ZGC since I'm on java21.

To estimate an order of magnitude for future readers stumbling upon this thread then I'd say loading the whole set using the JVM will use approx 4-5x the estimated size returned by Polars.

TL;DR : works : 

```
tasks.withType(Test) {
    maxHeapSize = "16g"
    jvmArgs '--add-opens', 'java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED', '-XX:+UseZGC', '-XX:+ZGenerational'
}
```

I understand KDF minimal java version is 8, and ZGC did not land under java <= 15 so ZGC may not be a viable option for massive datasets requiring a big heap (esp. if target is an Android device). I'm using kotlin on server side, so ZGC is ok for me.

Thanks !

**Comment by fb64** [comment]
After testing Parquet and CSV files, I also found that Kotlin Dataframe consumes 4-5 times more memory than Polars and Pandas. This maybe due to type boxing or other things related to the JVM, but not directly linked to parquet parsing.
IMHO there is a room for memory improvement and it should be addressed in a dedicated issue
Regarding the reading of parquet files, it seems quite stable. I know that there are some missing type mappings but I plan to work on it.
In any case, thank you @laurentperez  for taking the time to test and provide constructive feedback. 🙏

**Comment by fb64** [comment]
I just dig on to understand why the same DataFrame takes much space in JVM, and it is indeed due to JVM memory representation (mark word, class pointer, memory alignment). 
According to this [great article](https://dzone.com/articles/java-object-size-estimation-measuring-verifying), Java objects could take up to 16 times more memory than their primitive counterparts.

**Comment by Jolanrensen** [comment]
@fb64 We're aware of the JVM space usage of DataFrame. Some weeks ago I've actually been investigating whether it's possible to swap out the underlying ArrayLists containing the column's data with primitive arrays (https://github.com/Kotlin/dataframe/pull/712). While, yes, it's possible, I found it to be a large trade-off in terms of performance. Yes, we can store values as primitives, but then for each operation, since we use generic functions, they are boxed and unboxed one at a time. This is heavy and is very noticable for large operations. While we could, in theory, create primitive overloads for each operation we have, it may be worth to wait for Project Valhalla to help us here: https://openjdk.org/jeps/402.

**Comment by fb64** [comment]
@Jolanrensen Yes Valhalla sounds promising to improve memory, and combined with the new [Vector API ](https://openjdk.org/jeps/489)it it could improve the performances significantly. 
Another approach could be to use Arrow representation in memory (such as Polars, Pandas, etc.) and layer DataFrame methods on top of it. However, this may not be straightforward.

**Comment by zaleslaw** [comment]
Hi @fb64 I've tested your solution with such type of Parquet file, as serialized Spark models which I parsed manually with hadoop-mr and discovered that groups are not parsed correctly, could it be something it the Arrow implementation?

Could you please check with this file? It contains a schema 
```
message spark_schema {
  required int32 treeID;
  optional group nodeData {
    required int32 id;
    required double prediction;
    required double impurity;
    optional group impurityStats (LIST) {
      repeated group list {
        required double element;
      }
    }
    required double gain;
    required int32 leftChild;
    required int32 rightChild;
    optional group split {
      required int32 featureIndex;
      optional group leftCategoriesOrThreshold (LIST) {
        repeated group list {
          required double element;
        }
      }
      required int32 numCategories;
    }
  }
}
```
but only 
```
treeID: Int
nodeData: Map<String, Any?>
``` 
is discovered
 and as a result, you need to manually convert JSONs in each nodeData

[part-00000-290bdb9d-bc1b-411c-8811-c3205434f5fc-c000.snappy.zip](https://github.com/user-attachments/files/20794807/part-00000-290bdb9d-bc1b-411c-8811-c3205434f5fc-c000.snappy.zip)


P.S. highly lilkely you need to update code in `readField` method to convert StructVector/ListVector somehow in ColumnGroup
`val children = this.children.map { child ->
                child.toDataColumn(child.name)
            }
            ColumnGroup(name, children)
            `

**Comment by Jolanrensen** [comment]
@zaleslaw yes, we don't support nested types in Arrow at all yet https://github.com/Kotlin/dataframe/issues/1256

> P.S. highly lilkely you need to update code in readField method to convert StructVector/ListVector somehow in ColumnGroup

Also, `StructVector.values()` could return either `List<DataColumn<*>` or `List<AnyRow?>`. The first only works if you use `createColumnGuessingType(... allColsMakesColGroup = true)`. The second one works with almost any column creation function, so is easier to use.

**Comment by fb64** [comment]
I also created this ticket  about inner fields when I worked on Arrow Reader : https://github.com/Kotlin/dataframe/issues/536

File: dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReading.kt [hunk 1]
```diff
@@ -185,3 +186,11 @@ public fun DataFrame.Companion.readArrow(
  */
 public fun ArrowReader.toDataFrame(nullability: NullabilityOptions = NullabilityOptions.Infer): AnyFrame =
     DataFrame.Companion.readArrowImpl(this, nullability)
+
+/**
+ * Read [Parquet](https://parquet.apache.org/) data from existing [url] by using [Arrow Dataset](https://arrow.apache.org/docs/java/dataset.html)
+ */
+public fun DataFrame.Companion.readParquet(
+    url: URL,
```
Jolanrensen: Like other `readXxx()` functions in the library, we probably need overloads for providing the path/url in the types `Path`, `File`, `URL`, `String` (as url).

I also noticed Parquet files can be partitioned, so consist of multiple files. It seems like `FileSystemDatasetFactory` can also take an array of uris. If this is common, we should probably allow multiple uri's, so perhaps make it `vararg`? What do you think about that?

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
```diff
@@ -414,3 +419,27 @@ internal fun DataFrame.Companion.readArrowImpl(
         return flattened.concatKeepingSchema()
     }
 }
+
+internal fun DataFrame.Companion.readArrowDataset(
```
Jolanrensen: Would it be worth it to expose this one to the public API as well? That way people could also supply .orc files for instance. We could also keep it internal and add a `readArrowOrc` function later :)

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
```diff
@@ -414,3 +419,27 @@ internal fun DataFrame.Companion.readArrowImpl(
...
+    fileUri: String,
+    fileFormat: FileFormat,
+    nullability: NullabilityOptions = NullabilityOptions.Infer,
+): AnyFrame {
+    val scanOptions = ScanOptions(32768)
```
Jolanrensen: I saw this 32,768 number in the Arrow docs as well. Do you have any idea what this does and if we should allow the user to change it?

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
Jolanrensen: *readArrowDatasetImpl

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
```diff
@@ -414,3 +419,27 @@ internal fun DataFrame.Companion.readArrowImpl(
...
+        FileSystemDatasetFactory(
+            allocator,
+            NativeMemoryPool.createListenable(DirectReservationListener.instance()),
+            fileFormat,
+            fileUri,
+        ).use { datasetFactory ->
+            datasetFactory.finish().use { dataset ->
+                dataset.newScan(scanOptions).use { scanner ->
+                    scanner.scanBatches().use { reader ->
+                        return readArrow(reader, nullability)
```
Jolanrensen: can better call `readArrowImpl()`

Jolanrensen (changes requested):
The team decided to be okay with having Parquet support as a part of `dataframe-arrow`, when this branch is up-to-date and the notes are addressed we'll be ready to merge it :) If you don't have the time, we could also do it for you, let us know. We will write some documentation on the website about it as well, so people know Parquet is supported via Arrow.

fb64:
Good news ! 
@Jolanrensen, I should be able to consider the notes and make the modifications in few days.

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReading.kt
fb64: As far as I Know arrow-dataset does not support http URL for the moment but support many file systems like s3, gcs, hdfs ...
I propose to add methods with Path and File for local files. For http URL it's possible to download the file in the tmp file load the dataframe and delete the file. For other urls we can let arrow-dataset try to load the urls and throw an error if it doesn't. 
I think I can easily add a vararg for URL, File or Path and using the `FileSystemDatasetFactory` constructor that you mention with multiple uris.

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
fb64: According to the documenatation Batchsize is the `Maximum row number of each returned org.apache.arrow.vector.ipc.message.ArrowRecordBatch` I think in most of use case `32768` should be fine and maybe need to be tuned for edge cases. It can be relevant to expose this parameter in the API at least with a default value.

fb64:
I updated the PR according to the previous comments :
- Change method names
- Add support for multiple files with vararg
- Add an API for Path, File, URL, String
- Support `HTTP` URL by downloading temporary file. I don't find an easy way to support Streaming content with arrow-dataset as it's not supported (yet?) natively. 
- Add BatchSize as an optional parameter

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReading.kt
```diff
@@ -185,3 +189,55 @@ public fun DataFrame.Companion.readArrow(
...
 public fun ArrowReader.toDataFrame(nullability: NullabilityOptions = NullabilityOptions.Infer): AnyFrame =
     DataFrame.Companion.readArrowImpl(this, nullability)
+
+/**
+ * Read [Parquet](https://parquet.apache.org/) data from existing [urls] by using [Arrow Dataset](https://arrow.apache.org/docs/java/dataset.html)
+ */
+public fun DataFrame.Companion.readParquet(
+    vararg urls: URL,
+    nullability: NullabilityOptions = NullabilityOptions.Infer,
+    batchSize: Long = 32768,
```
Jolanrensen: forgot one hardcoded default batchSize

Jolanrensen: looks good otherwise :)

Jolanrensen (approved):

zaleslaw (approved):

zaleslaw:
Dear @fb64 thanks for your time and belief in this project, now this PR will be included in 1.0 release and will be available in Beta-3 at the early September

fb64:
> Dear @fb64 thanks for your time and belief in this project, now this PR will be included in 1.0 release and will be available in Beta-3 at the early September

Good news ! I hope this parquet integration will be useful for the community and boost to the library adoption 🚀

PR #579, opened 2024-02-06
add infer types parameter in DSL functions

koperagen:
I believe tests that i added speak for themselves. Type can often be erased to Any and we need an easy way to get a meaningful result

### DIFF
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
@@ -61,6 +61,21 @@ class CreateDataFrameTests {
     }
 
     @Test
+    fun `create column with infer type`() {
+        val data: List<Any> = listOf(1, 2, 3)
+        val res = data.toDataFrame {
+            "e" from inferType { it }
+            expr(infer = Infer.Type) { it } into "d"
+        }
+
+        res["e"].type() shouldBe typeOf<Int>()
+        res["e"].kind() shouldBe ColumnKind.Value
+
+        res["d"].type() shouldBe typeOf<Int>()
+        res["d"].kind() shouldBe ColumnKind.Value
+    }
+
+    @Test
     fun `preserve fields order`() {
         class B(val x: Int, val c: String, d: Double) {
             val b: Int = x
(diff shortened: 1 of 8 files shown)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/testSets/person/DataFrameTests.kt
```diff
@@ -919,6 +921,15 @@ class DataFrameTests : BaseTest() {
         }
     }
 
+    @Test
+    fun `add several columns with type inference`() {
+        val f: Any = 123
+        val df = typed.add {
+            expr(infer = Infer.Type) { f } into "f"
```
Jolanrensen, line 928: I actually expected this to be done by default. If you need to manually specify to infer types, I think people will just not do it. And then, it will be `Any`?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -60,6 +60,21 @@ class CreateDataFrameTests {
         res["e"].type() shouldBe typeOf<Int>()
     }
 
+    @Test
+    fun `create column with infer type`() {
+        val data: List<Any> = listOf(1, 2, 3)
+        val res = data.toDataFrame {
+            "e" from inferType { it }
```
Jolanrensen, line 67: same here, it's good it exists, but I think it should be done by default when using `expr`

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/testSets/person/DataFrameTests.kt
koperagen: It's consistent with `add` itself (default infer = Infer.Nulls) and most other usages of this enum. I think it makes sense. Looking at values to deduce type is a reflective operation and, while i didn't test it myself and decision wasn't made by me, i suppose it can be quite heavy + produce unexpected results (more precise type than needed or something). So i'd say keeping it as an option is a reasonable compromise

Jolanrensen:
I think this requires more explanation, as `expr` already has the ability to generate a `type`. However, this type is only known if it is supplied to `expr` as `reified` parameter. If a column needs to be added using a normal generic function, this type is not known and the newly added column will always be `Any`/`Any?`.
This PR adds the option to infer the (schema) type from the data even if the reified type is too broad.

koperagen:
This summarizes the issue how i initially found it. There was no way to create anything other than `Any` column from my list of Any that i got from a library. Changes in `AddDsl` were made for consistency of `expr` functions
```
 val data: List<Any> = listOf(1, 2, 3)
val res = data.toDataFrame {
      "d" from { it } // Any :( 
      "e" from inferType { it } // Int
       expr(infer = Infer.Type) { it } into "d" // Int
}
```

Jolanrensen (approved):

PR #590, opened 2024-02-14
Version bumps

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/581
Depends on https://github.com/Kotlin/dataframe/pull/592

## Changes:
- Replaced String dependencies to TOML accessors wherever possible
- Added `dependencyUpdates` task that can be run to display which updates are available in a neat DataFrame. `dependencyUpdateExclusions` can be extended to prevent certain dependencies from showing up in the table.
- Based on https://github.com/Kotlin/dataframe/pull/592
  - Added  `kotlin.dataframe.add.ksp=false` in gradle.properties
  - Added explicit ksp plugin in each module

## Version bumps:
 - kotlin `1.9.0` -> `1.9.22`
 - kotest-assertions `4.6.3` -> `5.5.4`
   - 5.8.0 is not possible due to https://github.com/Kotlin/dataframe/issues/595
 - kotlinJupyter `0.11.0-358` -> `0.12.0-139`
   -   replaced deprecated `exec()` calls with `execRendered()`
 - kotlinPoet `1.12.0` -> `1.16.0`
 - libsPublisher `0.0.60-dev-30` -> `1.8.10-dev-43`
 - kover `0.6.0-Beta` -> `0.6.1`
   - https://github.com/Kotlin/dataframe/issues/594
 - commonsCsv `1.8` -> `1.10.0`
 - commonsCompress `1.21` -> `1.26.0`
 - poi `5.2.2` -> `5.2.5`
 - mariadb `3.1.4` -> `3.3.2`
 - h2db `2.2.220` -> `2.2.224`
 - PostgreSQL `42.6.0` -> `42.7.1`
 - sqlite `3.42.0.1` -> `3.45.1.0`
 - kotlinx datetime `0.4.0` -> `0.5.0`
 - openapi swagger `2.1.13` -> `2.1.20`
 - kotlinlogging `5.0.1` -> `6.0.3`
 - sl4j `2.0.7` -> `2.0.12`
 - jsoup `1.14.3` -> `1.17.2`
 - arrow `11.0.0` -> `15.0.0`
 - JUnit jupiter to `5.10.2`
 - JUnit platform to `1.10.2`
- dependencyVersions new! `0.51.0`
- Gradle plugin-publish `0.15.0` -> `1.2.1`
- shadow `7.1.2` -> `8.1.1`
- ktor server netty `1.6.7` -> `2.3.8`
 
## Skipped version bumps:
- ktlint
[...]

PR #592, opened 2024-02-15
Ksp fix

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/pull/571 further.

With `kotlin.dataframe.add.ksp=false`, the user-specified KSP still needs to depend on symbol-processor-all like before. This PR configures KSP regardless of the add.ksp setting.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,8 +1,8 @@
 [versions]
-ksp = "1.9.21-1.0.15"
+ksp = "1.9.22-1.0.17"
 kotlinJupyter = "0.11.0-358"
 ktlint = "3.4.5"
-kotlin = "1.9.0"
+kotlin = "1.9.22"
 dokka = "1.8.10"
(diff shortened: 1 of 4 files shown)


Jolanrensen:
I updated the KSP version to 1.9.22 ahead of the [version bumps](https://github.com/Kotlin/dataframe/pull/590).
Last update of KSP in https://github.com/Kotlin/dataframe/pull/571 caused KSP to be unaware of `@DataSchema`s generated by `dataframes {}` tasks. 
This is done by modifying the creation of each individual `generateDataFrameX` task. Those tasks already added their new src directory to the kotlin sourceSets. While this used to be enough for KSP, it isn't anymore, so now the source is added to the relevant ksp task.
For instance: "build/generated/dataframe/main/kotlin" will be added to `kspKotlin` and  "build/generated/dataframe/test/kotlin" will be added to `kspTestKotlin`.

PR #597, opened 2024-02-21
generate "nullable properties" only where needed

koperagen:
The fact that we generate 4 (and not 2) extension properties for every column has been bothering me for some time now. It makes %trackExecution harder to comprehend, especially considering that we want to improve UX with kernel-generated code on Kotlin Notebook side.
After some thinking i came to believe that 4 properties are only needed for declared schemas (like interfaces, or something generated from open api) and thing that bothers me can be solved by adding an option.


zaleslaw (approved):

Jolanrensen (approved):
I had to look into it a bit deeper again... But I agree, it makes sense!
In REPL, the actual data in the DataFrame is used, meaning the nulls are "pushed down" to the value level.

Issue #312, opened 2023-03-20
Migrate the serialization from Klaxon to kotlinx-serialization library


devcrocod:
This will help integrate the dataframe better into other libraries where serialization is required, such as `ktor`, `ggdsl`, and others

devcrocod:
@belovrv had doubts about the performance of kotlinx.serialization. We discussed this issue and came to the conclusion that there are no obvious problems. However, since this task is quite time-consuming, even though part of the serialization/deserialization logic will be preserved from Klaxon, we decided to postpone its implementation

zaleslaw:
I'd like the serialization and io tasks and especially both libraries, klaxon and kotlinx.serialization, will assign it to me

devcrocod:
A bit more detail about serialization support.

### Benefits:

1. **Performance**. Based on my limited research, simple conversion to strings is performance-wise equivalent to `klaxon`, as it performs identical logic. Marking classes as serializable and writing custom serializers, the performance of `kotlinx-serialization` is expected to be higher than `klaxon`. However, the actual implementation of the serializers will also affect this. In general, the performance will be slightly better or the same as klaxon. Thus, I wouldn't consider this as the main reason for switching to `kotlinx-serialization` or for sticking with `klaxon`.
2. **Code Refactoring**. Support for `kotlinx-serialization` will help us improve the code responsible for json parsing. When I examined the code, I found some code quality issues. Serializers for dataframe will help alleviate this area, which will increase the codebase but will improve code readability and quality.
3. **Type Inference**. This will affect type parsing, as some of the load can be shifted to the `kotlinx-serialization` plugin. This, in turn, will help us partially eliminate the use of platform-dependent reflection. Also, kotlinx-serialization has better support for `JsonElement`. For example, it can identify `NaN` unlike klaxon.
4. **Kotlin Ecosystem**. `kotlinx-serialization` is an official library and is part of the Kotlin ecosystem, which will make it easier for us to work within it in the future. For example, with `Ktor`. There are also guarantees that it will be supported further.
5. **Other Formats**. While official support is limited, interestingly, there is support for `Protobuf`.
6. Multiplatform Support. kotlinx-serialization is a multiplatform library.
7. Flexibility and reliability. Writing custom serializers will allow better control over serialization and deserialization. For example, this can help avoid issues with serialization when a `Map` might be inside a column.


I envision the migration from `klaxon` to `kotlinx-serialization` as follows:
[...]

PR #603, opened 2024-02-23
Replace Klaxon with kotlinx-serialization

devcrocod:
This is the first step in migrating from klaxon to kotlinx-serialization #312 

- Removed klaxon dependency
- Added kotlinx-serialization
- Klaxon objects replaced with JsonElement from serialization
- All other behavior is preserved. Unlike klaxon, kotlinx-serialization works better with primitives. For example, klaxon does not work with float, only with double, so I had to specially handle the float case. Also, nulls are wrapped in serialization.

This PR should be accepted after #574, after which I will resolve the conflicts.


build.gradle.kts
```diff
@@ -71,7 +71,7 @@ private fun String.findVersion(): Version {
 // these names of outdated dependencies will not show up in the table output
 val dependencyUpdateExclusions = listOf(
     // 5.6 requires Java 11
-    libs.klaxon.get().name,
+//    libs.serialization.get().name,
```
Jolanrensen, line 74: you can remove this line altogether if klaxon is removed

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
```diff
@@ -1,60 +1,26 @@
...
-import org.jetbrains.kotlinx.dataframe.api.columnOf
-import org.jetbrains.kotlinx.dataframe.api.concat
-import org.jetbrains.kotlinx.dataframe.api.dataFrameOf
-import org.jetbrains.kotlinx.dataframe.api.firstOrNull
-import org.jetbrains.kotlinx.dataframe.api.getColumn
-import org.jetbrains.kotlinx.dataframe.api.indices
-import org.jetbrains.kotlinx.dataframe.api.isList
-import org.jetbrains.kotlinx.dataframe.api.mapIndexed
+import kotlinx.serialization.ExperimentalSerializationApi
+import kotlinx.serialization.json.*
```
Jolanrensen, line 6: please turn off *-imports in your IDE.

devcrocod:
> @devcrocod you wrote that this is a first step, what could be the next step?

Check related issue

devcrocod:
> What do you mean with "specially handle the float case". Looking at your code I see you skip it and convert it to a Double sometimes. I don't really understand why; DataFrame understands Floats just fine, so you can keep them as Floats

By default,[ klaxon translates all floating-point numbers into Double](https://github.com/cbeust/klaxon/blob/cfd3dac07614d27845001de0ca83583239ffbd73/klaxon/src/main/kotlin/com/beust/klaxon/KlaxonJson.kt#L45). kotlinx-serialization can work with float. So, the question here is not what DataFrame can do, but what klaxon could do. It was possible to support Float in JSON parsing, but then it would have been necessary to additionally change the logic in JSON parsing. I decided not to do this because these subsequent parsers should ultimately change, and because it maximally preserves the[ previous behavior of the library](https://github.com/Kotlin/dataframe/blob/b9d4f0c0ea2bf5a2f8ee5c7806ee87951e74be59/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt#L382).

Jolanrensen:
> > What do you mean with "specially handle the float case". Looking at your code I see you skip it and convert it to a Double sometimes. I don't really understand why; DataFrame understands Floats just fine, so you can keep them as Floats
> 
> By default,[ klaxon translates all floating-point numbers into Double](https://github.com/cbeust/klaxon/blob/cfd3dac07614d27845001de0ca83583239ffbd73/klaxon/src/main/kotlin/com/beust/klaxon/KlaxonJson.kt#L45). kotlinx-serialization can work with float. So, the question here is not what DataFrame can do, but what klaxon could do. It was possible to support Float in JSON parsing, but then it would have been necessary to additionally change the logic in JSON parsing. I decided not to do this because these subsequent parsers should ultimately change, and because it maximally preserves the[ previous behavior of the library](https://github.com/Kotlin/dataframe/blob/b9d4f0c0ea2bf5a2f8ee5c7806ee87951e74be59/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt#L382).

I think the previous behavior of the library was not intended, but rather a limit imposed on the library due to Klaxon. It seems odd to keep this broken behavior if both Kotlinx-serialization ánd DataFrame support Floats to convert them to Doubles. So yes, this might require updating some tests and logic, but it's ultimately for the best I believe.

devcrocod:
I agree that such behavior is not entirely correct, as it also carries computational errors when converting float to double. But I will repeat that this behavior will be changed in the future when serializers for dataframe objects are written.

Moreover, in the current version, there is a question:
During serialization, it's obvious how it should look:
```
1.1f -> 1.1 (float)
1.1 -> 1.100000 (double)
```

But during deserialization, we look at each element, and the logic is as follows: if the number fits within 32, then it's a float; if it's more than 32, then it's a double. And it turns out that our list can easily contain both Float and Double, and instead of the output type being Double, we get Number.

Jolanrensen:
@devcrocod Looks like you introduced 4 failing tests in `org.jetbrains.dataframe.ksp.DataFrameSymbolProcessorTest`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/readJson.kt
```diff
@@ -155,7 +167,7 @@ internal fun fromJsonListAnyColumns(
 
     @Suppress("KotlinConstantConditions")
     val columns: List<AnyCol> = when {
-        // Create one column of type Any? (or guessed primitive type) from all the records
+        // Create one column of type Any? (or guessed a primitive type) from all the records
```
Jolanrensen, line 170: "a" does not belong here

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/readJson.kt
```diff
@@ -189,9 +201,21 @@ internal fun fromJsonListAnyColumns(
                         )
                     }
 
-                    "NaN" -> {
-                        nanIndices.add(i)
-                        collector.add(null)
+                    is JsonPrimitive -> {
```
Jolanrensen, line 204: While this `xOrNull` conversion works, it's quite heavy as it works by throwing/catching exceptions. Ilya actually recently replaced similar logic from the notebook kernel because it was so heavy. It was replaced by Jackson. Now I know that's not an option here, but maybe we could try another approach if that's possible.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/readJson.kt
```diff
@@ -395,24 +419,24 @@ internal fun fromJsonListArrayAndValueColumns(
 
     // list element type can be JsonObject, JsonArray or primitive
     // So first, we gather all properties of objects to merge including "array" and "value" if needed
-    // so the resulting type of a property with instances 123, ["abc"], and { "a": 1, "b": 2 } will be
+    // so the resulting type of property with instances 123, ["abc"], and { "a": 1, "b": 2 } will be
```
Jolanrensen, line 422: "a" should not be removed here

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/readJson.kt
```diff
@@ -488,10 +512,23 @@ internal fun fromJsonListArrayAndValueColumns(
                         records.forEachIndexed { i, v ->
                             when (v) {
                                 is JsonObject -> collector.add(null)
-                                is JsonArray<*> -> collector.add(null)
-                                "NaN" -> {
-                                    nanIndices.add(i)
-                                    collector.add(null)
+                                is JsonArray -> collector.add(null)
+                                is JsonPrimitive -> {
```
Jolanrensen, line 516: Same note as before with `xOrNull` checks

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
```diff
@@ -61,91 +63,121 @@ internal object SerializationKeys {
     const val TYPES = "types"
 }
 
-// See docs/serialization_format.md for a description of
-// serialization versions and format.
-internal const val SERIALIZATION_VERSION = "2.1.0"
+private val valueTypes =
+    setOf(Boolean::class, Double::class, Int::class, Float::class, Long::class, Short::class, Byte::class)
```
Jolanrensen, line 67: Where's Char?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
```diff
@@ -61,91 +63,121 @@ internal object SerializationKeys {
...
+                    }
+                }
             }
+
             else -> encodeValue(col, index, imageEncodingOptions)
         }.let { col.name to it }
     }
     if (values.isEmpty()) return null
-    return obj(values)
+    JsonObject(mapOf("exampleKey" to JsonPrimitive("exampleValue")))
```
Jolanrensen, line 159: Is this a leftover?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
```diff
@@ -261,45 +293,48 @@ internal fun AnyFrame.extractArrayColumn(): DataColumn<*>? {
...
 
     val data = frame.indices().map { rowIndex ->
-        valueColumn
-            ?.get(rowIndex)
-            ?: arrayColumn?.get(rowIndex)
-                ?.let {
-                    if (arraysAreFrames) encodeFrame(it as AnyFrame) else null
-                }
-            ?: encodeRow(frame, rowIndex)
+        valueColumn?.get(rowIndex) ?: arrayColumn?.get(rowIndex)?.let {
```
Jolanrensen, line 303: This ?: chain is very hard to understand. I know you just reformatted it, but maybe you know a better way to write it? :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
```diff
@@ -107,8 +111,8 @@ public class JSON(
     }
 }
 
-public const val arrayColumnName: String = "array"
-public const val valueColumnName: String = "value"
+internal const val arrayColumnName: String = "array"
```
Jolanrensen, line 114: They were made public for external modules like OpenAPI I believe. Though they could probably be renamed to ARRAY_COLUMN_NAME for consistency

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
```diff
@@ -260,22 +265,24 @@ public fun DataRow.Companion.readJsonStr(
...
+        isLenient = true
+        allowSpecialFloatingPointValues = true
+    }
+    return json.encodeToString(JsonElement.serializer(), encodeFrame(this@toJson))
 }
 
 /**
  * Converts the DataFrame to a JSON string representation with additional metadata about serialized data.
- * It is heavily used to implement some integration features in Kotlin Notebook IntellJ IDEA plugin.
+ * It is heavily used to implement some integration features in Kotlin Notebook IntelliJ IDEA plugin.
```
Jolanrensen, line 279: lol

Jolanrensen (changes requested):
Overall a trivial conversion, thanks :) just some small notes

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/readJson.kt
devcrocod: I might be missing something
Could you please explain in more detail or provide an example from the notebook?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
Jolanrensen: ah, but would Char make sense here? or would that just become a String from json?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/readJson.kt
Jolanrensen: Sure, let's say we want to parse an array of `null`s. In this case, for each element it would check [`intOrNull`](https://github.com/Kotlin/kotlinx.serialization/blob/d2dc7d2978bd62971e8ab00de197d59d7e378ed2/formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt#L267), `longOrNull`, `doubleOrNull`, and `floatOrNull`. All of these work with `mapExceptionsToNull {}`, meaning, for each element in the array an exception is thrown and caught 4 times. This is very heavy, especially when it happens very often.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
Jolanrensen: That's a good idea :) I think I originally made it public because user-side generated code could reference it, but I cannot find that anymore.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
devcrocod: I think second options

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/readJson.kt
devcrocod: Hm, indeed this isn't the good place that should be fixed when supporting Serializable for DataFrame and DataRow. I also think that part of this might be optimized at runtime since our value is already wrapped in JsonPrimitive. There are doubts that jdk8 handles such optimizations well

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
devcrocod: By default, everything is read as string, kotlinx-serialization doesn't have the char option for JsonPrimitive

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/readJson.kt
devcrocod: I moved the null check higher up. Given the current capabilities of kotlinx-serialization, there are two options: either leave it as it is now, or write a custom serializer. However, the logic in the custom serializer would be quite similar. We can always take a String and then manually attempt to convert it to the desired type. This might be slightly better in terms of performance, and we would need to account for all edge cases for different types

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
devcrocod: I rewrote it using the `when` construct. it seems more readable now. Could you please take a look?

devcrocod:
I made some minor corrections and a bit of refactoring. I also fixed issues with the plugin tests. The problem was that we set a `rateLimit` (10000) on the incoming input stream. When reading, we check if the selected format is appropriate; if not, we reset the input stream. The issue was that the JSON format was being predicted among the first, and kotlinx-serialization reads all bytes from the incoming stream, ignoring our set limit. Consequently, when we perform a reset, it always results in an error because more than the limit was read.

devcrocod:
This is strange, I got errors with DataFrameBlackBoxCodegenTestGenerated after today's merge

Jolanrensen (approved):

devcrocod:
I did as you said, and locally everything worked for me. But as you can see, the tests on TC still failed:
```
org.opentest4j.AssertionFailedError: Actual data differs from file content: toDataFrame_dsl.fir.ir.txt
```

Could these files be platform or environment dependent? In any case, this approach of regenerating files for tests doesn't look as perfect decision

Jolanrensen:
> I did as you said, and locally everything worked for me. But as you can see, the tests on TC still failed:
> 
> ```
> org.opentest4j.AssertionFailedError: Actual data differs from file content: toDataFrame_dsl.fir.ir.txt
> ```
> 
> Could these files be platform or environment dependent? In any case, this approach of regenerating files for tests doesn't look as perfect decision

Try setting your project sdk + gradle's jvm to JDK 11, while we figure out how to make that the default for all contributors :)

PR #604, opened 2024-02-27
Add Vertica DB support on dataframe-jdbc

brunoteo:
Hi!
I've created this pull request to handle Vertica Database inside dataframe-jdbc

A new Vertica.kt class was added that implements DbType

I tested the solution in verticaTest.kt class running this local docker image -> [https://hub.docker.com/r/vertica/vertica-ce](https://hub.docker.com/r/vertica/vertica-ce)

The documentation is updated with the Vertica dependency

Thanks


zaleslaw:
Hi @brunoteo thanks for the contribution! We need approximately 2–4 weeks to test your solution and make a decision about including this feature in 0.14 release or not. After that, I will be able to return to you with some questions.

It's an excellent-prepared PR!

PR #611, opened 2024-03-01
Updating maven publication developers list

Jolanrensen:
I think it's about time we add @koperagen @Jolanrensen @zaleslaw @ermolenkodev to the developers list, don't you? ;).

@nikitinas Is there another email address we can use?

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -235,7 +235,11 @@ kotlinPublications {
             apache2()
         }
         developers {
-            developer("nikitinas", "Anatoly Nikitin", "Anatoly.Nikitin@jetbrains.com")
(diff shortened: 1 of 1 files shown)


koperagen (approved):

zaleslaw:
Could we post our nicknames without e-mail addresses (for security reasons?)

koperagen:
Repository rules require library authors to provide emails too

zaleslaw:
> > Could we post our nicknames without e-mail addresses (for security reasons?)
> 
> I think our Jetbrains' email addresses are public anyway (and if not, they're not hard to guess)

Ok then

Issue #585, opened 2024-02-12
KDoc rendering looks odd on 2024 Nightly IntelliJ


Jolanrensen:
![image](https://github.com/Kotlin/dataframe/assets/17594275/7cb0a1ae-471d-4a6e-a0b2-b82c32ab852a)

Looks like KDocs need to be checked such that no spaces exist outside inline code blocks.

In comparison, this is what it currently looks like:
![image](https://github.com/Kotlin/dataframe/assets/17594275/6ba87701-a35c-4da9-8a84-2b9265470e23)

Jolanrensen:
I suspect IntelliJ now leans more into their [Markdown Renderer](https://github.com/JetBrains/markdown/tree/master).

This renderer trims code spans:

https://github.com/JetBrains/markdown/blob/aa54cceca5c28b694dcb7c940e910612c82fa76e/src/commonMain/kotlin/org/intellij/markdown/html/GeneratingProviders.kt#L73

Making things like this look weird:
![image](https://github.com/Kotlin/dataframe/assets/17594275/4561f43d-406d-4eb5-9adc-4a9c15886950)

Jolanrensen:
Depends on https://github.com/JetBrains/markdown/pull/150

Jolanrensen:
https://youtrack.jetbrains.com/issue/IDEA-348571/Markdown-KDoc-rendering-errors 
This will fix at least some space-trimming errors, hopefully enough to make the KDocs look okay

Jolanrensen:
Another nice example of how broken it is. 
![image](https://github.com/Kotlin/dataframe/assets/17594275/3e2c969f-2633-4a81-af17-c02794fd5a2b)

Jolanrensen:
https://youtrack.jetbrains.com/issue/KTIJ-29221/Markdown-KDoc-space-rendering-errors is done! The KDocs look a lot better now :) (in the latest 2024.1 nightly)
![image](https://github.com/Kotlin/dataframe/assets/17594275/e8984982-5713-4cf8-abbc-a5a91279071b)

I did create a new YT issue to track the other rendering errors: https://youtrack.jetbrains.com/issue/KTIJ-29221/Markdown-KDoc-space-rendering-errors

Issue #605, opened 2024-02-27
Annotate non-user facing documentation interfaces with `@ExcludeFromSources`


Jolanrensen:
Ever since the doc processors was updated to [v0.3.2](https://github.com/Jolanrensen/docProcessorGradlePlugin/releases/tag/v0.3.2), we've had the `@ExcludeFromSources` annotation at our disposal.
This allows temporary doc-interfaces to be excluded from our preprocessed source files, cleaning them up a lot. It might also shrink our bytecode size, since there are fewer interfaces to be compiled.
Finally, it could help in other future applications. For instance, the `dokkaJekyll` task failed before I added `@ExcludeFromSources` to `api/all.kt`.

Jolanrensen:
Let's introduce this gradually when we make new doc interfaces or we spot a good place for it

zaleslaw:
definition of done = no empty interfaces in jar!!

PR #613, opened 2024-03-04
KDoc fixes for IntelliJ 2024.X

Jolanrensen:
Attempting to fix https://github.com/Kotlin/dataframe/issues/585 and some of https://github.com/Kotlin/dataframe/issues/605

Cannot fully fix until https://github.com/JetBrains/markdown/pull/150 is merged and inside IntelliJ

This fixes:
- Making sure all spaces in code examples are inside `` ` ` `` brackets. Otherwise, the background will not render.
- If a `` ` code block ` `` (like that) starts and ends with a space, one surrounding space will be removed. So add one extra like `` `  this  ` ``.
- Making sure that code periods in grammars, like **`.`**, are actually code and bold.
- `` `block`\n`block` `` is rendered like `` `block` `block` ``, but that again introduces a space that shouldn't be there, so fixing those.
- With the doc processor, putting all spaces inside code spans increases the change of getting `something like``this`. This can however be avoided by using `code`<code>`tags`</code> in intellij as those tags are removed anyways. Might be converted to `span` if they are ever rendered correctly.
- Same with `**`, alternating with `__`
- code in references should have `` ` `` tags (Only fixed in grammars for now, as it's invisible in IDE anyways)


Jolanrensen:
@zaleslaw @koperagen It's ready I think. Although, it might be hard to review.
I mainly applied the rules I specified in the top-comment. 
As described in the issue: https://github.com/Kotlin/dataframe/issues/585, the latest nightly of the IDE fixed some part of the rendering and the docs looks best at that version (well, actually they look better in the 2023 version, but that's not what the PR is about). There are still some space-trim discrepancies in intellij, but I just followed the GFM spec in those cases, so they should look good in the future.

zaleslaw:
@Jolanrensen what is the best way to test this? I suppose, looking at the rendering docs, isn't it?

Jolanrensen:
> @Jolanrensen what is the best way to test this? I suppose, looking at the rendering docs, isn't it?

Yep, and the best way to view those is by letting IntelliJ render them. That works best when the project is published to mavenLocal and you're viewing the source from another project

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/frameCol.kt
```diff
@@ -50,26 +50,26 @@ public interface FrameColColumnsSelectionDsl {
...
      * }
      */
     public interface Grammar {
 
-        /** [**frameCol**][ColumnsSelectionDsl.frameCol] */
+        /** [**`frameCol`**][ColumnsSelectionDsl.frameCol] */
         public interface PlainDslName
 
-        /** .[**frameCol**][ColumnsSelectionDsl.frameCol] */
+        /** __`.`__[**`frameCol`**][ColumnsSelectionDsl.frameCol] */
```
zaleslaw, line 69: How this works? What kind of formatting?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/update.kt
```diff
@@ -270,6 +277,7 @@ public fun <T, C> Update<T, C>.with(expression: UpdateExpression<T, C, C?>): Dat
     }
 
 /** [Update with][Update.with] to provide a new value for every selected cell giving its row. */
+@ExcludeFromSources
```
zaleslaw, line 280: What's happened when we exclude this from source? Or all documentation interfaces should be shipped with this annotation?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/SelectingColumns.kt
```diff
@@ -53,7 +53,8 @@ internal interface SelectingColumns {
      */
     interface OperationArg
 
-    /** {@set [OperationArg] operation} */
+    // Using <code>` notation to not create double `` when including
+    /** {@set [OperationArg] <code>`operation`</code>} */
```
zaleslaw, line 57: We need a table with all our special tricks:))

zaleslaw (approved):
I've just left a couple of comments to be familiar with some syntax

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/frameCol.kt
Jolanrensen: `__` is the same as `**` :) so in this case it makes the period bold+code style.
I switched to using `__` instead of `**` here because:
- kdocs don't like `*` at the start. It doesn't always recognize the difference between `*` indicating the KDoc or `*` for making stuff bold
- If this piece of text is copied over next to something else bold, like ``**`singleColumn`**{@include [ColumnGroupName]}``, the result would become ``**`singleColumn`****`.`**[**`frameCol`**]...`` and break markdown because of the `****`. 
Like this:
**`singleColumn`****`.`**[**`frameCol`**]()...
With dashes, instead it becomes  ``**`singleColumn`**__`.`__[**`frameCol`**]...`` which renders correctly: 
**`singleColumn`**__`.`__[**`frameCol`**]()...

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/update.kt
Jolanrensen: Yes, indeed. We should annotate all our documentation interfaces with `@ExcludeFromSources`. This will exclude them from the sources+compilation and thus shrink the bytecode (in theory). For this PR I started gradually introducing them on a few interfaces already, but far from all of them (https://github.com/Kotlin/dataframe/issues/605). There's no rush though, so we can just annotate them whenever we like.

Jolanrensen:
I bumped the docPreprocessor version to 0.3.5 for the project. 
You can now view the docs easier in the update IntelliJ plugin, found at the release: https://github.com/Jolanrensen/docProcessorGradlePlugin/releases/tag/v0.3.5

PR #617, opened 2024-03-07
then operation in pivot column selection DSL inside aggregate

koperagen:
Can be considered a feature, because pivoting inside groupBy `aggregate` is a bit more powerful. So `then` support unlocks some use cases

Additional aggregation per each group
```
val df = dataFrameOf(
    "category1" to List(12) { it % 3 },
    "category2" to List(12) { "category2_${it % 2}" },
    "category3" to List(12) { "category3_${it % 5}" },
    "value" to List(12) { it }
)

val df1 = df.groupBy("category1").aggregate {
    count() into "totalCount" // 
    pivot { "category2" then "category3" }.count() into "counts"
}
```
While this syntax wouldn't allow additional aggregation per each group
```
df.pivot { "category2" then "category3" }.groupBy("category1").count()
```


Jolanrensen:
Could you share some examples of what this will do?
(`df.toStaticHtml()` can be rendered on GH :) )

koperagen:
image

PR #618, opened 2024-03-08
Fixes in resize-iframes.js

Jolanrensen:
- `doObserveIFrame()`: fixed resized typo
- now running `resize_iframe_out()` on visibility change, so `<dataFrame>` can work inside `<tab>`
- cleaning up the file a bit

### DIFF
--- a/docs/StardustDocs/cfg/resize-iframes.js
+++ b/docs/StardustDocs/cfg/resize-iframes.js
@@ -1,4 +1,4 @@
-<meta name="google-site-verification" content="Lffz_ab-_S5cmA07ZXVbucHVklaRsnk8gEt8frHKjMk" />
+<meta name="google-site-verification" content="Lffz_ab-_S5cmA07ZXVbucHVklaRsnk8gEt8frHKjMk"/>
 <!-- Google Analytics -->
 <script>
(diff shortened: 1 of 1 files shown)


docs/StardustDocs/cfg/resize-iframes.js
```diff
@@ -46,59 +46,64 @@
...
+                let resized = false;
                 for (let mutation of mutations) {
                     // skip attribute changes except open to avoid loop
                     // (callback sees change in iframe, triggers resize, sees change...)
                     if (mutation.type === 'attributes') {
                         if (mutation.attributeName === 'open') {
-                            resize_iframe_out(el)
-                            resized = true
+                            resize_iframe_out(el);
+                            resized = true;
```
koperagen, line 58: on a second thought now `resized = true` triggers second call of `resize_iframe_out` which doesn't make much sense to me

Jolanrensen: so should we remove them altogether?

Jolanrensen: agreed, I'll also rename it to `resize` then, since it defines whether the resize will happen or not.

PR #625, opened 2024-03-15
Checking notebooks for 0.13.1

Jolanrensen:
Reran and fixed example notebooks


Jolanrensen:
in the process of fixing Datalore examples as well

koperagen:
I see changes in 40 puzzles and netflix, other are only different in outputs, apparently. Right?

koperagen (approved):

Jolanrensen:
> I see changes in 40 puzzles and netflix, other are only different in outputs, apparently. Right?

mostly yes. I also changed updated titanic with your PR

PR #636, opened 2024-03-22
Dev version bump to 0.14.0 & updating deprecations

Jolanrensen:
as the title says

### DIFF
--- a/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt
+++ b/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt
@@ -7,43 +7,6 @@ package org.jetbrains.kotlinx.dataframe.util
  * Level.ERROR -> Remove
  */
 
-// region WARNING in 0.12, ERROR in 0.13
-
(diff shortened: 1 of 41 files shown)


core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt
```diff
@@ -112,6 +75,12 @@ private const val message_0_15 = "Will be removed in 0.15."
 
 // endregion
 
+// region WARNING in 0.15, ERROR in 0.16
+
+private const val message_0_16 = "Will be removed in 0.16."
```
zaleslaw, line 80: MESSAGE_0_16? for constants (no needs to fix, but for future)

zaleslaw (approved):

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt
Jolanrensen: it's a private const so it doesn't really matter (the linter doesn't complain), but we could change it in the future if you want to

Issue #380, opened 2023-05-19
Known issues with Collection of Java classes


zaleslaw:

zaleslaw:
@belovrv could you share some details here and give some code snippets?

zaleslaw:
@belovrv

zaleslaw:
@belovrv

belovrv:
Repro: I have a Java class like this:
``` java
public class Item {
    private final int val;
    public Item(int value) {
        val = value;
    }

    int getValue() {
        return val;
    }
}
```

And following Kotlin code:
``` kotlin
class Item1(val value: Int)

data class Container(val item: Item, val item1: Item1)

fun main() {
    val arr = listOf(
        Container(Item(1), Item1(1)),
        Container(Item(2), Item1(3)),
        Container(Item(3), Item1(3))
        )
    println("${arr[0].item.value}, ${arr[0].item1.value}")

    println(arr.toDataFrame(maxDepth = 2).toString())

}
```

The output is
```
1, 1
   item       item1
 0  { } { value:1 }
 1  { } { value:3 }
 2  { } { value:3 }
```
But expected that java property will be traversed by `toDataFrame` as well.

Jolanrensen:
Java class traversing could be possible by scanning functions starting with 
"get" or "is" using reflection. This is similar to how the [Java Bean encoder from Spark](https://spark.apache.org/docs/latest/api/java/org/apache/spark/sql/Encoders.html#bean(java.lang.Class)) scans Java 'data' classes.

Not that difficult, but it requires some research. Might look into it because it does sound fun

PR #650, opened 2024-04-06
POJO toDataFrame support (and array improvements)

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/380

This PR relaxes, the `toDataFrame()` `properties()` DSL to allow for getter functions too.

If no `memberProperties` are available, it'll fall back to getter-like `memberFunctions`. 
Also, it now tries to sort by any constructor if there are multiple (common in java with a no-arg constructor).

EDIT: 
- Added that `toDataFrame { properties }` treats arrays as values using new `isArray` reflection helpers.
- Fixed rendering of Array types in output
- Added array value rendering to output

### DIFF
--- a/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
+++ b/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
@@ -13,6 +13,7 @@ import org.jetbrains.kotlinx.dataframe.impl.asList
 import org.jetbrains.kotlinx.dataframe.impl.columnName
 import org.jetbrains.kotlinx.dataframe.impl.columns.guessColumnType
 import org.jetbrains.kotlinx.dataframe.index
+import kotlin.reflect.KCallable
 import kotlin.reflect.KClass
 import kotlin.reflect.KProperty
 
@@ -115,24 +116,26 @@ public fun Iterable<Pair<String, Iterable<Any?>>>.toDataFrameFromPairs(): AnyFra
 public interface TraversePropertiesDsl {
 
     /**
-     * Skip given [classes] during recursive (dfs) traversal
+     * Skip given [classes] during recursive (dfs) traversal.
      */
     public fun exclude(vararg classes: KClass<*>)
 
     /**
-     * Skip given [properties] during recursive (dfs) traversal
+     * Skip given [properties] during recursive (dfs) traversal.
+     * These can also be getter-like functions.
      */
-    public fun exclude(vararg properties: KProperty<*>)
+    public fun exclude(vararg properties: KCallable<*>)
 
     /**
-     * Store given [classes] in ValueColumns without transformation into ColumnGroups or FrameColumns
+     * Store given [classes] in ValueColumns without transformation into ColumnGroups or FrameColumns.
      */
     public fun preserve(vararg classes: KClass<*>)
 
     /**
-     * Store given [properties] in ValueColumns without transformation into ColumnGroups or FrameColumns
+     * Store given [properties] in ValueColumns without transformation into ColumnGroups or FrameColumns.
+     * These can also be getter-like functions.
      */
-    public fun preserve(vararg properties: KProperty<*>)
+    public fun preserve(vararg properties: KCallable<*>)
 }
 
 public inline fun <reified T> TraversePropertiesDsl.preserve(): Unit = preserve(T::class)
@@ -148,7 +151,7 @@ public abstract class CreateDataFrameDsl<T> : TraversePropertiesDsl {
     public infix fun AnyBaseCol.into(path: ColumnPath): Unit = add(this, path)
 
     public abstract fun properties(
-        vararg roots: KProperty<*>,
+        vararg roots: KCallable<*>,
         maxDepth: Int = 0,
         body: (TraversePropertiesDsl.() -> Unit)? = null,
     )
--- a/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
+++ b/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
@@ -122,7 +122,7 @@ public interface TraversePropertiesDsl {
 
     /**
      * Skip given [properties] during recursive (dfs) traversal.
-     * These can also be getter-like functions.
+     * These can also be getter-like functions (like `getX()` or `isX()`).
      */
     public fun exclude(vararg properties: KCallable<*>)
 
@@ -133,7 +133,7 @@ public interface TraversePropertiesDsl {
 
     /**
      * Store given [properties] in ValueColumns without transformation into ColumnGroups or FrameColumns.
-     * These can also be getter-like functions.
+     * These can also be getter-like functions (like `getX()` or `isX()`).
      */
     public fun preserve(vararg properties: KCallable<*>)
 }
--- a/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/Utils.kt
+++ b/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/Utils.kt
@@ -17,6 +17,7 @@ import java.math.BigInteger
 import kotlin.reflect.KCallable
 import kotlin.reflect.KClass
 import kotlin.reflect.KFunction
+import kotlin.reflect.KProperty
 import kotlin.reflect.KType
 import kotlin.reflect.KTypeProjection
 import kotlin.reflect.KVariance
@@ -24,6 +25,7 @@ import kotlin.reflect.full.createType
 import kotlin.reflect.full.findAnnotation
 import kotlin.reflect.full.isSubtypeOf
 import kotlin.reflect.full.starProjectedType
+import kotlin.reflect.full.valueParameters
 import kotlin.reflect.full.withNullability
 import kotlin.reflect.jvm.jvmErasure
 import kotlin.reflect.typeOf
@@ -338,8 +340,22 @@ internal fun List<String>.joinToCamelCaseString(): String {
         .replaceFirstChar { it.lowercaseChar() }
 }
 
+internal fun KFunction<*>.isGetterLike(): Boolean =
+    (name.startsWith("get") || name.startsWith("is")) && valueParameters.isEmpty()
+
+internal fun KCallable<*>.isGetterLike(): Boolean =
+    when (this) {
+        is KProperty<*> -> true
+        is KFunction<*> -> isGetterLike()
+        else -> false
+    }
+
 @PublishedApi
-internal val <T> KCallable<T>.columnName: String
+internal val KProperty<*>.columnName: String
+    get() = findAnnotation<ColumnName>()?.name ?: name
+
+@PublishedApi
+internal val KCallable<*>.columnName: String
     get() = findAnnotation<ColumnName>()?.name
         ?: when (this) {
             // for defining the column names based on a getter-function, we use the function name minus the get/is prefix
@@ -348,6 +364,8 @@ internal val <T> KCallable<T>.columnName: String
                     .removePrefix("get")
                     .removePrefix("is")
                     .replaceFirstChar { it.lowercase() }
+
+            is KProperty<*> -> this.columnName
 
             else -> name
         }
--- a/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/Utils.kt
+++ b/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/Utils.kt
@@ -340,32 +340,61 @@ internal fun List<String>.joinToCamelCaseString(): String {
         .replaceFirstChar { it.lowercaseChar() }
 }
 
+/** Returns `true` if this callable is a getter-like function.
+ *
+ * A callable is considered getter-like if it is either a property getter,
+ * or it's a function with no (type) parameters that starts with "get"/"is". */
 internal fun KFunction<*>.isGetterLike(): Boolean =
-    (name.startsWith("get") || name.startsWith("is")) && valueParameters.isEmpty()
+    (name.startsWith("get") || name.startsWith("is")) &&
+        valueParameters.isEmpty() &&
+        typeParameters.isEmpty()
 
+/** Returns `true` if this callable is a getter-like function.
+ *
+ * A callable is considered getter-like if it is either a property getter,
+ * or it's a function with no (type) parameters that starts with "get"/"is". */
+internal fun KProperty<*>.isGetterLike(): Boolean = true
+
+/**
+ * Returns `true` if this callable is a getter-like function.
+ *
+ * A callable is considered getter-like if it is either a property getter,
+ * or it's a function with no (type) parameters that starts with "get"/"is".
+ */
 internal fun KCallable<*>.isGetterLike(): Boolean =
     when (this) {
-        is KProperty<*> -> true
+        is KProperty<*> -> isGetterLike()
         is KFunction<*> -> isGetterLike()
         else -> false
     }
 
+/** Returns the column name for this callable.
+ * If the callable contains the [ColumnName][org.jetbrains.kotlinx.dataframe.annotations.ColumnName] annotation, its [ColumnName.name][org.jetbrains.kotlinx.dataframe.annotations.ColumnName.name] is returned.
+ * Otherwise, the name of the callable is returned with proper getter-trimming iff it's a [KFunction]. */
+@PublishedApi
+internal val KFunction<*>.columnName: String
+    get() = findAnnotation<ColumnName>()?.name
+        ?: name
+            .removePrefix("get")
+            .removePrefix("is")
+            .replaceFirstChar { it.lowercase() }
+
+/** Returns the column name for this callable.
+ * If the callable contains the [ColumnName][org.jetbrains.kotlinx.dataframe.annotations.ColumnName] annotation, its [ColumnName.name][org.jetbrains.kotlinx.dataframe.annotations.ColumnName.name] is returned.
+ * Otherwise, the name of the callable is returned with proper getter-trimming iff it's a [KFunction]. */
 @PublishedApi
 internal val KProperty<*>.columnName: String
     get() = findAnnotation<ColumnName>()?.name ?: name
 
+/**
+ * Returns the column name for this callable.
+ * If the callable contains the [ColumnName] annotation, its [ColumnName.name] is returned.
+ * Otherwise, the name of the callable is returned with proper getter-trimming iff it's a [KFunction].
+ */
 @PublishedApi
 internal val KCallable<*>.columnName: String
-    get() = findAnnotation<ColumnName>()?.name
-        ?: when (this) {
-            // for defining the column names based on a getter-function, we use the function name minus the get/is prefix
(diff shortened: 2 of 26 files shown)


core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
```diff
@@ -127,18 +143,47 @@ internal fun <T> Iterable<T>.createDataFrameImpl(clazz: KClass<*>, body: CreateD
...
+    val primaryConstructorOrder = getPropertyOrderFromPrimaryConstructor(clazz)
+    val anyConstructorOrder = getPropertyOrderFromAnyConstructor(clazz)
+
+    val properties: List<KCallable<*>> = roots
+        .ifEmpty {
+            clazz.memberProperties
+                .filter { it.visibility == KVisibility.PUBLIC && it.valueParameters.isEmpty() }
+        }
+
+        // fall back to getter functions for pojo-like classes
```
koperagen, line 161: Is it possible to tell if class is Java or Kotlin? I'd prefer to have clear separation: for Kotlin only ever consider properties, and for Java take get/is

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
```diff
@@ -127,18 +143,47 @@ internal fun <T> Iterable<T>.createDataFrameImpl(clazz: KClass<*>, body: CreateD
...
+        .let {
+            val names = it.map { it.columnName }.toSet()
 
-    val properties = roots.ifEmpty {
-        clazz.memberProperties
-            .filter { it.visibility == KVisibility.PUBLIC && it.parameters.toList().size == 1 }
-    }.sortedBy { order[it.name] ?: Int.MAX_VALUE }
+            // use the primary constructor order if it's available,
+            // else try to find a constructor that matches all properties
+            val order = primaryConstructorOrder
```
koperagen, line 177: I suggest to move this and above
```
val primaryConstructorOrder = getPropertyOrderFromPrimaryConstructor(clazz)
val anyConstructorOrder = getPropertyOrderFromAnyConstructor(clazz)
```
to universal
`getPropertiesOrder`

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/Utils.kt
```diff
@@ -338,4 +339,15 @@ internal fun List<String>.joinToCamelCaseString(): String {
 }
 
 @PublishedApi
-internal val <T> KProperty<T>.columnName: String get() = findAnnotation<ColumnName>()?.name ?: name
+internal val <T> KCallable<T>.columnName: String
```
koperagen, line 342: Let's create two variants of this function: one for KProperty, one for KCallable. Having this case that is very specific to `toDataFrame` implementation in common util function that is used in many places will make it confusing

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -115,24 +116,26 @@ public fun Iterable<Pair<String, Iterable<Any?>>>.toDataFrameFromPairs(): AnyFra
...
     /**
-     * Skip given [classes] during recursive (dfs) traversal
+     * Skip given [classes] during recursive (dfs) traversal.
      */
     public fun exclude(vararg classes: KClass<*>)
 
     /**
-     * Skip given [properties] during recursive (dfs) traversal
+     * Skip given [properties] during recursive (dfs) traversal.
+     * These can also be getter-like functions.
```
koperagen, line 125: getter-like? c: i'd clarify: a function without value parameters

koperagen (changes requested):

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
koperagen: and maybe this should be checked and an exception be thrown

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
Jolanrensen: I looked for that but could not find it anywhere. That said, if we ever go multiplatform it's best to avoid references to java and just check whether it "quacks like a duck"

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
Jolanrensen: function without value parameters and starts with "is" or "get". 
We are ignoring Scala classes btw, there a getter is simply the "property()", but okay ;P

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/Utils.kt
```diff
@@ -337,5 +340,32 @@ internal fun List<String>.joinToCamelCaseString(): String {
         .replaceFirstChar { it.lowercaseChar() }
 }
 
+internal fun KFunction<*>.isGetterLike(): Boolean =
```
zaleslaw, line 343: Could you please document the idea behind what "isGetterLike" for us

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/Utils.kt
```diff
@@ -337,5 +340,32 @@ internal fun List<String>.joinToCamelCaseString(): String {
...
 @PublishedApi
-internal val <T> KProperty<T>.columnName: String get() = findAnnotation<ColumnName>()?.name ?: name
+internal val KProperty<*>.columnName: String
+    get() = findAnnotation<ColumnName>()?.name ?: name
+
+@PublishedApi
+internal val KCallable<*>.columnName: String
+    get() = findAnnotation<ColumnName>()?.name
+        ?: when (this) {
+            // for defining the column names based on a getter-function, we use the function name minus the get/is prefix
```
zaleslaw, line 361: probably we could extract this to the separate function with the clear function name (converted from this comment)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
```diff
@@ -208,18 +248,34 @@ internal fun convertToDataFrame(
...
+                    type = returnType.withNullability(nullable),
+                )
+
+            kclass == DataFrame::class && !nullable ->
+                DataColumn.createFrameColumn(
+                    name = it.columnName,
+                    groups = values as List<AnyFrame>
+                )
+
+            kclass == DataRow::class ->
```
zaleslaw, line 269: could you please add comments or extract to the separate function (with proper names) for each branch in when (sometimes the logic is unclear

zaleslaw (changes requested):
Add a little bit more work on clarification and refactoring for future generations

koperagen:
Let's add a check in isValueType for different arrays. Maybe there's a simple check? If there's no, below code from compiler for reference.
Also, functions in java class can in theory be static, we should filter them out.
Please check how Kotlin reflection sees a `List` and `Int`, `int`, `Number` returned from Java bean: it could be that they're not converted automatically to their Kotlin counterparts and our isValue or isIterable checks won't work.
Last thing that worth checking is type arguments. `Map<String, Integer>` from Java can become something like `(Mutable)Map<String!, Int!>` in Kotlin. Expected result here would be `Map<String?, Int?>`

```
private fun ConeKotlinType.isArrayType(isNullable: Boolean?): Boolean {
    return isBuiltinType(StandardClassIds.Array, isNullable) ||
            StandardClassIds.primitiveArrayTypeByElementType.values.any { isBuiltinType(it, isNullable) } ||
            StandardClassIds.unsignedArrayTypeByElementType.values.any { isBuiltinType(it, isNullable) }
}

val primitiveTypes = setOf(Boolean, Char, Byte, Short, Int, Long, Float, Double)

    val primitiveArrayTypeByElementType = primitiveTypes.associateWith { id -> id.shortClassName.primitiveArrayId() }

 val unsignedTypes = setOf(UByte, UShort, UInt, ULong)
    val unsignedArrayTypeByElementType = unsignedTypes.associateWith { id -> id.shortClassName.primitiveArrayId() }

private fun Name.primitiveArrayId() = ClassId(StandardClassIds.Array.packageFqName, Name.identifier(identifier + StandardClassIds.Array.shortClassName.identifier))
```

koperagen:
Let's filter out functions that have type arguments too
```
class B {
    fun <T> getA(): T {
        return TODO()
    }

    fun <T> getAa(): List<T> {
        return TODO()
    }
}

fun t(b: B) {
    b.getAa<Int>()
}
```

Jolanrensen:
> Let's add a check in isValueType for different arrays. Maybe there's a simple check? If there's no, below code from compiler for reference. Also, functions in java class can in theory be static, we should filter them out. Please check how Kotlin reflection sees a `List` and `Int`, `int`, `Number` returned from Java bean: it could be that they're not converted automatically to their Kotlin counterparts and our isValue or isIterable checks won't work. Last thing that worth checking is type arguments. `Map<String, Integer>` from Java can become something like `(Mutable)Map<String!, Int!>` in Kotlin. Expected result here would be `Map<String?, Int?>`
> 
> ```
> private fun ConeKotlinType.isArrayType(isNullable: Boolean?): Boolean {
>     return isBuiltinType(StandardClassIds.Array, isNullable) ||
>             StandardClassIds.primitiveArrayTypeByElementType.values.any { isBuiltinType(it, isNullable) } ||
>             StandardClassIds.unsignedArrayTypeByElementType.values.any { isBuiltinType(it, isNullable) }
> }
> 
> val primitiveTypes = setOf(Boolean, Char, Byte, Short, Int, Long, Float, Double)
> 
>     val primitiveArrayTypeByElementType = primitiveTypes.associateWith { id -> id.shortClassName.primitiveArrayId() }
> 
>  val unsignedTypes = setOf(UByte, UShort, UInt, ULong)
>     val unsignedArrayTypeByElementType = unsignedTypes.associateWith { id -> id.shortClassName.primitiveArrayId() }
> 
> private fun Name.primitiveArrayId() = ClassId(StandardClassIds.Array.packageFqName, Name.identifier(identifier + StandardClassIds.Array.shortClassName.identifier))
> ```

lucky for us, static functions aren't in `memberFunctions` so this works by default :) working on the rest...

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/Utils.kt
```diff
@@ -337,5 +340,61 @@ internal fun List<String>.joinToCamelCaseString(): String {
...
+internal fun KCallable<*>.isGetterLike(): Boolean =
+    when (this) {
+        is KProperty<*> -> isGetterLike()
+        is KFunction<*> -> isGetterLike()
+        else -> false
+    }
+
+/** Returns the column name for this callable.
+ * If the callable contains the [ColumnName][org.jetbrains.kotlinx.dataframe.annotations.ColumnName] annotation, its [ColumnName.name][org.jetbrains.kotlinx.dataframe.annotations.ColumnName.name] is returned.
+ * Otherwise, the name of the callable is returned with proper getter-trimming iff it's a [KFunction]. */
```
koperagen, line 373: Let's not introduce comments that are essentially exactly follow the code? Code is easier to read and is always factual, unlike plain text here

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
```diff
@@ -147,22 +188,23 @@ internal fun convertToDataFrame(
...
-                    ?: error("conversion of value class $kClass with multiple parameters in constructor is not yet supported")
-                // there's no need to unwrap if underlying field is nullable
-                if (parameter.type.isMarkedNullable) return@let null
-                // box and unbox impl methods are part of binary API of value classes
-                // https://youtrack.jetbrains.com/issue/KT-50518/Boxing-Unboxing-methods-for-JvmInline-value-classes-should-be-public-accessible
-                val unbox = kClass.java.getMethod("unbox-impl")
-                val box = kClass.java.methods.single { it.name == "box-impl" }
-                val valueClassConverter = ValueClassConverter(unbox, box)
-                valueClassConverter
+            if (!kClass.isValue) return@let null
```
koperagen, line 191: Is it really needed? It was reviewed and merged, not sure why it needs to be changed without actual functional changes. I don't like when PR has things like this

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/schema/Utils.kt
```diff
@@ -141,5 +144,53 @@ internal fun DataFrameSchema.createEmptyDataFrame(numberOfRows: Int): AnyFrame =
...
+                .mapIndexed { i, v -> v to i }
+                .toMap()
+        }
+
+/**
+ * Sorts [this] according to the order of them in the constructors of [klass].
+ * It prefers the primary constructor if it exists, else it falls back to the other constructors to do the sorting.
+ * Finally, it falls back to lexicographical sorting if a property does not exist in any constructor.
+ */
+internal fun <T> Iterable<KCallable<T>>.sortWithConstructors(klass: KClass<*>): List<KCallable<T>> {
```
koperagen, line 168: Honestly i still don't understand what to expect from this as a user. I'd like something simpler

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -193,6 +193,7 @@ class CreateDataFrameTests {
     fun treatErasedGenericAsAny() {
         class IncompatibleVersionErrorData<T>(val expected: T, val actual: T)
         class DeserializedContainerSource(val incompatibility: IncompatibleVersionErrorData<*>)
+
```
koperagen, line 196: can be removed

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/Utils.kt
Jolanrensen: I agree in the same file, but if you're calling the code from another place, all you see is the signature and the KDoc. I don't want to navigate to a piece of code to figure out what it does.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt
```diff
@@ -467,3 +469,103 @@ internal fun nothingType(nullable: Boolean): KType =
...
+ */
+internal val Any.isArray: Boolean
+    get() = this::class.isArray
+
+/**
+ * Returns this array object as a list of values.
+ *
+ * @throws IllegalArgumentException if this object is not an array
+ */
+internal fun Any.asArrayToList(): List<*> {
```
koperagen, line 551: A lot of new utilities here. Are all of them used?
This function can be changed so it returns `List<*>?` instead of throwing. It will combine check & conversion,  should easier to call
also `asList` (wrapper) instead of `toList` (copy)

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
Jolanrensen: Well, the complexity of this huge function is someting @zaleslaw stuggeled with and I agree. Lowering the cognitive complexity, like here, improves the codebase as a whole. And face it, we're not gonna do a refactor of everything anytime soon, so I think it's okay to do this along with PRs. I will try to do it in separate commits when I can though, so you can more easily skip it in reviews

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/Utils.kt
koperagen: ctrl+shift+i?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/schema/Utils.kt
Jolanrensen: nope :/ reflection returns them completely random.

The idea was to come as close to what a user expects the order of columns to be as possible. So if they have a weird class like

```kt
class A {
  constructor()
  constructor(foo: Int, bar: Int)
  constructor(bar: Int, too: Int)
  ...
}
```
well, you expect `foo` to be before `bar` and `too` after it, right? That's what this function achieves. It sorts the columns according to the order of all constructors. If there's a primary constructor, obviously that is used first.

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/Utils.kt
Jolanrensen: haha okay that might work, but still, I think there's no such thing as "unnecessary docs". If university programming has taught me anything is that documentation, no matter how useless it may seem to write, will help you in some way in the future.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt
Jolanrensen: Most of them are used, either in the implementation of `toDataFrame`, or in the type rendering for `print`, plus they will definitely help us in solving https://github.com/Kotlin/dataframe/issues/679 and https://github.com/Kotlin/dataframe/issues/678 (and probably even https://github.com/Kotlin/dataframe/issues/677). I thought it best to just write them now and test all cases so we have a good base to build upon and not reinvent the wheel again.

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
Jolanrensen: the linter did that, probably best to leave it like this otherwise it will keep reappearing every time we lint the file again

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt
koperagen: Ok, thanks! Good to know

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/schema/Utils.kt
Jolanrensen: Now simplified to just sorting lexicographically and by primary/single constructor

koperagen (approved):
Looks good, thank you!

Issue #635, opened 2024-03-21
Update extension properties API docs


AndreiKingsley:
I think there needs to be more explanation and examples here, topic looks seriously incomplete:
https://kotlin.github.io/dataframe/extensionpropertiesapi.html

Jolanrensen:
What else would you want to be there? Some example of what is generated perhaps?
There is already more at:
https://kotlin.github.io/dataframe/schemasgradle.html#execute-the-assemble-task-to-generate-type-safe-accessors-for-schemas

AndreiKingsley:
No, I don't want implementation details or other deep stuff. I'd like users, unfamiliar with DataFrame, to have a clearer understanding. For example, to say that "in this cell a dataframe appeared, it has these columns, and we can use them further like that".
So, in my personal opinion, there seems to be a lack of clarity regarding what happened between the lines of code.

PR #651, opened 2024-04-06
Changed documentation regarding extension properties api

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/635
more explicit explanation of what happens in between cell calls in the extension properties api in notebooks


AndreiKingsley (changes requested):
It seems to me that for very beginning users it may not be completely clear yet. I would add a DataFrame head (as table md) so the user would see its columns, then show 
1) that now df has properties with these names
2) show some simple example (select for example) with String API and say that now it can be rewritten like this with Column selection DSL.

docs/StardustDocs/topics/extensionPropertiesApi.md
```diff
@@ -28,12 +39,9 @@ df.add("lastName") { name.split(",").last() }
...
-In notebooks, extension properties are generated for [`DataSchema`](schemas.md) that is extracted from [`DataFrame`](DataFrame.md) 
-instance after REPL line execution. 
-After that [`DataFrame`](DataFrame.md)  variable is typed with its own [`DataSchema`](schemas.md), so only valid extension properties corresponding to actual columns in DataFrame will be allowed by the compiler and suggested by completion.
-
 Extension properties can be generated in IntelliJ IDEA using the [Kotlin Dataframe Gradle plugin](schemasGradle.md#configuration).
 
 <warning>
-In notebooks generated properties won't appear and be updated until the cell has been executed. It often means that you have to introduce new variable frequently to sync extension properties with actual schema
+In notebooks generated properties won't appear and be updated until the cell has been executed.
+It often means that you have to introduce new variable frequently to sync extension properties with actual schema.
```
zaleslaw, line 46: NOTE: ?

zaleslaw (approved):

docs/StardustDocs/topics/extensionPropertiesApi.md
Jolanrensen: No need, the warning tags already make it look like this:
image

AndreiKingsley (changes requested):
The paragraph assumes that the user 1) knows the "titanic" columns 2) has learned the String/CA API well. I think this is not a good approach. Instead of adding some explanations of the working mechanism, we should let the user understand what extension properties bring (and show that they are generated from columns). Imagine you make a notebook that you show to a newbie in DataFrame - first you load a table and display it, look at it, look at its columns. Then you can show that your `df` has extensional properties (i.e. write the code `df.age` and show what it will output). Then you show an example with String/Accessors API, and then rewrite it with EP API, and say that's how great it is.

Jolanrensen:
> The paragraph assumes that the user 1) knows the "titanic" columns 2) has learned the String/CA API well. I think this is not a good approach. Instead of adding some explanations of the working mechanism, we should let the user understand what extension properties bring (and show that they are generated from columns). Imagine you make a notebook that you show to a newbie in DataFrame - first you load a table and display it, look at it, look at its columns. Then you can show that your `df` has extensional properties (i.e. write the code `df.age` and show what it will output). Then you show an example with String/Accessors API, and then rewrite it with EP API, and say that's how great it is.

While that's a fantastic approach for, say, a demo, I'm not sure we can convey it like that on a static website. It might take some actual screenshots to show how generated accessors will look. Not a bad idea though, I'll think about it.

Jolanrensen:
@AndreiKingsley so more like here? https://kotlinlang.org/docs/data-analysis-work-with-data-sources.html#display-data

AndreiKingsley:
In this case there is no need to make a guide, a short chain explanation is enough: there was a dataframe with such columns -> such EPs were generated  -> they can be used instead of string/accessors and have a lot of advantages.

Jolanrensen:
I made it simpler, now I'm just sending the users to https://kotlin.github.io/dataframe/schemas.html which explains everything with examples and all :)

AndreiKingsley (approved):

Issue #562, opened 2024-01-17
The `split` operation documentation contains the example for another operations `merge`


devcrocod:
https://kotlin.github.io/dataframe/split.html#split-horizontally

<img width="753" alt="image" src="https://github.com/Kotlin/dataframe/assets/19648120/118194b7-5b2d-4067-af4c-85107716fce7">

NOTE: it should be deleted

Jolanrensen:
Also, the `limit =` argument does not exist on the site

Jolanrensen:
After investigation: it's not incorrect. The documentation just shows how the merged df is created before showing how it can be deconstructed again using Regex. That said, since users can now see the input DF, we can probably remove the first step for clarity.

PR #653, opened 2024-04-06
Updated split documentation

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/562

### DIFF
--- a/docs/StardustDocs/topics/split.md
+++ b/docs/StardustDocs/topics/split.md
@@ -2,12 +2,13 @@
 
 <!---IMPORT org.jetbrains.kotlinx.dataframe.samples.api.Modify-->
 
-Splits every value in the given columns into several values and optionally spreads them horizontally or vertically.
+This operation splits every value in the given columns into several values,
+and optionally spreads them horizontally or vertically.
 
 ```text
 df.split { columns }
     [.cast<Type>()]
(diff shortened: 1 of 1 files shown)


docs/StardustDocs/topics/split.md
```diff
@@ -131,18 +135,7 @@ df.split { "name"["lastName"]<String>() }
 <dataFrame src="org.jetbrains.kotlinx.dataframe.samples.api.Modify.split1.html"/>
 <!---END-->
 
-`String` columns can also be split into group matches of [`Regex`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/-regex/) pattern:
-
-<!---FUN splitRegex-->
-
-```kotlin
```
zaleslaw, line 138: What's happened with this example? What's its future here and in codebase?

zaleslaw (approved):

docs/StardustDocs/topics/split.md
Jolanrensen: Like I described here https://github.com/Kotlin/dataframe/issues/562, the example is not incorrect, but it show a bit too much information before getting to the point. It first merges string columns before splitting them again, while only the splitting part is relevant here. It should still exist in the code base to generate the right DFs tho.

PR #662, opened 2024-04-15
`groupBy {}.aggregate { keys }`

Jolanrensen:
Provides access to the keys of the `groupBy {}` operation in the `aggregate {}` step. Very useful when you're grouping by an `expr {}` column and want to use the value of a key column to influence how the aggregation happens.
For instance:

```kt
df.groupBy { expr { … } named "keyName" }
  .aggregate {
      keys["keyName"] into "valueName"
   }
```

`keys` will be provided as an `AnyRow`, since most it will just contain the key columns from `df`, a tiny subset. Giving it the same type as `df` would result in many breaking accessors on `keys`.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/GroupByReceiverImpl.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/GroupByReceiverImpl.kt
@@ -18,11 +18,21 @@ import org.jetbrains.kotlinx.dataframe.impl.createTypeWithArgument
 import org.jetbrains.kotlinx.dataframe.impl.getListType
 import kotlin.reflect.KType
 
-internal class GroupByReceiverImpl<T>(override val df: DataFrame<T>, override val hasGroupingKeys: Boolean) :
+internal class GroupByReceiverImpl<T>(
+    override val df: DataFrame<T>,
+    override val hasGroupingKeys: Boolean,
+    private val retrieveKey: () -> AnyRow = {
+        error("This property can only be used inside 'groupBy { }.aggregate { }' clause")
+    }
+) :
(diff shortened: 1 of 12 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/GroupByReceiverImpl.kt
```diff
@@ -18,12 +18,22 @@ import org.jetbrains.kotlinx.dataframe.impl.createTypeWithArgument
 import org.jetbrains.kotlinx.dataframe.impl.getListType
 import kotlin.reflect.KType
 
-internal class GroupByReceiverImpl<T>(override val df: DataFrame<T>, override val hasGroupingKeys: Boolean) :
+internal class GroupByReceiverImpl<T>(
+    override val df: DataFrame<T>,
+    override val hasGroupingKeys: Boolean,
+    private val retrieveKey: () -> AnyRow = {
```
koperagen, line 24: Hm, it has to be lambda here? And what about default parameter: can it somehow actually throw an exception?

koperagen: Oh yes, i remember now. aggregate probably should be an extension function on GroupBy
```
public interface GroupBy<out T, out G> : Grouped<G> {

    public val groups: FrameColumn<G>

    public val keys: DataFrame<T>
```
then keys can become `DataRow<T>`

Because now aggregate on GroupBy is resolved to this member and it simply doesn't know anything about keys
```
public interface Grouped<out T> : Aggregatable<T> {

    public fun <R> aggregate(body: AggregateGroupedBody<T, R>): DataFrame<T>
}
```

PR #663, opened 2024-04-17
Add GroupBy variable converter in Jupyter

koperagen:
This was a long standing minor inconvenience. Having this variable in a cell1 you couldn't access `k` in keys in the cell2
cell1:
```kt
val groupBy = dataFrameOf("a")("1", "11", "2", "22")
    .groupBy { expr { "a"<String>().length } named "k" }
```
cell2:
```kt
groupBy.keys.k
```

Other use case is adding a new column to `GroupBy`
```kt
val groupBy = dataFrameOf("a")("1", "11", "2", "22")
    .groupBy { expr { "a"<String>().length } named "k" }
    .add("newCol") { 42 }
```
```kt
groupBy.aggregate { newCol into "newCol" }
```

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/CodeWithConverter.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/CodeWithConverter.kt
@@ -7,14 +7,19 @@ import org.jetbrains.kotlinx.jupyter.api.VariableName
  * Class representing generated code declarations for a [Marker].
  *
  * @param declarations The generated code.
+ * @param converter Needs to provide additional info (name) from org.jetbrains.dataframe.impl.codeGen.CodeGenerator to its callers
+ * But at the same time name doesn't make sense for GroupBy where code to be executed contains two declarations
  * @param converter Optional converter for the [Marker], such as a [org.jetbrains.kotlinx.dataframe.api.cast], often used for Jupyter.
  */
-public data class CodeWithConverter(val declarations: Code, val converter: (VariableName) -> Code = EmptyConverter) {
+public data class CodeWithConverter<T : CodeConverter>(
+    val declarations: Code,
+    val converter: T
+) {
 
     public companion object {
         public const val EmptyDeclarations: Code = ""
-        public val EmptyConverter: (VariableName) -> Code = { it }
-        public val Empty: CodeWithConverter = CodeWithConverter(EmptyDeclarations, EmptyConverter)
+        public val EmptyConverter: CodeConverter = CodeConverter { it }
+        public val Empty: CodeWithConverter<CodeConverter> = CodeWithConverter(EmptyDeclarations, EmptyConverter)
     }
 
     val hasDeclarations: Boolean get() = declarations.isNotBlank()
@@ -25,5 +30,21 @@ public data class CodeWithConverter(val declarations: Code, val converter: (Vari
         !hasConverter -> declarations
         !hasDeclarations -> converter(name)
         else -> declarations + "\n" + converter(name)
+    }
+}
+
+public sealed interface CodeConverter : (VariableName) -> Code
+
+public class CodeConverterImpl(private val f: (VariableName) -> Code) : CodeConverter {
+    override fun invoke(p1: VariableName): Code {
+        return f(p1)
+    }
+}
+
+public fun CodeConverter(f: (VariableName) -> Code): CodeConverter = CodeConverterImpl(f)
+
+public class ProvidedCodeConverter(public val markerName: String) : CodeConverter {
+    override fun invoke(p1: VariableName): Code {
+        return "$p1.cast<$markerName>()"
     }
 }
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/CodeGenerationTests.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/CodeGenerationTests.kt
@@ -53,6 +53,22 @@ class CodeGenerationTests : DataFrameJupyterTest() {
     }
 
     @Test
+    fun `groupBy`() {
+        """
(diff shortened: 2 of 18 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/CodeWithConverter.kt
```diff
@@ -7,14 +7,19 @@ import org.jetbrains.kotlinx.jupyter.api.VariableName
...
  *
  * @param declarations The generated code.
+ * @param converter Needs to provide additional info (name) from org.jetbrains.dataframe.impl.codeGen.CodeGenerator to its callers
+ * But at the same time name doesn't make sense for GroupBy where code to be executed contains two declarations
  * @param converter Optional converter for the [Marker], such as a [org.jetbrains.kotlinx.dataframe.api.cast], often used for Jupyter.
  */
-public data class CodeWithConverter(val declarations: Code, val converter: (VariableName) -> Code = EmptyConverter) {
+public data class CodeWithConverter<T : CodeConverter>(
+    val declarations: Code,
+    val converter: T
```
Jolanrensen, line 16: maybe it's an idea to rename this variable or rework it a bit, because looking at this with new eyes, even with the documentation, I have no idea what's going on

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/CodeWithConverter.kt
```diff
@@ -7,14 +7,19 @@ import org.jetbrains.kotlinx.jupyter.api.VariableName
  * Class representing generated code declarations for a [Marker].
  *
  * @param declarations The generated code.
+ * @param converter Needs to provide additional info (name) from org.jetbrains.dataframe.impl.codeGen.CodeGenerator to its callers
```
Jolanrensen, line 10: there are two `@param converter`s now

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/CodeWithConverter.kt
```diff
@@ -27,3 +32,19 @@ public data class CodeWithConverter(val declarations: Code, val converter: (Vari
         else -> declarations + "\n" + converter(name)
     }
 }
+
+public sealed interface CodeConverter : (VariableName) -> Code
```
Jolanrensen, line 36: This probably needs docs to explain what it does/should do. As it's a foreign concept and essentially just a String->String, which doesn't tell much. Maybe it could steal some docs from `CodeWithConverter`

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/CodeGenerationTests.kt
```diff
@@ -52,6 +52,22 @@ class CodeGenerationTests : DataFrameJupyterTest() {
         """.checkCompilation()
     }
 
+    @Test
+    fun `groupBy`() {
+        """
+            val groupBy = dataFrameOf("a")("1", "11", "2", "22").groupBy { expr { "a"<String>().length } named "k" }
+            groupBy.keys.k
```
Jolanrensen, line 59: wait, this works in the same cell?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/CodeWithConverter.kt
```diff
@@ -7,14 +7,19 @@ import org.jetbrains.kotlinx.jupyter.api.VariableName
  * Class representing generated code declarations for a [Marker].
  *
  * @param declarations The generated code.
+ * @param converter Needs to provide additional info (name) from org.jetbrains.dataframe.impl.codeGen.CodeGenerator to its callers
+ * But at the same time name doesn't make sense for GroupBy where code to be executed contains two declarations
  * @param converter Optional converter for the [Marker], such as a [org.jetbrains.kotlinx.dataframe.api.cast], often used for Jupyter.
  */
-public data class CodeWithConverter(val declarations: Code, val converter: (VariableName) -> Code = EmptyConverter) {
+public data class CodeWithConverter<T : CodeConverter>(
```
Jolanrensen, line 14: I see most functions return a `CodeWithWithConverter<*>` and since both `CodeConverterImpl` and `ProvidedCodeConverter` are essentially just a `(VariableName) -> Code`, do we need the generic type `T`?

Jolanrensen:
I'll also check whether it influences https://github.com/Kotlin/dataframe/pull/662 or not

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/CodeWithConverter.kt
koperagen: Idk, hard to tell more than expressed in the code itself. Basically CodeWithConverter is all the information that dataframe integration needs to generate except name of the variable being updated. Variable name is provided by the caller due to separation of concerns

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/CodeWithConverter.kt
koperagen: For GroupBy i need to combine names of two generated interfaces together for cast, and so CodeWithConverter<ProvidedCodeConverter> provides this additional info. This parameter doesn't make sense for other usages, so there we use more generic CodeWithWithConverter<*>

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/CodeGenerationTests.kt
koperagen: No :) The test executes each line individually. It's like 2 separate cells here

Jolanrensen:
what needs to be done about this pr?

Jolanrensen:
Rewritten but used logic as base in https://github.com/Kotlin/dataframe/pull/1263

PR #672, opened 2024-04-22
Add nullability inference support to dataframe-jdbc

### LINKED ISSUE #541: DataColumn nullability in JDBC
I'd argue that KType nullability should always check actual column values. 
https://github.com/Kotlin/dataframe/blob/master/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt#L597
Which is done by `infer = Infer.Nulls`
My reasoning is mostly related to notebooks
Pros: you won't have to handle nullable values if given snapshot doesn't have any! Very convenient if you just want to work with specific fragment of data
Cons: Imagine you want to rerun the same notebook, but this time data has nulls. Now, you'll have to modify your code to handle it, or it will be compilation error 
So, depending on your use case: explore data once vs reuse notebook, desirable behavior can vary. 
My suggestion here: to support re-usability of notebooks, JDBC integration should have method to import data schema from DB schema the same way as `open api` support does. 

Things to consider here: it's already possible to write (or generate and edit) a data schema to rerun notebooks without problems. There're other operation that work like this: `add`, `convert` and other functions will create nullable KType only if there are nulls, as well as other data sources (discussion about this in context of Arrow: https://github.com/Kotlin/dataframe/issues/428 with additional argument about KType nullability)  
```
public inline fun <reified R, T> DataFrame<T>.add(
    name: String,
    noinline expression: AddExpression<T, R>
): DataFrame<T> = add(name, Infer.Nulls,  expression)
```

zaleslaw:
This update adds a new parameter `inferNullability` for various dataframe reading functions including `readSqlTable`, `readSqlQuery`, and `readResultSet`. It allows better control over how column nullability should be inferred. The `h2Test` file has been adjusted to test this new feature.

Fixes #541

### DIFF
--- a/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
+++ b/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
@@ -4,6 +4,7 @@ import io.github.oshai.kotlinlogging.KotlinLogging
 import org.jetbrains.kotlinx.dataframe.AnyFrame
 import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.api.Infer
 import org.jetbrains.kotlinx.dataframe.api.toDataFrame
 import org.jetbrains.kotlinx.dataframe.impl.schema.DataFrameSchemaImpl
 import org.jetbrains.kotlinx.dataframe.io.db.DbType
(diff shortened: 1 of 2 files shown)


dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -105,15 +106,18 @@ public data class DatabaseConfiguration(val url: String, val user: String = "",
...
  * @return the DataFrame containing the data from the SQL table.
  */
 public fun DataFrame.Companion.readSqlTable(
     dbConfig: DatabaseConfiguration,
     tableName: String,
-    limit: Int = DEFAULT_LIMIT
+    limit: Int = DEFAULT_LIMIT,
+    inferNullability: Infer = Infer.None,
 ): AnyFrame {
+    Infer.Nulls
```
koperagen, line 118: this line can be removed

PR #682, opened 2024-04-30
Update comments and doc to provide more details about ResultSet

### LINKED ISSUE #495: The method DataFrame.readResultSet(resultSet, PostgreSql) could not be read twice
From @Jolanrensen 

> DataFrame.readResultSet(resultSet, PostgreSql)
> cannot be called more than once! This is not expected behavior for me as dataframe calls are always functional. They do not modify the argument.
> DataFrame.readResultSet(statement.executeQuery(selectQuery), PostgreSql)
> this does work and can be repeated, as expected.

zaleslaw:
Closes #495


Jolanrensen (approved):
Very clear, thanks!

PR #692, opened 2024-05-07
Add JDBC credentials extraction from env variables and improve exception handling

### LINKED ISSUE #642: Add ability to reference environment variable in ImportDataSchema
JdbcOptions can have additional Boolean parameter that makes plugin interpret `password` (maybe username too?) parameter as a env variable name.
People won't have to fallback to Gradle plugin if they don't feel like storing these credentials in the source code 
@zaleslaw

zaleslaw:
This update introduces the ability to extract JDBC credentials from environment variables enhancing security features. Additionally, improvements have been made to the exception handling when generating schema for SQL table or result of SQL query.

Fixes #642

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/ImportDataSchema.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/ImportDataSchema.kt
@@ -49,9 +49,19 @@ public annotation class CsvOptions(
     public val delimiter: Char,
 )
 
+/**
+ * An annotation class that represents options for JDBC connection.
+ *
+ * @property [user] The username for the JDBC connection. Default value is an empty string.
+ * @property [password] The password for the JDBC connection. Default value is an empty string.
+ * @property [extractCredFromEnv] Whether to extract the JDBC credentials from environment variables. Default value is false.
+ * @property [tableName] The name of the table for the JDBC connection. Default value is an empty string.
+ * @property [sqlQuery] The SQL query to be executed in the JDBC connection. Default value is an empty string.
+ */
 public annotation class JdbcOptions(
-    public val user: String = "", // TODO: I'm not sure about the default parameters
-    public val password: String = "", // TODO: I'm not sure about the default parameters)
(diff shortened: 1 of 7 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/ImportDataSchema.kt
```diff
@@ -49,9 +49,19 @@ public annotation class CsvOptions(
     public val delimiter: Char,
 )
 
+/**
+ * An annotation class that represents options for JDBC connection.
+ *
+ * @property [user] The username for the JDBC connection. Default value is an empty string.
+ * @property [password] The password for the JDBC connection. Default value is an empty string.
+ * @property [extractCredFromEnv] Whether to extract the JDBC credentials from environment variables. Default value is false.
```
koperagen, line 57: Worth to mention that user and password will be used as keys in this case

plugins/symbol-processor/src/test/kotlin/org/jetbrains/dataframe/ksp/DataFrameJdbcSymbolProcessorTest.kt
```diff
@@ -212,6 +212,63 @@ class DataFrameJdbcSymbolProcessorTest {
...
+        val result = KspCompilationTestRunner.compile(
+            TestCompilationParameters(
+                sources = listOf(
+                    SourceFile.kotlin(
+                        "MySources.kt",
+                        """
+                @file:ImportDataSchema(
+                     "Customer",
+                     "$CONNECTION_URL",
+                     jdbcOptions = JdbcOptions("", "", extractCredFromEnv = true, tableName = "Customer")
```
koperagen, line 229: Is it expected here to use empty strings?

zaleslaw: Yes, for now, it's fine, especially for a few databases (something could be empty row)

PR #694, opened 2024-05-10
Serialize BufferedImages as base64

### LINKED ISSUE #675: Serialize BufferedImages as base64
The table UI component allows for the display of images within table cells, but the images need to be encoded as base64 strings.
TODO: Encode BufferedImages as base64 strings when serializing the dataframe for display in the Kotlin Notebook plugin.

ermolenkodev:
Fixes #675


core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
```diff
@@ -108,10 +116,45 @@ internal fun KlaxonJson.encodeValue(col: AnyCol, index: Int): Any? = when {
         } else v
     }
 
+    col.typeClass == BufferedImage::class -> col[index]?.let { image ->
```
koperagen, line 119: Hm, at first i didn't notice that given `ImageEncodingOptions(encodeAsBase64 = false)` BufferedImage will be serialized as `toString`. Maybe lift that condition up and, for example, if ImageEncodingOptions == null then call toString and otherwise proceed with encoding?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
```diff
@@ -203,6 +209,19 @@ internal class Integration(
             }
         }
 
+        notebook.renderersProcessor.registerWithoutOptimizing(
```
koperagen, line 212: It where a way to override context renderer only in JupyterHtmlRenderer? 
As i see it users will start to see images resized to 100 px if they use dataframe even outside HTML table

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
```diff
@@ -1077,4 +1086,141 @@ class JsonTests {
...
+
+        for (i in images.indices) {
+            val row = (json[KOTLIN_DATAFRAME] as JsonArray<*>)[i] as JsonObject
+            val img = row["imgs"] as String
+
+            img shouldContain "BufferedImage"
+        }
+    }
+
+    private fun testSerializeImagesAsBase64(encodingOptions: ImageEncodingOptions? = null) {
```
koperagen, line 1137: This test is quite abstract, somehow hard to tell what it tests given different options

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/image.kt
```diff
@@ -0,0 +1,60 @@
...
+import java.awt.Graphics2D
+import java.awt.RenderingHints
+import java.awt.image.BufferedImage
+import java.awt.image.ImageObserver
+import java.io.ByteArrayOutputStream
+import javax.imageio.ImageIO
+import kotlin.math.max
+import kotlin.math.min
+
+internal fun BufferedImage.resizeKeepingAspectRatio(
```
koperagen, line 12: Is this a new implementation, or maybe from kotlin dl?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
ermolenkodev: refactored a bit

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
ermolenkodev: I fixed that

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
ermolenkodev: agree, I fixed that

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/image.kt
ermolenkodev: No, it’s not from Kotlindl; it’s simply a wrapper around the Java API)

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ImageSerializationTests.kt
```diff
@@ -0,0 +1,174 @@
...
+
+@RunWith(Parameterized::class)
+class ImageSerializationTests(private val encodingOptions: Base64ImageEncodingOptions?) {
+    @Test
+    fun `serialize images as base64`() {
+        val images = readImagesFromResources()
+        val json = encodeImagesAsJson(images, encodingOptions)
+
+        if (encodingOptions == DISABLED) {
+            checkImagesEncodedAsStrings(json, images.size)
```
koperagen, line 34: `checkImagesEncodedAsToString` for clarify

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ImageSerializationTests.kt
Jolanrensen, line 1: @ermolenkodev wow these tests are very heavy! It takes my laptop about 2 full minutes to run just these ones.
image

Could you make the images like waaay smaller? Like 0.5KB or smaller instead of 2-4MB?

PR #695, opened 2024-05-14
Introduce OtherSamples.kt to generate tables for writerside docs outside Korro

koperagen:
Later, generated tables can be added anywhere. Example:
```
<dataFrame src="manual/extensionPropertiesApi1.html"/>
```

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -136,7 +136,8 @@ val clearSamplesOutputs by tasks.creating {
 
     doFirst {
         delete {
-            delete(fileTree(File(projectDir, "../docs/StardustDocs/snippets")))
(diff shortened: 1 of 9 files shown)


Jolanrensen (approved):
Thank you! Just what I needed for https://github.com/Kotlin/dataframe/issues/635 :) 
And indeed, for flexibility it's better to have the writerside styles separately like now, who knows what else we'll need it for.

Jolanrensen:
I merged it to try it out with https://github.com/Kotlin/dataframe/issues/635, and it seems to work great, aside from the scroll bars that are still visible. This might be due to resize-iframes.js, but I'll investigate it a little myself.
image

Issue #697, opened 2024-05-16
Generate sitemap.xml automatically


Jolanrensen:
While before, a temporary solution was needed (https://github.com/Kotlin/dataframe/pull/431), according to https://youtrack.jetbrains.com/issue/WRS-2357/Automatically-generate-a-sitemap we can now generate a sitemap using WriterSide automatically. This should be enabled.

PR #698, opened 2024-05-16
Enable sitemap.xml

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/697

enables sitemap according to https://www.jetbrains.com/help/writerside/generate-sitemap.html


docs/StardustDocs/cfg/buildprofiles.xml
```diff
@@ -17,6 +17,7 @@
         <enable-browser-edits>true</enable-browser-edits>
         <browser-edits-url>https://github.com/Kotlin/dataframe/edit/master/docs/StardustDocs/</browser-edits-url>
     </variables>
+    <sitemap priority="0.35" change-frequency="monthly"/>
```
zaleslaw, line 20: When do you define the priority number, how do you generate these numbers?

zaleslaw (approved):

zaleslaw:
@Jolanrensen thanks for carrying about this question

docs/StardustDocs/cfg/buildprofiles.xml
Jolanrensen: Nope, just what the docs says https://www.jetbrains.com/help/writerside/generate-sitemap.html. To be honest, I'm not sure what it does.

Jolanrensen: I could probably just leave out the priority, since it's optional

Jolanrensen:
We need to check with WriterSide that our version gets updated so the sitemap generation feature works. It probably won't hurt to merge this already, waiting for the update, but we need to take into account it won't work right away.

PR #706, opened 2024-05-27
KTNB-693 Send the full dataframe schema as metadata

cmelchior:
This part adds the infrastructure needed for https://youtrack.jetbrains.com/issue/KTNB-693/Enable-AI-Actions-for-DataFrames-in-Kotlin-Notebooks as we currently are not able to detect column types in a good way which is needed when creating prompts for the AI Assistant.

It adds a new "types" property to the top-level "metadata" as well as recursively on each row so it is possible to easily identify column types.

A `columns` property has also been added to `ColumnGroup` and `FrameColumn` metadata, it contains nested column names similar to the top-level `columns` property.

Example:

```
val col1 by columnOf("a", "b", "c")
val col2 by columnOf(1, 2, 3)
val col3 by columnOf("Foo", "Bar", null)
val df2 = dataFrameOf(Pair("header", listOf("A", "B", "C")))
val col4 by columnOf(df2, df2, df2)
var df = dataFrameOf(col1, col2, col3, col4)
df.group(col1, col2).into("group")            
```

```
{
   ...
             {
              "${'$'}version": "2.1.0",
              "metadata": {
                "columns": ["group", "col3", "col4"],
                "types": [{
                  "kind": "ColumnGroup"
                }, {
                  "kind": "ValueColumn",
                  "type": "kotlin.String?"
                }, {
                  "kind": "FrameColumn"
                }],
                "nrow": 3,
                "ncol": 3
              },
              "kotlin_dataframe": [{
                "group": {
                  "data": {
                    "col1": "a",
                    "col2": 1
                  },
                  "metadata": {
                    "kind": "ColumnGroup",
                    "columns": ["col1", "col2"],
                    "types": [{
                      "kind": "ValueColumn",
                      "type": "kotlin.String"
                    }, {
                      "kind": "ValueColumn",
                      "type": "kotlin.Int"
                    }]
                  }
                },
                "col3": "Foo",
                "col4": {
                  "data": [{
                    "header": "A"
                  }, {
                    "header": "B"
                  }, {
                    "header": "C"
                  }],
                  "metadata": {
                    "kind": "FrameColumn",
                    "columns": ["header"],
                    "types": [{
                      "kind": "ValueColumn",
                      "type": "kotlin.String"
                    }],
                    "ncol": 1,
                    "nrow": 3
                  }
                }
              }, {
                "group": {
                  "data": {
                    "col1": "b",
                    "col2": 2
                  },
                  "metadata": {
                    "kind": "ColumnGroup",
                    "columns": ["col1", "col2"],
                    "types": [{
                      "kind": "ValueColumn",
                      "type": "kotlin.String"
                    }, {
                      "kind": "ValueColumn",
                      "type": "kotlin.Int"
                    }]
                  }
                },
                "col3": "Bar",
                "col4": {
                  "data": [{
                    "header": "A"
                  }, {
                    "header": "B"
                  }, {
                    "header": "C"
                  }],
                  "metadata": {
                    "kind": "FrameColumn",
                    "columns": ["header"],
                    "types": [{
                      "kind": "ValueColumn",
                      "type": "kotlin.String"
                    }],
                    "ncol": 1,
                    "nrow": 3
                  }
                }
              }, {
                "group": {
                  "data": {
                    "col1": "c",
                    "col2": 3
                  },
                  "metadata": {
                    "kind": "ColumnGroup",
                    "columns": ["col1", "col2"],
                    "types": [{
                      "kind": "ValueColumn",
                      "type": "kotlin.String"
                    }, {
                      "kind": "ValueColumn",
                      "type": "kotlin.Int"
                    }]
                  }
                },
                "col3": null,
                "col4": {
                  "data": [{
                    "header": "A"
                  }, {
                    "header": "B"
                  }, {
                    "header": "C"
                  }],
                  "metadata": {
                    "kind": "FrameColumn",
                    "columns": ["header"],
                    "types": [{
                      "kind": "ValueColumn",
                      "type": "kotlin.String"
                    }],
                    "ncol": 1,
                    "nrow": 3
                  }
                }
              }]
            }
}
```

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
@@ -16,13 +16,11 @@ import org.jetbrains.kotlinx.dataframe.impl.io.SerializationKeys.KOTLIN_DATAFRAM
 import org.jetbrains.kotlinx.dataframe.impl.io.SerializationKeys.METADATA
 import org.jetbrains.kotlinx.dataframe.impl.io.SerializationKeys.NCOL
 import org.jetbrains.kotlinx.dataframe.impl.io.SerializationKeys.NROW
-import org.jetbrains.kotlinx.dataframe.impl.io.SerializationKeys.SCHEMA
+import org.jetbrains.kotlinx.dataframe.impl.io.SerializationKeys.TYPES
 import org.jetbrains.kotlinx.dataframe.impl.io.SerializationKeys.VERSION
 import org.jetbrains.kotlinx.dataframe.io.Base64ImageEncodingOptions
 import org.jetbrains.kotlinx.dataframe.io.arrayColumnName
 import org.jetbrains.kotlinx.dataframe.io.valueColumnName
-import org.jetbrains.kotlinx.dataframe.ncol
-import org.jetbrains.kotlinx.dataframe.nrow
 import org.jetbrains.kotlinx.dataframe.schema.ColumnSchema
 import org.jetbrains.kotlinx.dataframe.schema.DataFrameSchema
 import java.awt.image.BufferedImage
@@ -49,7 +47,7 @@ internal object SerializationKeys {
     const val VERSION = "\$version"
     const val COLUMNS = "columns"
     const val KOTLIN_DATAFRAME = "kotlin_dataframe"
-    const val SCHEMA = "schema"
+    const val TYPES = "types"
 }
 
 /**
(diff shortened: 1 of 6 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
```diff
@@ -3,15 +3,9 @@ package org.jetbrains.kotlinx.dataframe.impl.io
 import com.beust.klaxon.JsonArray
 import com.beust.klaxon.JsonObject
 import com.beust.klaxon.KlaxonJson
-import org.jetbrains.kotlinx.dataframe.AnyCol
-import org.jetbrains.kotlinx.dataframe.AnyFrame
-import org.jetbrains.kotlinx.dataframe.ColumnsContainer
-import org.jetbrains.kotlinx.dataframe.DataColumn
-import org.jetbrains.kotlinx.dataframe.api.indices
-import org.jetbrains.kotlinx.dataframe.api.isList
+import org.jetbrains.kotlinx.dataframe.*
```
Jolanrensen, line 6: DF doesn't use *-imports in the project. Yes, we need to fix the linter to give this advice :)

Jolanrensen (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
```diff
@@ -267,3 +274,29 @@ internal fun KlaxonJson.encodeDataFrameWithMetadata(
...
+ * - data frames: `{name: "<columnName>", kind: "FrameColumn", dataframe: [<InnerDataFrameSchema>,...] }`
+ */
+internal fun DataFrameSchema.toJson(): MutableList<MutableMap<String, Any?>> {
+    val list: MutableList<MutableMap<String, Any?>> = mutableListOf()
+    columns.forEach { (name: String, columnSchema: ColumnSchema) ->
+        val schemaData = mutableMapOf<String, Any?>()
+        schemaData["name"] = name
+        schemaData["kind"] = columnSchema.kind.toString()
+        when (columnSchema) {
+            is ColumnSchema.Value -> schemaData["type"] = columnSchema.type.toString()
```
koperagen, line 294: We also have a function to turn KType to String, it's used in HTML rendering and DataFrameSchema.toString

https://github.com/Kotlin/dataframe/blob/6be5e1ac9a5b2bac84fed690740adc1ebcc1d22c/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/Rendering.kt#L51

What do you think? Suits for AI actions?

ermolenkodev:
There is a problem when a FrameColumn contains frames with different schemas. I recommend attaching types to the metadata of each nested frame. This may lead to duplication if the schema of each nested frame is the same, but it will make it easier to work with on the Kotlin Notebook plugin side. We already have a lot of duplication because we pass column names for each value in rows, so this additional overhead will be minimal.
Here is the short reproducer of the problem:
`dataFrameOf("a", "b")(1, dataFrameOf("c", "d")(1,  2), 2, dataFrameOf("e", "f")(1, 2))
`

ermolenkodev (changes requested):
https://github.com/Kotlin/dataframe/pull/706#issuecomment-2133657186

cmelchior:
@ermolenkodev I see your point. I forgot to think about that each row could hold different schemas for data frame references. So you are right, it is probably better to have the schema as part of the metadata inside the data frame content.

I'll refactor it.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
cmelchior: Looking into that method, it seems to hide some of the type information in some cases. So I do not think it is suitable for AI Actions. At least if we want to be as specific as possible with the context information. Also, if others want to use the type information, we probably need the fully qualified type as well

cmelchior: Looking into this method, it seems to remove type information in some cases, which might make it problematic when we want to use the metadata to improve the context of AI actions. So for now I would prefer to keep the fully qualified names that also include nullability, e.g. `Kotlin.String?`

ermolenkodev (approved):

Issue #570, opened 2024-01-25
Documentation needed for Columns Selection DSL


Jolanrensen:
Aside from merging https://github.com/Kotlin/dataframe/pull/372 we also need to update the columns selection page on the website.
Need to look into getting the DSL grammar parts copied over neatly as well.

Jolanrensen:
We also probably need some more aside from the grammar:

- Big explanation for 3/4 function
  - (We could also take their grammar if we'd want to)
- Small mention of the other functions (probably grouped) with small explanation (more than just a comment in the examples)

PR #619, opened 2024-03-12
Column Selector Grammar in docs

Jolanrensen:
Works on https://github.com/Kotlin/dataframe/issues/570

Started the process for writing docs for the columns selection DSL. Since [the latest doc preprocessor update](https://github.com/Jolanrensen/docProcessorGradlePlugin/releases/tag/v0.3.4) we are able to `@ExportAsHtml` any KDoc and this PR aims to export the DSL Grammar from the ColumnsSelectionDSL directly on Writerside.

~This makes it dependent on previous KDoc fixes in https://github.com/Kotlin/dataframe/pull/613~

Aside from those general KDoc fixes, this PR removes some nbsp from type definitions (as those don't align in generated HTML), and it splits the ColumnsSelection DSL grammar in separate interfaces so they can be imported into writerside separately and put in a table (fixed by https://github.com/Kotlin/dataframe/pull/618).

PR #711, opened 2024-05-31
Column selector docs function descriptions

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/570

Must be merged/reviewed after https://github.com/Kotlin/dataframe/pull/619


Jolanrensen:
image

I created a overview for each function in the DSL with explanation and some examples where needed.
I was worried the page got too large with this much information, but making the header collapsible pretty much solves that :)

Issue #30, opened 2021-06-28
Add primitive arrays column wrappers


altavir:
Primitive array columns are required for optimized big-data applications. It is also possible to add numerical DataFrame integration with MultiK or KMath.

nikitinas:
My current idea is to use Arrow as a backend for primitive types.
https://github.com/Kotlin/dataframe/issues/78

altavir:
It is a great idea, but it will be worth it only in terms of interop with other platforms. For JVM-only, Arrow will give nothing new.

nikitinas:
Arrow should give significant performance increase for JVM due to nullable values types support. Current implementation generates quite a lot of boxing/unboxing. It can be solved without Arrow, but I expect Arrow implementation to be faster. We will do performance benchmarks before implementation.

And we need to support Arrow I/O anyway.

Jolanrensen:
I was experimenting with `asList()` wrappers. Maybe this could solve this long-standing issue:
https://github.com/Kotlin/dataframe/compare/master...primitive-array-value-columns
but it needs more testing.

daniel-rusu:
I recommend considering [Immutable Arrays](https://github.com/daniel-rusu/pods4k/tree/main/immutable-arrays) as they have some surprising memory and performance benefits.  Since they are immutable and their operations produce Immutable Arrays to maintain safety, this enables many optimizations that make them even faster and more efficient than regular primitive arrays.

PR #712, opened 2024-06-03
Research: `ColumnDataHolder`/primitive arrays

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/30, one of our oldest issues.

I introduced `ColumnDataHolder` to replace the `List` in `DataColumnImpl`. This interface can define how the data of columns is stored. `ColumnDataHolderImpl` was created as default implementation and it defaults to store data in primitive arrays whenever possible. Other implementations might be possible in the future as well (to make DF act on top of an existing DB for instance).

Things to be done:

- [ ] Let data sources create `ColumnDataHolder`s directly wherever possible instead of `List`s.
- [ ] Fix cases where `DataColumnImpl.type` mismatches `DataColumnImpl.values`
- [ ] Test performance/memory differences
- [ ] Improve API

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/DataColumnImpl.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/DataColumnImpl.kt
@@ -1,27 +1,29 @@
 package org.jetbrains.kotlinx.dataframe.impl.columns
 
+import org.jetbrains.kotlinx.dataframe.ColumnDataHolder
 import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.api.dataFrameOf
 import kotlin.reflect.KType
 
 internal abstract class DataColumnImpl<T>(
-    protected val values: List<T>,
+    protected val values: ColumnDataHolder<T>,
     val name: String,
-    val type: KType,
-    distinct: Lazy<Set<T>>? = null
+    val type: KType
 ) : DataColumn<T>, DataColumnInternal<T> {
 
-    protected val distinct = distinct ?: lazy { values.toSet() }
+    protected val distinct
+        get() = values.distinct
 
     override fun name() = name
 
-    override fun values() = values
+    override fun values(): List<T> = values.toList()
 
     override fun type() = type
 
     override fun toSet() = distinct.value
 
-    override fun contains(value: T): Boolean = if (distinct.isInitialized()) distinct.value.contains(value) else values.contains(value)
+    override fun contains(value: T): Boolean =
+        if (distinct.isInitialized()) distinct.value.contains(value) else values.contains(value)
 
     override fun toString() = dataFrameOf(this).toString() // "${name()}: $type"
 
@@ -37,7 +39,8 @@ internal abstract class DataColumnImpl<T>(
 
     override fun hashCode() = hashCode
 
-    override operator fun get(range: IntRange) = createWithValues(values.subList(range.first, range.last + 1))
+    override operator fun get(range: IntRange) =
+        createWithValues(values[range])
 
     protected abstract fun createWithValues(values: List<T>, hasNulls: Boolean? = null): DataColumn<T>
 }
--- a/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/ColumnDataHolder.kt
+++ b/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/ColumnDataHolder.kt
@@ -1,0 +1,72 @@
+package org.jetbrains.kotlinx.dataframe
+
+import org.jetbrains.kotlinx.dataframe.impl.columns.ColumnDataHolderImpl
+import kotlin.reflect.KType
+import kotlin.reflect.typeOf
+
+public interface ColumnDataHolder<T> : Iterable<T> {
+
+    public val size: Int
+
+    public fun toSet(): Set<T>
+
+    public fun toList(): List<T>
+
+    public fun contains(value: T): Boolean
+
+    public operator fun get(index: Int): T
+
+    public operator fun get(range: IntRange): List<T>
+
+    public val distinct: Lazy<Set<T>>
+
+    public companion object
+}
+
+public fun <T> Collection<T>.toColumnDataHolder(type: KType, distinct: Lazy<Set<T>>? = null): ColumnDataHolder<T> =
+    ColumnDataHolderImpl.of(this, type, distinct)
+
+public inline fun <reified T> Collection<T>.toColumnDataHolder(distinct: Lazy<Set<T>>? = null): ColumnDataHolder<T> =
+    this.toColumnDataHolder(typeOf<T>(), distinct)
+
+public fun <T> Array<T>.toColumnDataHolder(type: KType, distinct: Lazy<Set<T>>? = null): ColumnDataHolder<T> =
+    ColumnDataHolderImpl.of(this, type, distinct)
+
+public inline fun <reified T> Array<T>.toColumnDataHolder(distinct: Lazy<Set<T>>? = null): ColumnDataHolder<T> =
+    this.toColumnDataHolder(typeOf<T>(), distinct)
+
+public fun BooleanArray.asColumnDataHolder(distinct: Lazy<Set<Boolean>>? = null): ColumnDataHolder<Boolean> =
+    ColumnDataHolderImpl.of(this, typeOf<Boolean>(), distinct)
+
+public fun ByteArray.asColumnDataHolder(distinct: Lazy<Set<Byte>>? = null): ColumnDataHolder<Byte> =
+    ColumnDataHolderImpl.of(this, typeOf<Byte>(), distinct)
+
+public fun ShortArray.asColumnDataHolder(distinct: Lazy<Set<Short>>? = null): ColumnDataHolder<Short> =
+    ColumnDataHolderImpl.of(this, typeOf<Short>(), distinct)
+
+public fun IntArray.asColumnDataHolder(distinct: Lazy<Set<Int>>? = null): ColumnDataHolder<Int> =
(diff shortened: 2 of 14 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/DataColumnImpl.kt
```diff
@@ -1,27 +1,55 @@
...
+            this == null -> type.isMarkedNullable
+            this.isPrimitiveArray -> type.isPrimitiveArray &&
+                this!!::class.qualifiedName == type.classifier?.let { (it as KClass<*>).qualifiedName }
+
+            this.isArray -> type.isArray // cannot check the precise type of array
+            else -> this!!::class.isSubclassOf(type.classifier as KClass<*>)
+        }
+
+    init {
+        if (DEBUG) {
```
zaleslaw, line 32: 4e4e008a64d0a6fe0100b96eabdd48cd
We are close to the Android macroses!

zaleslaw:
Very intersting idea and performance step forward, I suggest to start with a synteticly generated DataFrame with 1-10 columns with Ints, Longs, or something (better with the same type) and measure the average time/memory footprint of some performant operations before deep implementation.

I see how we could economy on memory, but not sure about speed on operations.

Also interesting to compare some non-default implementations as Multik or DirectByteBuffers

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/DataColumnImpl.kt
Jolanrensen: hahah, yeah it was just for research :)

Jolanrensen:
I've been doing some small performance/size tests [here](https://github.com/Kotlin/dataframe/blob/bfaaec825fd3eb46d47b2b287328db33d0b9ddc0/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/columns/ColumnDataHolder.kt) and I've come to the following results: 

```
⌌-------------------------------------------------------------------⌍
|  |                  type|   creation|  processing|            size|
|--|----------------------|-----------|------------|----------------|
| 0| BOXED_ARRAY_WITH_NULL| 1.668690ms| 40.072489ms| 14500481.813333|
| 1|        LIST_WITH_NULL| 9.142612ms| 41.064332ms| 14509001.813333|
| 2|                  LIST| 2.710987ms| 42.268814ms| 11496455.760000|
| 3|           BOXED_ARRAY| 2.415740ms| 42.270087ms| 11502541.520000|
| 4|          DOUBLE_ARRAY| 1.840757ms| 42.354001ms| 11499172.666667|
⌎-------------------------------------------------------------------⌏
```
In terms of creation, the two fastest are `Array<Double?>` and `DoubleArray`.
It makes sense `Array<Double>` and `List<Double>` are slower in this setup, as they're converted to `DoubleArray` in the `ColumnDataHolder`. Why `List<Double?>` is thát much slower, I honestly don't know, it's not converted to anything, just another list.

Processing speed is interesting, the arrays are the slowest of the bunch. I assume this is because they are stored as primitive arrays and need to box all their values before being able to calculate anything with them. This can also be seen in the `null`-containing columns, which are processed the quickest.

In terms of size, however, the world is flipped. All dataframes contain 3 columns of 50k rows. The ones that use primitive arrays to store their doubles hover around 11.5MB for the entire dataframe, while the others use about 14.5MB. This difference will likely increase when having more rows.

I'll run the same test on the master-branch in a minute...

Jolanrensen:
I've updated `ColumnDataHolder` in such a way that it can store a collection/array of nullable primitives:
- nulls are filled with `0` (or the zero-equivalent of that type)
- indices of where the nulls were are stored in an `IntArray`
- the null-filled data is stored in a primitive array
- `ColumnDataHolder` implements `List` and the right values are returned at the correct indices (needs some more tests)

### 10M rows
```
⌌-------------------------------------------------------------------⌍
|  |                  type|     creation|    processing|        size|
|--|----------------------|-------------|--------------|------------|
| 0|          DOUBLE_ARRAY| 452.407441ms| 10.661419664s| 250,303,579.2|
| 1| BOXED_ARRAY_WITH_NULL| 1.246602198s| 10.876937912s| 250,303,867.2|
| 2|                  LIST| 1.075708642s| 10.987466189s| 250,303,651.2|
| 3|           BOXED_ARRAY| 1.109656324s| 11.206449292s| 250,308,171.2|
| 4|        LIST_WITH_NULL| 1.878721075s| 11.211828024s| 250,294,786.4|
⌎-------------------------------------------------------------------⌏
```

Jolanrensen:
The previous tests only contained a single `null` value, which seems a bit wrong. I now have tests where half of the values are null in the `BOXED_ARRAY_WITH_NULL` and `LIST_WITH_NULL` tests.

I also experimented using a boolean array to store null-indices instead of int arrays. The results are the following:

### 10M rows, lots of nulls, boolean array
```
⌌-------------------------------------------------------------------⌍
|  |                  type|     creation|    processing|        size|
|--|----------------------|-------------|--------------|------------|
| 0|        LIST_WITH_NULL| 1.896135155s| 12.906753380s| 280,393,248.0|
| 1| BOXED_ARRAY_WITH_NULL| 1.622306469s| 13.093053168s| 280,320,763.2|
| 2|          DOUBLE_ARRAY| 535.327248ms| 13.494416201s| 280,330,497.6|
| 3|                  LIST| 1.395451763s| 13.647339781s| 280,372,962.4|
| 4|           BOXED_ARRAY| 1.240805238s| 14.096025326s| 280,339,035.2|
⌎-------------------------------------------------------------------⌏
```
### 10M rows, lots of nulls, int array
```
⌌-------------------------------------------------------------------⌍
|  |                  type|     creation|    processing|        size|
|--|----------------------|-------------|--------------|------------|
| 0|          DOUBLE_ARRAY| 472.084569ms| 13.341951593s| 250,313,040.0|
| 1|                  LIST| 1.395223809s| 13.447386786s| 250,312,961.6|
| 2| BOXED_ARRAY_WITH_NULL| 1.672050297s| 13.528234068s| 310,318,894.4|
| 3|           BOXED_ARRAY| 1.379209011s| 13.646054496s| 250,312,883.2|
| 4|        LIST_WITH_NULL| 2.950703003s| 14.230182141s| 310,293,660.8|
⌎-------------------------------------------------------------------⌏
```

As we can see, the performance differences are negligible but the difference in size is not.

Using boolean arrays results in the same size for every case, using int arrays has the size dependent on the number of nulls. While the number of nulls in every day cases is quite low, I think we should prefer the latter.

Jolanrensen:
I've furthered my investigations. I wanted to see whether it would make a difference if the ColumnDataHolder could also be used to collect data that will make up new columns (so what (Typed)ColumnDataCollector does).
This required migrating to the primitive ArrayLists of [fastutil](https://github.com/vigna/fastutil/tree/master): 

I made a [wrapper around all primitive arraylists](https://github.com/Kotlin/dataframe/blob/63d5e0a636e12edfb718271290dfb5709867e0a5/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/PrimitiveArrayList.kt) (which, yes (un)boxes the values it works with, but stores them primitively at least) that can be used as a normal mutable list. If given no type, the first element entered will dictate it. Null values are not allowed.

[ColumnDataHolderImpl](https://github.com/Kotlin/dataframe/blob/63d5e0a636e12edfb718271290dfb5709867e0a5/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ColumnDataHolderImpl.kt) now takes this `PrimitiveArrayList` wrapper as its base list, but it can fall back to a normal `ArrayList`/`MutableList` when needed. The `strictTypes` argument can be set to `true` to prevent this fallback to happen, but usually it's allowed.
If the primitive arraylist is in use, `nullIndices` keeps track of the null values in the data and `zeroValue` will be put in the primitive arraylist instead.

[ColumnDataCollector](https://github.com/Kotlin/dataframe/blob/63d5e0a636e12edfb718271290dfb5709867e0a5/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/ColumnDataCollector.kt#L68) now uses `ColumnDataHolder` to collect its values. When the result of this is turned into a column, no conversion is necessary, since it's a `ColumnDataHolder` already :).

And now for the test results:

### 5M rows, mutable ColumnDataHolder, also in Column Data Collector
```
⌌-------------------------------------------------------------------⌍
|  |                  type|     creation|    processing|        size|
|--|----------------------|-------------|--------------|------------|
| 0|             COLLECTOR| 5.811827943s|  2.642763677s| 379,996,177|
| 1|        LIST_WITH_NULL| 3.856582576s|  3.515039598s| 372,159,419|
| 2| BOXED_ARRAY_WITH_NULL| 4.368591388s|  4.029615844s| 372,119,758|
| 3|         NON_PRIMITIVE| 4.674218094s|  4.254802090s| 492,129,366|
| 4|                  LIST| 1.026079164s|  9.910889437s| 132,255,918|
| 5|          DOUBLE_ARRAY| 393.596507ms| 10.135781208s| 132,201,040|
| 6|           BOXED_ARRAY| 652.903191ms| 10.343283330s| 132,197,921|
⌎-------------------------------------------------------------------⌏
```
and on master:
```
⌌-----------------------------------------------------------⌍
|  |           type|     creation|   processing|        size|
|--|---------------|-------------|-------------|------------|
| 0| LIST_WITH_NULL| 907.783588ms| 1.995115957s| 250,323,646|
| 1|      COLLECTOR| 2.318079639s| 2.004851234s| 254,182,020|
| 2|  NON_PRIMITIVE| 620.808474ms| 3.236005222s| 250,482,091|
| 3|           LIST| 645.147782ms| 8.784381588s| 430,039,760|
⌎-----------------------------------------------------------⌏
```

Let's look at the sizes first, as that result is the easier to explain. Using `CDH`, we get the smallest sizes for non-null values: 132MB versus 250MB up-to 430MB. This does come with a performance hit

WIP.... TODO...

PR #718, opened 2024-06-04
Update SQL all table/schemas reading functions to return maps with table names

### LINKED ISSUE #671: Extract from SQL tables names of the initial tables where it's possible
It could be done somehow for the 

- [ ] readSQLTable
- [ ] readAllSQLTables

zaleslaw:
Fixes #671 

Refactored the SQL table reading functions to return map data structures instead of list. This change helps to easily correlate each dataframe with its underlying table. Additionally, the function comments and test cases were updated to match this change.

### DIFF
--- a/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
+++ b/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
@@ -290,33 +290,35 @@ public fun DataFrame.Companion.readResultSet(
 }
 
 /**
- * Reads all tables from the given database using the provided database configuration and limit.
+ * Reads all non-system tables from a database and returns them
+ * as a map of SQL tables and corresponding dataframes using the provided database configuration and limit.
  *
  * @param [dbConfig] the database configuration to connect to the database, including URL, user, and password.
  * @param [limit] the maximum number of rows to read from each table.
  * @param [catalogue] a name of the catalog from which tables will be retrieved. A null value retrieves tables from all catalogs.
  * @param [inferNullability] indicates how the column nullability should be inferred.
- * @return a list of [AnyFrame] objects representing the non-system tables from the database.
+ * @return a map of [String] to [AnyFrame] objects representing the non-system tables from the database.
  */
 public fun DataFrame.Companion.readAllSqlTables(
     dbConfig: DatabaseConfiguration,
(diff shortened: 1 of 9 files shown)


dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -325,20 +327,20 @@ public fun DataFrame.Companion.readAllSqlTables(
...
     val tables = metaData.getTables(catalogue, null, null, arrayOf("TABLE"))
 
-    val dataFrames = mutableListOf<AnyFrame>()
+    val dataFrames = mutableMapOf<String, AnyFrame>()
 
     while (tables.next()) {
         val table = dbType.buildTableMetadata(tables)
         if (!dbType.isSystemTable(table)) {
-            // we filter her second time because of specific logic with SQLite and possible issues with future databases
+            // we filter her a second time because of specific logic with SQLite and possible issues with future databases
```
Jolanrensen, line 343: *here

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -500,14 +502,15 @@ public fun DataFrame.Companion.getSchemaForAllSqlTables(connection: Connection):
...
-    val dataFrameSchemas = mutableListOf<DataFrameSchema>()
+    val dataFrameSchemas = mutableMapOf<String, DataFrameSchema>()
 
     while (tables.next()) {
         val jdbcTable = dbType.buildTableMetadata(tables)
         if (!dbType.isSystemTable(jdbcTable)) {
-            // we filter her second time because of specific logic with SQLite and possible issues with future databases
-            val dataFrameSchema = getSchemaForSqlTable(connection, jdbcTable.name)
-            dataFrameSchemas += dataFrameSchema
+            // we filter her a second time because of specific logic with SQLite and possible issues with future databases
```
Jolanrensen, line 510: *here

Jolanrensen (approved):
Looks good :) 

The only concern I have is with notebooks. Since we load the tables and we know the names, we should be able to generate safe accessors for them (like for a `DataRow`). Maybe something like https://github.com/Kotlin/dataframe/issues/719 could help with that. But that's for the future :) for now it's a good change

Issue #704, opened 2024-05-22
Kotlin DataFrame compiler plugin


koperagen:
Place for discussion and questions about Kotlin DataFrame compiler plugin

Idea behind it is to make such code compile, provide coding assistance in project files and later in Kotlin Notebooks - on top of already existing code generation in-between notebook cells

```
data class @DataSchema
class WikiData(val name: String, val paradigms: List<String>)

fun main() {
    val df = dataFrameOf(
        WikiData("Kotlin", listOf("object-oriented", "functional", "imperative")), // Kotlin
        WikiData("Haskell", listOf("Purely functional")), // Haskell
        WikiData("C", listOf("imperative")), // C)
    )
    val df1 = df.add("size") { 
        paradigms.size // `paradigms` is generated based on WikiData class structure
    }
    // `size` property is generated based on `add` argument
    df1.size.print()
}

```

Implementation lives here
https://github.com/Kotlin/dataframe/tree/compiler-plugin

Demo project that you can clone and run
https://github.com/koperagen/df-plugin-demo

Jolanrensen:
We might need to do some additional research with regard to the maintainability of the implementation, mainly the cases where we have to write the same DataFrame logic in two places.

Doing operations on DataFrames with the plugin happens in two places:
- The library itself
  - This works mostly on runtime
  - Is based on both the structure, types, and names of the DataFrame, but also on its data
- The compiler plugin
  - This works during code analysis
  - Is purely based on structure, types and names of the DataFrame
  - It could carry some information under-the-hood, for example:
    - `@Import` json data
    - The state of a DSL scope, such as `groupBy {}`
    - `df.transpose()` -> The df will have a `keys: String` column containing the previous column names
    - Etc.

I believe we should try, wherever we can, to share the logic between these two scopes. This can only be done in places where the logic is exclusively dependent on the structure, types, or names of the DataFrame. Sharing the logic will help us (and future contributors) to a) fix bugs more easily and b) keep ensuring consistency between the plugin and the library.

I see 3 options for us:

[...]

Issue #120, opened 2022-06-14
Enhance support for @DataSchema in gradle projects


koperagen:
Currently, a KSP preprocessor generates DataSchema-boilerplate. 
We want to address the following:
1. You need to build a project to generate boilerplate. We want the same experience as kotlinx.serialization (generate as you type, rename / move refactorings)
2. Better error reporting. Report warning for local classes annotated with `@DataSchema`
3. Add `companion object` to a data schema interface automatically https://github.com/Kotlin/dataframe/issues/113
4. Resolution of the generated code without extra gradle configurations on the user side.

If 1 and 4 are solved on the KSP side with the release of K2, then we only need a small compiler plugin for 2 and 3. Otherwise, codegen should be moved to a compiler plugin as well.

koperagen:
Everything described here is done. Use https://kotlin.github.io/dataframe/compiler-plugin.html

PR #729, opened 2024-06-10
Compiler plugin

koperagen:
There are changes in core module itself to be aware of:
1. DataRowApi.kt is related to https://github.com/Kotlin/dataframe/issues/113. Compiler plugin automatically adds DataRowSchema as a supertype to data schemas, enabling those APIs. To be continued later
2. Changes in insert.kt. An attempt to make it work over a generic tree, in order to share implementation between compiler plugin and core library
3. Across library added annotations that provide required meta information for compiler plugin
4. aggregate, toDataFrame functions user to be member functions in Grouped, GroupBy interfaces. Now they are extension functions, in order for plugin to work + it's not clear why they should be member function after all

Compiler plugin module consists of three main parts:
1. Plugin that generates extension properties, that will replace KSP plugin as it generates declarations in IDE without running a Gradle task. Related to https://github.com/Kotlin/dataframe/issues/120
2. Plugin that adds DataRowSchema supertype to a data schema annotated declarations. 
3. Most importantly, plugin that makes it possible to elevate compile time information from function arguments to type level. Plugin updates types of DataFrame objects whenever supported (=annotated with Refine) function is called with constant arguments. Most of the code here is its implementation

Everything related to compiler plugins aimed at K2 compiler
https://github.com/Kotlin/dataframe/issues/704


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/columns/BaseColumn.kt
```diff
@@ -18,7 +19,7 @@ import kotlin.reflect.KType
  *
  * @param T type of values contained in column.
  */
-public interface BaseColumn<out T> : ColumnReference<T> {
+public interface BaseColumn<out T> : ColumnReference<T>, GenericColumn {
```
Jolanrensen, line 22: @koperagen What's `GenericColumn` doing here btw? 
It's fine to have the interface as its supertype to just get the name of a column, but it lives in ./impl/api/insert.kt which breaks our structure. Probably best to move `GenericColumn(Group)` to ./columns

Jolanrensen: Ah I see it's mentioned in the comment! thanks

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/SimpleCol.kt
```diff
@@ -0,0 +1,127 @@
...
+
+    fun changeType(type: TypeApproximation): SimpleDataColumn {
+        return SimpleDataColumn(name, type)
+    }
+
+}
+
+data class SimpleFrameColumn(
+    override val name: String,
+    private val columns: List<SimpleCol>
```
Jolanrensen, line 76: If SimpleColumnGroup is a List<SimpleColumn>, then SimpleFrameColumn should be a List<List<SimpleColumn>> or a List<ColumnGroup> right?

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/data/toPluginDataFrameSchema.kt
```diff
@@ -0,0 +1,198 @@
...
+import org.jetbrains.kotlinx.dataframe.schema.ColumnSchema
+import org.jetbrains.kotlinx.dataframe.schema.DataFrameSchema
+import org.jetbrains.kotlinx.dataframe.plugin.extensions.wrap
+import org.jetbrains.kotlinx.dataframe.plugin.impl.SimpleDataColumn
+import kotlin.reflect.KClass
+import kotlin.reflect.KType
+import kotlin.reflect.KTypeProjection
+import kotlin.reflect.KVariance
+
+val KotlinTypeFacade.toPluginDataFrameSchema: DataFrameSchema.() -> PluginDataFrameSchema
```
Jolanrensen, line 29: Nice! very useful :)

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/data/toPluginDataFrameSchema.kt
```diff
@@ -0,0 +1,198 @@
...
+                }
+                is ColumnSchema.Group -> SimpleColumnGroup(
+                    name,
+                    columnSchema.schema.toPluginDataFrameSchema().columns()
+                )
+                is ColumnSchema.Frame -> SimpleFrameColumn(
+                    name,
+                    columnSchema.schema.toPluginDataFrameSchema().columns()
+                )
+                else -> TODO()
```
Jolanrensen, line 50: probably throw a `NotImplementedError` manually to not clog up the TODO scope

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/SimpleCol.kt
Jolanrensen: Ah no I see, it treats it like that because it tries to generalize the schema's, interesting

PR #737, opened 2024-06-14
Rename private properties with names matching library classes

koperagen:
Sometimes these private properties appear in import quick fix
image


Jolanrensen:
maybe they could be renamed to `_dataFrame` even? That would further hint that they're not supposed to appear and maybe make them appear even further down the list

koperagen:
I wouldn't go as far because it seems like a bug on IDEA side to suggest private properties anyway, and lowercase properties do not appear when DataFrame is typed

Jolanrensen:
> I wouldn't go as far because it seems like a bug on IDEA side to suggest private properties anyway, and lowercase properties do not appear when DataFrame is typed

true, but I think a lot of users will create variables named `df` or `dataFrame`, which will now mix with these private ones.

Jolanrensen (approved):

PR #738, opened 2024-06-14
Documentation for replaceAll

koperagen:
I decided to experiment with a new dataset


koperagen:
image

docs/StardustDocs/topics/update.md
```diff
@@ -42,6 +42,21 @@ df.update { city }.where { name.firstName == "Alice" }.with { "Paris" }
 <dataFrame src="org.jetbrains.kotlinx.dataframe.samples.api.Modify.updateWithConst.html"/>
 <!---END-->
 
+Update with mapping:
+
+<!---FUN replaceAll-->
+
+```kotlin
+functions.replaceAll(
```
Jolanrensen, line 50: the `replaceAll` call looks a bit out of place in between all these `update` functions, it should be in `replace`, shouldn't it? https://kotlin.github.io/dataframe/replace.html

Fillnulls https://kotlin.github.io/dataframe/fill.html is also an update "flavour", just like `replaceAll`, but it also has a separate page

Jolanrensen: you could provide a link from here to replaceAll though, but probably at the bottom

Jolanrensen (changes requested):
as commented :)

zaleslaw:
@koperagen what's is the future of this PR? Seems like outdated and hard to maintain?

PR #742, opened 2024-06-18
Extend "unfold" operation and support it in the compiler plugin

koperagen:
It covers two interesting use cases:
1. Replace column with multiple, potentially nested. Provides DSL similar to `add` for that 
2. More fine-grained toDataFrame. Instead of converting 20-30 properties to 2-3 level of nesting all at once user can choose to convert toDataFrame(maxDepth = 0) and unfold required properties to whatever level they need 

On the compiler plugin side i will continue to support other overloads later in different PR

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/unfold.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/unfold.kt
@@ -5,6 +5,8 @@ import org.jetbrains.kotlinx.dataframe.AnyColumnReference
 import org.jetbrains.kotlinx.dataframe.ColumnsSelector
 import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.annotations.Interpretable
+import org.jetbrains.kotlinx.dataframe.annotations.Refine
 import org.jetbrains.kotlinx.dataframe.columns.toColumnSet
 import org.jetbrains.kotlinx.dataframe.impl.api.unfoldImpl
 import kotlin.reflect.KProperty
@@ -18,6 +20,8 @@ public inline fun <reified T> DataColumn<T>.unfold(noinline body: CreateDataFram
 public inline fun <T, reified C> ReplaceClause<T, C>.unfold(vararg props: KProperty<*>, maxDepth: Int = 0): DataFrame<T> =
     with { it.unfold(props = props, maxDepth) }
 
+@Refine
+@Interpretable("ReplaceUnfold1")
 public inline fun <T, reified C> ReplaceClause<T, C>.unfold(noinline body: CreateDataFrameDsl<C>.() -> Unit): DataFrame<T> =
     with { it.unfoldImpl(skipPrimitive = false, body) }
 
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/unfold.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/unfold.kt
@@ -5,22 +5,21 @@ import org.jetbrains.kotlinx.dataframe.AnyColumnReference
 import org.jetbrains.kotlinx.dataframe.ColumnsSelector
 import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.DataFrame
-import org.jetbrains.kotlinx.dataframe.columns.ColumnKind
 import org.jetbrains.kotlinx.dataframe.columns.toColumnSet
-import org.jetbrains.kotlinx.dataframe.impl.api.createDataFrameImpl
-import org.jetbrains.kotlinx.dataframe.typeClass
+import org.jetbrains.kotlinx.dataframe.impl.api.unfoldImpl
 import kotlin.reflect.KProperty
 
-public inline fun <reified T> DataColumn<T>.unfold(): AnyCol =
-    when (kind()) {
-        ColumnKind.Group, ColumnKind.Frame -> this
-        else -> when {
-            isPrimitive() -> this
-            else -> values().createDataFrameImpl(typeClass) {
-                (this as CreateDataFrameDsl<T>).properties()
-            }.asColumnGroup(name()).asDataColumn()
-        }
-    }
+public inline fun <reified T> DataColumn<T>.unfold(vararg props: KProperty<*>, maxDepth: Int = 0): AnyCol =
+    unfoldImpl(skipPrimitive = true) { properties(roots = props, maxDepth = maxDepth) }
+
+public inline fun <reified T> DataColumn<T>.unfold(noinline body: CreateDataFrameDsl<T>.() -> Unit): AnyCol =
+    unfoldImpl(skipPrimitive = false, body)
+
+public inline fun <T, reified C> ReplaceClause<T, C>.unfold(vararg props: KProperty<*>, maxDepth: Int = 0): DataFrame<T> =
+    with { it.unfold(props = props, maxDepth) }
+
+public inline fun <T, reified C> ReplaceClause<T, C>.unfold(noinline body: CreateDataFrameDsl<C>.() -> Unit): DataFrame<T> =
+    with { it.unfoldImpl(skipPrimitive = false, body) }
 
 public fun <T> DataFrame<T>.unfold(columns: ColumnsSelector<T, *>): DataFrame<T> = replace(columns).with { it.unfold() }
 
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/unfold.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/unfold.kt
@@ -1,0 +1,23 @@
+package org.jetbrains.kotlinx.dataframe.impl.api
+
+import org.jetbrains.kotlinx.dataframe.AnyCol
+import org.jetbrains.kotlinx.dataframe.DataColumn
+import org.jetbrains.kotlinx.dataframe.api.CreateDataFrameDsl
+import org.jetbrains.kotlinx.dataframe.api.asColumnGroup
+import org.jetbrains.kotlinx.dataframe.api.asDataColumn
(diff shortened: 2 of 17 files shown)


Jolanrensen (changes requested):
I'm wondering, why did you decide to put `unfold` after `replace`?
Unfold itself, by definition, already replaces a column with a new column. I think we're making the API more complicated than it needs to be by making our users write `df.replace { a }.unfold {}` instead of keeping it inside the unfold operation, like `df.unfold { a }.by {}`. You can even allow `df.unfold(maxDepth = 2) { a }`. It would keep "replace with" simple and "unfold" more powerful.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/unfold.kt
```diff
@@ -0,0 +1,23 @@
...
+import org.jetbrains.kotlinx.dataframe.api.isPrimitive
+import org.jetbrains.kotlinx.dataframe.columns.ColumnKind
+import org.jetbrains.kotlinx.dataframe.typeClass
+
+@PublishedApi
+internal fun <T> DataColumn<T>.unfoldImpl(skipPrimitive: Boolean, body: CreateDataFrameDsl<T>.() -> Unit): AnyCol {
+    return when (kind()) {
+        ColumnKind.Group, ColumnKind.Frame -> this
+        else -> when {
+            skipPrimitive && isPrimitive() -> this
```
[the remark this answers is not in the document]
koperagen: You can't. Have a look `unfold primitive`. skipPrimitive = false is needed to make it work, and skipPrimitive = true is needed to avoid unpacking for example a column of String to `ColumnGroup, size: Int`, the same as we do for toDataFrame with overloads

koperagen:
> Unfold itself, by definition, already replaces a column with a new column
What definition? 
Originally i wanted a add `replace by` function with AddDsl. Then i remembered that toDataFrame DSL is pretty similar to AddDsl and we already have `unfold` function, it just lacks an overload with a DSL. Such an overload must be a multiplex operation, right?
`unfold by` is interesting, but it will have exactly one operation probably forever. Sounds good, right, but its semantics is no different from replace. 
I'd rather have one entry point (replace) for situations when you want to, like, replace a column with a different one.

Jolanrensen:
> What definition? Originally i wanted a add `replace by` function with AddDsl. Then i remembered that toDataFrame DSL is pretty similar to AddDsl and we already have `unfold` function, it just lacks an overload with a DSL. Such an overload must be a multiplex operation, right? `unfold by` is interesting, but it will have exactly one operation probably forever. Sounds good, right, but its semantics is no different from replace. I'd rather have one entry point (replace) for situations when you want to, like, replace a column with a different one.

I meant the definition of "unfolding", you're unfolding a column with its contents, so its type is bound to change, aka, the column is replaced.

I can see where you're coming from, but it still may be hard for users to have two different ways to unfold, namely `.replace {}.unfold {}` and `.unfold()` and both work a little bit different and have different arguments.

So I'd either: 
- make `.replace {}.byUnfolding {}` (or a name like that) the only version of unfold and deprecate the old one
- or make just `unfold {}` more powerful
- or keep both for discoverability, but both should be equally powerful

wdyt?

koperagen:
I'd say `df.unfold` should stay, because use case for *simply* unfolding a column with objects stays. Worth to add `df.unfold(maxDepth = , roots = )` overload too, missed it.

> So I'd either:
> or make just df.unfold {} more powerful

Please write this API with needed overloads and a few examples of its usage then

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/unfold.kt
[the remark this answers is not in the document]
Jolanrensen: Yes, but can you take a look at the `isPrimitive()` function? That function also returns true when you run in on a collection and an array. My suggestion was only to rename the `isPrimitive()` function.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/unfold.kt
```diff
@@ -5,22 +5,25 @@ import org.jetbrains.kotlinx.dataframe.AnyColumnReference
...
-            }.asColumnGroup(name()).asDataColumn()
-        }
-    }
+public inline fun <reified T> DataColumn<T>.unfold(vararg props: KProperty<*>, maxDepth: Int = 0): AnyCol =
+    unfoldImpl(skipPrimitive = true) { properties(roots = props, maxDepth = maxDepth) }
+
+public inline fun <reified T> DataColumn<T>.unfold(noinline body: CreateDataFrameDsl<T>.() -> Unit): AnyCol =
+    unfoldImpl(skipPrimitive = false, body)
+
+public inline fun <T, reified C> ReplaceClause<T, C>.unfold(vararg props: KProperty<*>, maxDepth: Int = 0): DataFrame<T> =
```
Jolanrensen, line 20: oh, we should also use `KCallable` instead of `KProperty` for java classes support :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/replace.kt
```diff
@@ -13,4 +15,61 @@ class ReplaceTests {
...
+        val c = conv["a"]["c"]["prop"]
+        c.type() shouldBe typeOf<Int>()
+        c.values() shouldBe listOf(1)
+    }
+
+    @Test
+    fun `unfold properties`() {
+        val col by columnOf(A("1", 123, B(3.0)))
+        val df1 = dataFrameOf(col)
+        val conv = df1.replace { col }.unfold(maxDepth = 2)
```
Jolanrensen, line 42: Not specifying maxDepth now breaks, while it worked before.
Try running `df1.replace { col }.with { it.unfold() }` before the PR and after it.
It works before, but now it gives: `java.lang.UnsupportedOperationException: Can not get nested column 'd' from ValueColumn 'bb'`

Jolanrensen:
> I'd say `df.unfold` should stay, because use case for _simply_ unfolding a column with objects stays. Worth to add `df.unfold(maxDepth = , roots = )` overload too, missed it.
> 
> > So I'd either:
> > or make just df.unfold {} more powerful
> 
> Please write this API with needed overloads and a few examples of its usage then

I made a little sample of `unfold {}.by {}` with the same options as your version. Both `df.unfold {}` and `df.unfold {}.by {}` can be used. See the commit here for more details:
https://github.com/Kotlin/dataframe/commit/20cab7a12968babcb7f9d8ce8a24e3ca0917eda1

koperagen:
UnfoldingDataFrame looks good, we can use it. Supporting it on the plugin side will require some changes, but it's ok. I'm only worried with return type being different than DataFrame people will be tempted to call `df.unfold { col }.by()`. It's first time intermediate object in a multiplex operation is also a `DataFrame`
But if we go this route, i'd also add
`df.replace {}.by(CreateDataFrameDsl)` (only this, without `vararg props: KProperty<*> and maxDepth: Int = 0` overloads)

Jolanrensen:
@koperagen Yes, I couldn't find another way to have a notation with 2 selectors and the second one being optional while keeping the DataFrame DSL style :/ but indeed, it's something new. A bit like `.recursively()`.

Actually, since `by` is defined on `UnfoldingDataFrame` it should appear quite high on the list:
image

We could also put it inside the class to make it even more discoverable.

`replace {}.by {}` also looks interesting :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/replace.kt
```diff
@@ -13,4 +15,61 @@ class ReplaceTests {
...
+        b.values() shouldBe listOf("123")
+
+        val c = conv["a"]["c"]["prop"]
+        c.type() shouldBe typeOf<Int>()
+        c.values() shouldBe listOf(1)
+    }
+
+    @Test
+    fun `unfold properties`() {
+        val col by columnOf(A("1", 123, B(3.0)))
```
zaleslaw, line 40: Is this case "More fine-grained toDataFrame. Instead of converting 20-30 properties to 2-3 level of nesting all at once user can choose to convert toDataFrame(maxDepth = 0) and unfold required properties to whatever level they need" covered here, in this test?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/unfold.kt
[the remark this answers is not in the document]
zaleslaw: If it's possible, let's avoid motion or gravity to the native language, I believe, it's not a goal

zaleslaw:
Honetsly, I like the idea from the use-case **"More fine-grained toDataFrame. Instead of converting 20-30 properties to 2-3 level of nesting all at once user can choose to convert toDataFrame(maxDepth = 0) and unfold required properties to whatever level they need"**, defining a level of our unfolding (that is a special situation of `convert`)

Also I found that we have a lack of docs/examples for this operation https://kotlin.github.io/dataframe/unfold.html

Some crazy ideas

```
replace {}.by { ::unfold }
replace {}.by { ::unfoldDeeply }
```

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/replace.kt
koperagen: Technically, yes. I intend to have a more representative example as a part of compiler plugin demo. There's a tree of objects with many properties and potentially deep nesting from konsist library. It will be a good illustration. But here it merely unfolds one specific column up to 2 levels.

Jolanrensen:
Since it's a WIP I made it a draft for now

Jolanrensen:
Just a thought :) We can actually have something like `replace by unfolding` without changing the `replace` API. Replace already works like `replace with`, meaning we can already write something like `df.replace { "data"<ColumnGroup<*>>() }.with { it.unfold() }` in the current state of the library. Maybe we could expand on `DataColumn.unfold` to provide the Add-DSL like notation you suggest :) 

Something like:
```kt
df.replace { data }.with { 
    it.unfold {
         "b" from { it }
         "c" from { DataRow.readJsonStr("""{"prop": 1}""") }
    }
}
```

koperagen:
Partially merged as part of other PR

PR #743, opened 2024-06-19
Add delimiter parameter to readDelimStr

koperagen:


tests/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Write.kt
```diff
@@ -46,7 +46,8 @@ class Write : TestBase() {
     @Test
     fun writeCsvStr() {
         // SampleStart
-        val csvStr = df.toCsv(CSVFormat.DEFAULT.withDelimiter(';').withRecordSeparator(System.lineSeparator()))
+        val format = CSVFormat.DEFAULT.builder().setDelimiter(';').setRecordSeparator(System.lineSeparator()).build()
```
Jolanrensen, line 49: long line, but the linter will fix all these soon anyway

Jolanrensen (approved):
Yes! I missed that parameter too. Looks like TC still fails, but after the test is fixed, should be good to go

dave08:
Little comment, maybe since this is a function to especially read delimeters, it might be useful to have an override that takes in a Regex as a delimiter... this might be used for command line output tables that are usually space separated but sometimes inside a column value there might be a single space, so I need to use "\\s\\s+" to correctly read it in.

Jolanrensen:
@dave08 sounds like a good idea :) could you make an issue for that?

dave08:
https://github.com/Kotlin/dataframe/issues/746

PR #744, opened 2024-06-19
Update Kotlin version for compiler plugin

koperagen:
In the process of updating i found that kotlinx.serialization is more permissive compared to klaxon and parses value to JSON where it shouldn't. Here's the issue https://github.com/Kotlin/kotlinx.serialization/issues/2511


Jolanrensen (approved):
Actually two pr's haha, but I'll allow it :)

koperagen:
Thank you!

PR #745, opened 2024-06-19
Add an option to read Excel cell values as a String regardless of their content type

### LINKED ISSUE #669: Support an option to get formatCellValue in readExcel
You might want to get cell value not how it is represented in the xlsx file and seen by Apache Poi, but as a String
There's method that does exactly this
We can provide an argument to specify what columns should be read by value methods like stringCellValue, numericCellValue, etc. and by formatCellValue
```
val style: CellStyle = getCellStyle()
val dataFormatter = DataFormatter()
val formattedCellValue = dataFormatter.formatCellValue(this, style)
return formattedCellValue
```

koperagen:
fixes #669
Once again based on user feedback, and also was mentioned previously


dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -93,17 +97,21 @@ public fun DataFrame.Companion.readExcel(
     sheetName: String? = null,
     skipRows: Int = 0,
     columns: String? = null,
+    formattingOptions: FormattingOptions? = null,
```
Jolanrensen, line 100: If this argument is just a "range of columns to read as string", "formattingOptions" is not a good name. it should be something like "stringColumns". That said, if formattingOptions can do something more, like define other colTypes as well, then the name is better, but I would try to make the API similar to csv, where you can also define colTypes.

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -83,6 +84,9 @@ private fun setWorkbookTempDirectory() {
 /**
  * @param sheetName sheet to read. By default, the first sheet in the document
  * @param columns comma separated list of Excel column letters and column ranges (e.g. “A:E” or “A,C,E:F”)
+ * @param formattingOptions range of columns to read as String regardless of a cell type.
```
Jolanrensen, line 87: It's not just that right? You can format a range of columns as string using different options as defined by a `DataFormatter`. The description should probably reflect this.

If this argument was just to format columns as string, the argument should be `colRangeAsString` or better, mirror the `colTypes` argument from CSV.

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -173,18 +193,29 @@ public fun DataFrame.Companion.readExcel(
...
         ?.let { wb.getSheet(it) ?: error("Sheet with name $sheetName not found") }
         ?: wb.getSheetAt(0)
-    return readExcel(sheet, columns, skipRows, rowsCount, nameRepairStrategy)
+    return readExcel(sheet, columns, formattingOptions, skipRows, rowsCount, nameRepairStrategy)
+}
+
+/**
+ * @param range comma separated list of Excel column letters and column ranges (e.g. “A:E” or “A,C,E:F”)
+ */
+public class FormattingOptions(range: String, public val formatter: DataFormatter = DataFormatter()) {
```
Jolanrensen, line 209: I do have one concern exposing `DataFormatter` to the public API as that makes the excel implementation stuck to the `org.apache.poi.ss.usermodel` package which is not great if we want to swap excel implementations

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -83,6 +84,9 @@ private fun setWorkbookTempDirectory() {
 /**
  * @param sheetName sheet to read. By default, the first sheet in the document
  * @param columns comma separated list of Excel column letters and column ranges (e.g. “A:E” or “A,C,E:F”)
+ * @param formattingOptions range of columns to read as String regardless of a cell type.
+ * For example, by default numeric cell with value "3" will be parsed as Double with value being 3.0. With this option, it will be simply "3"
+ * Useful
```
Jolanrensen, line 89: dangling new sentence :)

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
koperagen: Sort of, yes. Thing is that Sheet is already a part of public API. I thought potential new implementation could live alongside existing one

koperagen: I'll make this dataFormatter argument only for overload with Sheet parameter

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
koperagen: Changed the parameter for some overloads to StringColumns class and provided see also where it's still FormattingOptions

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
Jolanrensen: that may be better indeed, to limit how much we expose. We did the same with Klaxon and JSON some time back.
I wouldn't have two Excel implementations side by side though. Best to keep it simple for the user and just have 1 `DataFrame.readExcel()` implementation

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -83,6 +84,8 @@ private fun setWorkbookTempDirectory() {
 /**
  * @param sheetName sheet to read. By default, the first sheet in the document
  * @param columns comma separated list of Excel column letters and column ranges (e.g. “A:E” or “A,C,E:F”)
+ * @param stringColumns range of columns to read as String regardless of a cell type.
+ * For example, by default numeric cell with value "3" will be parsed as Double with value being 3.0. With this option, it will be simply "3"
```
Jolanrensen, line 88: I'd surround types with `[]` to make them clickable :). *their type or *a cell's type

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -173,18 +196,37 @@ public fun DataFrame.Companion.readExcel(
...
         ?.let { wb.getSheet(it) ?: error("Sheet with name $sheetName not found") }
         ?: wb.getSheetAt(0)
-    return readExcel(sheet, columns, skipRows, rowsCount, nameRepairStrategy)
+    return readExcel(sheet, columns, formattingOptions, skipRows, rowsCount, nameRepairStrategy)
+}
+
+/**
+ * @param range comma separated list of Excel column letters and column ranges (e.g. “A:E” or “A,C,E:F”)
+ */
+public class StringColumns(public val range: String)
```
Jolanrensen, line 212: good name :)

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -173,18 +196,37 @@ public fun DataFrame.Companion.readExcel(
...
+
+public fun StringColumns.toFormattingOptions(formatter: DataFormatter = DataFormatter()): FormattingOptions =
+    FormattingOptions(range, formatter)
+
+/**
+ * @param range comma separated list of Excel column letters and column ranges (e.g. “A:E” or “A,C,E:F”)
```
Jolanrensen, line 218: probably add a @param for formatter too so people can click through to DataFormatter

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -173,18 +196,37 @@ public fun DataFrame.Companion.readExcel(
...
+ */
+public class FormattingOptions(range: String, public val formatter: DataFormatter = DataFormatter()) {
+    public val columnIndices: Set<Int> = getColumnIndices(range).toSet()
 }
 
 /**
  * @param sheet sheet to read.
  * @param columns comma separated list of Excel column letters and column ranges (e.g. “A:E” or “A,C,E:F”)
+ * @param formattingOptions range of columns to read as String regardless of a cell type.
```
Jolanrensen, line 227: This should probably say "as [String], formatted by the given [FormattingOptions]" or something like that. Otherwise it's not clear what the difference is by the overload with stringColumns and this one.

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
Jolanrensen: could be a value class tho :) since it's just a String essentially

Jolanrensen (changes requested):

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
koperagen: What types? String and Double?

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
koperagen: I think "see also" below and the fact that type is different is enough

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
koperagen: done

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
koperagen: done

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
Jolanrensen: yes, that's what I do to all types across the kdocs :) I think it's nice for interactivity and consistency.

Jolanrensen (approved):

Issue #364, opened 2023-04-26
Updating the linter


Jolanrensen:
The linter plugin we use (https://github.com/jeremymailen/kotlinter-gradle) can be updated to v3.14.0.

However, two things:

 - The excluding of rules is now handled by the .editorconfig file and this seems to be broken atm: https://github.com/jeremymailen/kotlinter-gradle/issues/324
 - Currently only changed files are linted, but the project is linted relatively poorly, so adding an updated linter would require a complete linting of every file in the project.
 - We need to check if we still get unused imports for extension functions referenced in KDocs.

Jolanrensen:
After that PR is merged, the linter can be updated relatively easily.

- We need to wait till there are not so many PRs anymore
- The linter will format the entire project (!), most things can be done automatically, but the rest can be done in about a day's work, as seen in this example branch: https://github.com/Kotlin/dataframe/tree/k2-linter-attempt
- This branch also contains a sample .editorconfig file which disables only the rule "standard:filename" and it configures IntelliJ to not have *-imports at all :)
- Generated code can be excluded
- Using the [IntelliJ KtLint plugin](https://pinterest.github.io/ktlint/latest/install/setup/) we can more easily see errors and `@Suppress()` rules if we need to deviate somewhere.
- We can commit-hook the formatKotlin task so only clean code gets pushed :)
- We can even write custom rules if we need them (I don't expect we do) https://pinterest.github.io/ktlint/latest/api/custom-rule-set/

Jolanrensen:
I've been tweaking some settings in .editorconfig to what I think looks best (and most in-line with DataFrame):
https://github.com/Kotlin/dataframe/blob/k2-linter-attempt/.editorconfig

We just need to figure out how to NOT run on generated-sources by default (as it's way to heavy) but allow the linter to format generated-sources from github actions instead.

Jolanrensen:
Some updates:

I've had difficulties using Kotlinter due to its limited customization options. After trying Spotless, which also wasn't fit for the job, I found https://github.com/JLLeitschuh/ktlint-gradle, which is also used by kotlin-jupyer.

This branch contains a working project with that plugin: https://github.com/Kotlin/dataframe/tree/k2-ktlint-jlleitschuh-attempt.

[...]

PR #754, opened 2024-06-25
Replacing the linter

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/364, the last stopper for updating to Kotlin 2.0.

- replaced kotlinter by jlleitschuh's ktlint plugin
- .editorconfig is used to set the rules of the linter and the code style
- Please use https://plugins.jetbrains.com/plugin/15057-ktlint from now on :)
- Individual cases can be `@Suppress`-ed easily with that plugin
- I mostly-automatically refactored the entire project, but I checked all changes
- I fixed some manual things: *-imports, names of consts, `@file:ExcludeFromSources` to prevent "empty" files in generated-sources
- Tests in samples/api are cleaned for now and ignored by ktlint on a per-file basis (`@file:Suppress("ktlint")`)
- See the issue for more information


dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/MsSql.kt
```diff
@@ -17,9 +17,7 @@ public object MsSql : DbType("sqlserver") {
     override val driverClassName: String
         get() = "com.microsoft.sqlserver.jdbc.SQLServerDriver"
 
-    override fun convertSqlTypeToColumnSchemaValue(tableColumnMetadata: TableColumnMetadata): ColumnSchema? {
-        return null
-    }
+    override fun convertSqlTypeToColumnSchemaValue(tableColumnMetadata: TableColumnMetadata): ColumnSchema? = null
```
zaleslaw, line 20: is it really linter setting? could we change it? I don't like these short forms without brackets, it looks ugly and make me slower

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -139,15 +140,16 @@ public fun DataFrame.Companion.readSqlTable(
     val url = connection.metaData.url
     val dbType = extractDBTypeFromConnection(connection)
 
-    val selectAllQuery = if (limit > 0) dbType.sqlQueryLimit("SELECT * FROM $tableName", limit)
-    else "SELECT * FROM $tableName"
+    val selectAllQuery = if (limit > 0) {
+        dbType.sqlQueryLimit("SELECT * FROM $tableName", limit)
```
zaleslaw, line 144: and again, here we don't need these brackets in the code

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -523,16 +520,14 @@ private fun buildSchemaByTableColumns(tableColumns: MutableList<TableColumnMetad
...
 
-private fun generateColumnSchemaValue(
-    dbType: DbType,
-    tableColumnMetadata: TableColumnMetadata
-): ColumnSchema = dbType.convertSqlTypeToColumnSchemaValue(tableColumnMetadata) ?: ColumnSchema.Value(
-    makeCommonSqlToKTypeMapping(tableColumnMetadata)
-)
+private fun generateColumnSchemaValue(dbType: DbType, tableColumnMetadata: TableColumnMetadata): ColumnSchema =
+    dbType.convertSqlTypeToColumnSchemaValue(tableColumnMetadata) ?: ColumnSchema.Value(
+        makeCommonSqlToKTypeMapping(tableColumnMetadata),
```
zaleslaw, line 529: sorry, but it's fine with signatures, but not fine with the client code for passing parameters, it breaks the reading for me

CONTRIBUTING.md
```diff
@@ -81,13 +85,22 @@ so do familiarize yourself with the following guidelines.
 
 ## Environment requirements
 
-JDK >= 11 referred to by the `JAVA_HOME` environment variable.
+* JDK >= 11 referred to by the `JAVA_HOME` environment variable.
+
+  * Note, any version above 11 should work in theory, but JDK 11 is the only version we test with,
```
zaleslaw, line 90: what's with the formatting here?

zaleslaw:
@Jolanrensen you did a great job leading us forward to K2, but let's discuss that when you return and change if possible some rules

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/MsSql.kt
Jolanrensen: It is https://pinterest.github.io/ktlint/latest/rules/standard/#function-expression-body.
Because of the official style guide: https://kotlinlang.org/docs/coding-conventions.html#functions.
To me, block bodies with just one return statement take up unnecessary space and feel like java-boilerplate. We could change it, it's a matter of taste. But I think the most important thing is consistency; Mixing single-line block body's and expression bodies is messy.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
Jolanrensen: It's more readable for multiline if-statements: https://pinterest.github.io/ktlint/latest/rules/standard/#if-else-bracing and https://kotlinlang.org/docs/coding-conventions.html#control-flow-statements.
No brackets in if-statements are only allowed/clear when it fits on the same line.

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/h2/h2Test.kt
```diff
@@ -716,10 +717,18 @@ class JdbcTest {
 
         connection.createStatement().execute(createTestTable1Query)
 
-        connection.createStatement().execute("INSERT INTO TestTable1 (id, name, surname, age) VALUES (1, 'John', 'Crawford', 40)")
-        connection.createStatement().execute("INSERT INTO TestTable1 (id, name, surname, age) VALUES (2, 'Alice', 'Smith', 25)")
-        connection.createStatement().execute("INSERT INTO TestTable1 (id, name, surname, age) VALUES (3, 'Bob', 'Johnson', 47)")
-        connection.createStatement().execute("INSERT INTO TestTable1 (id, name, surname, age) VALUES (4, 'Sam', NULL, 15)")
+        connection.createStatement().execute(
+            "INSERT INTO TestTable1 (id, name, surname, age) VALUES (1, 'John', 'Crawford', 40)",
+        )
```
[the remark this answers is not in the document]
Jolanrensen: so like:
```kt
 connection
    .createStatement()
    .execute("INSERT INTO TestTable1 (id, name, surname, age) VALUES (4, 'Sam', NULL, 15)")
```
?

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
Jolanrensen: Yes, this can indeed look better. How about:
```kt
dbType.convertSqlTypeToColumnSchemaValue(tableColumnMetadata) 
    ?: ColumnSchema.Value(makeCommonSqlToKTypeMapping(tableColumnMetadata))
```
or
```kt
dbType.convertSqlTypeToColumnSchemaValue(tableColumnMetadata) 
    ?: ColumnSchema.Value(
          makeCommonSqlToKTypeMapping(tableColumnMetadata),
    )
```
?

CONTRIBUTING.md
Jolanrensen: It's a double indent 
* something
  * like this

Jolanrensen:
> @Jolanrensen you did a great job leading us forward to K2, but let's discuss that when you return and change if possible some rules

Thanks for reviewing!

I think some of your suggestions are in conflict with the [official Kotlin styleguide](https://kotlinlang.org/docs/coding-conventions.html) which are honored by [Ktlint](https://pinterest.github.io/ktlint/latest/rules/standard/). We could deviate as I've done with a small number of rules in the [.editorconfig](https://github.com/Kotlin/dataframe/blob/ktlint-migration/.editorconfig) file but we would need to be sure it'd be in the best interest of the library and our users.

For example, I've turned off the filename check, as we name our API files after the function-family it represents. I also turned off `chain-method-continuation` so DataFrame notations like:
```kt
df.update { .. }.where { .. }.with { .. }
  .update { .. }.with { .. }
```
are allowed and not replaced with:
```kt
df
  .update { .. }
  .where { .. }
  .with { .. }
  .update { .. }
  .with { .. }
```

Jolanrensen:
As for the trailing-comma case, there are many viewpoints on what's best.
The [Kotlin coding convention](https://kotlinlang.org/docs/coding-conventions.html#trailing-commas) specifies many benefits:

> It makes version-control diffs cleaner – as all the focus is on the changed value.
> 
> It makes it easy to add and reorder elements – there is no need to add or delete the comma if you manipulate elements.
> 
> It simplifies code generation, for example, for object initializers. The last element can also have a comma.
> 
> It even encourages it at declaration site (so defining classes, functions, etc.) and leaves it at our own discretion for the call site.

Ktlint enables [trailing commas on call site](https://pinterest.github.io/ktlint/latest/rules/standard/#trailing-comma-on-call-site) by default for the same reasons as on the declaration site, however, while IntelliJ can be configured to _allow_ or _disallow_ trailing commas, Ktlint can only be configured to _enforce_ or _disallow_ them. Consistency is preferred, which is honestly a good decision I think.

Now, enforcing trailing commas does make notation like
```kt
myFunction(
  singleArg,
)
```
appear. Which is debatably ugly and can hurt our ability to read it well, however it can often be rewritten like:
```kt
myFunction(singleArg)
```
or
```kt
myFunction(
  arg = singleArg,
)
```

Having trailing commas for function calls with many (optional) named arguments is still a must for me:
```kt
createValueColumn(
    name = name,
    values = values,
    type = getValuesType(
        values = values,
        type = typeOf<T>(),
        infer = infer,
    ),
)
```
Being able to reorder arguments like these or add new ones is something that occurs a lot and it can save headaches with merge conflicts if every line simply ends with a comma.

So, to conclude, I think we should follow [kotlin-juypter](https://github.com/Kotlin/kotlin-jupyter/blob/master/.editorconfig) and just enable them both on the call- and the declaration site.

zaleslaw (approved):

PR #755, opened 2024-06-26
Compiler plugin update

koperagen:
Review commits one by one, they are not related to each other
Note that i provided explanations b15493785ddf0377e354773f55b5e7e1e48b399e in commit message 
Here you can find how to enable fir dump for an individual test
plugins/kotlin-dataframe/testData/box/propertiesOrder.kt


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MarkersExtractor.kt
```diff
@@ -54,7 +55,8 @@ internal object MarkersExtractor {
 
     private fun getFields(markerClass: KClass<*>, nullableProperties: Boolean): List<GeneratedField> {
         val order = getPropertyOrderFromPrimaryConstructor(markerClass) ?: emptyMap()
-        return markerClass.memberProperties.sortedBy { order[it.name] ?: Int.MAX_VALUE }.mapIndexed { _, it ->
+        val structuralProperties = markerClass.memberProperties.filter { it.findAnnotation<ScopeProperty>() == null }
```
Jolanrensen, line 58: you could probably just use .hasAnnotation<>() :)

plugins/kotlin-dataframe/testData/box/propertiesOrder.fir.txt
```diff
@@ -1,6 +1,4 @@
-FILE: read.kt
-    package org.jetbrains.kotlinx.dataframe
-
+FILE: propertiesOrder.kt
     public final fun box(): R|kotlin/String| {
```
Jolanrensen, line 2: You wanted to exclude these files right?

plugins/kotlin-dataframe/testData/diagnostics/selectDuringTyping.fir.txt
```diff
@@ -28,9 +28,8 @@ FILE: selectDuringTyping.kt
             }
 
             local abstract class ExplodeSchema_94 : R|<local>/ExplodeSchema_94I| {
-                public abstract var scope0: R|<local>/Scope0|
+                @R|org/jetbrains/kotlinx/dataframe/annotations/ScopeProperty|() public abstract val scope0: R|<local>/Scope0|
                     public get(): R|<local>/Scope0|
```
Jolanrensen, line 32: should the fir files in diagnostics also be included, or are they fine?

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/Plugin.kt
```diff
@@ -28,5 +28,18 @@ public annotation class Import
...
+ * Compiler plugin materializes schemas as classes.
+ * These classes have two kinds of properties:
+ * 1. Scope properties that only serve as a reference for internal property resolution
+ * 2. Schema properties that reflect dataframe structure
+ * Scope properties need
+ * to be excluded in IDE plugin and in [org.jetbrains.kotlinx.dataframe.codeGen.MarkersExtractor.get]
+ * This annotation serves to distinguish between the two where needed
+ */
+@Target(AnnotationTarget.PROPERTY)
+public annotation class ScopeProperty
```
Jolanrensen, line 42: do you use this annotation already somewhere? I'm curious to see how/where it's used :)

Jolanrensen (approved):
just some small things :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MarkersExtractor.kt
koperagen: Didn't know, thanks!

plugins/kotlin-dataframe/testData/box/propertiesOrder.fir.txt
koperagen: I opt in fir dump for this file specifically. Would be good to have a few of them present to track changes in generated code, like for example there's this `@Order` annotation and IDE plugin relies on it

plugins/kotlin-dataframe/testData/diagnostics/selectDuringTyping.fir.txt
koperagen: For diagnostics tests, FIR dumps are needed to verify that warnings/errors were actually reported, it seems

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/Plugin.kt
koperagen: What do you mean by use? :o 
Here generated property is annotated 
plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/extensions/TokenGenerator.kt

koperagen: Plus other things from the same commit

Jolanrensen: Ooh so this annotation class is only used in generated properties. That wasn't clear to me

PR #760, opened 2024-06-27
Enable parallel forks for tests

koperagen:
Let's see how it affects build server performance. I tried to benchmark locally, but execution time is just way too unstable to tell if it actually helps or not 
Configured according to https://docs.gradle.org/current/userguide/performance.html#execute_tests_in_parallel

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -364,6 +364,7 @@ tasks.withType<KotlinCompile> {
 
 tasks.test {
     maxHeapSize = "2048m"
+    maxParallelForks = (Runtime.getRuntime().availableProcessors() / 2).coerceAtLeast(1)
     extensions.configure(kotlinx.kover.api.KoverTaskExtension::class) {
         excludes.set(
             listOf(
@@ -372,6 +373,10 @@ tasks.test {
             )
         )
     }
+}
+
+tasks.getByName<Test>("samplesTest") {
+    maxParallelForks = 1
 }
 
 tasks.processJupyterApiResources {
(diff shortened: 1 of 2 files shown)


Jolanrensen:
That is a good idea!

core/build.gradle.kts
```diff
@@ -374,6 +375,10 @@ tasks.test {
     }
 }
 
+tasks.getByName<Test>("samplesTest") {
```
zaleslaw, line 378: Is this referred to all of our tests?

core/build.gradle.kts
```diff
@@ -364,6 +364,7 @@ tasks.withType<KotlinCompile> {
 
 tasks.test {
     maxHeapSize = "2048m"
+    maxParallelForks = (Runtime.getRuntime().availableProcessors() / 2).coerceAtLeast(1)
```
zaleslaw, line 367: it's worth to know, how many processors/cores are available in average on TC

zaleslaw (commented):
Imagine that we have 4 cores, could take only 2 of them, not all parts are parallelized: @BeforeTest leads to the max.parallelization < 2 and plus overhead of running in different threads. From the other hand it's doing nothing bad, let's play

I also saw, that it worth to set up env variables  
systemProperties["junit.jupiter.execution.parallel.enabled"] = true
systemProperties["junit.jupiter.execution.parallel.mode.default"] = "concurrent" 

but also a lot of threads that with these settings JUnit don't want to run tests in parallel
https://stackoverflow.com/questions/74599354/junit-not-running-tests-in-parallel

Probably we need to log/println on that core it was executed

Jolanrensen:
I converted the PR to a draft for now since we didn't get the results we wanted (yet). Feel free to change it if it improves :)

koperagen:
I think we have a better chance of improving performance, after we eventually get rid of gradle integration tests

Issue #344, opened 2023-04-06
Streamline use of toListOf by adding method to print data class definition of table schema


holgerbrandl:
To streamline the transition from table rows to objects using `df.toListOf<Something>()`, it would be wonderful if kdf could print a schema (as data class definition) to standard out.

Currently, this is only possible by converting a data-frame into krangl-df (using https://github.com/holgerbrandl/kdfutils) and and using `printDataClassSchema ` to print the schema:

```kotlin

import com.github.holgerbrandl.kdfutils.toKranglDF
import krangl.printDataClassSchema
import org.jetbrains.kotlinx.dataframe.api.toListOf
import org.jetbrains.kotlinx.dataframe.datasets.flightsData

fun main() {

    flightsData.toKranglDF().printDataClassSchema("FlightRecord")

    // note: the data class definition is created and printed by printDataClassSchema to stdout  
    data class FlightRecord(
        val year: Int,
        val month: Int,
        val day: Int,
        val dep_time: String?,
        val dep_delay: String?,
        val arr_time: String?,
        val arr_delay: String?,
        val carrier: String,
        val tailnum: String,
        val flight: Int,
        val origin: String,
        val dest: String,
        val air_time: String?,
        val distance: Int,
        val hour: String?,
        val minute: String?
    )

    val records = flightsData.toListOf<FlightRecord>()
    println(records.first())
}
```
which prints just
```
data class FlightRecord(val year: Int, val month: Int, val day: Int, val dep_time: Any?, val dep_delay: Any?, val arr_time: Any?, val arr_delay: Any?, val carrier: String, val tailnum: String, val flight: Int, val origin: String, val dest: String, val air_time: Any?, val distance: Int, val hour: Any?, val minute: Any?)

FlightRecord(year=2013, month=1, day=1, dep_time=517, dep_delay=2, arr_time=830, arr_delay=11, carrier=UA, tailnum=N14228, flight=1545, origin=EWR, dest=IAH, air_time=227, distance=1400, hour=5, minute=17)
```
Technically the approach suggested in https://github.com/Kotlin/dataframe/issues/340#issuecomment-1497310501 points into the same direction. However, having the ability to print a data-class definition of a data-frame would work more generally also in a simple kotlin jvm application without any compiler plugin or jupyter.

Jolanrensen:
Solved by https://github.com/Kotlin/dataframe/pull/763

PR #763, opened 2024-07-03
Improve codegen for stdlib <-> df interop workflow

koperagen:
I recommend to review commits individually. First two were small changes made to improve debugging experience, and other two are separate features
solves https://github.com/Kotlin/dataframe/issues/344

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MarkersExtractor.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MarkersExtractor.kt
@@ -17,14 +17,33 @@ import kotlin.reflect.full.withNullability
 import kotlin.reflect.jvm.jvmErasure
 import kotlin.reflect.typeOf
 
-internal fun KType.shouldBeConvertedToFrameColumn(): Boolean = when (jvmErasure) {
-    DataFrame::class -> true
-    List::class -> arguments[0].type?.jvmErasure?.hasAnnotation<DataSchema>() == true
-    else -> false
+internal fun KType.getFieldKind(): FieldKind = when {
+    jvmErasure == DataFrame::class -> Frame
+    jvmErasure == List::class && (arguments[0].type?.jvmErasure?.hasAnnotation<DataSchema>() == true) -> ListToFrame
+    jvmErasure == DataRow::class -> Group
+    jvmErasure.hasAnnotation<DataSchema>() -> ObjectToGroup
+    else -> Default
 }
 
-internal fun KType.shouldBeConvertedToColumnGroup(): Boolean = jvmErasure.let {
-    it == DataRow::class || it.hasAnnotation<DataSchema>()
+internal sealed interface FieldKind {
+    val shouldBeConvertedToColumnGroup: Boolean get() = false
+    val shouldBeConvertedToFrameColumn: Boolean get() = false
+}
+internal data object Frame : FieldKind {
+    override val shouldBeConvertedToFrameColumn: Boolean = true
+}
+internal data object ListToFrame : FieldKind {
+    override val shouldBeConvertedToFrameColumn: Boolean = true
+}
+
+internal data object Default : FieldKind
+
+internal data object Group : FieldKind {
+    override val shouldBeConvertedToColumnGroup: Boolean = true
+}
+
+internal data object ObjectToGroup : FieldKind {
+    override val shouldBeConvertedToColumnGroup: Boolean = true
 }
 
 private fun String.toNullable(): String = if (endsWith("?")) this else "$this?"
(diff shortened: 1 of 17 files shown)


dataframe-openapi/src/test/kotlin/OpenApiTests.kt
```diff
@@ -478,507 +479,609 @@ class OpenApiTests : JupyterReplTestCase() {
...
+                    public companion object {
+                      public val keyValuePaths: kotlin.collections.List<org.jetbrains.kotlinx.dataframe.api.JsonPath>
                         get() = listOf()
-
-                    public fun org.jetbrains.kotlinx.dataframe.DataFrame<*>.convertToDog(convertTo: org.jetbrains.kotlinx.dataframe.api.ConvertSchemaDsl<Dog>.() -> kotlin.Unit = {}): org.jetbrains.kotlinx.dataframe.DataFrame<Dog> = convertTo<Dog> { 
-                        convertDataRowsWithOpenApi() 
-                        convertTo()
+                
+                      public fun org.jetbrains.kotlinx.dataframe.DataFrame<*>.convertToPet(convertTo: org.jetbrains.kotlinx.dataframe.api.ConvertSchemaDsl<Pet>.() -> kotlin.Unit = {}): org.jetbrains.kotlinx.dataframe.DataFrame<Pet> = convertTo<Pet> { 
+                          convertDataRowsWithOpenApi()
```
zaleslaw, line 517: Something wrong with formatting in the strings with code

dataframe-openapi/src/test/kotlin/OpenApiTests.kt
```diff
@@ -478,507 +479,609 @@ class OpenApiTests : JupyterReplTestCase() {
...
+                
+                }
+            }
+            typealias PetRef = `Other advanced test`.Pet
+            typealias AlsoCat = `Other advanced test`.Cat
+            typealias Integer = kotlin.Int
+            typealias SomeArray = org.jetbrains.kotlinx.dataframe.DataFrame<`Other advanced test`.SomeArrayContent>
+            typealias SomeArrayArray = kotlin.collections.List<SomeArray>
+
+            val org.jetbrains.kotlinx.dataframe.ColumnsContainer<`Other advanced test`.Pet>.`value`: org.jetbrains.kotlinx.dataframe.DataColumn<kotlin.Any?> @JvmName("Pet_value") get() = this["value"] as org.jetbrains.kotlinx.dataframe.DataColumn<kotlin.Any?>
```
zaleslaw, line 921: How was this snippet of code was generated (approach/test code)?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/CodeGenerationTests.kt
```diff
@@ -305,6 +309,38 @@ class CodeGenerationTests : BaseTest() {
...
+            @DataSchema
+            data class Person(
+                val group: List<Person1>,
+                val name: String
+            )
+        """.trimIndent().toCodeString()
+    }
+
+    @Test
+    fun `normalize name`() {
```
zaleslaw, line 334: renameTo -> `check name normalization for generated data classes`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MarkersExtractor.kt
```diff
@@ -17,14 +17,33 @@ import kotlin.reflect.full.withNullability
 import kotlin.reflect.jvm.jvmErasure
 import kotlin.reflect.typeOf
 
-internal fun KType.shouldBeConvertedToFrameColumn(): Boolean = when (jvmErasure) {
-    DataFrame::class -> true
-    List::class -> arguments[0].type?.jvmErasure?.hasAnnotation<DataSchema>() == true
-    else -> false
+internal fun KType.getFieldKind(): FieldKind = when {
+    jvmErasure == DataFrame::class -> Frame
+    jvmErasure == List::class && (arguments[0].type?.jvmErasure?.hasAnnotation<DataSchema>() == true) -> ListToFrame
```
zaleslaw, line 22: Could be here only List type, is it possible to meet a Map, Array?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MarkersExtractor.kt
```diff
@@ -17,14 +17,33 @@ import kotlin.reflect.full.withNullability
...
+    val shouldBeConvertedToFrameColumn: Boolean get() = false
+}
+internal data object Frame : FieldKind {
+    override val shouldBeConvertedToFrameColumn: Boolean = true
+}
+internal data object ListToFrame : FieldKind {
+    override val shouldBeConvertedToFrameColumn: Boolean = true
+}
+
+internal data object Default : FieldKind
```
zaleslaw, line 39: It will be a good to explain the bunch of these descendants of FieldKind

zaleslaw (changes requested):
Do we have an analogue of suggested `printDataClassSchema` from https://github.com/holgerbrandl/kdfutils?tab=readme-ov-file#data-frame-schema?

I could not understand this and find in tests in proposed PR, sorry

koperagen:
`df.generateDataClasses().print()`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MarkersExtractor.kt
koperagen: Only `List`, i.e. given
```
@DataSchema
data class Record(val a: String)
```

`data class MyClass(val l: LIst<Record>)` and `data class MyClass(val l: DataFrame<Record>)`
are equivalent declarations from `toDataFrame` and `MarkersExtractor.get` point of view.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MarkersExtractor.kt
koperagen: 1 for each branch of getFieldKind :) Previously this function returned just Boolean, but for rendering purposes we need to return more info.

dataframe-openapi/src/test/kotlin/OpenApiTests.kt
koperagen: It just a refactoring of an existing test

dataframe-openapi/src/test/kotlin/OpenApiTests.kt
koperagen: should haveSubstring doesn't show diff between two texts, and with generated code being 500 lines or so it's impossible to find why assertion fails.
so i refactored it to use shouldBe
In the editor code looks ok

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/CodeGenerationTests.kt
```diff
@@ -305,6 +309,38 @@ class CodeGenerationTests : BaseTest() {
         }
     }
 
+    @Test
+    fun `data classses`() {
```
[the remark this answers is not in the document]
koperagen: done

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/CodeGenerationTests.kt
koperagen: done, thank you!

zaleslaw:
@koperagen could you add a page in our docs or chapter with simple description and small example?

koperagen:
Let's do it later. I want to find a better form and place for it

zaleslaw:
We could add documentation later, but in this case, please create a doc ticket for this, link to this issue and assign on your-self

PR #767, opened 2024-07-04
Adding utils to help ensure that compile time schema ~ runtime schema

koperagen:


plugins/kotlin-dataframe/tests/org/jetbrains/kotlin/fir/dataframe/AbstractDataFrameBlackBoxCodegenTest.kt
```diff
@@ -65,5 +66,15 @@ open class AbstractDataFrameBlackBoxCodegenTest : AbstractFirLightTreeBlackBoxCo
     override fun createKotlinStandardLibrariesPathProvider(): KotlinStandardLibrariesPathProvider {
         return EnvironmentBasedStandardLibrariesPathProvider
     }
+
+    class SourceProvider(testServices: TestServices) : AdditionalSourceProvider(testServices) {
+        companion object {
+            const val COMMON_SOURCE_PATH = "testData/common.kt"
```
zaleslaw, line 72: what is containing in this file? could you explain or add a comment

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -397,4 +398,30 @@ class CreateDataFrameTests {
...
     }
+
+    @DataSchema
+    data class Person(val firstName: String, val lastName: String, val age: Int, val city: String?) : DataRowSchema
+
+    @DataSchema
+    data class Group(val id: String, val participants: List<Person>) : DataRowSchema
+
+    @Test
+    fun `deeply convert data schema and list of data schema`() {
```
zaleslaw, line 409: Is it worth to compare here, for example, a circular references?

Group depends on Person, Person depends on Group

plugins/kotlin-dataframe/tests/org/jetbrains/kotlin/fir/dataframe/AbstractDataFrameBlackBoxCodegenTest.kt
koperagen: I'll rename SourceProvider -> TestUtilsSourceProvider and testData/common.kt to testData/testUtils.kt. Ok?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
koperagen: I started working on this PR with this circular reference idea in mind. Because regular Kotlin object can have them quite easily. But since we ended up converting only DataSchema objects deeply, i think it's not an issue anymore. You really need to do it on purpose to have a cycle between DataSchema objects. So i think it's not worth to bother here

plugins/kotlin-dataframe/tests/org/jetbrains/kotlin/fir/dataframe/AbstractDataFrameBlackBoxCodegenTest.kt
koperagen: Makes it possible to call compareSchemas from any test

PR #772, opened 2024-07-09
Add extension functions for the ResultSet

### LINKED ISSUE #458: Create extension functions for java.sql main objects to extract data `toDF()`
- [x] Connection
- [x] ResultSet

@Jolanrensen suggested thinking about creation extension functions for some of the main java.sql.* classes

zaleslaw:
Fixes #458

### DIFF
--- a/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
+++ b/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
@@ -265,6 +265,32 @@ public fun DataFrame.Companion.readResultSet(
  *
  * NOTE: Reading from the [ResultSet][java.sql.ResultSet] could potentially change its state.
  *
+ * @param [dbType] the type of database that the [ResultSet] belongs to.
+ * @param [limit] the maximum number of rows to read from the [ResultSet][java.sql.ResultSet].
+ * @param [inferNullability] indicates how the column nullability should be inferred.
+ * @return the DataFrame generated from the [ResultSet][java.sql.ResultSet] data.
+ *
+ * [java.sql.ResultSet]: https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html
+ */
+public fun ResultSet.toDF(
+    dbType: DbType,
+    limit: Int = DEFAULT_LIMIT,
+    inferNullability: Boolean = true,
+): AnyFrame {
+    return DataFrame.Companion.readResultSet(this, dbType, limit, inferNullability)
+}
+
+/**
+ * Reads the data from a [ResultSet][java.sql.ResultSet] and converts it into a DataFrame.
+ *
+ * A [ResultSet][java.sql.ResultSet] object maintains a cursor pointing to its current row of data.
+ * By default, a ResultSet object is not updatable and has a cursor that can only move forward.
+ * Therefore, you can iterate through it only once, from the first row to the last row.
+ *
+ * For more details, refer to the official Java documentation on [ResultSet][java.sql.ResultSet].
+ *
+ * NOTE: Reading from the [ResultSet][java.sql.ResultSet] could potentially change its state.
+ *
  * @param [resultSet] the [ResultSet][java.sql.ResultSet] containing the data to read.
  * Its state may be altered after the read operation.
  * @param [connection] the connection to the database (it's required to extract the database type)
@@ -284,6 +310,33 @@ public fun DataFrame.Companion.readResultSet(
     val dbType = extractDBTypeFromConnection(connection)
 
     return readResultSet(resultSet, dbType, limit, inferNullability)
+}
+
+/**
+ * Reads the data from a [ResultSet][java.sql.ResultSet] and converts it into a DataFrame.
+ *
+ * A [ResultSet][java.sql.ResultSet] object maintains a cursor pointing to its current row of data.
+ * By default, a ResultSet object is not updatable and has a cursor that can only move forward.
+ * Therefore, you can iterate through it only once, from the first row to the last row.
+ *
+ * For more details, refer to the official Java documentation on [ResultSet][java.sql.ResultSet].
+ *
+ * NOTE: Reading from the [ResultSet][java.sql.ResultSet] could potentially change its state.
+ *
+ * @param [connection] the connection to the database (it's required to extract the database type)
+ * that the [ResultSet] belongs to.
+ * @param [limit] the maximum number of rows to read from the [ResultSet][java.sql.ResultSet].
+ * @param [inferNullability] indicates how the column nullability should be inferred.
+ * @return the DataFrame generated from the [ResultSet][java.sql.ResultSet] data.
+ *
+ * [java.sql.ResultSet]: https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html
+ */
+public fun ResultSet.toDF(
+    connection: Connection,
+    limit: Int = DEFAULT_LIMIT,
+    inferNullability: Boolean = true,
+): AnyFrame {
+    return DataFrame.Companion.readResultSet(this, connection, limit, inferNullability)
 }
 
 /**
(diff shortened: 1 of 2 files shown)


dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/h2/h2Test.kt
```diff
@@ -579,6 +660,51 @@ class JdbcTest {
...
+        @Language("SQL")
+        val sqlQuery = """
+            SELECT c.name as customerName, SUM(s.amount) as totalSalesAmount
+            FROM Sale s
+            INNER JOIN Customer c ON s.customerId = c.id
+            WHERE c.age > 35
+            GROUP BY s.customerId, c.name
+        """.trimIndent()
+
+        val df = connection.toDF(sqlQuery).cast<CustomerSales>()
```
koperagen, line 674: Looks similar to File.readDataFrame, what do you think about this name?
https://github.com/Kotlin/dataframe/blob/f631f68cd62ee563356e23943fe4ac883aea959d/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/guess.kt#L291

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/h2/h2Test.kt
```diff
@@ -579,6 +660,51 @@ class JdbcTest {
...
+
+        df.rowsCount() shouldBe 2
+        df.filter { it[CustomerSales::totalSalesAmount]!! > 100 }.rowsCount() shouldBe 1
```
koperagen, line 677: Use extension properties API if  possible

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -254,6 +315,32 @@ public fun DataFrame.Companion.readResultSet(
...
+ *
+ * @param [dbType] the type of database that the [ResultSet] belongs to.
+ * @param [limit] the maximum number of rows to read from the [ResultSet][java.sql.ResultSet].
+ * @param [inferNullability] indicates how the column nullability should be inferred.
+ * @return the DataFrame generated from the [ResultSet][java.sql.ResultSet] data.
+ *
+ * [java.sql.ResultSet]: https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html
+ */
+public fun ResultSet.toDF(
+    dbType: DbType,
```
koperagen, line 337: Hmmm, i wonder, so this DbType is a mandatory parameter. Usually i find it very convenient when such classes live in a common scope and expect something like `DbType.` to show me all possible options. It's not in the scope of this PR, so please treat it as a note / suggestion

koperagen: Would you like to name it toDF or toDataFrame is ok?

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -216,6 +216,67 @@ public fun DataFrame.Companion.readSqlQuery(
     }
 }
 
+/**
+ * Converts the result of an SQL query to the DataFrame.
```
Jolanrensen, line 220: *a DataFrame

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -216,6 +216,67 @@ public fun DataFrame.Companion.readSqlQuery(
...
+ * Also, it could be a just name of the SQL table.
```
Jolanrensen, line 221: I don't understand this sentence

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -216,6 +216,67 @@ public fun DataFrame.Companion.readSqlQuery(
...
+ *
+ * NOTE: It should be a name of one of the existing SQL tables,
```
Jolanrensen, line 223: Shouldn't this be part of the `@param [sqlQueryOrTableName]` part? Or at least specify that this is information regarding that parameter. It's a bit unclear to me now.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
Jolanrensen: toDataFrame() is better IMO too

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -216,6 +216,67 @@ public fun DataFrame.Companion.readSqlQuery(
...
+ * or the SQL query should start from SELECT and contain one query for reading data without any manipulation.
+ * It should not contain `;` symbol.
+ *
+ * @param [sqlQueryOrTableName] the SQL query to execute or name of the SQL table.
+ * @param [limit] the maximum number of rows to retrieve from the result of the SQL query execution.
+ * @param [inferNullability] indicates how the column nullability should be inferred.
+ * @return the DataFrame containing the result of the SQL query.
+ */
+public fun DatabaseConfiguration.toDF(
```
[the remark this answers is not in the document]
Jolanrensen: Actually, it might even be a good idea to get different names per overload:

- `ResultSet.toDataFrame()` makes sense, because we're converting a resultset to a dataframe.
- For `DataBaseConfiguration.toDataFrame()` we aren't converting the configuration to a dataframe, so the name is misleading. We'd better consider something like `DataBaseConfiguration.getDataFrameFor(sqlQueryOrTableName)` or something similar.
- Same for `Connection`

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -437,6 +551,38 @@ public fun DataFrame.Companion.getSchemaForSqlQuery(connection: Connection, sqlQ
     }
 }
 
+/**
+ * Retrieves the schema of an SQL query result or the SQL table using the provided database configuration.
+ *
+ * @param [sqlQueryOrTableName] the SQL query to execute and retrieve the schema from.
+ * @return the schema of the SQL query as a [DataFrameSchema] object.
+ */
+public fun DatabaseConfiguration.getDataFrameSchema(
```
Jolanrensen, line 560: maybe `getDataFrameSchemaFor(sqlQueryOrTableName)`, otherwise a good name :)

Jolanrensen (changes requested):
Just some concerns about the names, but they're good additions :)

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
zaleslaw: better be unified with other schema names (and rename them together)

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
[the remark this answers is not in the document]
zaleslaw: The problem is following:
- extract data from Connection via extension function is normal (it's a pipe)
- extract data from DbConfig - weird, but we're calling methods on DataFrame.readSQLTable (we are not reading SQL table from from DataFrame? It's just a place for method)

Different names make it's more and more complicated 

Much more better will be to have something like DatabaseReader.readSqlTable instead of DataFrame.readSqlTable

Also, as an alternative - I could remove extension functions for Connection/DbConfig and stay only with ResultSet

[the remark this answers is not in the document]
zaleslaw: Let's live a little bit, if it will be bad on Demos, I will remove

zaleslaw:
@koperagen @Jolanrensen thanks for review, it's not a final version, I see, the next step - revisit obtaining the Schema and change the place of all these methods

PR #779, opened 2024-07-15
Make sortBy(ColumnReference) accept pathOf without extra cast

koperagen:
Before this change only `df.sortBy(pathOf().cast())` would work. `df.sortByDesc(pathOf().cast())` also didn't! Was very hard to figure out correct way to call these functions. PR solves it.
I found this problem while working with this dataset in Datalore which can be found in their sample database. It's public domain and seems nice, let's add it. 
https://www.kaggle.com/datasets/ruchi798/data-science-job-salaries. It shows a good use case


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -67,4 +72,27 @@ class SortDataColumn {
         col.sortWith { df1, df2 -> df1[a] - df2[a] } shouldBe sortedCol
         col.sortWith(compareBy { it[a] }) shouldBe sortedCol
     }
+
+    @Test
```
zaleslaw, line 76: Cool new dataset for unit tests! Thanks!

zaleslaw: Great to validate on that some high-level integration tests with more complex logic instead of low-level A.a.b

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/testSets/DsSalaries.kt
```diff
@@ -0,0 +1,31 @@
+package org.jetbrains.kotlinx.dataframe.testSets
+
+import org.jetbrains.kotlinx.dataframe.annotations.ColumnName
+import org.jetbrains.kotlinx.dataframe.annotations.DataSchema
+
+@Suppress("unused")
+@DataSchema
+interface DsSalaries {
```
zaleslaw, line 8: We need to add somewhere the link to the initial source of dataset (for example here, on that data schema), the origin, the license, the possible transformation which we did 

It's a gentleman set of actions for copying of external dataset to your codebase. If we missed it earlier, let's start from that

zaleslaw (approved):
I approved in advance, but please add minor information about dataset

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/testSets/DsSalaries.kt
koperagen: I added a link here. It was downloaded as is, without modifications and is public domain. So, do we need to add anything else here?

PR #782, opened 2024-07-16
Refactor toDataFrame implementation in compiler plugin

koperagen:


plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/loadInterpreter.kt
```diff
@@ -161,6 +164,9 @@ internal inline fun <reified T> String.load(): T {
         "ReadDelimStr" -> ReadDelimStr()
         "GroupByToDataFrame" -> GroupByToDataFrame()
         "ToDataFrameFrom0" -> ToDataFrameFrom()
```
zaleslaw, line 166: To and to?, again, could they be a CONSTANTS?

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/interpret.kt
```diff
@@ -269,6 +270,10 @@ fun <T> KotlinTypeFacade.interpret(
     }
 }
 
+fun interpretationFrameworkError(message: String): Nothing = throw InterpretationFrameworkError(message)
```
zaleslaw, line 273: the name looks like object name, not a function

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/interpret.kt
```diff
@@ -89,13 +89,14 @@ fun <T> KotlinTypeFacade.interpret(
...
     if (conflictingKeys.isNotEmpty()) {
-        error("Conflicting keys: $conflictingKeys")
+        interpretationFrameworkError("Conflicting keys: $conflictingKeys")
     }
     val expectedArgsMap = processor.expectedArguments
         .filterNot { it.name.startsWith("typeArg") }
         .associateBy { it.name }.toSortedMap().minus(additionalArguments.keys)
 
-    if (expectedArgsMap.keys - defaultArguments != actualArgsMap.keys - defaultArguments) {
+    val unexpectedArguments = expectedArgsMap.keys - defaultArguments != actualArgsMap.keys - defaultArguments
```
zaleslaw, line 98: very interesting algebra of arguments (great to explain somehow this kind of logic)

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/analyzeRefinedCallShape.kt
```diff
@@ -38,44 +32,6 @@ fun KotlinTypeFacade.analyzeRefinedCallShape(call: FirFunctionCall, reporter: In
...
-                    toDataFrame(maxDepth, call, TraverseConfiguration())
-                } else {
-                    PluginDataFrameSchema(emptyList())
-                }
-            }
-            "toDataFrameDefault" -> {
-                val maxDepth = 0
-                toDataFrame(maxDepth, call, TraverseConfiguration())
-            }
             "Aggregate" -> {
```
zaleslaw, line 35: sorry, again better to have CONSTANTS, believe that these names are repeated in codebase

zaleslaw (approved):

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/interpret.kt
koperagen: Can't argue, but i was inspired by "error" function from standard library

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/interpret.kt
koperagen: This check is needed to simplify debug if someone writes an interpreter for `fun a(myArgument: String)` but misspells myArgument as myArg123123ument. Expected argument (by interpreter) is myArg123123ument, actual myArgument.
Easy to check.
But with default value:
fun a(myArgument: String, myArgument1: String = "defaultValue")
a("abc")
actualArg = myArgument, expectedArg = myArg123123ument, myArgument1.
We know that for myArgument1 default value is provided, so we exclude such arguments from comparison: it's not a mistake.

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/analyzeRefinedCallShape.kt
koperagen: Answered in other PR :)

PR #783, opened 2024-07-16
Add initial support for CS DSL in the compiler plugin

koperagen:


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/all.kt
```diff
@@ -300,6 +301,7 @@ public interface AllColumnsSelectionDsl<out _UNUSED> {
      *
      * `df.`[select][DataFrame.select]`  {  `[all][ColumnsSelectionDsl.all]`() }`
      */
+    @Interpretable("All0")
```
zaleslaw, line 304: ALLO!

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/group.kt
```diff
@@ -12,21 +12,21 @@ import org.jetbrains.kotlinx.dataframe.plugin.impl.dataFrame
...
-class GroupClauseApproximation(val df: PluginDataFrameSchema, val columns: List<ColumnWithPathApproximation>)
+class GroupClauseApproximation(val df: PluginDataFrameSchema, val columns: ColumnsResolver)
 
 class Into0 : AbstractSchemaModificationInterpreter() {
     val Arguments.receiver: GroupClauseApproximation by arg()
     val Arguments.column: String by arg()
 
     override fun Arguments.interpret(): PluginDataFrameSchema {
-        val grouped = groupImpl(receiver.df.columns(), receiver.columns.mapTo(mutableSetOf()) { it.path.path }, column)
+        val grouped = groupImpl(receiver.df.columns(), receiver.columns.resolve(receiver.df).mapTo(mutableSetOf()) { it.path.path }, column)
```
zaleslaw, line 29: that is reciever.df here?

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/loadInterpreter.kt
```diff
@@ -161,6 +166,11 @@ internal inline fun <reified T> String.load(): T {
         "ReadDelimStr" -> ReadDelimStr()
         "GroupByToDataFrame" -> GroupByToDataFrame()
         "ToDataFrameFrom0" -> ToDataFrameFrom()
+        "All0" -> All0()
+        "ColsOf0" -> ColsOf0()
+        "ColsOf1" -> ColsOf1()
+        "ColsAtAnyDepth0" -> ColsAtAnyDepth0()
+        "FrameCols0" -> FrameCols0()
```
zaleslaw, line 173: Cool list, looks like a good candidate to be a list of CONSTANTS or objects? Could you please remind? Could we reuse CONSTANTS in the @Interpretable annotations? (it makes the navigation easier)

zaleslaw (approved):

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/group.kt
koperagen: Given this call:
df.group { a }.into("abc")
receiver = df.group { a }. Because it's a function call, its result is also interpreted. It becames GroupClauseApproximation, and so `receiver.df` is *df*.group { a }
Receiver is a special name for "this"

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/loadInterpreter.kt
koperagen: Hm, my concern was that they should live in a dataframe-core, but also not available in public API. But looks like it's not a problem with friend module. I'll write this down and try to refactor later

PR #786, opened 2024-07-17
google analytics script

azhebel:
also:
- meta name="google-site-verification"
- iframe resize function

not sure if they are necessary though

### DIFF
--- a/docs/StardustDocs/cfg/resize-iframes.js
+++ b/docs/StardustDocs/cfg/resize-iframes.js
@@ -1,13 +1,14 @@
 <meta name="google-site-verification" content="Lffz_ab-_S5cmA07ZXVbucHVklaRsnk8gEt8frHKjMk"/>
 <!-- Google Analytics -->
 <script>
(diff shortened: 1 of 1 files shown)


koperagen:
Hi! We're using a bit outdated custom writerside build, 2023.1.4. This tag doesn't seem to work there. 
image
So let's keep analytics-head-script-file. These are needed too, indeed
```
meta name="google-site-verification"
iframe resize function
```

azhebel:
@koperagen sure, you can keep the old tag `<analytics-head-script-file>` and even the old script file if you want, just make sure the actual analytics script is the one I added here, not the UA one you had previously

Do you want me to modify this MR or will you add the necessary changes manually?

koperagen:
Modify the MR please :)

azhebel:
@koperagen I kept everything else except the actual script. Let's see if this helps.

koperagen:
Thank you! I'll deploy docs now

Issue #785, opened 2024-07-16
isOpenApiStr logger leaks to Gradle 


koperagen:
https://www.kaggle.com/datasets/ruchi798/data-science-job-salaries
Try to generate a schema:
```
schema {
    sourceSet = "test"
    visibility = org.jetbrains.dataframe.gradle.DataSchemaVisibility.IMPLICIT_PUBLIC
    data = file("src/test/resources/ds_salaries.csv")
    name = "org.jetbrains.kotlinx.dataframe.testSets.DsSalaries"
}
```
![image](https://github.com/user-attachments/assets/92b17041-4c1d-409e-b042-9761e8342ccb)

It appears as an error, while code generation was in fact successful. Looks bad, i'd revisit why this needs logging and remove if possible
```
logger.debug(e) { "Attempt to read input as YAML/JSON OpenAPI specification failed." }
false
```

Jolanrensen:
Unfortunately, the `logger.debug(e)` line you point to is not the actual place the logging occurs. It's in the swagger-parser library, [here](https://github.com/swagger-api/swagger-parser/blob/0ab2e8a7774b147e7b8c2c75b929089677d1b246/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/OpenAPIV3Parser.java#L197). I haven't found a way yet to turn off the sl4j warning logs of `io.swagger.v3.parser`.

Jolanrensen:
Well... That was a long journey. For some reason, the logger of `io.swagger.v3.parser.OpenAPIV3Parser` gets the name `"i.swagger.v3.parser.OpenAPIV3Parser"`. So, to disable it, you need that exact name.
The logger of `io.swagger.v3.parser.util.DeserializationUtils` becomes `"i.s.v.p.util.DeserializationUtils"` for some reason... That explains why I couldn't disable it.

koperagen:
gradle plugin is no longer published

PR #793, opened 2024-07-25
Disable the `OpenAPIV3Parser` logs

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/785

Needs some more testing to see if this actually disables it and keeps other logs alive


Jolanrensen:
this does not solve the issue

Issue #713, opened 2024-06-03
☂ `type: KType` in `DataColumnImpl` mismatches actual values sometimes


Jolanrensen:
`type: KType` in `DataColumnImpl` mismatches actual values in some cases. This can result in runtime exceptions and makes life difficult attempting to fix https://github.com/Kotlin/dataframe/issues/30 or https://github.com/Kotlin/dataframe/issues/704 where we assume the `type` always correctly represents the data.
This issue relates to https://github.com/Kotlin/dataframe/issues/701 as well.

To discover these bugs, we can introduce a (debug-only!!) check in `DataColumnImpl`, like:

```kt
private infix fun <T> T?.matches(type: KType) =
    when {
        this == null -> type.isMarkedNullable
        this.isPrimitiveArray -> type.isPrimitiveArray &&
            this!!::class.qualifiedName == type.classifier?.let { (it as KClass<*>).qualifiedName }

        this.isArray -> type.isArray // cannot check the precise type of array
        else -> this!!::class.isSubclassOf(type.classifier as KClass<*>)
    }

init {
    if (DEBUG) {
        require(values.all { it matches type }) {
            val types = values.map { if (it == null) "Nothing?" else it!!::class.simpleName }.distinct()
            "Values of column '$name' have types '$types' which are not compatible given with column type '$type'"
        }
    }
}
```
At the moment of testing, I can find 8+ breaking tests in `:core`:
- `org.jetbrains.kotlinx.dataframe.io.PlaylistJsonTest#aggregate by column`
  - data has types `[DataFrameImpl]`, `kType` is: `org.jetbrains.kotlinx.dataframe.aggregation.AggregateGroupedDsl<org.jetbrains.kotlinx.dataframe.io.PlaylistJsonTest.DataFrameType1>`
  - This exception is probably fine to ignore, as `AggregateGroupedDsl` is a `DataFrame`
- `org.jetbrains.kotlinx.dataframe.samples.api.Analyze#pivotDefault_accessors` / `#pivotDefault_strings` / `#pivotDefault_properties`
  - data has types `[Boolean, Int]`, `kType` is: `kotlin.Boolean`
  - Not really sure what's going on here, something with `concatImpl` I think
- `org.jetbrains.kotlinx.dataframe.samples.api.Modify#customConverters`
  - data has types `[Nothing?]`, `kType` is `kotlin.Int`
  - Seems to originate from `convertToImpl$convertToSchema`
[...]

PR #796, opened 2024-07-30
Build config and Debug mode

Jolanrensen:
This PR adds the [buildconfig](https://github.com/gmazzo/gradle-buildconfig-plugin) plugin to the project.

This achieves two things:
- It allows us to access the gradle project version via `BuildConfig.VERSION`. We can use this well in the Jupyter integration where we're often interested which dataframe version is pulled into the notebook when doing a `%use dataframe`. 
- It helps with https://github.com/Kotlin/dataframe/issues/713. 
  By simply adding `-Pkotlin.dataframe.debug=true` to the build on TC, we can run the extra checks from the issue. By default this check should be off, as it costs performance, but it's very valuable for catches bugs. We can use `BuildConfig.DEBUG` in other places in the future as well.

See each commit for more info.

We should add the flag to TC only after https://github.com/Kotlin/dataframe/issues/713 is solved completely, otherwise all builds will fail.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/DataColumnImpl.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/DataColumnImpl.kt
@@ -1,8 +1,13 @@
 package org.jetbrains.kotlinx.dataframe.impl.columns
 
+import org.jetbrains.kotlinx.dataframe.BuildConfig
 import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.api.dataFrameOf
+import org.jetbrains.kotlinx.dataframe.impl.isArray
+import org.jetbrains.kotlinx.dataframe.impl.isPrimitiveArray
+import kotlin.reflect.KClass
 import kotlin.reflect.KType
+import kotlin.reflect.full.isSubclassOf
 
 internal abstract class DataColumnImpl<T>(
     protected val values: List<T>,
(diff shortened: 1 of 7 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/DataColumnImpl.kt
```diff
@@ -12,6 +17,32 @@ internal abstract class DataColumnImpl<T>(
...
+                    this!!::class.qualifiedName == type.classifier?.let { (it as KClass<*>).qualifiedName }
+
+            this.isArray -> type.isArray
+
+            // cannot check the precise type of array
+            else -> this!!::class.isSubclassOf(type.classifier as KClass<*>)
+        }
+
+    init {
+        /* Check for [Issue #713](https://github.com/Kotlin/dataframe/issues/713).
```
zaleslaw, line 35: strange formatting of KDoc

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/DataColumnImpl.kt
```diff
@@ -12,6 +17,32 @@ internal abstract class DataColumnImpl<T>(
...
+         * This only runs with `kotlin.dataframe.debug=true` in gradle.properties
+         */
+        if (BuildConfig.DEBUG) {
+            require(values.all { it matches type }) {
```
zaleslaw, line 39: It checks all values, right? Is it worth to check random 10-20 numbers from the column?

gradle.properties
```diff
@@ -11,3 +11,8 @@ org.gradle.jvmargs=-Xmx4G
 # This makes it mandatory to explicitly apply your own version of the
 # KSP plugin in the modules that use it.
 kotlin.dataframe.add.ksp=false
+
+# Enables debug mode for dataframe.
+# This can make certain tests run that should not be run in production.
+# It can also be turned on from the command line with `-Pkotlin.dataframe.debug=true`
+kotlin.dataframe.debug=false
```
zaleslaw, line 18: does it mean, that we could add more modes, with different names, it's not a strongly-defined schema?

zaleslaw (approved):

gradle.properties
Jolanrensen: not strong typed indeed, just an argument I came up with. Similar to the add.ksp argument.
It's not strongly typed, but when read it tries to parse the argument as Boolean and if failed it defaults to `false`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/DataColumnImpl.kt
Jolanrensen: If we only check a handful of values there's a chance we won't catch the outlier and get flaky tests.

Jolanrensen:
Thanks!

PR #800, opened 2024-07-30
Small convertTo fix

Jolanrensen:
With the debug mode enabled, one failing test of https://github.com/Kotlin/dataframe/issues/713 was in `convertTo`.

It likely didn't cause any user issues, but it could prevent https://github.com/Kotlin/dataframe/pull/712, so best to solve it.

It was a relatively small fix. I renamed `createEmptyColumn` to `createNullFilledColumn` as that better reflects is behavior. Then, instead of creating the erroneous column, it throws an exception. This eliminates the need for the large boolean check in the `convertToImpl` function. Finally, since `update` cannot target non-existing columns, I needed to update the `fill` option of `convertTo`, to `update` or `add` columns depending on whether they were there already or not.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convertTo.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convertTo.kt
@@ -276,8 +276,8 @@ internal fun AnyFrame.convertToImpl(
 
     dsl.fillers.forEach { filler ->
         val paths = result.getColumnPaths(UnresolvedColumnsPolicy.Create, filler.columns).toSet()
-        missingPaths -= paths
         val (newPaths, existingPaths) = paths.partition { it in missingPaths }
+        missingPaths -= paths
 
         // first fill cols that are already in the df
         result = result.update { existingPaths.toColumnSet() }.with { filler.expr(this, this) }
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convertTo.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convertTo.kt
@@ -1,5 +1,6 @@
 package org.jetbrains.kotlinx.dataframe.impl.api
 
+import io.github.oshai.kotlinlogging.KotlinLogging
 import org.jetbrains.kotlinx.dataframe.AnyFrame
 import org.jetbrains.kotlinx.dataframe.AnyRow
 import org.jetbrains.kotlinx.dataframe.ColumnsSelector
@@ -46,6 +47,8 @@ import org.jetbrains.kotlinx.dataframe.size
 import kotlin.reflect.KType
 import kotlin.reflect.full.withNullability
 import kotlin.reflect.jvm.jvmErasure
+
+private val logger = KotlinLogging.logger {}
 
 private open class Converter(val transform: ConverterScope.(Any?) -> Any?, val skipNulls: Boolean)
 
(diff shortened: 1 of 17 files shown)


zaleslaw:
@Jolanrensen could you please check do we need to update a documentation for that? If yes, please include changes in the PR

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convertTo.kt
```diff
@@ -252,22 +254,15 @@ internal fun AnyFrame.convertToImpl(
...
-                    // like DataRow<Something?> for a group column (all columns in the group will be nullable)
-                    targetColumn.contentType?.isMarkedNullable == true ||
-                    // frame column can be filled with empty dataframes
-                    targetColumn.kind == ColumnKind.Frame
-
             if (name !in visited) {
-                newColumns += targetColumn.createEmptyColumn(name, size)
-                if (!isNullable) {
+                try {
+                    newColumns += targetColumn.createNullFilledColumn(name, size)
```
zaleslaw, line 263: Could we improve the logic in that place a little bit? I have a feeling that we need to log something in catch section, or use check()/error() for createNullFilledColumn if it's possible?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convertTo.kt
```diff
@@ -280,10 +275,16 @@ internal fun AnyFrame.convertToImpl(
     var result = convertToSchema(marker.schema, emptyPath())
 
     dsl.fillers.forEach { filler ->
-        val paths = result.getColumnPaths(filler.columns)
-        missingPaths.removeAll(paths.toSet())
-        result = result.update { paths.toColumnSet() }.with {
-            filler.expr(this, this)
+        val paths = result.getColumnPaths(UnresolvedColumnsPolicy.Create, filler.columns).toSet()
```
zaleslaw, line 278: could you please add a comment with the high-level idea of this algorithm starting from dsl.fillers

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/schema/Utils.kt
```diff
@@ -110,14 +113,22 @@ internal fun ColumnSchema.createEmptyColumn(name: String): AnyCol =
...
+internal fun ColumnSchema.createNullFilledColumn(name: String, numberOfRows: Int): AnyCol =
     when (this) {
-        is ColumnSchema.Value -> DataColumn.createValueColumn(
-            name = name,
-            values = List(numberOfRows) { null },
-            type = type,
-        )
+        is ColumnSchema.Value -> {
+            if (!type.isMarkedNullable && numberOfRows > 0) {
+                error("Cannot create a null-filled value column of type $type as it's not nullable.")
```
zaleslaw, line 124: this is that error() call!

zaleslaw (changes requested):
I could understand very high-level what happens in changes, but it could be difficult in the future!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convertTo.kt
Jolanrensen: Sure, that makes sense

Jolanrensen: Though, a simple log of the exception should be enough, as any `missingPaths` left will inform the user with  an IllegalArgumentException in line 296

Jolanrensen:
> I could understand very high-level what happens in changes, but it could be difficult in the future!

`convertToImpl` (as well as many of the other internal implementations of functions in DF) are hard to follow indeed. I hope I now made it at least a bit less complicated :)

Jolanrensen:
Interestingly, adding kotlinLogging breaks the compiler plugin in the black box tests wherever it is used with:
`java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory`

@koperagen can we add this dependency to the plugin too?

zaleslaw (approved):

PR #801, opened 2024-07-30
Small implode fix

Jolanrensen:
With the debug mode enabled, two failing tests of https://github.com/Kotlin/dataframe/issues/713 were found in `implode`.

This PR makes sure the temporary column created in the function adheres to the given kType.

PR #803, opened 2024-07-31
GroupBy aggregate fix

Jolanrensen:
With the debug mode enabled, one failing test of https://github.com/Kotlin/dataframe/issues/713 was in `GroupBy.aggregate { this into "" }`.

I narrowed it down to a `NamedValue` being created in `GroupByReceiverImpl` with KType `AggregateGroupedDsl<DfType>`. Since this value is an `AggregateInternalDsl<*>` too, it is unpacked in the `yield()` function to get the underlying dataframe (by copying the `NamedValue` and replacing its `value` with its `value.df`), however, the KType of this `NamedValue` was left unchanged and thus mismatched the value.

This PR fixes that by updating that KType value too


Jolanrensen:
@zaleslaw the last commit can be reviewed if you like. It builds on top of https://github.com/Kotlin/dataframe/pull/801 which should be merged first

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/GroupByReceiverImpl.kt
```diff
@@ -118,7 +119,19 @@ internal class GroupByReceiverImpl<T>(override val df: DataFrame<T>, override va
                 pivot.aggregator.values.clear()
             }
 
-            is AggregateInternalDsl<*> -> yield(value.copy(value = value.value.df))
+            is AggregateInternalDsl<*> -> {
+                // Attempt to create DataFrame<Type> from AggregateInternalDsl<Type>
+                val dfType = value.type?.arguments?.firstOrNull()?.type
```
zaleslaw, line 124: I'm a little bit worried about this expression with 6 ? - what if something goes wrong and will be null, I could not explain myself the logic, should we handle the alternative (or common alternative)?

Jolanrensen: That's what the `?:` is for at the end. We try to create `DataFrame<Type1>` from `AggregateInternalDsl<Type1>`. However, in theory, any KType can be supplied by the user. KTypes like `GroupByReceiverImpl<Type1>` , `DataFrame<Type1>` (which would both still work), but unfortunately also KTypes like `CompletelyCustomType` without arguments. So in cases where this type argument cannot be fetched, we default to `DataFrame<*>` with the `?:`.

PR #812, opened 2024-08-07
[Compiler plugin] silently abort interpretation in case of invariant errors

koperagen:


Jolanrensen:
We should probably create a "compiler plugin" label as well :)

PR #823, opened 2024-08-12
Support for the serialization of DataframeConvertible values in ValueColumns has been added to enhance visualization in the KTNB plugin.

ermolenkodev:
If the value within the ValueColumn can be converted to a dataframe, it will be serialized in a manner similar to the FrameColumn values. Consequently, it will be visualized as a nested table in the KTNB plugin.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
@@ -22,8 +22,6 @@ import org.jetbrains.kotlinx.dataframe.impl.io.encodeDataFrameWithMetadata
 import org.jetbrains.kotlinx.dataframe.impl.io.encodeFrame
 import org.jetbrains.kotlinx.dataframe.impl.io.encodeRow
 import org.jetbrains.kotlinx.dataframe.impl.io.readJson
-import org.jetbrains.kotlinx.dataframe.io.Base64ImageEncodingOptions.Companion.GZIP_ON
-import org.jetbrains.kotlinx.dataframe.io.Base64ImageEncodingOptions.Companion.LIMIT_SIZE_ON
 import org.jetbrains.kotlinx.dataframe.io.JSON.TypeClashTactic
 import org.jetbrains.kotlinx.dataframe.io.JSON.TypeClashTactic.ANY_COLUMNS
 import org.jetbrains.kotlinx.dataframe.io.JSON.TypeClashTactic.ARRAY_AND_VALUE_COLUMNS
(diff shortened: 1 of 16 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
```diff
@@ -312,12 +312,19 @@ public fun AnyFrame.toJsonWithMetadata(
...
 }
 
 internal const val DEFAULT_IMG_SIZE = 600
 
+/**
+ * Interface representing encoding options that can be applied when converting a data structure to JSON format.
+ * Implementations of this interface can provide specific behaviors for encoding various data types,
+ * such as images or data frames, when serializing to JSON.
+ */
+public interface EncodingOptions
```
Jolanrensen, line 326: I like the abstraction, maybe we could even take this a step further by defining what needs to be done for the current encoding option in the implementations of this interface (So what's currently in the when-statement above). We could also do this later of course, if we get more options than 2 :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ImageSerializationTests.kt
```diff
@@ -62,7 +62,11 @@ class ImageSerializationTests(private val encodingOptions: Base64ImageEncodingOp
         encodingOptions: Base64ImageEncodingOptions?,
     ): JsonObject {
         val df = dataFrameOf(listOf("imgs"), images)
-        val jsonStr = df.toJsonWithMetadata(20, nestedRowLimit = 20, imageEncodingOptions = encodingOptions)
+        val jsonStr = df.toJsonWithMetadata(
+            20,
+            nestedRowLimit = 20,
+            encodingOptions = if (encodingOptions != null) listOf(encodingOptions) else emptyList(),
```
Jolanrensen, line 68: `listOfNotNull(encodingOptions)` is probably easier :)

Jolanrensen (approved):
Just some small notes, but LGTM

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
ermolenkodev: Actually, I’ve realized that it would be nice to replace EncodingOptions with the Encoder interface, which both checks if a type needs to be encoded and actually performs the encoding.
something like
interface Encoder {
 fun suitable(value: Any): Boolean
 fun encode(value: Any): JsonObject
}

ermolenkodev:
I’ve pushed a refactoring that replaces EncodingOption with a CustomEncoder interface. You might want to take a look, @Jolanrensen @koperagen.

Jolanrensen (approved):
Great job! I really like this modular approach :)

PR #825, opened 2024-08-19
Add toDataFrame(KClass) overload to convert type erased lists

koperagen:
It's not new to have KType / KClass overloads next to inline reified APIs, so i believe this addition would be an easy one.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
@@ -28,6 +28,11 @@ public inline fun <reified T> Iterable<T>.toDataFrame(): DataFrame<T> =
         properties()
     }
 
+public fun Iterable<*>.toDataFrame(klass: KClass<*>): DataFrame<*> =
+    createDataFrameImpl(klass) {
+        properties()
+    }
+
 @Refine
 @Interpretable("toDataFrameDsl")
 public inline fun <reified T> Iterable<T>.toDataFrame(noinline body: CreateDataFrameDsl<T>.() -> Unit): DataFrame<T> =
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
@@ -452,4 +452,14 @@ class CreateDataFrameTests {
             df.participants[0].city
         }
     }
+
+    @Test
+    fun `convert type erased list to dataframe`() {
+        val data = listOf(Person("Alice", "Cooper", 15, "London"))
+        val erased: List<Any?> = data
+        val df = erased[0]?.let {
+            erased.toDataFrame(it::class)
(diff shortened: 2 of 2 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -28,6 +28,11 @@ public inline fun <reified T> Iterable<T>.toDataFrame(): DataFrame<T> =
         properties()
     }
 
+public fun Iterable<*>.toDataFrame(klass: KClass<*>): DataFrame<*> =
```
Jolanrensen, line 31: I'd add some small KDocs for new public functions :) Also, maybe people want to specify roots/maxdepth as well, would that be possible here?

Jolanrensen: Also, since we supply a `KClass<T>` we can get the type information. If we have an `Iterable<Any?>`, we know we'll get a `DataFrame<T>`.

Jolanrensen: Finally, would it make sense to offer overloads with `body: CreateDataFrameDsl<T>.() -> Unit` too?

Jolanrensen (changes requested):
good idea but missing some typing and overloads I think

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
koperagen: > Also, since we supply a KClass<T> we can get the type information. If we have an Iterable<Any?>, we know we'll get a DataFrame<T>.
If you have KClass with specific T, i'd say you might as well use list.toDataFrame() without KClass?

> Finally, would it make sense to offer overloads with body: CreateDataFrameDsl<T>.() -> Unit too?
Well, this overload is `CreateDataFrameDsl<T>.() -> Unit = { properties() }` so we'd need to combine them. Also it's supposed to be used when only runtime KClass<*> is available. What useful can be done with `CreateDataFrameDsl<Any?>.() -> Unit`?

> Also, maybe people want to specify roots/maxdepth as well, would that be possible here?
roots - no, because Any?. maxDepth - yes

Jolanrensen: > If you have KClass with specific T, i'd say you might as well use list.toDataFrame() without KClass?

Hmm, I guess you're right. I first thought it might be possible to use another class as argument of `toDataFrame()`, then it would make sense to get `T`... However, that results in `java.lang.IllegalArgumentException: object is not an instance of declaring class`. It might be worth mentioning in attached KDocs that all elements in the iterable must be exact instances the KClass given.

koperagen:
```
val entityManager = springContext.getBean(EntityManager::class.java)
entityManager.createQuery("select e from Owner e").resultList.toDataFrame()
```
another case when Iterable has Any? type

PR #828, opened 2024-08-20
Hide properties of intermediate objects and remove data class attributes

koperagen:
In the image you can see how it used to be for some of those objects. copy, properties, generated component functions. These are useless, and yet they appear in the completion.
I left "data" for those classes where implementation uses copy more than once, and for some reason component* functions do not appear for Gather and Merge, so idk. 
toString for now is just the same as it was with data class, but in theory it can even be improved in the future for better debugger experience. 

`df.convert { a }.`

image

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
@@ -91,11 +91,13 @@ public inline fun <T, C, reified R> Convert<T, C?>.notNull(
     }
 
 @HasSchema(schemaArg = 0)
-public data class Convert<T, out C>(val df: DataFrame<T>, val columns: ColumnsSelector<T, C>) {
+public class Convert<T, out C>(internal val df: DataFrame<T>, internal val columns: ColumnsSelector<T, C>) {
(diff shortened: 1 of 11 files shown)


Jolanrensen:
It's a good idea, but we need to be careful changing data classes to normal classes. There might be parts which rely on the equality of those classes being based solely on its components. Removing `data` makes two classes with the same arguments not equal anymore.

Also, is this safe from a jupyter standpoint? aka, can we still access the internal `df` value, even if jupyter is moved to another module?

koperagen:
> Also, is this safe from a jupyter standpoint? aka, can we still access the internal df value, even if jupyter is moved to another module?

Migrating jupyter to a different module requires it to use friend module / suppress hidden members regardless of that

> It's a good idea, but we need to be careful changing data classes to normal classes. There might be parts which rely on the equality of those classes being based solely on its components. Removing data makes two classes with the same arguments not equal anymore. (this can pop up in strange places, such as Sets, tests etc)

All tests are green and it doesn't seem like intermediate objects are used anywhere except respective terminal operations. That's why i decided to not generate hashCode + equals. I think they were "data" not for this, but for "copy" and "toString", and other things are just a side effect

Jolanrensen (approved):

Issue #687, opened 2024-05-05
Reading CSV with byte order mark bleeds into first Column Name


janseeger:
If you try to read a CSV file that begins with a BOM, the first column name will be all weird.

<img width="122" alt="Screenshot 2024-05-05 at 20 20 56" src="https://github.com/Kotlin/dataframe/assets/1467318/9603741b-043f-4f1f-a959-b87848f77f8a">

Jolanrensen:
Do you have a sample CSV so we can make some tests too?

janseeger:
Unfortunately none that I can easily share. I'll try to create a small reproducer this evening.

janseeger:
Here's the [CSV File](https://github.com/Kotlin/dataframe/files/15234290/with-bom.csv). I noticed that the Column Names are correct when accessed with `columnNames()` but the generated code does not strip the BOM character.

<img width="584" alt="Screenshot 2024-05-07 at 12 37 21" src="https://github.com/Kotlin/dataframe/assets/1467318/73756a1c-55ee-4ccb-a2b2-03ab13daf7e1">

Jolanrensen:
Can reproduce, thanks :) Looks like the column name is read incorrectly entirely, since getting by name (without zero width non breaking space character) also breaks.

![image](https://github.com/Kotlin/dataframe/assets/17594275/7cc3cd32-002a-4330-9d7e-a7788ffff097)

PR #831, opened 2024-08-21
Fixes reading CSV files with BOM characters

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/687

The solution is based on https://stackoverflow.com/questions/46296827/java-spring-returning-csv-file-encoded-in-utf-8-with-bom

Passing csv reader through BOMInputStream to catch BOM characters if they exist in the file.


koperagen:
Code is different than that in the SO answers, is this ok?

Jolanrensen:
> Code is different than that in the SO answers, is this ok?

Yeah, if I had it like there, it didn't like the `.reset()` command. But the `BOMInputStream` changed in the meantime; the old constructor was deprecated, so I assume it now works like this. The test worked in any case :)

koperagen (approved):

PR #832, opened 2024-08-21
Update compiler plugin

koperagen:


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt
```diff
@@ -436,6 +436,8 @@ internal fun guessValueType(values: Sequence<Any?>, upperBound: KType? = null, l
                 collectionClasses.add(it.javaClass.kotlin)
             }
 
+            is Function<*> -> classes.add(Function::class)
```
Jolanrensen, line 439: why not :) Maybe we should change the rendering for functions in dataframes though. After a quick test I found it looks like:
```
⌌---------------------------------------------------------------------------------------------------------------⌍
|  |                                                                                        a:Function<*>| b:Int|
|--|-----------------------------------------------------------------------------------------------------|------|
| 0| org.jetbrains.kotlinx.dataframe.testSets.person.DataFrameTests$$Lambda$60/0x000000010013a040@64ee819|     2|
⌎---------------------------------------------------------------------------------------------------------------⌏
```

plugins/kotlin-dataframe/testData/box/fillNulls.kt
```diff
@@ -0,0 +1,11 @@
+import org.jetbrains.kotlinx.dataframe.*
+import org.jetbrains.kotlinx.dataframe.annotations.*
+import org.jetbrains.kotlinx.dataframe.api.*
+import org.jetbrains.kotlinx.dataframe.io.*
+
+fun box(): String {
+    val df = dataFrameOf("a", "b")(1, null, null, "")
+    val df1 = df.fillNulls { b }.with { "empty" }
+    val b: DataColumn<String> = df1.b
```
Jolanrensen, line 9: this is beautiful behavior :) Literally how DataFrame is meant to be! Anatoly would be proud I'm sure

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/flatten.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.jetbrains.kotlinx.dataframe.plugin.impl.dataFrame
+import org.jetbrains.kotlinx.dataframe.plugin.impl.toPluginDataFrameSchema
+
+class FlattenDefault : AbstractSchemaModificationInterpreter() {
+    val Arguments.receiver by dataFrame()
+    val Arguments.keepParentNameForColumns: Boolean by arg(defaultValue = Present(false))
+    val Arguments.separator: String by arg(defaultValue = Present("."))
+
+    override fun Arguments.interpret(): PluginDataFrameSchema {
+        return receiver.asDataFrame().flatten(keepParentNameForColumns, separator).toPluginDataFrameSchema()
```
Jolanrensen, line 20: we can use the actual flatten function :D awesome!

Jolanrensen (approved):
Nice to have a lot more functions supported in the plugin :) I just have a few small comments, mostly regarding clarification of the new toDataFrame overload, otherwise, it's great. Could you mention which new functions are now supported in the PR notes? That'll help us find it back later

Jolanrensen:
oh btw, can we already update the kotlin version of the compiler plugin to 2.0.20-RC2? That's the latest now, the beta version is not available anymore

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/update.kt
```diff
@@ -273,7 +275,9 @@ public typealias UpdateExpression<T, C, R> = AddDataRow<T>.(C) -> R
  * - {@include [SeeAlsoPerRowCol]}
  * @param [expression] The {@include [ExpressionsGivenRow.RowValueExpressionLink]} to update the rows with.
  */
-public fun <T, C> Update<T, C>.with(expression: UpdateExpression<T, C, C?>): DataFrame<T> =
+@Refine
+@Interpretable("UpdateWith0")
+public fun <T, C, R : C?> Update<T, C>.with(expression: UpdateExpression<T, C, R>): DataFrame<T> =
```
Jolanrensen, line 280: In my testrun of the compiler plugin I now cannot use `update {}.with {}` anymore, just `fillNulls {}.with {}`. It gives ```
[NONE_APPLICABLE] None of the following candidates is applicable: val DataRow<Into_93I>.age: Int? val ColumnsContainer<Into_93I>.age: DataColumn<Int?>
```

File: core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt (hunk 1 above)
Comment (koperagen) [reply core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt:439]: Sadly for such lambda objects toString is weird. I tried to look at the object in the debugger, but there's literally nothing that hints at signature or anything useful

File: core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/update.kt (hunk 4 above)
Comment (koperagen) [reply core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/update.kt:280]: For now yes, there's such issue because plugin fails to interpret update { }.with { } (update not supported) and fallback to an empty schema. Will fix

File: docs/StardustDocs/topics/createDataFrame.md [hunk 5]
```diff
@@ -151,6 +151,19 @@ df.add("length") { value.length }
 
 <!---END-->
 
+Creates a [`DataFrame`](DataFrame.md) with one column
+made from [`Iterable`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-iterable/) of values:
```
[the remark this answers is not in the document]
Jolanrensen: Yes! But probably also specify that the properties are not unfolded, aka, you get one "value column" and not a "column group".

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt
Jolanrensen: hmm you'd think there was a way in kotlin to detect it's a `() -> Int` or something :/

Jolanrensen: actually...

image

Jolanrensen: It already renders correctly often!

image

might just be a fluke in the tests if the lambda is serialized as interface

koperagen: what kernel version do you use?
image

Jolanrensen: I run the dev version of this PR's branch in the notebook

koperagen: Ah, ok, so the fix is needed anyway

Issue #834, opened 2024-08-23
Reading empty CSV bug


Jolanrensen:
Reading an empty CSV results in an index out of bounds exception instead of providing a helpful error

Jolanrensen:
This can also be reproduced with `DataFrame.readDelimStr("")`

PR #835, opened 2024-08-23
Empty csv fix

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/834

See test for expected behavior

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/csv.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/csv.kt
@@ -345,7 +345,7 @@ public fun DataFrame.Companion.readDelim(
     }
 
     val columnNames = csvParser.headerNames.takeIf { it.isNotEmpty() }
-        ?: (1..records[0].count()).map { index -> "X$index" }
+        ?: (1..(records.firstOrNull()?.count() ?: 0)).map { index -> "X$index" }
 
     val generator = ColumnNameGenerator()
     val uniqueNames = columnNames.map { generator.addUnique(it) }
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/CsvTests.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/CsvTests.kt
@@ -13,6 +13,7 @@ import org.jetbrains.kotlinx.dataframe.api.dataFrameOf
 import org.jetbrains.kotlinx.dataframe.api.group
 import org.jetbrains.kotlinx.dataframe.api.groupBy
 import org.jetbrains.kotlinx.dataframe.api.into
+import org.jetbrains.kotlinx.dataframe.api.isEmpty
 import org.jetbrains.kotlinx.dataframe.api.schema
 import org.jetbrains.kotlinx.dataframe.api.toStr
 import org.jetbrains.kotlinx.dataframe.ncol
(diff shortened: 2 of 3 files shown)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/CsvTests.kt
```diff
@@ -282,6 +282,32 @@ class CsvTests {
...
+        emptyDelimFile shouldBe DataFrame.empty()
+
+        val emptyCsvFile = DataFrame.readCSV(File.createTempFile("empty", "csv"))
+        emptyCsvFile shouldBe DataFrame.empty()
+
+        val emptyCsvFileManualHeader = DataFrame.readCSV(
+            file = File.createTempFile("empty", "csv"),
+            header = listOf("a", "b", "c"),
+        )
+        emptyCsvFileManualHeader shouldBe dataFrameOf("a", "b", "c").fill(0) { "" }
```
koperagen, line 300: Why zeros?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/csv.kt
```diff
@@ -345,7 +345,7 @@ public fun DataFrame.Companion.readDelim(
     }
 
     val columnNames = csvParser.headerNames.takeIf { it.isNotEmpty() }
-        ?: (1..records[0].count()).map { index -> "X$index" }
+        ?: (1..(records.firstOrNull()?.count() ?: 0)).map { index -> "X$index" }
```
koperagen, line 348: If records.firstOrNull() == null, this will be (1..0)?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/CsvTests.kt
koperagen: ah, wait, it's not zeros but 0 rows of empty strings?

Jolanrensen: yes, this generates a DF with 3 named empty String columns

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/csv.kt
Jolanrensen: actually that will be an empty range :)

Jolanrensen: `IntRange.isEmpty()` is actually defined as `first > last` and `IntRange.EMPTY = IntRange(1, 0)`

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/CsvTests.kt
koperagen: i'd prefer checks like above, because it takes time to imagine result of df constructor rather than see expected assertions

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/csv.kt
koperagen: Didn't know that. Ok, cool!

koperagen:
Please make a rebase and squash fixup before merge

koperagen (approved):

koperagen:
Thank you! Don't forget the milestone and other things

Jolanrensen:
> Thank you! Don't forget the milestone and other things

they are already on the issue, as @zaleslaw recommended :)

Issue #822, opened 2024-08-08
DataFrame.describe() Sortable by Type


jeffgnpc:
When I run df.describe() I would like to group columns of like types together.  When I try and click the sort button on the dataframe I get this error:
Column 'type' has type 'kotlin.Any' that is not Comparable java.lang.RuntimeException: Column 'type' has type 'kotlin.Any' that is not Comparable at org.jetbrains.kotlinx.dataframe.impl.columns.UtilsKt.assertIsComparable(Utils.kt:98) at org.jetbrains.kotlinx.dataframe.impl.api.SortKt.createComparator(sort.kt:60) at org.jetbrains.kotlinx.dataframe.impl.api.SortKt.sortByImpl(sort.kt:49) at

It would seem to me that sorting based on toString() when a Comparable is not available.  

I was able to get around the problem with this code:
```kt
df.describe().sortWith { r1, r2 -> r1["type"].toString().compareTo(r2["type"].toString()) }[0]["type"]
```

Jolanrensen:
Thanks for letting us know! That seems like a bug indeed. I like the idea of sorting by String for non comparable columns too. We can do this in two different places though:
- on the notebook side 
  - keep `sortBy` in the library as is: fail when you're sorting a non-comparable column, but fallback to string sorting in the notebook, then the button work as expected
- on the library side
  - changes the behavior of dataframe, might be unexpected if some random type suddenly has order

Jolanrensen:
@ermolenkodev is this something you would like to do from the notebook/table side?

ermolenkodev:
It won’t work as a fallback on the KTNB plugin side because, in general, only a subset of rows is loaded for visualization. This could confuse users, as they might not realize that the sorting is applied only to the local subset of rows rather than the entire column’s values.
However, we can try to implement a fallback in the sorting command sent to the kernel for execution.

PR #836, opened 2024-08-23
Improve dataframe sorting in KTNB UI by handling non-comparable columns

ermolenkodev:
Fixes [#822](https://github.com/Kotlin/dataframe/issues/822)

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
@@ -3,6 +3,7 @@ package org.jetbrains.kotlinx.dataframe.jupyter
 import org.jetbrains.kotlinx.dataframe.AnyCol
 import org.jetbrains.kotlinx.dataframe.AnyFrame
 import org.jetbrains.kotlinx.dataframe.AnyRow
+import org.jetbrains.kotlinx.dataframe.DataRow
 import org.jetbrains.kotlinx.dataframe.api.Convert
 import org.jetbrains.kotlinx.dataframe.api.FormatClause
 import org.jetbrains.kotlinx.dataframe.api.FormattedFrame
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
```diff
@@ -79,27 +88,80 @@ public object KotlinNotebookPluginUtils {
...
+        val secondValue = (row2.getValueOrNull(key)?.toString() ?: "")
+
+        return if (desc) {
+            secondValue.compareTo(firstValue)
+        } else {
+            firstValue.compareTo(secondValue)
+        }
+    }
+
+    private fun KType.isComparable(): Boolean = this.isSubtypeOf(typeOf<Comparable<*>>())
```
Jolanrensen, line 164: This check breaks whenever you have, say, String and Int in a single column. Yes, they are both `Comparable`, but not comparable to each other. They are `Comparable<Nothing>` and calling the `compare` there will break.

Check out the fix I made here for `AnyCol.isComparable()`: https://github.com/Kotlin/dataframe/pull/726/files#diff-68de50cc685a57fef49b535d54b8dc925caecc39ba0bc1de3d312ffeb25e8a60

In this case, if we get `Comparable<Nothing>` we should probably just compare them as `String` again, so returning `false` here should be fine (I think).

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
```diff
@@ -79,27 +88,80 @@ public object KotlinNotebookPluginUtils {
...
 
-            val sortKeys = columnPaths.map { path ->
-                ColumnPath(path)
+        return df.sortWith(comparator)
+    }
+
+    private fun createComparator(sortKeys: List<ColumnPath>, isDesc: List<Boolean>): Comparator<DataRow<*>> {
+        return Comparator { row1, row2 ->
+            for ((key, desc) in sortKeys.zip(isDesc)) {
+                val comparisonResult = if (row1.df().getColumn(key).type.isComparable()) {
```
Jolanrensen, line 116: actually, we already have an `AnyCol.isComparable()` function :) it includes the fix I describe in the other comment

Jolanrensen (changes requested):
Nicely done! A good number of tests too :). Aside from my note about isComparable it looks good to me

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
ermolenkodev: Thank you for noticing the problem!

Jolanrensen (approved):
Thanks for the extra test :)

PR #840, opened 2024-08-26
replacing often-used `typeOf<>()` calls with constants and enforcing kotlinx.datetime bias

Jolanrensen:
As described in [this YT issue](https://youtrack.jetbrains.com/issue/KT-53508), `typeOf<>()` is a very heavy operation. A single call to it creates a completely fresh reflection environment and can result in hundreds of function calls.

image

I found a couple calls (like `typeOf<String>()`) which are called 30+ times across the library!

This PR tries to minimize these calls by collecting the most-used KTypes into a single place. They are lazily evaluated when needed, so they shouldn't take up much memory when unused. I expect runtime performance improvements across the library and especially in the tests, as lots of type-checks are done there.

I've converted core/src and core/test already but I'll need to look into the other modules to replace some of the typeOf calls there too.

While working on the PR I found some inconsistencies with kotlinx datetime and java datetime classes. I've addressed those in a separate commit, make sure to check that too :).


Jolanrensen:
Fun fact, https://youtrack.jetbrains.com/issue/KT-53508 the issue has already been addressed and fixed two years ago...

image

and a quick test shows that, indeed, `typeOf<>()` caches its results by default.

I'll move the kotlinx.datetime fixes to a new pr

Issue #841, opened 2024-08-26
KSP generated code doesn't work with custom source sets


mgroth0:
I have a custom intermediate common source set called commonJvmAndroid. As a result, I get a gradle configuration called `kspCommonJvmAndroidMainMetadata`. I must apply ksp to this otherwise kotlin.dataframes doesn't behave as expected in common jvm/android sources.

PR #647, opened 2024-04-05
apply ksp to multiplatform configs in multiplatform modules

mgroth0:
Fixes https://github.com/Kotlin/dataframe/issues/645

PR #842, opened 2024-08-26
allow applying ksp to custom configs

mgroth0:
Fixes https://github.com/Kotlin/dataframe/issues/841 

As described in the issue, I needed to modify the ConvenienceSchemeGeneratorPlugin again (this is sort of part 2 after https://github.com/Kotlin/dataframe/pull/647). This PR is basically what I've been using for myself all of this time so I could correctly use dataframes in my custom source set. 

I think my solution is hacky and I am not sure this PR should be accepted, but it does get the job done for me.

I allow setting custom configs as either a gradle property or an extra property. The reason why is because while in my own code I set the extra property programatically based on some custom build logic that detects if a module has both android and jvm source sets, the extra property can only be set programatically wheras the regular property can be set in a `gradle.properties` file. I only use the extra property, but I expect if this PR is merged that others might want to do it via `gradle.properties`. `gradle.properties` would not work for me, however, because I need to set it programatically which is only supported by extra properties. Therefore, unforunately, the best I could think of was allow both.

### DIFF
--- a/plugins/dataframe-gradle-plugin/src/main/kotlin/org/jetbrains/dataframe/gradle/ConvenienceSchemaGeneratorPlugin.kt
+++ b/plugins/dataframe-gradle-plugin/src/main/kotlin/org/jetbrains/dataframe/gradle/ConvenienceSchemaGeneratorPlugin.kt
@@ -5,6 +5,7 @@ import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.artifacts.UnknownConfigurationException
 import org.gradle.kotlin.dsl.getByType
+import org.jetbrains.kotlin.gradle.plugin.extraProperties
 import java.util.Properties
 
 @Suppress("unused")
@@ -32,7 +33,7 @@ class ConvenienceSchemaGeneratorPlugin : Plugin<Project> {
         // configure it to depend on symbol-processor-all
         target.plugins.whenPluginAdded {
             if ("com.google.devtools.ksp" in this.javaClass.packageName) {
-                val isMultiplatform =
+                val isMultiplatform by lazy {
                     when {
                         target.plugins.hasPlugin("org.jetbrains.kotlin.jvm") -> false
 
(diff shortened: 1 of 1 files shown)


Jolanrensen:
I think it's a good idea. The method we used was already hacky at the time of writing https://github.com/Kotlin/dataframe/pull/647, at least something like this could solve the case where the `ksp` config has a non-standard name. 

I would try to improve the logic a bit. Does `target.findProperty()` work to find both properties and extraProperties?

We also probably should make `"kotlin.dataframe.add.ksp"` and `"kotlin.dataframe.ksp.configs"` constant with some documentation at the top of the file.

Finally, don't forget to run `ktlintFormat`/`ktlintCheck` :)

plugins/dataframe-gradle-plugin/src/main/kotlin/org/jetbrains/dataframe/gradle/ConvenienceSchemaGeneratorPlugin.kt
```diff
@@ -45,29 +57,25 @@ class ConvenienceSchemaGeneratorPlugin : Plugin<Project> {
...
-                    target.configurations.getByName(testKspCfg).dependencies.add(
-                        target.dependencies.create(
-                            "org.jetbrains.kotlinx.dataframe:symbol-processor-all:$preprocessorVersion",
-                        ),
-                    )
-                } catch (e: UnknownConfigurationException) {
-                    target.logger.warn(
-                        "Configuration '$testKspCfg' not found. Please make sure the KSP plugin is applied.",
-                    )
+                val overriddenConfigs = target.findProperty(PROP_KSP_CONFIGS)?.let { (it as String) }?.split(",")
```
Jolanrensen, line 61: Maybe we should also trim the entries in the list. When a user writes: "config1, config2", we now get the configs "config1" and " config2". I'm not sure `getByName` can find it like that

Jolanrensen (approved):
Just one small suggestion, the rest looks good, thank you very much!

plugins/dataframe-gradle-plugin/src/main/kotlin/org/jetbrains/dataframe/gradle/ConvenienceSchemaGeneratorPlugin.kt
Jolanrensen: I'll add it quickly so we can merge :)

PR #843, opened 2024-08-27
Enforcing kotlinx datetime bias

Jolanrensen:
Different parts of the library had different biases towards using Java's datetime libraries or Kotlin's.

Since DF is a Kotlin library, we should prefer Kotlin's.

This PR addresses that, while also filling in some missing conversions/parsers.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convert.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convert.kt
@@ -3,6 +3,7 @@ package org.jetbrains.kotlinx.dataframe.impl.api
 import kotlinx.datetime.Instant
 import kotlinx.datetime.LocalDate
 import kotlinx.datetime.LocalDateTime
+import kotlinx.datetime.LocalTime
 import kotlinx.datetime.TimeZone
 import kotlinx.datetime.atStartOfDayIn
 import kotlinx.datetime.atTime
@@ -10,6 +11,7 @@ import kotlinx.datetime.toInstant
 import kotlinx.datetime.toJavaInstant
 import kotlinx.datetime.toJavaLocalDate
 import kotlinx.datetime.toJavaLocalDateTime
+import kotlinx.datetime.toJavaLocalTime
 import kotlinx.datetime.toKotlinInstant
 import kotlinx.datetime.toKotlinLocalDate
 import kotlinx.datetime.toKotlinLocalDateTime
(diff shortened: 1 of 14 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convert.kt
```diff
@@ -299,13 +304,19 @@ internal fun createConverter(from: KType, to: KType, options: ParserOptions? = n
...
+                LocalTime::class -> convert<Int> { it.toLong().toLocalTime(defaultTimeZone) }
+
+                Instant::class -> convert<Int> { Instant.fromEpochMilliseconds(it.toLong()) }
+
+                JavaLocalDateTime::class -> convert<Int> {
+                    it.toLong().toLocalDateTime(defaultTimeZone).toJavaLocalDateTime()
                 }
 
-                java.time.LocalDate::class -> convert<Long> { it.toLocalDate(defaultTimeZone).toJavaLocalDate() }
+                JavaLocalDate::class -> convert<Int> { it.toLong().toLocalDate(defaultTimeZone).toJavaLocalDate() }
```
koperagen, line 315: It appears like Long converter was removed. Why?

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/ArrowWriterImpl.kt
```diff
@@ -243,29 +243,30 @@ internal class ArrowWriterImpl(
             is TimeNanoVector ->
                 column.convertToLocalTime()
                     .forEachIndexed { i, value ->
-                        value?.also { vector.set(i, value.toNanoOfDay()) }
+                        value?.also { vector.set(i, value.toJavaLocalTime().toNanoOfDay()) }
```
koperagen, line 246: is this conversion needed?

koperagen:
I wouldn't assign "bug" to this. Instead, something like "migration-needed" would be good. For user it will be a breaking change

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convert.kt
Jolanrensen: this is the `when (fromClass) { Int::class -> }` branch, not the `Long` branch :) it was in the wrong place

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/ArrowWriterImpl.kt
Jolanrensen: actually, it isn't :) thanks

PR #851, opened 2024-09-02
Compiler plugin update

koperagen:
This PR contains two parts:
1. 5 new operations in first 5 commits, can be reviewed one-by-one
2. Major plugin feature. Before this PR it was only possible to update function call that returns DataFrame<T>, but it also makes sense for at least GroupBy (df.groupBy, groupBy.add) and DataRow (df.aggregate, DataRow.readJsonStr) types. So here's support for GroupBy type along with refactoring that makes whole transformation pipeline extensible in terms of supported types similar to DataFrame


plugins/kotlin-dataframe/testData/box/groupBy_refine.kt
```diff
@@ -0,0 +1,12 @@
+import org.jetbrains.kotlinx.dataframe.*
+import org.jetbrains.kotlinx.dataframe.annotations.*
+import org.jetbrains.kotlinx.dataframe.api.*
+import org.jetbrains.kotlinx.dataframe.io.*
+
+
+fun box(): String {
+    val df = dataFrameOf("a", "b", "c")(1,2,3)
```
Jolanrensen, line 8: at one point, some linter is going to complain about the lack of spaces after the comma's XD

plugins/kotlin-dataframe/testData/box/groupBy_refine.kt
```diff
@@ -0,0 +1,12 @@
...
+    val df1 = df.groupBy { a }
+    df1.keys.compareSchemas(strict = true)
```
Jolanrensen, line 10: what does this test achieve actually?

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/extensions/FunctionCallTransformer.kt
```diff
@@ -421,6 +429,110 @@ class FunctionCallTransformer(
         return newCall1
     }
 
+    private fun materialize(
+        dataFrameSchema: PluginDataFrameSchema,
```
Jolanrensen, line 435: I can't say I have any idea what's going on in this massive file. Might it be an idea to split it up eventually? I can't review it if I don't understand it.

Jolanrensen (approved):
the small new implementations are welcome and understandable :) The large file changes I don't understand at all at the moment, so it's hard to review

plugins/kotlin-dataframe/testData/box/groupBy_refine.kt
koperagen: compareSchemas calls df.compileTimeSchema (=MarkersExtractor.get()), this function creates a DataFrameSchema from class declaration. In case with plugin, we expect that it will be a generated schema. Then it compares it to actual runtime schema.
So, previously this test would fail, because by default groupBy returns GroupBy<T, T>, so compileTimeSchema for keys
a: Int
b: Int
c: Int
and it's different from runtime
a: Int 
In short, this is the way to make sure there's no extra columns in the schema

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
```diff
@@ -207,6 +207,7 @@ public class AddDsl<T>(
 
     public infix fun <R> ColumnReference<R>.into(column: KProperty<R>): Boolean = into(column.name)
 
+    @Interpretable("AddDslStringInvoke")
     public operator fun String.invoke(body: AddDsl<T>.() -> Unit): Unit = group(this, body)
```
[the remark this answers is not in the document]
koperagen: yeah i thought about it, to me invoke one seems like preferred one :) i decided to keep group in a queue for now

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/extensions/FunctionCallTransformer.kt
koperagen: as a starting point i recommend to read kdocs for FirFunctionCallRefinementExtension

PR #855, opened 2024-09-06
Improved SQL<->JDBC mapping

### LINKED ISSUE #701: JDBC H2, `CHAR(10)` column read as `Char?` instead of `String?`
@zaleslaw 
From an H2 database, a `CHAR(10)` column is read as `Char?` instead of `String?`

This issue can be reproduced easily by editing the ``fun `read from huge table`()`` in `h2Test.kt` by adding the line:
`schema.columns["characterCol"]!!.type shouldBe typeOf<String?>()`
The column contains `["ABC", "DEF", "GHI"]` so I gotta agree it's not supposed to be a `Char`.

I discovered it by exploring whether the `DataColumnImpl.type` actually matches the types found in `DataColumnImpl.values`, which mismatches in this case. It might be a good check to do for all type conversions :).

### LINKED ISSUE #804: JDBC KType mismatches in "debug" mode
Related to https://github.com/Kotlin/dataframe/issues/713

Running the JDBC tests with `kotlin.dataframe.debug=true` results in 15 failing tests:

- [x] SqliteTest
    - [x] read from all tables
    - [x] read from sql query
    - [x] read from tables
- [x] JdbcTest
    - [x] read from huge table
- [x] MSSQLH2Test
    - [x] read from all tables
    - [x] basic test for reading sql tables
- [x] MariadbH2Test
    - [x] read from all tables
    - [x] basic test for reading sql tables
    - [x] reading numeric types
- [x] MySqlH2Test
    - [x] read from all tables
    - [x] basic test for reading sql tables
    - [x] reading numeric types
- [x] PostgresH2Test
    - [x] read from all tables
    - [x] read from tables
    - [x] read columns of different types to check type mapping

### LINKED ISSUE #859: Dataframe with PostgreSQL: Class declares 1 type parameters, but 0 were provided.
Plugin-Version: 241.18034.95

First of all: Thanks for the brilliant project! I just had to get started, but ran directly into an issue: When I read a table of a Postgresql Database using`readSqlTable` with a column of the type `integer[]`, I'll get the error `Class declares 1 type parameters, but 0 were provided.`.

## Code
```kotlin
USE {
    dependencies("org.postgresql:postgresql:42.7.3")
}

%use dataframe
import java.sql.DriverManager
import java.util.*

val dbConfig = DatabaseConfiguration(url = "jdbc:postgresql://localhost:5499/demo", user = "username", password = "password")

val table = DataFrame.readSqlTable(dbConfig = dbConfig, "example")
```

## DDL
```sql
CREATE TABLE example (
    names integer[]
);
```

## Stacktrace
```
Class declares 1 type parameters, but 0 were provided.
java.lang.IllegalArgumentException: Class declares 1 type parameters, but 0 were provided.
	at kotlin.reflect.full.KClassifiers.createType(KClassifiers.kt:53)
	at kotlin.reflect.full.KClassifiers.createType$default(KClassifiers.kt:42)
	at org.jetbrains.kotlinx.dataframe.io.ReadJdbcKt.makeCommonSqlToKTypeMapping(readJdbc.kt:688)
	at org.jetbrains.kotlinx.dataframe.io.ReadJdbcKt.generateKType(readJdbc.kt:635)
	at org.jetbrains.kotlinx.dataframe.io.ReadJdbcKt.fetchAndConvertDataFromResultSet(readJdbc.kt:576)
	at org.jetbrains.kotlinx.dataframe.io.ReadJdbcKt.readSqlTable(readJdbc.kt:148)
	at org.jetbrains.kotlinx.dataframe.io.ReadJdbcKt.readSqlTable(readJdbc.kt:116)
	at org.jetbrains.kotlinx.dataframe.io.ReadJdbcKt.readSqlTable$default(readJdbc.kt:110)
	at Line_38_jupyter.<init>(Line_38.jupyter.kts:1)
	at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:67)
```

zaleslaw:
Fixes #804 Fixes #701 
Closed #859


dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -775,6 +781,9 @@ private fun manageColumnNameDuplication(columnNameCounter: MutableMap<String, In
     return name
 }
 
+// Utility function to cast arrays based on the type of elements
+public fun <T : Any> castArray(array: Array<*>, elementType: KClass<T>): List<T> = array.map { elementType.cast(it) }
```
Jolanrensen, line 785: does this need to be public? (also probably change the comment to kdoc so it shows up in the IDE)

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -831,6 +846,25 @@ private fun fetchAndConvertDataFromResultSet(
...
+        (it as java.sql.Array).array ?: emptyArray<Any>()
+    }
+
+    // Determine the type based on the first element
+    val firstArray = sqlArrays.firstOrNull()
+    val firstElementType = firstArray?.javaClass?.componentType?.kotlin ?: Any::class
+
+    // Cast arrays to the detected type
+    return if (firstArray != null && firstArray is Array<*>) {
+        val castedArrays = sqlArrays.map { castArray(it as Array<*>, firstElementType).toTypedArray() }
```
Jolanrensen, line 861: *castArrays, the past participle form of "to cast" is "cast"

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -831,6 +846,25 @@ private fun fetchAndConvertDataFromResultSet(
...
 
+private fun handleArrayValues(values: MutableList<Any?>): List<Any> {
+    // Intermediate variable for the first mapping
+    val sqlArrays = values.map {
+        (it as java.sql.Array).array ?: emptyArray<Any>()
+    }
+
+    // Determine the type based on the first element
+    val firstArray = sqlArrays.firstOrNull()
+    val firstElementType = firstArray?.javaClass?.componentType?.kotlin ?: Any::class
```
Jolanrensen, line 857: can we be sure this never breaks? Say the array consists of Longs but the first element is "123" which can be read as an integer, or even a short, will this break things? Or are the types decided elsewhere based on the entire contents of the array?

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -882,7 +917,7 @@ private fun makeCommonSqlToKTypeMapping(tableColumnMetadata: TableColumnMetadata
         Types.DOUBLE to Double::class,
         Types.NUMERIC to BigDecimal::class,
         Types.DECIMAL to BigDecimal::class,
-        Types.CHAR to Char::class,
+        Types.CHAR to String::class,
```
Jolanrensen, line 920: So we cannot parse `char(1)` as Char? just `char(n)` as String

Jolanrensen: May be for the best probably actually, if we cannot check `n`

Jolanrensen (changes requested):
just some small questions

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
zaleslaw: Because JDBC has no special readChar/writeChar methods and returns String value for Char(1), we leave for a user the decision to convert.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -831,6 +847,31 @@ private fun fetchAndConvertDataFromResultSet(
     return dataFrame
 }
 
+private fun handleArrayValues(values: MutableList<Any?>): List<Any> {
+    // Intermediate variable for the first mapping
+    val sqlArrays = values.mapNotNull {
+        (it as? java.sql.Array)?.array?.let { array -> array as? Array<*> }
```
Jolanrensen, line 853: This usage of `let` here has negligible benefit, `(it as? java.sql.Array)?.array as? Array<*>`, does exactly the same AFAIK

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -831,6 +847,31 @@ private fun fetchAndConvertDataFromResultSet(
...
+    }
+
+    // Flatten the arrays to iterate through all elements and filter out null values, then map to component types
+    val allElementTypes = sqlArrays
+        .flatMap { array ->
+            (array.javaClass.componentType?.kotlin?.let { listOf(it) } ?: emptyList())
+        } // Get the component type of each array and convert it to a Kotlin class, if available
```
Jolanrensen, line 860: Why is this put into a singular list and flatMapped? Doesn't this do exactly the same thing?
```kt
.mapNotNull { array ->
   array.javaClass.componentType?.kotlin
}
```

PR #856, opened 2024-09-06
Compiler plugin bugfixes

koperagen:
toDataFrame and explode produced invalid schemas and now fixed, making sure using generated extension properties won't lead to exceptions at runtime
Second commit is about ClassCast exception at compilation, or rather in IDE at analysis. All plugin code should not throw exceptions because they all appear in IDE

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
@@ -251,11 +251,9 @@ internal fun convertToDataFrame(
         val kClass = returnType.classifier as KClass<*>
         val fieldKind = returnType.getFieldKind()
 
-        val shouldCreateValueCol = (
-            maxDepth <= 0 &&
-                !fieldKind.shouldBeConvertedToFrameColumn &&
-                !fieldKind.shouldBeConvertedToColumnGroup
-        ) ||
+        val keepSubtree =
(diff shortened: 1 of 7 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
```diff
@@ -251,11 +251,9 @@ internal fun convertToDataFrame(
         val kClass = returnType.classifier as KClass<*>
         val fieldKind = returnType.getFieldKind()
 
-        val shouldCreateValueCol = (
-            maxDepth <= 0 &&
-                !fieldKind.shouldBeConvertedToFrameColumn &&
-                !fieldKind.shouldBeConvertedToColumnGroup
-        ) ||
+        val keepSubtree =
+            maxDepth <= 0 && !fieldKind.shouldBeConvertedToFrameColumn && !fieldKind.shouldBeConvertedToColumnGroup
```
zaleslaw, line 255: Is it fine that in this logic exp `maxDepth <= 0` and in impl/api/toDataFrame is `depth>maxDepth`?

zaleslaw (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
koperagen: I think so. I didn't want to follow the same logic in compiler implementation because incrementing depth in recursive calls seems decreasing maxDepth. Maybe i should re-evaluate and change one to work as the other

Issue #594, opened 2024-02-20
Update Kover to 0.7.6


Jolanrensen:
The project uses `org.jetbrains.kotlinx.kover:org.jetbrains.kotlinx.kover.gradle.plugin:0.6.1` but it requires more work to update it to 0.7.0: https://github.com/Kotlin/kotlinx-kover/blob/v0.7.0/docs/gradle-plugin/migrations/migration-to-0.7.0.md

fb64:
I think this one can be closed

PR #858, opened 2024-09-07
Bump Kover version

fb64:
#594

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -55,6 +55,13 @@ dependencies {
     api(project(":dataframe-excel"))
     api(project(":dataframe-openapi"))
     api(project(":dataframe-jdbc"))
+
+    kover(project(":core"))
+    kover(project(":dataframe-arrow"))
+    kover(project(":dataframe-excel"))
+    kover(project(":dataframe-openapi"))
+    kover(project(":dataframe-jdbc"))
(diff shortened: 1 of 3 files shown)


koperagen:
Hi! Thank you for the PR :) Please fix :ktlintKotlinScriptCheck

fb64:
> Hi! Thank you for the PR :) Please fix :ktlintKotlinScriptCheck

Fixed,  is `:ktlintKotlinScriptCheck` supposed to run during a `clean build` because I didn't get any error...

koperagen:
> > Hi! Thank you for the PR :) Please fix :ktlintKotlinScriptCheck
> 
> Fixed, is `:ktlintKotlinScriptCheck` supposed to run during a `clean build` because I didn't get any error...

No, it should run without clean build. We need to investigate this problem, thanks for letting me know

fb64:
I mean, I didn't get any error with `clean build` but I got it with `:ktlintKotlinScriptCheck` single task... Maybe this task should be included in `build` task or git pre-commit to let developers know that there is an issue before committing.

koperagen:
Ah, ok, i misunderstood. But i thought build already depends check which in turn depends on ktlintKotlinScriptCheck.. need to verify. I noticed that sometimes linkCheck tasks don't print errors even when they run

PR #868, opened 2024-09-18
Fix duplicated jvm signature

koperagen:
Problem appears when multiple nested DataSchema declarations have property with the same name. Using parent declarations in JvmName argument solves it. I decided to not use package names there

computeNestedName is slightly modified computeFqName from compiler

### DIFF
--- a/plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/extensions/IrBodyFiller.kt
+++ b/plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/extensions/IrBodyFiller.kt
@@ -5,7 +5,6 @@
 
 package org.jetbrains.kotlinx.dataframe.plugin.extensions
 
-import kotlinx.serialization.decodeFromString
 import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
 import org.jetbrains.kotlin.backend.common.FileLoweringPass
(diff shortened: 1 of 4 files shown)


plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/extensions/IrBodyFiller.kt
```diff
@@ -218,6 +220,21 @@ private class DataFrameFileLowering(val context: IrPluginContext) : FileLowering
         return declaration
     }
 
+    private fun IrDeclarationWithName.nestedName() = StringBuilder().let { computeNestedName(this, it); it.toString() }
```
Jolanrensen, line 223: Would `buildString { computeNestedName(this@nestedName, this) }` not be a lot shorter? :)

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/extensions/IrBodyFiller.kt
```diff
@@ -218,6 +220,21 @@ private class DataFrameFileLowering(val context: IrPluginContext) : FileLowering
...
+
+    private fun computeNestedName(declaration: IrDeclarationWithName, result: StringBuilder): Boolean {
```
Jolanrensen, line 225: How does this work when parents have backticks in the name? I suspect `` `Parent name`_`child name` `` won't compile, right?

Jolanrensen (approved):

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/extensions/IrBodyFiller.kt
koperagen: Not sure yet. There are a few places where backtick identifiers can break things. I plan to review it later

Jolanrensen: alright, don't forget it ;) haha

PR #869, opened 2024-09-18
[Compiler plugin] Lower frontend generated implicit receivers

koperagen:
For extension properties to work, frontend plugin inserts implicit receivers. But they are just non-existing references needed for resolve. Backend must find them and replace with something that will be executed

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/aggregation/AggregateDsl.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/aggregation/AggregateDsl.kt
@@ -1,6 +1,7 @@
 package org.jetbrains.kotlinx.dataframe.aggregation
 
 import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.annotations.HasSchema
 import org.jetbrains.kotlinx.dataframe.annotations.Interpretable
 import org.jetbrains.kotlinx.dataframe.api.ColumnSelectionDsl
 import org.jetbrains.kotlinx.dataframe.api.pathOf
@@ -11,6 +12,7 @@ import org.jetbrains.kotlinx.dataframe.impl.columnName
 import kotlin.reflect.KProperty
 import kotlin.reflect.typeOf
 
+@HasSchema(schemaArg = 0)
 public abstract class AggregateDsl<out T> :
     DataFrame<T>,
     ColumnSelectionDsl<T> {
--- a/plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/extensions/IrBodyFiller.kt
+++ b/plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/extensions/IrBodyFiller.kt
@@ -30,6 +30,7 @@ import org.jetbrains.kotlin.ir.expressions.IrConst
 import org.jetbrains.kotlin.ir.expressions.IrErrorCallExpression
 import org.jetbrains.kotlin.ir.expressions.IrExpression
 import org.jetbrains.kotlin.ir.expressions.IrTypeOperator
+import org.jetbrains.kotlin.ir.expressions.IrTypeOperatorCall
 import org.jetbrains.kotlin.ir.expressions.impl.IrCallImpl
 import org.jetbrains.kotlin.ir.expressions.impl.IrConstImpl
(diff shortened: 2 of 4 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/aggregation/AggregateDsl.kt
```diff
@@ -11,6 +12,7 @@ import org.jetbrains.kotlinx.dataframe.impl.columnName
 import kotlin.reflect.KProperty
 import kotlin.reflect.typeOf
 
+@HasSchema(schemaArg = 0)
```
Jolanrensen, line 15: What does HasSchema do again? There's no docs :/

plugins/kotlin-dataframe/testData/box/wrongReceiver.kt
```diff
@@ -0,0 +1,14 @@
...
+import org.jetbrains.kotlinx.dataframe.api.*
+
+
+@DataSchema
+data class Record(val a: String, val b: Int)
+
+fun box(): String {
+    val df = List(10) { Record(it.toString(), it) }.let { dataFrameOf(*it.toTypedArray()) }
+    val aggregate = df.pivot { b }.aggregate {
+        this.add("c") { 123 }.c
```
Jolanrensen, line 11: which part failed before and is now solved?

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/extensions/IrBodyFiller.kt
```diff
@@ -218,16 +220,33 @@ private class DataFrameFileLowering(val context: IrPluginContext) : FileLowering
...
+        val origin = (type.classifierOrNull?.owner as? IrClass)?.origin ?: return false
+        val fromPlugin = origin is IrDeclarationOrigin.GeneratedByPlugin && origin.pluginKey is DataFramePlugin
+        val scopeReference = type.classFqName?.shortName()?.asString()?.startsWith("Scope") ?: false
+        return fromPlugin || scopeReference
+    }
+
+    @OptIn(UnsafeDuringIrConstructionAPI::class)
+    private fun IrExpression.replaceWithConstructorCall(): IrConstructorCallImpl {
+        val constructor = type.getClass()!!.constructors.toList().single()
+        val type = type
```
Jolanrensen, line 249: unnecessary assignment

Jolanrensen (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/aggregation/AggregateDsl.kt
koperagen: To help identify type parameter that represents dataframe schema
org.jetbrains.kotlinx.dataframe.plugin.InterpretKt#getSchema
org/jetbrains/kotlinx/dataframe/plugin/interpret.kt:255

plugins/kotlin-dataframe/testData/box/wrongReceiver.kt
koperagen: execution of this expression `this.add("c") { 123 }.c` failed due to wrong receiver in generated bytecode. `c` extension property used `this@aggregate` instead of scope it comes from

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/extensions/IrBodyFiller.kt
koperagen: thanks!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/aggregation/AggregateDsl.kt
Jolanrensen: right, so if we would take the type parameter at `schemaArg`, which in `AggregateDsl<out T>` is `out T`, and put it in `DataFrame<>` like `DataFrame<T>`, a schema generated for thát would represent this object. And this could be different for types like `GroupedDataRow<out T, out G>` where `G` might represent the data schema?

Could you explain this in a small piece of KDoc?

koperagen: I'll think of something

PR #870, opened 2024-09-18
Rerunning notebooks for 0.14.0-RC1

Jolanrensen:
Reran all notebooks with a version that should be compatible with the to-be-released version 0.14 and the latest Kandy version. Will do the same for the example notebooks on datalore


zaleslaw (approved):
Hard to track changes, but possible

Jolanrensen:
> Hard to track changes, but possible

It's easier to check in the IDE, tho many notebooks have 0 visual changes, which is a good thing :) it means we didn't break anything

PR #880, opened 2024-09-23
Update GeneratedField.kt

GeorgCantor:
**Helper Function:** The `toNotNullableName()` extension function encapsulates the logic for determining how to transform a `markerName` or `typeFqName`. This reduces code duplication across the different field types.

**Readability:** By using a helper function, we make it clear what transformation is being applied to the names, improving overall readability.

**Maintainability:** If the logic for determining how to handle names changes in the future, you only need to update it in one place.


Jolanrensen:
Hi! Thanks for your small refactor.

For this PR it seems like you edited the `generated-sources` directory instead of the `src` one. The generated sources are, as the name suggests, auto-generated based on the sources. This means that editing them will do nothing as they will be overwritten when pushed to master.

I would like to point you to our [contribution guide](https://github.com/Kotlin/dataframe/blob/master/CONTRIBUTING.md) which describes how to contribute, run tests, the kdocs preprocessor etc.

For now, if you edit `src` instead, I'll approve it.

PR #882, opened 2024-09-24
Adding contracts for `Anycol.isValueColumn` etc. for smart-casting

Jolanrensen:
Small change to reduce verbose code like:

```kt
if (col.isColumnGroup()) callFunction(col.asColumnGroup())
```
to
```kt
if (col.isColumnGroup()) callFunction(col)
```
thanks to contracts :).

Unfortunately, we cannot yet do this with enum entries, so when we're using notations like
```kt
when (col.kind) {
    ColumnKind.Group -> callFunction(col.asColumnGroup())
    ....
}
```
are still necessary.


koperagen (approved):
Cool!

PR #888, opened 2024-09-26
Partially inline AnyFrame typealias in return type position

koperagen:
I noticed that completion pop-up still shows that AnyFrame = DataFrame<*> and it makes it look cluttered
It makes me want to remove all such type aliases :c But, we can start from dataFrameOf

Before:
image

Looks even worse on a narrow screen:
image

After:
image

image


Jolanrensen:
Looks better indeed :) I do see the same behavior for argument types defined with type aliases in your screenshot. Should we keep those? or also replace them

koperagen:
> Looks better indeed :) I do see the same behavior for argument types defined with type aliases in your screenshot. Should we keep those? or also replace them

let's make an issue and merge this part

Issue #827, opened 2024-08-20
☂ CSV rework


Jolanrensen:
I've collected several issues regarding CSV reading / readDelim to be reworked. Most are marked with the new [CSV label](https://github.com/Kotlin/dataframe/labels/csv).
I marked them all, because most are related and solving one will require solving others and vice versa.

Let this issue be an umbrella issue linking to the relevant issues:

- [x] https://github.com/Kotlin/dataframe/issues/787
  - [x] Split readDelim and readCsv into different files, possibly a different module?
  - [x] Deprecate .read() overload, this should only exist in guess.kt https://github.com/Kotlin/dataframe/pull/902
  - [x] KDocs for each overload
  - [x] Check which CSVFormat options we can put in the readDelim and readCsv functions to create overloads without it. (Like setting the quote mark)
  - [x] Keep one overload for readDelim and readCsv with CSVFormat as option so people can set advanced settings with it.
- [x] https://github.com/Kotlin/dataframe/issues/826
- [x] https://github.com/Kotlin/dataframe/issues/508
- [x] https://github.com/Kotlin/dataframe/issues/746
- [x] https://github.com/deephaven/deephaven-csv/issues/212
- [x] https://github.com/Kotlin/dataframe/issues/687
- [x] https://github.com/Kotlin/dataframe/issues/607 Covered by https://github.com/Kotlin/dataframe/pull/935
- [x] ~https://github.com/Kotlin/dataframe/issues/593~ Covered by https://github.com/Kotlin/dataframe/pull/925
- [x] https://github.com/Kotlin/dataframe/issues/589
- [x] https://github.com/Kotlin/dataframe/issues/507
- [x] https://github.com/Kotlin/dataframe/issues/469
- [x] https://github.com/Kotlin/dataframe/issues/383
- [x] https://github.com/Kotlin/dataframe/issues/100
- [x] https://github.com/Kotlin/dataframe/issues/834
- [x] https://github.com/Kotlin/dataframe/issues/921
- [x] https://github.com/Kotlin/dataframe/issues/527
- [x] Remove opt-in
- [ ] https://github.com/Kotlin/dataframe/issues/765
- [ ] Extend CsvOption in `ImportDataSchema` annotation and Gradle Task
- [ ] Change drag-and-drop behavior in notebooks
- [ ] Documentation about new CSV implementation on the website/examples when it's no longer experimental.
- [ ] Move Compression to :core and apply to other io's

Jolanrensen:
[...]

PR #902, opened 2024-10-01
Deprecating `DataFrame.read("", delimiter =)`

Jolanrensen:
First tiny piece of https://github.com/Kotlin/dataframe/issues/827.

All our `DataFrame.read()` overloads exist in guess.kt and go over all available formats based on the file or url given to the function.

`DataFrame.read(..., delimiter =)` can be replaced by the already existing `DataFrame.readCSV(..., delimiter =)` function family. This deprecates the old function with a WARNING in favor of `readCSV`.

### DIFF
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/OtherSamples.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/OtherSamples.kt
@@ -5,7 +5,7 @@ import org.jetbrains.kotlinx.dataframe.DataFrame
 import org.jetbrains.kotlinx.dataframe.api.take
 import org.jetbrains.kotlinx.dataframe.explainer.WritersideFooter
 import org.jetbrains.kotlinx.dataframe.explainer.WritersideStyle
-import org.jetbrains.kotlinx.dataframe.io.read
+import org.jetbrains.kotlinx.dataframe.io.readCSV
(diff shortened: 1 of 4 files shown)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/OtherSamples.kt
```diff
@@ -16,7 +16,7 @@ class OtherSamples {
 
     @Test
     fun example() {
-        val df = DataFrame.read("../data/titanic.csv", delimiter = ';').take(5)
+        val df = DataFrame.readCSV("../data/titanic.csv", delimiter = ';').take(5)
```
zaleslaw, line 19: Do we need to replace onlhy this call? Did we have something in the docs folder?

zaleslaw (approved):

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/OtherSamples.kt
Jolanrensen: in theory "no" because all our docs are generated from runnable code, but I'll check anyway :)

Jolanrensen: ha, there was one! Thanks :D

Issue #787, opened 2024-07-17
Improve and document CSV reading options 


koperagen:
https://youtrack.jetbrains.com/issue/KT-69798
Unlike readCSV, readDelim exposes underlying apache common CSVFormat. There're a lot more configuration options. From the linked issue it seems arguments that readCSV provide are not enough to read every possible kind of file. 
So at least readDelim needs documentation here
https://kotlin.github.io/dataframe/read.html
Or readCSV should have more options. But, i'd avoid exposing CSVFormat in readCSV

zaleslaw:
We agreed to split the issue
- answer the use-case
- provide a default

unrec:
Thanks everyone for working on this issue.

I agree that exposing `CSVFormat` to `readCSV` fun is not good, but there should be a flexibility in CSV setup configuration.
I faced it with a .csv file with quote character in some row, so as I understand there are 2 possible options:
- be able to set specific quote character instead of `"`
- have an opportunity to disable quote char completely

This should be added as a optional param to `readCsv` fun instead of `readDelim`.

Jolanrensen:
Will be solved by the new CSV implementation: "dataframe-csv". Setting a quote character is supported by default.

The new read functions will also expose a single overload where you can adjust additional `CsvSpecs` if we ever miss out on a parameter in the future.

Issue #589, opened 2024-02-13
Alternative CSV reader


Jolanrensen:
should be investigated: https://github.com/doyaaaaaken/kotlin-csv

koperagen:
I tried FastCSV and want to utilize it on JVM for performance that several times better than existing one and beats pandas too

devcrocod:
Keep in mind that you can always write your own interface and hide the platform implementation later

Jolanrensen:
I've been experimenting with different implementations to find the fastest one in combination with DataFrame.

Each test has two versions of the implementation:
- The default version first loads the entire CSV into memory. This is usually the fastest for smaller CSVs since the right amount of memory for the columns can be created right away. However, this can run into memory issues more quickly for larger CSV files.
- That's why each test is accompanied with a "sequential" version. This version uses data collectors to stream the csv rows into separate string-columns directly. The downside of this is that we don't know the right amount of memory yet, so the ArrayLists need to grow accordingly, but we never get a full `List<SomeCsvRowClass>`, saving memory in the long run :)

We test:
- [Apache Commons CSV](https://commons.apache.org/proper/commons-csv/) (the current implementation)
- [Fast CSV](https://fastcsv.org/)
- [Kotlin-CSV](https://github.com/jsoizo/kotlin-csv)

**Small CSV: 65.4 kB**
(Higher score is better)
![image](https://github.com/user-attachments/assets/d26d5949-c7e0-41ac-80f4-df7619e212a7)

**Large CSV: 857.7 MB**
(Higher score is better)
![image](https://github.com/user-attachments/assets/bfaa1a75-0dfd-4c5c-9634-971eaca0660b)

Jolanrensen:
I now added Deephaven-csv:

(s/op: Lower is better)
```
Benchmark                                    Mode  Cnt   Score    Error  Units
CsvBenchmark.apacheCsvReader                   ss   10   0.007 ±  0.003   s/op
CsvBenchmark.apacheCsvReaderSequential         ss   10   0.008 ±  0.003   s/op
CsvBenchmark.deephavenCsvReader                ss   10   0.009 ±  0.011   s/op
CsvBenchmark.fastCsvReader                     ss   10   0.004 ±  0.001   s/op
CsvBenchmark.fastCsvReaderSequential           ss   10   0.004 ±  0.002   s/op
CsvBenchmark.kotlinCsvReader                   ss   10   0.008 ±  0.001   s/op
[...]

Issue #508, opened 2023-11-20
Bug: `DataFrame.readCSV("")` returns files in the folder


Jolanrensen:
![image_720](https://github.com/Kotlin/dataframe/assets/17594275/e7b8cdf1-4fa0-4dfd-b15e-29ca9bd1c92f)

Also happens with `readTSV` and other functions that use `catchHttpResonse`.
`readJson` crashes when trying to read it: `com.beust.klaxon.KlaxonException: Unexpected character at position 0: 'b' (ASCII: 98)'`, but this too should be caught.

zaleslaw:
Could it be a default behavior for dplyr/pandas?

Jolanrensen:
No, it makes no sense to display list the files in the directory and interpret that String as csv.
It should either throw an exception, or concatenate all CSVs in the directory and read them all :)

Jolanrensen:
Will be solved in the new CSV implementation: "dataframe-csv"

Issue #469, opened 2023-10-14
readCSV fails for *.zip


koperagen:
This file extension is treated in a special way: there's a `isCompressed` method, and depending on it `readCSV` wraps `InputStream`. But it doesn't work for *.zip because InputStream is wrapped in a GZIPInputStream. Apparently it's also not enough to just wrap an InputStream, because ZIP has more complex structure and you need to call methods of ZipInputStream:
```
val zipInputStream = ZipInputStream(
    File("data.csv.zip").inputStream(),
    Charsets.UTF_8
)
zipInputStream.nextEntry
val df1 = DataFrame.readCSV(zipInputStream)
zipInputStream.closeEntry()
```

koperagen:
Another issue is that file ending with *.gz can be *.tar.gz, and we cannot read it properly without some special handling. So, i suggest to either support it or at least provide an exception message that file should be just an archive and not a *.tar

koperagen:
After the fix it needs to be mentioned in the docs

Jolanrensen:
There's actually a lot of places where DataFrame assumes a type based on the file extension, but we should avoid that, as file extensions can be changed while the contents of the file are not.

Jolanrensen:
Will be solved in the new CSV implementation: "dataframe-csv". I will probably also migrate its new `Compression` class to the :core module in the future to solve reading zips from other read functions too.

PR #903, opened 2024-10-02
New CSV implementation

Jolanrensen:
See the issue for the progress: https://github.com/Kotlin/dataframe/issues/827

~TODO:~
- [x] Split double parser into separate PR when FastDoubleParser publishes ConfigurableDoubleParser: https://github.com/Kotlin/dataframe/pull/935
- [x] Clean git history after merging fast double parser

Some high level overview of the changes:
- introduces new `:dataframe-csv` module
- new functions, requiring `@OptIn(ExperimentalCsv::class)`:
  - `read(Csv|Tsv|Delim)(Str)` based on Deephaven
  - `write(Csv|Tsv|Delim)`, and `to(Csv|Tsv|Delim)Str` based on Apache commons csv
- Added `ParserOptions.skipTypes`
- `parse` can now recognize `Char`
- Extended supported `ColTypes` to all the types we can parse
- Added OOM warning to old `readCSV` function informing the user about our new experimental functions :)
- See all checked off boxes in https://github.com/Kotlin/dataframe/issues/827 for all the bugs this fixes or issues it takes in mind.

Fixes https://github.com/Kotlin/dataframe/issues/787
Fixes https://github.com/Kotlin/dataframe/issues/508
Fixes https://github.com/Kotlin/dataframe/issues/589
Fixes https://github.com/Kotlin/dataframe/issues/469
Fixes https://github.com/Kotlin/dataframe/issues/921

Example of the new KDocs and all supported parameters:
image
image
image


Jolanrensen:
NOTE: maybe we can `try {} catch(OutOfMemoryError) {}` around the apache commons read csv and recommend the new experimental csv reader.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/parse.kt
```diff
@@ -64,8 +67,17 @@ public data class ParserOptions(
     val dateTimeFormatter: DateTimeFormatter? = null,
     val dateTimePattern: String? = null,
     val nullStrings: Set<String>? = null,
+    val skipTypes: Set<KType> = emptySet(),
```
Jolanrensen, line 70: Maybe `skipTypes` be better be inverted to `checkTypes`. But that would mean exposing all `Parsers` so users/csv readers can filter which types to consider when parsing and which to skip. WDYT?

Jolanrensen:
Released dev version to test: `0.15.0-dev-4759`

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/readDelim.kt
```diff
@@ -0,0 +1,345 @@
...
+    ignoreExcessColumns: Boolean = IGNORE_EXCESS_COLUMNS,
+    quote: Char = QUOTE,
+    ignoreSurroundingSpaces: Boolean = IGNORE_SURROUNDING_SPACES,
+    trimInsideQuoted: Boolean = TRIM_INSIDE_QUOTED,
+    parseParallel: Boolean = PARSE_PARALLEL,
+    additionalCsvSpecs: CsvSpecs? = ADDITIONAL_CSV_SPECS,
+): DataFrame<*> {
+    // set up the csv specs
+    val csvSpecs = with(CsvSpecs.builder()) {
+        if (additionalCsvSpecs != null) from(additionalCsvSpecs)
```
Jolanrensen, line 102: Parameters set in `additionalCsvSpecs` are now overridden by the other (default) parameters. Do we think this is intended behavior? Or should we let `additionalCsvSpecs` overwrite all parameters given here?

Additionally, we could make a separate public function which takes _just_ a `CsvSpecs` instance and skips all our default parameters and DF logic. WDYT?

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/ListSink.kt
```diff
@@ -0,0 +1,202 @@
...
+internal interface SinkSource<T : Any> :
+    Sink<T>,
+    Source<T>
+
+/**
+ * Implementation of Deephaven's [Sink] and [Source] that stores data in an [ArrayList].
+ *
+ * The implementation is based on [Writing Your Own Data Sinks](https://github.com/deephaven/deephaven-csv/blob/main/ADVANCED.md).
+ *
+ * If we ever store column data unboxed / primitively, this needs to be modified.
```
Jolanrensen, line 32: Actually, we could also create a `ListSink` for each primitive type and `.asList()` the result. DataFrame can handle such columns. However, this could slow down processing after reading (as all values visited will need to be boxed) and potentially postpone the OOM exception. Unless the user filters first maybe...

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/readDelim.kt
Jolanrensen: Fixed by making it a lambda running on the builder :)

Jolanrensen:
Benchmark results (in terms of speed; seconds per operation, lower is better):

| benchmark | type | score | scoreError | scoreUnit |
| :--- | :--- | :--- | :--- | :--- |
| apache | small | 0.005 | 0.001 | s/op |
| apache | medium | 3.397 | 0.224 | s/op |
| apache | large | OOM |  | s/op |
| deephaven | small | 0.009 | 0.002 | s/op |
| deephaven | medium | 1.278 | 0.109 | s/op |
| deephaven | large | 134.84 | 16.472 | s/op |

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/ImportDataSchema.kt
```diff
@@ -47,6 +47,7 @@ public enum class DataSchemaVisibility {
     EXPLICIT_PUBLIC,
 }
 
+// TODO add more options
```
zaleslaw, line 50: TODO: convert to issues or fix

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/parse.kt
zaleslaw: Really weird feature, let's live a little bit with that solution to reflect later in code

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/ListSink.kt
```diff
@@ -0,0 +1,202 @@
...
+            LONG -> {
+                dest as LongArray
+                for ((srcIndex, destIndex) in (srcBeginAsInt..<srcEndAsInt).withIndex()) {
+                    val value = data[srcIndex] as Long?
+                    if (value != null) dest[destIndex] = value
+                    isNull[destIndex] = value == null
+                }
+            }
+
+            else -> error("unsupported sink state")
```
zaleslaw, line 197: Could you please improve the error message! Or give some advice

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/readDelim.kt
```diff
@@ -0,0 +1,343 @@
...
+        skipHeaderRows(skipLines)
+    } else {
+        skipRows(skipLines)
+    }
+
+/**
+ * Sets the correct parsers for the csv, based on [colTypes] and [ParserOptions.skipTypes].
+ * If [ColType.DEFAULT] is present, it sets the default parser.
+ */
+private fun CsvSpecs.Builder.parsers(
```
zaleslaw, line 253: I also have one question, @Jolanrensen did you measure a JUnit test coverage here? When I am starting to think about multiple variety and combinatorial explosion of programming paths here, it makes me crazy.

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/Compression.kt
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Compression algorithm to use when reading csv files.
+ * We support [GZIP][Compression.Gzip] and [ZIP][Compression.Zip] compression out of the box.
+ *
+ * Custom decompression algorithms can be added by creating an instance of [Custom].
+ *
+ * @param wrapStream function that wraps any [InputStream] into a decompressing [InflaterInputStream] stream
+ */
+public sealed class Compression<I : InputStream>(public open val wrapStream: (InputStream) -> I) {
```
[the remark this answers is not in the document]
zaleslaw: I think our core users will be happy with compression support

dataframe-csv/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/BenchmarkTest.kt
```diff
@@ -0,0 +1,54 @@
...
+@State(Scope.Benchmark)
+open class BenchmarkTest {
+
+    @Param("small", "medium", "large")
+    var type = ""
+    var file: File? = null
+
+    @Setup
+    fun setup() {
+        System.setProperty("org.slf4j.simpleLogger.defaultLogLevel", "info")
```
zaleslaw, line 29: it's really cool!

dataframe-csv/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/DelimCsvTsvTests.kt
```diff
@@ -0,0 +1,657 @@
...
+        df["duplicate1"].type() shouldBe typeOf<Char?>()
+        df["double"].type() shouldBe typeOf<Double?>()
+        df["number"].type() shouldBe typeOf<Double>()
+        df["time"].type() shouldBe typeOf<LocalDateTime>()
+
+        df.print(columnTypes = true, borders = true, title = true)
+    }
+
+    @Test
+    fun `read 2 compressed Csv`() {
```
zaleslaw, line 135: Do you have a test for reading something zipped, but that is not CSV or has an extension, but in reality has a weird format internally

dataframe-csv/src/test/resources/largeCsv.csv.gz
zaleslaw, line 1: I'm not sure that we should have this file in our repo just for performance test, this is a reason why I've excluded SQL dump from the repo

Let's discuss what it's possible to do, but it's fine
1) write a csv generator programm, which creates a file in tmp directory
2) save it externally and add a link for loading

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/DataFrameCustomDoubleParser.kt
```diff
@@ -0,0 +1,24 @@
...
+    override fun parse(bs: ByteSlice): Double =
+        try {
+            fastDoubleParser.parseOrNull(bs.data(), bs.begin(), bs.size())
+        } catch (e: Exception) {
+            null
+        } ?: throw NumberFormatException("Failed to parse double")
+
+    override fun parse(cs: CharSequence): Double =
+        fastDoubleParser.parseOrNull(cs.toString())
+            ?: throw NumberFormatException("Failed to parse double")
```
koperagen, line 23: maybe it will be useful to include substring to exception message

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/common.kt
```diff
@@ -1,28 +1,46 @@
...
-internal fun catchHttpResponse(url: URL, body: (InputStream) -> AnyFrame): AnyFrame {
+/**
+ * Opens a stream to [url] to create a [DataFrame] from it.
+ * If the URL is a file URL, the file is read directly.
+ * If the URL is an HTTP URL, it's also read directly, but if the server returns an error code,
+ * the error response is read as JSON and parsed as [DataFrame] too.
+ *
+ * Public so it may be used in other modules.
+ */
+public fun catchHttpResponse(url: URL, body: (InputStream) -> AnyFrame): AnyFrame {
```
koperagen, line 24: Did you consider other ways to expose this API. Is it somehow possible to keep in a separate artifact so it's only available with explicit dependency?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/csv.kt
```diff
@@ -331,58 +355,72 @@ public fun DataFrame.Companion.readDelim(
...
-    val records = if (readLines == null) {
-        csvParser.records
-    } else {
-        require(readLines >= 0) { "`readLines` must not be negative" }
-        val records = ArrayList<CSVRecord>(readLines)
-        val iter = csvParser.iterator()
-        var count = readLines ?: 0
-        while (iter.hasNext() && 0 < count--) {
-            records.add(iter.next())
+    try {
```
koperagen, line 358: Please extract try body in a readDelimImpl

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/CsvTests.kt
```diff
@@ -70,7 +70,7 @@ class CsvTests {
         df.nrow shouldBe 5
         df.columnNames()[5] shouldBe "duplicate1"
         df.columnNames()[6] shouldBe "duplicate11"
-        df["duplicate1"].type() shouldBe typeOf<String?>()
+        df["duplicate1"].type() shouldBe typeOf<Char?>()
```
koperagen, line 73: So it converts columns with single characters to Char?

dataframe-csv/build.gradle.kts
```diff
@@ -0,0 +1,170 @@
...
+        artifactId = project.name
+        description = "CSV support for Kotlin Dataframe"
+        packageName = artifactId
+    }
+}
+
+kotlin {
+    explicitApi()
+    sourceSets.all {
+        languageSettings {
```
koperagen, line 167: leftovers

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/DelimParams.kt
```diff
@@ -0,0 +1,245 @@
...
+     */
+    const val TRIM_INSIDE_QUOTED: Boolean = false
+
+    /**
+     * @param parseParallel Whether to parse the data in parallel. Default: `true`.
+     *
+     *   If `true`, the data will be parsed in parallel.
+     *   This is usually faster, but can be turned off for debugging.
+     */
+    const val PARSE_PARALLEL: Boolean = true
```
koperagen, line 195: i feel it's a big ambiguous. So it's not about deephaven parser, rather it's about DataColumn.parse?

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/Compression.kt
```diff
@@ -0,0 +1,105 @@
...
```
koperagen, line 19: Is this type parameter needed? In the end readCsv has Compression<*>, so it's the same as `Compression(public open val wrapStream: (InputStream) -> InputStream)`

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/Compression.kt
```diff
@@ -0,0 +1,105 @@
...
+    compression.doFirst(wrappedStream)
+
+    try {
+        return block(wrappedStream)
+    } finally {
+        compression.doFinally(wrappedStream)
+    }
+}
+
+public fun compressionStateOf(fileOrUrl: String): Compression<*> =
```
koperagen, line 94: Not sure about "state". 
I'd expect this factory to be in companion object of Compression, not on top level

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readCsv.kt
```diff
@@ -0,0 +1,292 @@
...
+ * @include [PATH_READ]
+ * @include [CSV_DELIMITER]
+ * @include [COMPRESSION]
+ * @include [CommonReadDelimDocs.CommonReadParams]
+ */
+public fun DataFrame.Companion.readCsv(
+    path: Path,
+    delimiter: Char = CSV_DELIMITER,
+    header: List<String> = HEADER,
+    hasFixedWidthColumns: Boolean = HAS_FIXED_WIDTH_COLUMNS,
```
koperagen, line 51: Are hasFixedWidthColumns and fixedColumnWidths connected? What will happen if i provide false and non-empty list?

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/Compression.kt
```diff
@@ -0,0 +1,105 @@
...
+                throw IllegalArgumentException("Zip file contains more than one entry")
+            }
+            inputStream.close()
+        }
+    }
+
+    /**
+     * No compression.
+     */
+    public data object None : Compression<InputStream>(wrapStream = { it })
```
koperagen, line 60: Seems like "None" is actually kind of "Detect", no? At least according to kdocs it seems like it is

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/readDelim.kt
```diff
@@ -0,0 +1,343 @@
...
+                /* specs = */ csvSpecs,
+                /* stream = */ decompressedInputStream.skippingBomCharacters(),
+                /* sinkFactory = */ ListSink.SINK_FACTORY,
+            )
+        } catch (e: CsvReaderException) {
+            // catch case when the file is empty and header needs to be inferred from it.
+            if (e.message ==
+                "Can't proceed because hasHeaderRow is set but input file is empty or shorter than skipHeaderRows"
+            ) {
+                return@readDelimImpl DataFrame.empty()
```
koperagen, line 150: Are you sure about it? I incline to think that this is an error that needs to be fixed, having and empty dataframe instead of the exception will be not-so-clear

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/parse.kt
```diff
@@ -388,6 +388,8 @@ internal object Parsers : GlobalParserOptions {
                 null
             }
         },
+        // Char
+        stringParser<Char> { it.singleOrNull() },
```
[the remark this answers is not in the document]
Jolanrensen: not necessarily, columns with single characters can be quite common, especially when reading pivoted data, plus it could save memory, which is always a good thing, right? :)

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/readDelim.kt
Jolanrensen: It was to align with the old behavior: feeding an empty file/string should result in an empty dataframe (we could change this of course).

Deephaven can do this successfully if you supply a header yourself (it simply returns empty columns), however it throws an exception when it needs to guess a header from an empty file, like here. But if the file is empty anyway, it makes no sense to require the user to supply a header to avoid this exception, so that's why I caught it like this.

I also tried detecting whether the inputstream is empty before passing it to CsvReader, but I didn't find a reliable way (`.available()` can return `0`, even though the stream still has data).

dataframe-csv/src/test/resources/abc.tsv
```diff
@@ -0,0 +1,3 @@
+A	B	C
```
koperagen, line 1: Maybe it'd be ok to just read files from `../core/src/test/resources`?

dataframe-csv/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/DelimCsvTsvTests.kt
```diff
@@ -0,0 +1,657 @@
...
+
+    @Test
+    fun `read standard CSV with floats when user has alternative locale`() {
+        val currentLocale = Locale.getDefault()
+        try {
+            Locale.setDefault(Locale.forLanguageTag("ru-RU"))
+            val df = DataFrame.readCsv(wineCsv, delimiter = ';')
+            val schema = df.schema()
+
+            fun assertColumnType(columnName: String, kClass: KClass<*>) {
```
koperagen, line 183: duplicated function

dataframe-csv/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/DelimCsvTsvTests.kt
```diff
@@ -0,0 +1,657 @@
...
+        producedFile.exists() shouldBe true
+        producedFile.readText() shouldBe "1,2,3\r\n1,3,2\r\n"
+        producedFile.delete()
+    }
+
+    @Test
+    fun `check integrity of example data`() {
+        val df = DataFrame.readCsv(
+            "../data/jetbrains_repositories.csv",
+            skipLines = 1, // now needs this, ignoreEmptyLines cannot catch it
```
koperagen, line 323: let's update the data?

plugins/symbol-processor/src/main/kotlin/org/jetbrains/dataframe/ksp/DataSchemaGenerator.kt
```diff
@@ -155,6 +155,7 @@ class DataSchemaGenerator(
             codeGenerator.createNewFile(Dependencies(true, importStatement.origin), packageName, "$name.Generated")
 
         val formats = listOf(
+            // TODO new Csv() and Tsv()
```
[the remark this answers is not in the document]
Jolanrensen: It's already part of the umbrella issue todo-list: https://github.com/Kotlin/dataframe/issues/827

dataframe-csv/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/DelimCsvTsvTests.kt
```diff
@@ -0,0 +1,657 @@
...
+        df["double"].type() shouldBe typeOf<Double?>()
+        df["number"].type() shouldBe typeOf<Double>()
+        df["time"].type() shouldBe typeOf<LocalDateTime>()
+
+        df.print(columnTypes = true, borders = true, title = true)
+    }
+
+    @Test
+    fun `read GZ Csv`() {
+        val df = DataFrame.readCsv(simpleCsvGz)
```
koperagen, line 101: I think a lot of these assertions aren't needed. We only need to test that reading .gz works the same as reading plain csv

dataframe-csv/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/DelimCsvTsvTests.kt
```diff
@@ -0,0 +1,657 @@
...
+
+        val dfThree = DataFrame.readCsv(simpleCsv, readLines = 3)
+        dfThree.rowsCount() shouldBe 3
+
+        val dfFull = DataFrame.readCsv(simpleCsv, readLines = 10)
+        dfFull.rowsCount() shouldBe 5
+    }
+
+    @Test
+    fun `if string starts with a number, it should be parsed as a string anyway`() {
```
koperagen, line 238: I wanted to see actual file to understand what name implies. Maybe it'd be better to inline the file and use readCsvStr

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/ListSink.kt
Jolanrensen: The datatype supplied here can only be byte, short, int, or long, as described by https://github.com/deephaven/deephaven-csv/blob/main/ADVANCED.md#supporting-the-fast-path-for-numeric-type-inference, so this case never occurs, unless deephaven updates and does something weird. But I'll add a comment

dataframe-csv/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/DelimCsvTsvTests.kt
```diff
@@ -0,0 +1,657 @@
...
+import java.io.StringWriter
+import java.math.BigDecimal
+import java.net.URL
+import java.util.Locale
+import java.util.zip.GZIPInputStream
+import kotlin.reflect.KClass
+import kotlin.reflect.typeOf
+
+@Suppress("ktlint:standard:argument-list-wrapping")
+class DelimCsvTsvTests {
```
koperagen, line 33: We already have a lot of CSV tests in core tests. Do you copy them here or this is a completely written from scratch? I understand that API is different, but at least for default case when only path argument is supplied, it'd be good to compare results and to be able to say they're identical or not

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/parse.kt
Jolanrensen: It's mostly a performance thing, but the problem is that we don't have a user-facing way to refer to parsers, so we cannot say: "please parse this string column and try this type but not this type". I'm not sure how the API should look here

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/common.kt
Jolanrensen: friend modules would work, but I'm not entirely a fan of those. A separate module could also work maybe. I could think of some more functions/classes that would benefit from being in a separate internal module (code generation, the double parser, these file/url/path/string helper functions, etc.) But that might be better solved in another issue.

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/CsvTests.kt
Jolanrensen: Deephaven has the same functionality, so I thought it would be a good idea to make our parsers behave the same

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/DelimParams.kt
Jolanrensen: no it's deephaven, our parser is not parallel atm. I'll specify.

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/DataFrameCustomDoubleParser.kt
Jolanrensen: the message is already debug-logged in parseOrNull.
This `NumberFormatException` is just for Deephaven to catch. Actually, I can even remove the message altogether, as Deephaven doesn't do anything with it either. (they just catch it and return false)

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/readDelim.kt
```diff
@@ -0,0 +1,343 @@
...
+        trim(trimInsideQuoted)
+        concurrent(parseParallel)
+        header(header)
+        hasFixedWidthColumns(hasFixedWidthColumns)
+        if (hasFixedWidthColumns && fixedColumnWidths.isNotEmpty()) fixedColumnWidths(fixedColumnWidths)
+        skipLines(takeHeaderFromCsv = header.isEmpty(), skipLines = skipLines)
+
+        // Deephaven's LocalDateTime parser is unconfigurable, so if the user provides a locale, pattern, or formatter,
+        // we must use our own parser for LocalDateTime and let Deephaven read them as Strings.
+        val useDeepHavenLocalDateTime = with(parserOptions) {
```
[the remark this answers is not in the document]
Jolanrensen: Probably yes, however, in theory the user should not be able to tell the difference, only in speed.

- If they provide no configuration, it will use the fast LocalDateTime parser of deephaven and if that fails, it will remain a String and our slower LocalDateTime parser will try it too.

- If they do provide configuration, then deephaven's parser cannot be used and our slower LocalDateTime parser will be used.

In both cases, the user will get the expected result and where possible we save some performance. But we cannot "recommend" to not specify a configuration to gain performance, because that's dependent on the data the user has to read. All we can do is mention it somewhere in the documentation.

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/Compression.kt
Jolanrensen: no, it's actually "none". If a user supplies `compression = Compression.None`, the inputStream will be supplied to deephaven-csv directly.

If a user does _not_ supply a compression argument, the default will be "`Compression.None`, unless detected otherwise from the input file or url." So `compressionStateOf()` will be called (if possible) and the right one will be chosen based on the file extension, defaulting to `None` if that fails.

Hope that clarifies it :) If not, please tell me which part can be written clearer

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/Compression.kt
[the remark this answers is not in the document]
Jolanrensen: Will move it after dataframe-csv becomes non-experimental

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readCsv.kt
Jolanrensen: Then `fixedColumnWidths` will be ignored. They are only set if `hasFixedWidthColumns == true`.

Deephaven throws an error in this case. We could also do that if you like

dataframe-csv/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/DelimCsvTsvTests.kt
Jolanrensen: I think it's good that we know the difference in behavior for the new implementation; what happens if we try to read a csv file with a blank line at the top.

dataframe-csv/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/DelimCsvTsvTests.kt
Jolanrensen: yeah that's fair

dataframe-csv/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/DelimCsvTsvTests.kt
Jolanrensen: yes I copied all of them and added a couple more, so I could see they behave the same.

Jolanrensen: and actually, I changed the default arguments such that the behavior of not specifying any additional settings is as close as possible as the old behavior. Things like `allowMissingColumns` ans `ignoreExcessColumns` for instance.

dataframe-csv/src/test/resources/abc.tsv
Jolanrensen: maybe yeah, but I think once we remove the old csv implementation from :core we can better remove the tests and resources there. Then they're not duplicate anymore

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/readDelim.kt
Jolanrensen: For this specific function, yes, all cases are covered.
If you narrow it down, there are fewer paths than you think:

- if no colTypes are given
  - let deephaven use all its default parsers
  - subtract parsers of skipTypes if those are supplied
- if colTypes are supplied
  - if DEFAULT is among the values
    - set the parser for each supplied column+coltype
    - let deephaven use _only_ the parser given as DEFAULT type
  - if DEFAULT is not among the values
    - set the parser for each supplied column+coltype
    - let deephaven use all its default parsers
    - subtract parsers of skipTypes if those are supplied

(you may notice that `skipTypes` will never skip a type explicitly set by `colTypes`. This is intended.)

I will add this overview to the kdoc too, It may be helpful when debugging.

I will add a failing date test though. (if it fails, it remains String)

dataframe-csv/src/test/resources/largeCsv.csv.gz
Jolanrensen: generating it sounds interesting. 
Storing it externally, while I'm normally against doing that for tests, might work for this case, since benchmarks are not run as part of the normal test suite.

Jolanrensen: that said, I can also just remove the benchmarking test altogether. What do you think?

zaleslaw: Let's remove!

zaleslaw (approved):

Jolanrensen:
@zaleslaw @koperagen My two latest commits (https://github.com/Kotlin/dataframe/compare/703338a7042c30f8629bfc9e5979c0a6560a067f..53e2f64daa1b8167c56d5c4a4fd33a66cf78bfe2) contain a lot of new changes because I figured out how `ParserOptions` actually work in the library:
- There's a global setting `DataFrame.parser` which should be respected if either: `ParserOptions == null` or any of its parameters are `null`. I made sure the csv readers respect this too.
- I added my new `ParserOptions` arguments to the global one as well (and I made it not a data class anymore for better future binary compatibility).
- "skip types except" makes no sense, we have `convertTo` for that. But there was no overload for `stringColumn.convertTo` that took any `ParserOptions`, so I added that.
- Moved `convertToDouble` to `convertToDoubleImpl` and added explanation in KDocs for why (I think) it exists (and it now accepts `nullStrings` as well).
- Added more relevant tests

Could you rereview if you like?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/parse.kt
```diff
@@ -293,14 +306,19 @@ internal object Parsers : GlobalParserOptions {
...
         parser
     }
 
-    private val posixDoubleParser = FastDoubleParser(
-        ParserOptions(locale = Locale.forLanguageTag("C.UTF-8")),
-    )
+    // same as parserToDoubleWithOptions, but overrides the locale to C.UTF-8
+    private val posixParserToDoubleWithOptions = stringParserWithOptions { options ->
+        val parserOptions = (options ?: ParserOptions()).copy(locale = Locale.forLanguageTag("C.UTF-8"))
+        TODO()
```
koperagen, line 317: something is not right here

Jolanrensen: oops thanks

zaleslaw:
Is it a first global setting in KDF, or we had something earlier? Should user set up it manually and if it will be replaced is it replaced for the whole project, using our library? In that way it's closer to the CONFIG file which is visible or Spring Bean.

I'm ok, but looks like a temporary solution, could you create a ticket with research tag for that

zaleslaw (approved):

Jolanrensen:
> Is it a first global setting in KDF, or we had something earlier? Should user set up it manually and if it will be replaced is it global for the whole project, using our library? In that way it's closer to the CONFIG file which is visible or Spring Bean.
> 
> I see that it's inherited from external component, isn't it?
> 
> I'm ok, but looks like a temporary solution, could you create a ticket with research tag for that

`DataFrame.parser` has been part of the library for years, I just didn't know about it yet (it's also part of the documentation https://kotlin.github.io/dataframe/parse.html, at the bottom). It's similar to `Locale.setDefault()` on the JVM, so yes, it's global, like a config file, but it can be changed programmatically because it's stored in a singleton object.

In all cases, the global default can also be overridden by providing a `ParserOptions` instance to the function, so we embrace both a functional and imperative style which I think is cool.

PR #907, opened 2024-10-03
Generate valid code in transform(call) when interpret(call) fails

koperagen:
It fixes `Caused by: java.lang.IllegalStateException: Local class should have its name computed in InventNamesForLocalClasses: <stub>.Invoke_76`


Jolanrensen (approved):
Nice! Hopefully it fixes this crashing
image
and plenty other similar cases :)

PR #908, opened 2024-10-03
[Compiler plugin] Support dataFrameOf(Pair<String, List<T>)

koperagen:
Change in `buildLetCall` is needed because this function has a unique `(...) -> DataFrame` shape, while most other have some kind of `Receiver.(...) -> DataFrame`
`df.convert { }.with { }` -> `df.convert { }.let { it.with { } as ... }`

for  `(...) -> DataFrame` it will be
`dataFrameOf(pairs)` -> `run { dataFrameOf(pairs) as ... }`

### DIFF
--- a/plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/loadInterpreter.kt
+++ b/plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/loadInterpreter.kt
@@ -60,13 +60,20 @@ import org.jetbrains.kotlin.fir.expressions.FirFunctionCall
 import org.jetbrains.kotlin.fir.expressions.FirGetClassCall
 import org.jetbrains.kotlin.fir.expressions.FirLiteralExpression
 import org.jetbrains.kotlin.fir.expressions.FirResolvedQualifier
+import org.jetbrains.kotlin.fir.expressions.UnresolvedExpressionTypeAccess
 import org.jetbrains.kotlin.fir.references.FirResolvedNamedReference
+import org.jetbrains.kotlin.fir.references.resolved
+import org.jetbrains.kotlin.fir.references.symbol
+import org.jetbrains.kotlin.fir.references.toResolvedNamedFunctionSymbol
 import org.jetbrains.kotlin.fir.resolve.fqName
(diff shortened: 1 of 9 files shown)


koperagen:
image

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/loadInterpreter.kt
```diff
@@ -91,8 +100,16 @@ import org.jetbrains.kotlinx.dataframe.plugin.impl.api.ToDataFrameFrom
 import org.jetbrains.kotlinx.dataframe.plugin.impl.api.ToTop
 import org.jetbrains.kotlinx.dataframe.plugin.impl.api.Update0
 import org.jetbrains.kotlinx.dataframe.plugin.impl.api.UpdateWith0
+import org.jetbrains.kotlinx.dataframe.plugin.utils.Names
 
+@OptIn(UnresolvedExpressionTypeAccess::class)
 internal fun FirFunctionCall.loadInterpreter(session: FirSession): Interpreter<*>? {
+    if (
+        calleeReference.toResolvedNamedFunctionSymbol()?.callableId == Names.TO &&
```
Jolanrensen, line 108: so.. does this mean it fails for `dataFrameOf(Pair("name", listOf(1, 2, 3)))`?

Jolanrensen: It's probably fine, most people will use `to` anyway, but it might be something we should mention in a doc somewhere, for that one person who's going to be very confused as to why their Pair is not recognized by the plugin if the function only accepts pairs as the type argument.

Jolanrensen (approved):

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/loadInterpreter.kt
koperagen: Good point. Let's support Pair after i put in place more generic way for handling stdlib functions other than "if". And indeed with such things it's much easier to run into something plugin won't understand. One of the things that can be addressed with custom Checker that will highlight it as warning

Issue #775, opened 2024-07-11
Update library descriptor


koperagen:
New dataframe-jupyter module should be added here in dependencies
https://github.com/Kotlin/kotlin-jupyter-libraries/blob/master/dataframe.json
However, after kernel is updated in the Kotlin Notebook to use this descriptor (for example in 2024.2)
`%use dataframe(0.13.1)` will produce an error
![image](https://github.com/Kotlin/dataframe/assets/12936457/753de517-a2c2-4c8f-9e66-2460fe7adbec)

So we either need to
1. rollback the change (but still likely will have to do it in the future)
2. depend on dataframe-jupyter from dataframe, but it requires JVM 11 bump
3. publish empty kotlin-jupyter for 0.13.1, 0.12.0 and some other versions. So it will be resolved and cause no errors.
To do this we need to create a branch from commit before the refactoring, create an empty module, then create a few branches from it and set `version=` in gradle.properties. Then execute publishDataframe-jupyterPublicationToMavenRepository on teamcity

koperagen:
Even bigger problem is that on older kernels `%use dataframe(0.14.0)` will resolve only "dataframe", without jupyter integration.
So https://github.com/Kotlin/dataframe/pull/776.
And these should be re-open
https://github.com/Kotlin/dataframe/issues/241
https://github.com/Kotlin/dataframe/issues/700

koperagen:
> Even bigger problem is that on older kernels `%use dataframe(0.14.0)` will resolve only "dataframe", without jupyter integration.

There can be a migration period when both core module and dataframe-jupyter apply the same integration (just making sure to do it once)

Jolanrensen:
We could also tweak the .json descriptor file such that it resolves a different set of dependencies depending on the version.

https://github.com/Kotlin/kotlin-jupyter/blob/master/docs/libraries.md#descriptor-api-3

It's not the neatest, but in theory it could make sure the dataframe-jupyter dependency is added only when the version of dataframe is 0.14.0+. Something like:
```json
{
  "init": [
    "if ($v >= 0.14.0) USE { dependencies(\"...dataframe-jupyter:$v\") }"
  ]
}
```
(not sure if we can get the $v, but there's probably a way)

Alternatively, we could just prompt users to update their kernel versions with: https://github.com/Kotlin/kotlin-jupyter/blob/master/docs/libraries.md#minimal-kernel-version-supported-by-the-library

@ileasile What do you think is the wisest solution?

[...]

PR #909, opened 2024-10-04
GeoDataFrame init

AndreiKingsley:


settings.gradle.kts
```diff
@@ -41,3 +41,4 @@ plugins {
 }
 include("dataframe-excel")
 include("core")
+include("dataframe-geo")
```
Jolanrensen, line 44: if you want dataframe-geo to be imported by default when users depend on the generic dataframe package, also add dataframe-geo in build.gradle.kts

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/IntegrationGeo.kt
```diff
@@ -0,0 +1,84 @@
...
+    private fun KotlinKernelHost.execute(
+        codeWithConverter: CodeWithConverter,
+        property: KProperty<*>,
+        type: KType,
+    ): VariableName? {
+        val variableName = "(${property.name}${if (property.returnType.isMarkedNullable) "!!" else ""} as $type)"
+        return execute(codeWithConverter, variableName)
+    }
+
+    override fun Builder.onLoaded() {
```
Jolanrensen, line 50: what does this jupyter integration do exactly? could you add some docs?

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/io/read.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.geotools.data.simple.SimpleFeatureCollection
+import org.geotools.geojson.feature.FeatureJSON
+import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.geo.GeoDataFrame
+import org.jetbrains.kotlinx.dataframe.geo.geotools.toGeoDataFrame
+import org.jetbrains.kotlinx.dataframe.io.asURL
+import java.net.URL
+
+
+fun GeoDataFrame.Companion.readGeoJSON(path: String): GeoDataFrame<*> {
```
Jolanrensen, line 13: readGeoJson is proper camel casing

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/geotools/toSimpleFeatureCollection.kt
```diff
@@ -0,0 +1,46 @@
...
+
+fun GeoDataFrame<*>.toSimpleFeatureCollection(
+    name: String? = null,
+    singleGeometryType: Boolean = false
+): SimpleFeatureCollection {
+    val typeBuilder = SimpleFeatureTypeBuilder()
+    typeBuilder.name = name ?: "geodata"
+    typeBuilder.setCRS(crs)
+    val geometryClass = if (singleGeometryType) {
+        // todo singleOrNull() ?: error()
```
Jolanrensen, line 22: many todo's haha

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/geotools/ToGeoDataFrame.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.geotools.data.simple.SimpleFeatureCollection
+import org.jetbrains.kotlinx.dataframe.DataColumn
+import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.api.Infer
+import org.jetbrains.kotlinx.dataframe.api.toDataFrame
+import org.jetbrains.kotlinx.dataframe.geo.GeoDataFrame
+import org.jetbrains.kotlinx.dataframe.geo.GeoFrame
+import org.locationtech.jts.geom.Geometry
+
+fun SimpleFeatureCollection.toGeoDataFrame(): GeoDataFrame<*> {
```
Jolanrensen, line 16: please rename the file to "toGeoDataFrame" like you did with the other file

Jolanrensen (changes requested):
good initial version :) just some comments here and there

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/io/write.kt
```diff
@@ -0,0 +1,58 @@
...
+import org.geotools.api.data.Transaction
+import org.geotools.geojson.feature.FeatureJSON
+import org.jetbrains.kotlinx.dataframe.geo.GeoDataFrame
+import org.jetbrains.kotlinx.dataframe.geo.geotools.toSimpleFeatureCollection
+import java.io.File
+
+
+fun GeoDataFrame<*>.writeGeoJson(path: String): Unit = writeGeoJson(File(path))
+
+fun GeoDataFrame<*>.writeGeoJson(file: File) {
```
zaleslaw, line 14: Please add for IO part some JUnit tests for reading/writing files close to your examples https://gist.github.com/AndreiKingsley/8eada06f174c681e953f8d939248d153

zaleslaw (changes requested):
Generally, I am approving these changes, but want to summarize some thoughts and make this piece of code more mature and maintanable

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/GeoDataFrame.kt
```diff
@@ -0,0 +1,29 @@
+package org.jetbrains.kotlinx.dataframe.geo
+
+import org.geotools.api.referencing.crs.CoordinateReferenceSystem
+import org.geotools.geometry.jts.JTS
+import org.geotools.referencing.CRS
+import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.api.update
+import org.jetbrains.kotlinx.dataframe.api.with
+
+class GeoDataFrame<T : GeoFrame>(val df: DataFrame<T>, val crs: CoordinateReferenceSystem?) {
```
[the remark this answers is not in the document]
Jolanrensen: Every function returning a DataFrame would have to be wrapped with `GeoDataFrame`. Think about it, if we make `class GeoDataFrame : DataFrame<*>`, we could call:
```kt
val geoDf: GeoDataFrame = ...
val result = geoDf.filter { a in 0..2 }
```
But `result` will be `DataFrame<*>`, losing access to `crs`. It's the nature of the extension function DSL of DataFrame.

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/jts/geometryExtenstions.kt
```diff
@@ -0,0 +1,24 @@
...
+    val scale = AffineTransformation.scaleInstance(value, value)
+
+    val moveBack = AffineTransformation.translationInstance(centroid.x, centroid.y)
+
+    val transformation = moveToOrigin.compose(scale).compose(moveBack)
+
+    return transformation.transform(this)
+}
+
+fun Geometry.translate(valueX: Double, valueY: Double): Geometry {
```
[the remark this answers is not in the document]
zaleslaw: the main issue, that we are doing this not for math guys, but for Kotlin developers who are not familiar with all these concepts

settings.gradle.kts
AndreiKingsley: No, I'd like to be it separate

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/IntegrationGeo.kt
AndreiKingsley: Hmm, ok, but there's no docs for core integration

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/geotools/toSimpleFeatureCollection.kt
AndreiKingsley: I believe it's ok since it's very alpha.

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/GeoDataFrame.kt
```diff
@@ -0,0 +1,29 @@
...
+    fun update(updateBlock: DataFrame<T>.() -> DataFrame<T>): GeoDataFrame<T> {
```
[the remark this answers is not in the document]
AndreiKingsley: @koperagen @zaleslaw

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/GeoDataFrame.kt
```diff
@@ -0,0 +1,29 @@
...
+        val sourceCRS: CoordinateReferenceSystem = this.crs ?: DEFAULT_CRS
+        val transform = CRS.findMathTransform(sourceCRS, targetCRS, true)
+        return GeoDataFrame(
+            df.update { geometry }.with { JTS.transform(it, transform) },
+            targetCRS
+        )
+    }
+
+    companion object {
+        val DEFAULT_CRS = CRS.decode("EPSG:4326", true)
```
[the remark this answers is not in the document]
AndreiKingsley: Yes, it even exists in some form in GeoTools (see [DefaultEngineeringCRS](https://docs.geotools.org/stable/javadocs/org/geotools/referencing/crs/DefaultEngineeringCRS.html)).  But here is a more global question - whether it is necessary to use geotools API directly or should we wrap it completely.

dataframe-geo/build.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+        alias(jupyter.api)
+    }
+}
+
+group = "org.jetbrains.kotlinx"
+
+repositories {
+    // geo repositories should come before Maven Central
+    maven("https://maven.geotoolkit.org")
+    maven("https://repo.osgeo.org/repository/release")
```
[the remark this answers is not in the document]
AndreiKingsley: Unfortunately, the user will need to add these repositories. However, I believe we can fix this with shadow jar.

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/GeoFrame.kt
```diff
@@ -0,0 +1,35 @@
...
+
+import org.jetbrains.kotlinx.dataframe.ColumnsContainer
+import org.jetbrains.kotlinx.dataframe.DataColumn
+import org.jetbrains.kotlinx.dataframe.annotations.DataSchema
+import org.locationtech.jts.geom.Geometry
+import org.locationtech.jts.geom.MultiPolygon
+import org.locationtech.jts.geom.Polygon
+
+@DataSchema
+interface GeoFrame {
```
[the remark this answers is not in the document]
AndreiKingsley: `WithGeometry`?

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/GeoDataFrame.kt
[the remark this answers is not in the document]
koperagen: In my opinion "Df" is redundant and doesn't look good. Plus it's already clear from signature that it applies to DataFrame

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/IntegrationGeo.kt
koperagen: Maybe a test like below will make things clearer
for reference:
org.jetbrains.kotlinx.dataframe.jupyter.DataFrameJupyterTest and org.jetbrains.kotlinx.dataframe.jupyter.CodeGenerationTests

test:
```
val geo = **create geo**
val updatedGeo = geo.update { add("col") { 1 }  }
updatedGeo.df.col
```

Jolanrensen: > Hmm, ok, but there's no docs for core integration

That's not a good argument haha. We also need kdocs there :)

dataframe-geo/build.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+    implementation(libs.jts.core)
+    implementation(libs.jts.io.common)
+
+    implementation(libs.ktor.client.core)
+    implementation(libs.ktor.client.cio)
+    implementation(libs.ktor.client.content.negotiation)
+    implementation(libs.ktor.serialization.kotlinx.json)
+
+}
+
```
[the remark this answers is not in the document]
Jolanrensen: Only for the jupyter integration?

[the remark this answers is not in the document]
AndreiKingsley: Yes

dataframe-geo/build.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+tasks.processJupyterApiResources {
+    libraryProducers = listOf("org.jetbrains.kotlinx.dataframe.jupyter.IntegrationGeo")
+}
+
+
+tasks.test {
+    useJUnitPlatform()
+}
+kotlin {
+    jvmToolchain(11)
```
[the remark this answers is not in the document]
AndreiKingsley: We have a global problem - DataFrame uses 8 and the Jupyter API uses 11. Therefore, in the future it will be necessary to allocate separate module(s) for integrations.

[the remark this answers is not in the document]
AndreiKingsley: https://github.com/Kotlin/dataframe/issues/775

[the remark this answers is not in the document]
AndreiKingsley: but for now, for convenience, I'd keep it 11

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/jts/geometryExtenstions.kt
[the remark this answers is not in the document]
AndreiKingsley: it's a well-established concept, so I can't agree: 
https://en.wikipedia.org/wiki/Translation_(geometry)

dataframe-geo/build.gradle.kts
```diff
@@ -0,0 +1,73 @@
+import org.jetbrains.kotlin.gradle.tasks.BaseKotlinCompile
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+    with(libs.plugins) {
+        alias(kotlin.jvm)
+        alias(publisher)
+        alias(jupyter.api)
```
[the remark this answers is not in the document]
AndreiKingsley: Ok, but comment it because it adds too many linting warnings

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/GeoDataFrame.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.api.update
+import org.jetbrains.kotlinx.dataframe.api.with
+
+class GeoDataFrame<T : GeoFrame>(val df: DataFrame<T>, val crs: CoordinateReferenceSystem?) {
+    fun update(updateBlock: DataFrame<T>.() -> DataFrame<T>): GeoDataFrame<T> {
+        return GeoDataFrame(df.updateBlock(), crs)
+    }
+
+    fun applyCRS(targetCRS: CoordinateReferenceSystem? = null): GeoDataFrame<T> {
```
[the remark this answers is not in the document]
AndreiKingsley: Done

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/IntegrationGeo.kt
AndreiKingsley: Done

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/io/read.kt
AndreiKingsley: Done

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/geotools/ToGeoDataFrame.kt
AndreiKingsley: Done

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/GeoFrame.kt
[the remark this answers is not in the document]
AndreiKingsley: Done

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/jts/geometryExtenstions.kt
```diff
@@ -0,0 +1,24 @@
+package org.jetbrains.kotlinx.dataframe.geo.jts
+
+import org.locationtech.jts.geom.Geometry
+import org.locationtech.jts.geom.util.AffineTransformation
+
+fun Geometry.scale(value: Double): Geometry {
```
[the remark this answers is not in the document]
AndreiKingsley: Done

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/io/write.kt
AndreiKingsley: Done

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/jts/geometryExtenstions.kt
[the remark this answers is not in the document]
Jolanrensen: I agree with Andrei on this one. "translate" is common enough I think

dataframe-geo/build.gradle.kts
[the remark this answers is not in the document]
Jolanrensen: I mean, there's a reason they show up. Please fix them. Most of the time you can run `gradlew ktLintFormat` and it will auto-fix. For other cases, I recommend installing https://plugins.jetbrains.com/plugin/15057-ktlint which can read the .editorconfig file and highlight exactly what's not linted.

dataframe-geo/build.gradle.kts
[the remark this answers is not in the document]
Jolanrensen: Actually, you can probably remove this piece, as `kotlin.jvmToolChain(11)` is already called from the main `build.gradle.kts` file; It configures all subprojects automatically to target jvm 8 in the published jars while setting the jvm toolchain (so the version gradle uses to build and test the project) to 11.

dataframe-geo/build.gradle.kts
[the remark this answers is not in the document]
Jolanrensen: please check whether you can do that before this is merged

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/geocode/Geocoder.kt
```diff
@@ -0,0 +1,121 @@
...
+    "region_query_counties" : null,
+    "ambiguity_resolver" : {
+      "ambiguity_resolver_ignoring_strategy" : null,
+      "ambiguity_resolver_box" : null,
+      "ambiguity_resolver_closest_coord" : null
+    }
+    }
+    """.trimIndent()
+
+    private fun geocodeQuery(countries: List<String>) = """
```
Jolanrensen, line 42: This is exactly where the linter will help :)

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/jts/geometryExtenstions.kt
```diff
@@ -0,0 +1,79 @@
+package org.jetbrains.kotlinx.dataframe.geo.jts
```
Jolanrensen, line 1: typo in filename: *Extensions

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/io/read.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.geotools.geojson.feature.FeatureJSON
+import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.geo.GeoDataFrame
+import org.jetbrains.kotlinx.dataframe.geo.geotools.toGeoDataFrame
+import org.jetbrains.kotlinx.dataframe.io.asURL
+import java.net.URL
+
+
+fun GeoDataFrame.Companion.readGeoJson(path: String): GeoDataFrame<*> {
+    return readGeoJson(asURL(path))
```
Jolanrensen, line 14: One thing the linter will complain about is that these should be expression functions. One `ktLintFormat` can replace them all at once tho

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/GeoDataFrame.kt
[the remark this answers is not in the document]
Jolanrensen: maybe `.modify {}` or `.transform {}` are okay?

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/GeoDataFrame.kt
```diff
@@ -0,0 +1,66 @@
+package org.jetbrains.kotlinx.dataframe.geo
+
+import org.geotools.api.referencing.crs.CoordinateReferenceSystem
+import org.geotools.geometry.jts.JTS
+import org.geotools.referencing.CRS
+import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.api.update
+import org.jetbrains.kotlinx.dataframe.api.with
+
+/**
```
Jolanrensen, line 10: good docs :D

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/GeoDataFrame.kt
```diff
@@ -0,0 +1,66 @@
...
+        // Use WGS 84 by default TODO
+        val sourceCRS: CoordinateReferenceSystem = this.crs ?: DEFAULT_CRS
+        val transform = CRS.findMathTransform(sourceCRS, targetCrs, true)
+        return GeoDataFrame(
+            df.update { geometry }.with { JTS.transform(it, transform) },
+            targetCrs
+        )
+    }
+
+    override fun equals(other: Any?): Boolean {
```
Jolanrensen, line 52: missing hashCode and toString functions

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/WithGeometry.kt
```diff
@@ -0,0 +1,35 @@
...
+    override val geometry: Polygon
+}
+
+@DataSchema
+interface WithMultiPolygon : WithGeometry {
+    override val geometry: MultiPolygon
+}
+
+@get:JvmName("geometry")
+val <T : WithGeometry> ColumnsContainer<T>.geometry: DataColumn<Geometry>
```
Jolanrensen, line 26: I think the `T` is not needed, right? `ColumnsContainer<WithGeometry>` has the same effect

Jolanrensen: Also, for the generated accessor we have both `ColumnsContainer` and `DataRow` extensions, so probably you should add `DataRow<WithGeometry>.geometry: Geometry` etc. too :)

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/IntegrationGeo.kt
Jolanrensen: thanks!

Jolanrensen:
Left a couple more comments :) Thanks for fixing some stuff already!
Also, don't forget to update from master. Most important recent changes include:
- a README.md file in each module with a small description of what the module does
- binary compatibility checker, you can add the plugin, run `apiDump` and add the api file to git just before merging this PR

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/GeoDataFrame.kt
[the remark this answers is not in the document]
zaleslaw: I suggest not wrap for now, we could not avoid the learning of that

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/geotools/toGeoDataFrame.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.geotools.data.simple.SimpleFeatureCollection
+import org.jetbrains.kotlinx.dataframe.DataColumn
+import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.api.Infer
+import org.jetbrains.kotlinx.dataframe.api.toDataFrame
+import org.jetbrains.kotlinx.dataframe.geo.GeoDataFrame
+import org.jetbrains.kotlinx.dataframe.geo.WithGeometry
+import org.locationtech.jts.geom.Geometry
+
+fun SimpleFeatureCollection.toGeoDataFrame(): GeoDataFrame<*> {
```
zaleslaw, line 16: Please document it with KDocs, looks like it will not change too much in future

zaleslaw: It's worth to add in comment the summarization of the underlying algorithm

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/geotools/toSimpleFeatureCollection.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.geotools.data.simple.SimpleFeatureCollection
+import org.geotools.feature.simple.SimpleFeatureBuilder
+import org.geotools.feature.simple.SimpleFeatureTypeBuilder
+import org.jetbrains.kotlinx.dataframe.api.forEach
+import org.jetbrains.kotlinx.dataframe.api.map
+import org.jetbrains.kotlinx.dataframe.api.single
+import org.jetbrains.kotlinx.dataframe.geo.GeoDataFrame
+import org.locationtech.jts.geom.Geometry
+
+fun GeoDataFrame<*>.toSimpleFeatureCollection(
```
zaleslaw, line 14: Please add a KDoc, with tiny explanation (especially if some edge cases are not working for now)

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/GeoDataFrame.kt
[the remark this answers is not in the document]
AndreiKingsley: `transform` doesn't look good to me.

zaleslaw (approved):
For me it looks much more mature today, please fix comments/linter and it could be merged, I believe

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/GeoDataFrame.kt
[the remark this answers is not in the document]
AndreiKingsley: @koperagen omitting "df" make function not so obvious that it's the **dataframe** update and not something else.

dataframe-geo/build.gradle.kts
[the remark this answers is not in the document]
AndreiKingsley: I'd leave it that way for now, then I'll try to get rid of it. When using a descriptor in notebook, the user won't need to do this.

dataframe-geo/build.gradle.kts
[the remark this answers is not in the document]
AndreiKingsley: ok

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/geocode/Geocoder.kt
```diff
@@ -0,0 +1,121 @@
...
+import org.jetbrains.kotlinx.dataframe.geo.GeoDataFrame
+import org.jetbrains.kotlinx.dataframe.geo.toGeo
+import org.locationtech.jts.geom.Geometry
+import org.locationtech.jts.geom.GeometryFactory
+import org.locationtech.jts.io.geojson.GeoJsonReader
+
+
+object Geocoder {
+
+    private val url = "https://geo2.datalore.jetbrains.com/map_data/geocoding"
```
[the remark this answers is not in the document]
AndreiKingsley: Geocoder is not part of the main API and will be removed/reworked completely (via future Kotlin geocoding library we discussed). I kept it, exclusively to play with it. So no effort should be wasted on it.

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/geocode/Geocoder.kt
AndreiKingsley: Done

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/GeoDataFrame.kt
AndreiKingsley: ?

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/GeoDataFrame.kt
AndreiKingsley: image

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/geocode/Geocoder.kt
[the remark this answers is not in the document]
Jolanrensen: Probably fine then :) You could add a small note in KDocs too telling this is experimental/temporary, so potential users won't get used to it.

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/WithGeometry.kt
AndreiKingsley: @koperagen could plugin help here, to generate all these extensions?

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/GeoDataFrame.kt
Jolanrensen: I gave you a compliment haha

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/GeoDataFrame.kt
Jolanrensen: Well, they are redundant if you write them like that yeah, haha. I think you've gotten a bit used to `data class`'es and what they do for us.

Given `public data class Person(val name: String, val age: Int)`, Kotlin generates:

```kt
public class Person(public val name: String, public val age: Int) {
    // irrelevant for equality
    public operator fun component1(): String = name
    public operator fun component2(): Int = age
    public fun copy(name: String = this.name, age: Int = this.age): Person = Person(name, age)

    // for rendering the class neatly
    public override fun toString(): String = "Person(name=$name, age=$age)"

    // for checking equality based on the properties
    override fun equals(other: Any?): Boolean {
        if (this === other) return true
        if (javaClass != other?.javaClass) return false

        other as Person

        if (age != other.age) return false
        if (name != other.name) return false

        return true
    }

    // if two objects have equals() == true, then their hashCodes must be equal too
    override fun hashCode(): Int {
        var result = age
        result = 31 * result + name.hashCode()
        return result
    }
}
```
If you don't use a data class, and you want equality of an object to be defined by its properties, you'll have to write all of this manually. Sets, maps etc. often only check `hashCode()`, instead of the slower `equals()` to verify whether an element is already present.

`toString()`, `hashCode()`, and `equals()` can be generated by the IDE with Alt+Insert.

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/WithGeometry.kt
Jolanrensen: do you mean the gradle plugin or the compiler plugin?

The compiler plugin is highly experimental still and it doesn't work yet in notebooks. So I wouldn't rely on that for now.

The gradle plugin, however, could generate these automatically based on the `@DataSchema` interfaces you defined. You'd have to add the dataframe gradle plugin to the geo module, just like we did in core, and make sure the generated files are part of sources to they are published along with the rest of the code.

koperagen: Sources generated by the plugin are automatically included. Makes sense to double check, but it's expected that you don't need to do anything extra. Just add this and it will generate required properties for annotated interfaces
plugins {
  id("org.jetbrains.kotlinx.dataframe")
 }

AndreiKingsley: Oops, forgot to add it, thank you!

Jolanrensen (approved):
Great job :D Only thing missing is a README.md in the module folder like:
```md
## :dataframe-geo

This module, published as `dataframe-geo`, contains all logic and tests for DataFrame to be able to work with geographical data... etc.
```
So people looking through the codebase know what the folder is for. You also don't have the `binary.compatibility.validator` gradle plugin yet, but I think that can wait for now, as it's still experimental. Otherwise I approve

PR #913, opened 2024-10-09
[Compiler plugin] Generate ColumnName annotations on frontend for all names that contain illegal characters

koperagen:
Now whenever column has name with illegal characters, they get replaced as in jupyter and ColumnName annotation is generated for the

The way ColumnName is used:
1. When extracting schemas. PluginDataFrameSchema always contains actual names.
2. When generating getter of extension property.

### DIFF
--- a/plugins/kotlin-dataframe/testData/box/playground.kt
+++ b/plugins/kotlin-dataframe/testData/box/playground.kt
@@ -1,0 +1,9 @@
+import org.jetbrains.kotlinx.dataframe.*
+import org.jetbrains.kotlinx.dataframe.annotations.*
+import org.jetbrains.kotlinx.dataframe.api.*
+import org.jetbrains.kotlinx.dataframe.io.*
+
+fun box(): String {
+
+    return "OK"
+}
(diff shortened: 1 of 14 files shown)


plugins/kotlin-dataframe/testData/box/playground.kt
```diff
@@ -0,0 +1,9 @@
+import org.jetbrains.kotlinx.dataframe.*
+import org.jetbrains.kotlinx.dataframe.annotations.*
+import org.jetbrains.kotlinx.dataframe.api.*
+import org.jetbrains.kotlinx.dataframe.io.*
+
+fun box(): String {
```
Jolanrensen, line 6: this is just so you don't have to create this file every time you need to quickly test something? ;P

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/SchemaProperty.kt
```diff
@@ -0,0 +1,62 @@
...
+
+data class SchemaProperty(
+    val marker: ConeTypeProjection,
+    val propertyName: PropertyName,
+    val dataRowReturnType: ConeKotlinType,
+    val columnContainerReturnType: ConeKotlinType,
+    val override: Boolean = false
+)
+
+data class PropertyName(val identifier: Name, val columnNameAnnotation: FirAnnotation?) {
```
Jolanrensen, line 23: might deserve its own file? Makes it easier to find, especially since it's used very often

Jolanrensen (approved):

plugins/kotlin-dataframe/testData/box/playground.kt
koperagen: Yes indeed :) You need to create a file and run generateTests, it will modify BlackBoxTestsGenerated and you'd also need to rollback it if it was only a temporary test. Very inconvenient :c

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/SchemaProperty.kt
koperagen: Moved it under extensions.impl package

PR #914, opened 2024-10-09
[Compiler plugin] Add a mechanism to handle function calls to stdlib that can appear as df api arguments

koperagen:
This change is needed because argument for readJsonStr is almost always a function call to either trimMargin or trimIndent with a constant String argument. Bonus: easy to add `Pair("a", listOf())`


plugins/kotlin-dataframe/testData/box/dataFrameOf_to.kt
```diff
@@ -5,7 +5,7 @@ import org.jetbrains.kotlinx.dataframe.io.*
 
 fun box(): String {
     val df = dataFrameOf(
-        "a" to listOf(1, 2),
+        Pair("a", listOf(1, 2))
```
Jolanrensen, line 8: very nice :)

Jolanrensen (approved):

PR #917, opened 2024-10-10
Implemented custom SQL DB registration

zaleslaw:
This pull request introduces several enhancements to the `dataframe-jdbc` module, primarily focusing on adding support for custom database types (`DbType`) in various methods. Additionally, it includes a minor change to make the `H2` class extensible.

Also, this [PR](https://github.com/zaleslaw/KotlinDataFrame-SQL-Examples/pull/2) in the client application is based on 0.15-dev built on this branch 

### Enhancements to database type support:

* `dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt`:
  - Added optional `dbType` parameter to multiple methods, allowing custom database type objects to be passed. This parameter defaults to `null` and, if not provided, the database type is determined from the connection. 

### Minor changes:

* `dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/H2.kt`:
  - Changed `H2` class to be `open`, allowing it to be extended.

For finalization, it requires 
- [ ] the update of [readFromSQL.md](https://github.com/Kotlin/dataframe/blob/master/docs/StardustDocs/topics/readSqlDatabases.md)
- [x] Linter check


dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -142,12 +145,13 @@ public fun DataFrame.Companion.readSqlTable(
     tableName: String,
     limit: Int = DEFAULT_LIMIT,
     inferNullability: Boolean = true,
+    dbType: DbType? = null,
```
Jolanrensen, line 148: While having a dbType argument here is great for building and debugging support for new DbTypes, I do feel like it's not "complete custom support" of new datatypes.

To be fully extensible, like with `DataFrame.read()`, I think we should support having service-registered DbTypes. Then a user should be able to do:
```kts
implementation("org.jetbrains.kotlinx.dataframe...")
implementation("com.custom.dbtype-for-df")
```
and `DataFrame.readSqlTable(MyCustomDbConnection)` should work, recognizing the custom DBType that's on the classpath.

Still, it's just an idea and having this dbType argument here is still a must for when the automatic dbtype detection fails.

Jolanrensen (changes requested):

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/h2/h2Test.kt
```diff
@@ -967,4 +972,127 @@ class JdbcTest {
         }
         exception.message shouldBe "H2 database could not be specified with H2 dialect!"
     }
+
+    // helper object created for API testing purposes
+    object CustomDB: H2(MySql)
```
koperagen, line 977: If it's only for testing, i'd recommend to keep H2 final and use `object CustomDB : DbType by H2(MySql)`

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/h2/h2Test.kt
```diff
@@ -967,4 +972,127 @@ class JdbcTest {
...
+
+    @Test
+    fun `read from table from custom database`() {
+        val tableName = "Customer"
```
koperagen, line 981: Most of assertions test correct reading of data, but actually functionality under test is correct propagation of DbType, as i understand. So most of it is duplicated code? If so, changes that will trigger H2 tests assertions to fail, will fail assertions here too. For this reason please consider a different approach here

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/h2/h2Test.kt
zaleslaw: Delegation is supported only for interfaces.

Issue #911, opened 2024-10-08
`flatten` separator for `keepParentNameForColumns` should not be `.`


Jolanrensen:
Related to https://github.com/Kotlin/dataframe/issues/666

Since we're generating accessors for columns, having a `.` in a column name might cause unexpected behavior. (Such as column accessors with ``` `` ``` around their name). `_` might be a better default.

zaleslaw:
Should we forbid a wide range of separators, or propose something from "white list" only? Let's answer ourselves

Jolanrensen:
We should follow the kotlin spec:

- `(Letter | '_') {Letter | '_' | UnicodeDigit}` is allowed without `` ` ``
  - Letter: any unicode character of categories Lu, Ll, Lt, Lm or Lo
  - UnicodeDigit: any unicode character of category Nd
- ``'`' QuotedSymbol {QuotedSymbol} '`'``
  - any character excluding CR, LF and `` '`' `` (well except the last part, we cannot write `` ` `` inside a name with backticks
- `., ;, [, ], /, <, >, :, \\` are never allowed

Source: https://kotlinlang.org/spec/syntax-and-grammar.html#identifiers

That said, these never-allowed characters might appear from sources like csv/json, so I'm wondering whether we should replace them immediately or, like now, fix it only when generating accessors?

koperagen:
I vote to keep name String as is and only replace characters in generated properties names. So, to not change this logic

Jolanrensen:
> I vote to keep name String as is and only replace characters in generated properties names. So, to not change this logic

Alright, but can the generated property then always use `_` instead of ` `?

koperagen:
I wouldn't want to change to `_`, `group col` looks cleaner than `group_col` imo. Completion even inserts backticks automatically when you type, you don't have to start with `.

koperagen:
> > I wouldn't want to change to `_`, `group col` looks cleaner than `group_col` imo. Completion even inserts backticks automatically when you type, you don't have to start with `.
> 
> It works in notebooks in the IDE yes, but I don't know about other environments. Also I've got to disagree about cleaner looks:
> 
> * `df.parent_child`
> * `` df.`parent child` ``
> 
> I feel like backticks are a "last resort" in Kotlin. Like "okay you're not supposed to use this character, but if you really want to you can do it in backticks". If a space already exists in the name, then yes, we should just generate backticks around the name and call it a day. But I don't think we should explicitly introduce characters that need backticks to work.

[...]

PR #920, opened 2024-10-14
Change default flatten parent-child separator to "_"

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/911

While the discussion of separators in generated accessors is for later, this PR at least fixes the case for flatten.

Given a DataFrame with `df.group.child1` and `df.group.child2`. Calling `flatten` with `keepParentNameForColumns` will introduce `flatDf.group_child1` and `flatDf.group_child2` by default now.

This makes it a bit easier compared to the previous default with "." which made the flattened column name "group.child1" and the accessor `` flatDf`.`group child1` ``.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/flatten.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/flatten.kt
@@ -9,37 +9,41 @@ import org.jetbrains.kotlinx.dataframe.columns.toColumnSet
 import org.jetbrains.kotlinx.dataframe.impl.api.flattenImpl
 import kotlin.reflect.KProperty
 
+public const val FLATTEN_DEFAULT_SEPARATOR: String = "_"
+
 // region DataFrame
 
 @Refine
 @Interpretable("FlattenDefault")
-public fun <T> DataFrame<T>.flatten(keepParentNameForColumns: Boolean = false, separator: String = "."): DataFrame<T> =
-    flatten(keepParentNameForColumns, separator) { all() }
+public fun <T> DataFrame<T>.flatten(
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/flatten.kt
```diff
@@ -9,37 +9,41 @@ import org.jetbrains.kotlinx.dataframe.columns.toColumnSet
 import org.jetbrains.kotlinx.dataframe.impl.api.flattenImpl
 import kotlin.reflect.KProperty
 
+public const val FLATTEN_DEFAULT_SEPARATOR: String = "_"
```
koperagen, line 12: It makes parameter info popup less informative. I think it's better to have it inlined 
image

image

image

Jolanrensen: Come on IntelliJ XD. Don't make good coding practices harder

Jolanrensen: There https://youtrack.jetbrains.com/issue/KTIJ-31662/Show-value-of-const-default-argument-when-calling-a-function

koperagen: Good c: Let's inline it before this issue is fixed?

Jolanrensen: yes, a KDoc would also help, but it still doesn't show the value in the popup you show :/

PR #923, opened 2024-10-16
Revert insertGenericTreeImpl

koperagen:
It was an experimental attempt to reuse common logic for both plugin and dataframe, but implementation proved to be too tricky so i simply reverted it to how it used to be. Now plugin, with insert being data-agnostic, will use PluginDataFrameSchema -> DataFrame adapter - same as `move`


Jolanrensen (approved):
Nice :) It looks like a good generic solution indeed to convert to DataFrame, perform the exact operation, convert back to PluginDataFrameSchema. It's easy to follow and check :)

PR #924, opened 2024-10-16
Make it clear that dataframes are immutable

dmcg:


zaleslaw:
Thanks for suggestion @dmcg ! We will revisit our examples of code and see how better to say about immutability!

Jolanrensen (approved):
It does mention "immutable", right above this example part, haha. But it doesn't hurt to change it like this. People often only look at examples and recommending to put DataFrame operation results in a new variable does hint at them being immutable.

docs/StardustDocs/topics/overview.md
```diff
@@ -55,16 +55,16 @@ val df = DataFrame.readCSV("titanic.csv", delimiter = ';')
 
 ```kotlin
 // filter rows
-df.filter { survived && home.endsWith("NY") && age in 10..20 }
+val someSurvivors = df.filter { survived && home.endsWith("NY") && age in 10..20 }
     
 // add column
-df.add("birthYear") { 1912 - age }
+val withBirthYearColumen = df.add("birthYear") { 1912 - age }
```
Jolanrensen, line 61: typo

Issue #593, opened 2024-02-15
NPE in FrameColumnImpl.schema property


koperagen:
NPE happens in this line. 

`values.mapNotNull { it.takeIf { it.nrow > 0 }?.schema() }.intersectSchemas()`

Code that leads to it:
![image](https://github.com/Kotlin/dataframe/assets/12936457/2e3b154c-c3bd-4047-ab47-1a93cfbf38f9)

I briefly looked into csv.kt and found that only tryParseImpl method could potentially create FrameColumn and provide null there. Need to confirm that it's possible. 
Another thing that could cause problem is `read` method itself that actually tries to parse the file as JSON, CSV, TSV, Excel and others until it succeeds. So, if that file cannot be parsed as CSV, it continues and can produce strange result too

Full stack trace below
```
The problem is found in one of the loaded libraries: check library converters (fields callbacks)
java.lang.NullPointerException: Parameter specified as non-null is null: method org.jetbrains.kotlinx.dataframe.DataFrameKt.getNrow, parameter <this>
org.jetbrains.kotlinx.jupyter.exceptions.ReplLibraryException: The problem is found in one of the loaded libraries: check library converters (fields callbacks)
	at org.jetbrains.kotlinx.jupyter.exceptions.CompositeReplExceptionKt.throwLibraryException(CompositeReplException.kt:50)
	at org.jetbrains.kotlinx.jupyter.codegen.FieldsProcessorImpl.process(FieldsProcessorImpl.kt:68)
	at org.jetbrains.kotlinx.jupyter.repl.impl.CellExecutorImpl$execute$1$1.invoke(CellExecutorImpl.kt:94)
	at org.jetbrains.kotlinx.jupyter.repl.impl.CellExecutorImpl$execute$1$1.invoke(CellExecutorImpl.kt:93)
	at org.jetbrains.kotlinx.jupyter.config.LoggingKt.catchAll(logging.kt:42)
	at org.jetbrains.kotlinx.jupyter.config.LoggingKt.catchAll$default(logging.kt:41)
	at org.jetbrains.kotlinx.jupyter.repl.impl.CellExecutorImpl.execute(CellExecutorImpl.kt:93)
	at org.jetbrains.kotlinx.jupyter.repl.CellExecutor$DefaultImpls.execute$default(CellExecutor.kt:14)
	at org.jetbrains.kotlinx.jupyter.ReplForJupyterImpl$evalEx$1.invoke(repl.kt:500)
[...]

PR #925, opened 2024-10-17
Fixes nulls in framecols and improves column creation situation

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/593

This PR has turned into a collection of a couple smaller fixes:
- a debug=true check has been added to [FrameColumnImpl](https://github.com/Kotlin/dataframe/blob/nulls-in-framecols/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/FrameColumnImpl.kt#L31) to catch cases where `null` leaks in the values like before.
- [`DataColumn.Companion`](https://github.com/Kotlin/dataframe/blob/nulls-in-framecols/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataColumn.kt#L46) cleaning
  - All functions gained KDocs explaining these functions do no conversions or checks and just instantiate the columns
  - `create()` was renamed to `createUnsafe()` for the same reason. Together with `.toColumnKind()`, giving the type `AnyFrame?` will now instantiate a value column instead of a frame column.
  - `createFrameColumn()` with `startIndices` has been deprecated. This is not the "normal" way to instantiate a `FrameColumn` and thus belongs somewhere else, in this case, in `chunked.kt`
- `createColumn()` and `guessColumnType()` have been merged into `createColumnGuessingType()`
  - All cases of `createColumn` were already covered by `guessColumnType()` aside from `Iterable<DataColumn>` -> `ColumnGroup`
  - `allColsMakesColGroup` argument was added to `createColumnGuessingType()` to keep this behavior optionally. It is used by `columnOf` for instance, but doesn't make sense everywhere. `guessValueType()` was modified to handle this too.
  - We now have just one place where type guessing/unifying happens
  - I added tests to check the behavior is unchanged.
- `dataFrameOf` constructors had some fixes to align behavior with other column creation functions, added tests

TODO: check whether other uses of `DataColumn.create(Unsafe)` need to be swapped to createWithTypeInference/createColumnGuessingType


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/chunked.kt
```diff
@@ -7,12 +7,28 @@ import org.jetbrains.kotlinx.dataframe.columns.ColumnGroup
...
 import org.jetbrains.kotlinx.dataframe.impl.getListType
+import org.jetbrains.kotlinx.dataframe.impl.splitByIndices
 import org.jetbrains.kotlinx.dataframe.nrow
 import org.jetbrains.kotlinx.dataframe.type
 
+/**
+ * Creates a [FrameColumn] from [this] by splitting the dataframe into
+ * smaller ones, based on the given [startIndices].
+ */
+public fun <T> DataFrame<T>.chunked(startIndices: Iterable<Int>, name: String = "groups"): FrameColumn<T> =
```
koperagen, line 18: I always saw DataColumn.createFrameColumn(Iterable<Int>) was mostly an internal API, i wouldn't expose it like this because it's hard to imagine a use case for users

Jolanrensen: I can make it `chunkedImpl` maybe :)

koperagen (changes requested):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
```diff
@@ -341,30 +360,39 @@ public class DataFrameBuilder(private val header: List<String>) {
...
         withColumns { name ->
-            DataColumn.create(
+            DataColumn.createWithTypeInference(
                 name = name,
                 values = List(nrow, init),
             )
         }
 
-    private inline fun <reified C> fillNotNull(nrow: Int, crossinline init: (Int) -> C) =
+    private inline fun <reified C> fillNotNull(nrow: Int, crossinline init: (Int) -> C & Any) =
```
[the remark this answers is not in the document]
Jolanrensen: oh that's even simpler

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/parse.kt
```diff
@@ -525,7 +526,10 @@ internal fun DataColumn<String?>.tryParseImpl(options: ParserOptions?): DataColu
     if (type.jvmErasure == String::class && !nullStringParsed) {
         return this // nothing parsed
     }
-    return DataColumn.create(name(), parsedValues, type)
+
+    // Create a new column with the parsed values,
+    // createColumnGuessingType is used to handle unifying values if needed
+    return createColumnGuessingType(name(), parsedValues, type)
```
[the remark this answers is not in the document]
Jolanrensen: We could, but then we'd need to expand it further with `type` arguments etc. Also potentially adding somewhat confusing arguments like `suggestedTypeIsUpperbound` etc., things that seemed more internal to me. WDYT?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/constructors.kt
```diff
@@ -217,32 +176,84 @@ internal fun Array<out String>.toNumberColumns() = toColumnsSetOf<Number>()
...
-internal fun <T> guessColumnType(name: String, values: List<T>) = guessColumnType(name, values, null)
-
+/**
+ * Creates a new column by doing type inference on the given values and
+ * some conversions to unify the values if necessary.
+ *
+ * @param values values to create a column from
+ * @param suggestedType optional suggested type for values.
+ *   If set to `null` (default) the type will be inferred.
+ * @param guessTypeWithSuggestedAsUpperbound Only relevant when [suggestedType]` != null`.
```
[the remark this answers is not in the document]
Jolanrensen: good idea! that's way better indeed

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/readJson.kt
```diff
@@ -298,11 +299,12 @@ internal fun fromJsonListAnyColumns(
...
                 }
 
-                else -> DataColumn.createFrameColumn(
-                    name = ARRAY_COLUMN_NAME, // will be erased
-                    df = parsed.unwrapUnnamedColumns(),
-                    startIndices = startIndices,
-                )
+                else ->
+                    parsed.unwrapUnnamedColumns()
+                        .chunkedImpl(
```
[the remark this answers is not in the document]
Jolanrensen: Hmm, for this one, I don't agree. I don't see splitting a DataFrame by indices of rows as "a default way to instantiate a FrameColumn", but more as "an operation on DataFrame which produces a FrameColumn"

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/parse.kt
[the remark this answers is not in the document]
Jolanrensen: might be okay though, if we simplify suggestedTypeIsUpperbound with a sealed interface as suggested in the other file :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataColumn.kt
```diff
@@ -56,11 +62,20 @@ public interface DataColumn<out T> : BaseColumn<T> {
...
+                values = values,
+                name = name,
+                type = getValuesType(values, type, infer),
+                defaultValue = defaultValue,
+            )
 
         /**
          * Creates [ValueColumn] using given [name], [values] and reified column [type].
          *
+         * Be careful; values are NOT checked to adhere to [type] for efficiency,
```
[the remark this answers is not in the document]
Jolanrensen: yes, unless you cast, like we did, and it caused issues :/ That's why the warning is in place, still I'll clean it up a bit

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataColumn.kt
```diff
@@ -74,33 +89,92 @@ public interface DataColumn<out T> : BaseColumn<T> {
...
+            replaceWith = ReplaceWith(CREATE_FRAME_COLUMN_REPLACE, CREATE_FRAME_COLUMN_IMPORT),
+            level = DeprecationLevel.WARNING,
+        )
         public fun <T> createFrameColumn(name: String, df: DataFrame<T>, startIndices: Iterable<Int>): FrameColumn<T> =
             FrameColumnImpl(name, df.splitByIndices(startIndices.asSequence()).toList(), lazy { df.schema() })
 
+        /**
+         * Creates [FrameColumn] using the given [name] and list of dataframes [groups].
+         *
+         * Be careful; [groups] must be a non-null list of [DataFrames][DataFrame].
```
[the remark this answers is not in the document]
Jolanrensen: yes, unless you cast it, which is exactly what caused our bug with nulls in frame cols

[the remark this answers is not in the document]
Jolanrensen: I'll hide the warning as an explanation so it seems less redundant :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt
```diff
@@ -390,14 +392,24 @@ internal fun <T> getValuesType(values: List<T>, type: KType, infer: Infer): KTyp
...
+ *   that a column group should be created instead of a [DataColumn][DataColumn]`<`[AnyCol][AnyCol]`>`,
+ *   so the function will return [DataRow].
  */
 @PublishedApi
-internal fun guessValueType(values: Sequence<Any?>, upperBound: KType? = null, listifyValues: Boolean = false): KType {
+internal fun guessValueType(
+    values: Sequence<Any?>,
+    upperBound: KType? = null,
+    listifyValues: Boolean = false,
+    allColsMakesRow: Boolean = false,
```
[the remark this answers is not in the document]
Jolanrensen: In all other cases, allColsMakesRow can be false. I put it here so that we don't check all values for types multiple times in different places, like we did before, we just have this one place which makes the behavior easy to reason about and debug. `listifyValues` is a similar case, it was only used in Json I believe? (though I'm not sure it still is). `guessValueType()` is already a very difficult function, however, all the logic of "values -> columnType" for the entire library lives here so let's not make it more difficult by spreading logic across multiple files and then end up with another `createColumn` function which works exactly like `createColumnGuessingType` except when it doesn't

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/types/UtilTests.kt
```diff
@@ -151,6 +154,19 @@ class UtilTests {
...
+        guessValueType(
+            sequenceOf(DataColumn.empty(), columnOf(1)),
+            allColsMakesRow = true,
+        ) shouldBe typeOf<DataRow<*>>()
+
+        guessValueType(
+            sequenceOf(columnOf("a"), columnOf(1)),
+            allColsMakesRow = true,
+        ) shouldBe typeOf<DataRow<*>>()
+        guessValueType(
```
zaleslaw, line 167: split two guessValueType?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ParserTests.kt
```diff
@@ -215,4 +220,17 @@ class ParserTests {
             Locale.setDefault(currentLocale)
         }
     }
+
+    @Test
+    fun `Issue #593, mixing null and json`() {
```
zaleslaw, line 225: better add link to the Github issue

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/columns/DataColumns.kt
```diff
@@ -28,4 +32,20 @@ class DataColumns {
         val col = listOf(URI.create("http://example.com"), null).toColumn("a")
         col.type().toString() shouldBe "java.net.URI?"
     }
+
+    @Test
+    fun `allow no nulls in frame columns`() {
+        // enable kotlin.dataframe.debug=true for this
```
zaleslaw, line 38: emmm, strange that we have different behaviour here..it was a goal just to discover weird things

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt
```diff
@@ -383,21 +386,35 @@ internal fun <T> getValuesType(values: List<T>, type: KType, infer: Infer): KTyp
     }
 
 /**
- * Returns the value type of the given [values] sequence.
+ * Returns the guessed value type of the given [values] sequence.
+ *
+ * This function analyzes all [values] once and returns the expected column type.
```
zaleslaw, line 391: Did we cover in tests the situation when one column in file or in List has different types?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataColumn.kt
```diff
@@ -74,47 +89,134 @@ public interface DataColumn<out T> : BaseColumn<T> {
...
 
-        public inline fun <reified T> create(name: String, values: List<T>, infer: Infer = Infer.None): DataColumn<T> =
-            create(name, values, typeOf<T>(), infer)
+        /**
+         * Calls [createColumnGroup], [createFrameColumn], or [createValueColumn] based on
+         * type [T].
+         *
+         * This is generally safe, as [T] can be inferred, and more efficient than [createWithTypeInference].
+         *
+         * Be careful when casting occurs; Values in [values] are NOT checked to adhere to the given/inferred type [T],
```
zaleslaw, line 203: It is unsafe because of type? For me createUnsafe creates weird reference to something like `sun.misc.Unsafe`. Could we rename it somehow to the `createWithoutTypeChecking` or` createAndFailFast`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/FrameColumnImpl.kt
```diff
@@ -26,6 +28,14 @@ internal open class FrameColumnImpl<T> constructor(
     ),
     FrameColumn<T> {
 
+    init {
+        // Checks for nulls in the `values` list.
+        // This only runs with `kotlin.dataframe.debug=true` in gradle.properties.
+        if (BuildConfig.DEBUG) {
+            require(!values.anyNull()) { "FrameColumn cannot null values." }
```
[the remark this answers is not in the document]
Jolanrensen: Then each time we create a frame column, we'll have to check all values for nulls, which at worst is `O(n)`. For efficiency it may be best to at least have a zero-checks path available for library calls.

zaleslaw (changes requested):
I've suggested just rename or answer some my questions before merging

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataColumn.kt
Jolanrensen: >createAndFailFast

haha, another title could be `createRaw`, but `createWithoutTypeChecking` might also be clear. It's unsafe in the sense that it solely uses the given type `T` to decide which type of column to instantiate. It doesn't check any of the values, which makes it fast, but, well, unsafe.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt
Jolanrensen: We do have a lot of tests of this specific function: https://github.com/Kotlin/dataframe/blob/nulls-in-framecols/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/types/UtilTests.kt.
These cases are probably covered by other column creation tests, like in https://github.com/Kotlin/dataframe/blob/nulls-in-framecols/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt

Jolanrensen:
@zaleslaw Thanks! I'll have a look. I'll also check the binary compatibility now that plugin is merged. I suspect I broke some stuff.

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/columns/DataColumns.kt
Jolanrensen: True, all add `if (BuildConfig.DEBUG)` to this specific check. We don't want to enforce users to enable debug mode for local tests.

Jolanrensen: It's basically testing that if debug mode is enabled, it will catch this case

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/types/UtilTests.kt
Jolanrensen: then it would need to be split in 4 because we also have `listifyValues: Boolean`. I think in this case it's good to have all logic of `values -> column type` of the entire library handled by one function `guessValueType()`. This makes it easier to debug and reason about.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataColumn.kt
Jolanrensen: Hmm, on the other hand, then we can have `DataColumn.createWithoutTypeInference(..., infer = Infer.Type)`. That's gonna be confusing haha. We need a same that says it will choose createValueColumn, createColumnGroup, or createFrameColumn based on the type T

Jolanrensen: Alright, I think I'm gonna go with:
- createColumnByInference
- createColumnByType

Jolanrensen:
Alright, thanks for the reviews :) they were most helpful! I think I covered all now, so I'll merge so we won't get further out of sync with other branches. Let me know if I missed something major, then I'll fix it :)

Issue #926, opened 2024-10-17
Support Parquet UTC timestamp vectors (`TimeStampXXXTZVector`) in Arrow reader


laurentperez:
### Problem

Reading Parquet files with timestamp columns marked as `isAdjustedToUTC=true` currently fails with:

`NotImplementedError("reading from TimeStampXXXTZVector is not implemented")`

This affects common Parquet timestamp variants with millisecond, microsecond, and nanosecond precision.

### Expected

Implement reading support for `TimeStampXXXTZVector` timestamp columns.

### Acceptance criteria

- Parquet timestamp columns with `isAdjustedToUTC=true` are read successfully
- Supported precisions include:
  - milliseconds
  - microseconds
  - nanoseconds
- Values are converted to the expected Kotlin datetime type
- Tests cover UTC timestamp columns generated by common tools such as PyArrow / Polars
- Existing non-timezone timestamp behavior is not broken

### Motivation

Parquet is a common data exchange format.

UTC-adjusted timestamp columns are standard in real-world Parquet files. Failing with `NotImplementedError` makes Parquet support incomplete and can block users from reading valid datasets.

This should be fixed before 1.0 to make Parquet IO reliable.

**The user proposed a solution for a problem which blocked him**

--------------------------------------------------- initial comment from user --------------------------------------------------

Hi. I will PR this.

The following python code will generate a parquet with timestamp columns in `us ns ms` adjusted to UTC (1)

Reading it using https://github.com/Kotlin/dataframe/pull/577 , `org.jetbrains.kotlinx.dataframe.io.ArrowReadingImplKt#readField` will throw `NotImplementedError("reading from TimeStampXXXTZVector is not implemented")`

This PR implements `TimeStampXXXTZVector` following https://github.com/apache/parquet-format/blob/master/LogicalTypes.md

Please note parquet file format has no _seconds_ precision but only `MILLIS or MICROS, NANOS` so my implementation of `TimeStampSecTZVector` finds seconds from milliseconds.

Applying this code under my local checkout of #577, precisions return correctly for `us ns ms` as : 

```
    @Test
    fun testReadTimestamp() {
        val frame = DataFrame.readParquet(
            URL("file:/home/lperez/Bureau/work/pocs/lavaret/python/timestamps_with_utc_and_local.parquet")
        )
        val columnTypes = frame.columnTypes()
        println("columnTypes: $columnTypes")
        println(frame)
    }
[...]

PR #927, opened 2024-10-17
Implement TimeStampMicroTZVector for parquet isAdjustedToUTC timestam…

laurentperez:
#926 #576

### DIFF
--- a/dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
+++ b/dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
@@ -1,11 +1,14 @@
 package org.jetbrains.kotlinx.dataframe.io
 
+import kotlinx.datetime.Instant
 import kotlinx.datetime.LocalDate
 import kotlinx.datetime.LocalDateTime
 import kotlinx.datetime.LocalTime
+import kotlinx.datetime.TimeZone
 import kotlinx.datetime.toKotlinLocalDate
 import kotlinx.datetime.toKotlinLocalDateTime
 import kotlinx.datetime.toKotlinLocalTime
+import kotlinx.datetime.toLocalDateTime
 import org.apache.arrow.memory.RootAllocator
 import org.apache.arrow.vector.BigIntVector
 import org.apache.arrow.vector.BitVector
@@ -21,6 +24,7 @@ import org.apache.arrow.vector.LargeVarBinaryVector
 import org.apache.arrow.vector.LargeVarCharVector
 import org.apache.arrow.vector.NullVector
 import org.apache.arrow.vector.SmallIntVector
+import org.apache.arrow.vector.TimeStampMicroTZVector
 import org.apache.arrow.vector.TimeMicroVector
 import org.apache.arrow.vector.TimeMilliVector
 import org.apache.arrow.vector.TimeNanoVector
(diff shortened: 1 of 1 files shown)


laurentperez:
@fb64 might interest you ;)

fb64:
> @fb64 might interest you ;)

Yes, it's part of missing mapping for some Arrow Types.
It would be nice to extend it to the others missing TimeStamp*TZVector Types : `TimeStampSecTZVector`, `TimeStampNanoVector`, `TimeStampMilliTZVector`

Also you can add those types to this [test](https://github.com/Kotlin/dataframe/blob/54d00e69ae7c06ff254c6591242413eaf7894f01/dataframe-arrow/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ArrowKtTest.kt#L526)

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
```diff
@@ -345,6 +359,8 @@ private fun readField(root: VectorSchemaRoot, field: Field, nullability: Nullabi
 
             is TimeStampMicroVector -> vector.values(range).withTypeNullable(field.isNullable, nullability)
 
+            is TimeStampMicroTZVector -> vector.values(range).withTypeNullable(field.isNullable, nullability)
```
fb64, line 362: It would be nice to also add the missing Arrow TimeStamp timezoned Types : 
- `TimeStampSecTZVector` 
- `TimeStampNanoVector`
- `TimeStampMilliTZVector`

laurentperez: I've added these and updated https://github.com/Kotlin/dataframe/pull/926 with the python code used to create the parquet file plus a note regarding TimeStampSecTZVector because parquet format does not have seconds precision.

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
```diff
@@ -197,6 +235,16 @@ private fun TimeStampSecVector.values(range: IntRange): List<LocalDateTime?> =
         }
     }
 
+private fun TimeStampSecTZVector.values(range: IntRange): List<LocalDateTime?> =
+    range.mapIndexed { i, it ->
+        if (isNull(i)) {
+            null
+        } else {
+            DateUtility.getLocalDateTimeFromEpochMilli(TimeUnit.SECONDS.toMillis(getObject(it)), this.timeZone)
```
laurentperez, line 243: at second glance I'm not so sure about this one.  
Parquet does not have seconds precision and my PR is for Parquet, but pyarrow feather can `.floor('S')` its datetimes.

I'm not sure what's going to be present in its .feather file, either seconds or milliseconds.

I need to test this against a `.feather` file together with seconds precision and timezone awareness, perhaps one from https://github.com/Kotlin/dataframe/tree/master/dataframe-arrow/src/test/resources

fb64: Your pull request focuses on Parquet, but it would be beneficial to incorporate upgrades for all Arrow TimeStamp types as well, because Parquet reading (in my PR) relies totally on Arrow.
The test [testTimeStamp](https://github.com/Kotlin/dataframe/blob/54d00e69ae7c06ff254c6591242413eaf7894f01/dataframe-arrow/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ArrowKtTest.kt#L526) covers both cases for IPC format and Feather format (by using `ArrowStreamWriter` and `ArrowFileWriter`), so you only should  add `TimeStampNanoTZVector`,`TimeStampMicroVector` and `TimeStampSecTZVector` to the [writeArrowTimestamp](https://github.com/Kotlin/dataframe/blob/54d00e69ae7c06ff254c6591242413eaf7894f01/dataframe-arrow/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ArrowKtTest.kt#L544) method and to the expected DataFrame in `testTimeStamp` as you did for `TimeStampMicroTZVector`.
Finally as your PR improves Arrow types compatibility it could be merge independently of https://github.com/Kotlin/dataframe/pull/577 (IMHO) 😃

Issue #928, opened 2024-10-21
Local classes inferred as `Any`


Jolanrensen:
As reported on Slack https://kotlinlang.slack.com/archives/C4W52CFEZ/p1729318547736259

```kt
fun main()
    data class Car(val type: String, val model: String)

    val cars: DataFrame<*> = dataFrameOf("owner", "car")(
        "Max", Car("audi", "a8"),
        "Tom", Car("toyota", "corolla")
    )
    cars.schema().print()
}
```
the result is:
```
owner: String
car: Any
```
while putting the class outside of the `main` function works fine. This breaks functions like `unfold` which rely on the `kType` to unpack the properties.

PR #929, opened 2024-10-21
Fixed local classes being inferred as `Any` by changing visibility check

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/928

We can assume "no visibility" means "public". This fixes local classes not being inferred.


koperagen (commented):
Very nice! Similar case:
`listOf(Car("audi", "a8")).toDataFrame()`
Can you check it too?

Jolanrensen:
> Very nice! Similar case: `listOf(Car("audi", "a8")).toDataFrame()` Can you check it too?

yep, it works too :)

koperagen (approved):

PR #931, opened 2024-10-22
Open File in readExcel in read-only mode

koperagen:
WorkBook is used for reading, so it generally makes sense. It should fix the following error: 
```
OpenXML4JRuntimeException: Fail to save an error occurs while saving the package: The part/docProps/app.ant failed to be saved in the stream marshaller org.apache.pos.openxml4j
```


Jolanrensen (approved):

Jolanrensen:
As reported on Slack: https://kotlinlang.slack.com/archives/C4W52CFEZ/p1729161693795839

PR #934, opened 2024-10-23
Specify slf4j-api instead of slf4j-simple.

erikogenvik:
When used as a library in other components slf4j-api should be used, allowing the component to select the logging backend themselves (for example logback).

This fixes an issue where slf4j would complain that there were two competing logging backends in any application using this along with somehing like logback.

### DIFF
--- a/dataframe-jdbc/build.gradle.kts
+++ b/dataframe-jdbc/build.gradle.kts
@@ -27,7 +27,7 @@ dependencies {
     testImplementation(libs.h2db)
     testImplementation(libs.mssql)
     testImplementation(libs.junit)
-    testImplementation(libs.sl4j)
+    testImplementation(libs.sl4jsimple)
     testImplementation(libs.jts)
     testImplementation(libs.kotestAssertions) {
         exclude("org.jetbrains.kotlin", "kotlin-stdlib-jdk8")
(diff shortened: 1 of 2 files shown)


Jolanrensen:
Thanks!
However, I see you added the new dependency only as `testImplementation()`, meaning it will only be used when running our dataframe tests, not when it's published as a library.

erikogenvik:
Sorry, I should have explained more thoroughly. 
I changed so that the default dependency, that's part of what's exported, is now "slf4j-api". However, I noticed that you also had some tests, and I guessed that you do want logging output when running these tests.
Therefore I added a new "slf4j-simple" artifact (which is the one that contains a logging appender, and thus shouldn't be part of the distributed artifact) so that you'll get log output when running your tests.

Jolanrensen:
Aaah, now I see :)

Jolanrensen:
@zaleslaw could you also check everything works as expected? :)

zaleslaw:
Thanks for contribution, @erikogenvik

Jolanrensen:
It's released along the hot-fix version [0.14.2](https://github.com/Kotlin/dataframe/releases/tag/v0.14.2).

Issue #607, opened 2024-02-28
Incorrect parsing of CSV file. Locale is ignored for some columns


antonarhipov:
I have encountered a funny bug when parsing a CSV file. The floating point numbers in that file are specified with a comma `,` instead of dot `.`, and this should be a subject of specifying the correct locale. However, it seems that Locale is respected for one column and ignored for the other. For instance, the comma is ignored for the number `68,83`, and the result is `6883`.

![image](https://github.com/Kotlin/dataframe/assets/177047/390dc62a-cfc9-4bf2-8bb5-379c4d90f825)

CSV file: https://github.com/antonarhipov/kotlin-sandbox/blob/master/energy-consumption.csv
Notebook: https://github.com/antonarhipov/kotlin-sandbox/blob/master/demo.ipynb

koperagen:
Ok, this is interesting. Here's the culprit:
```
val nf = NumberFormat.getInstance(Locale("et", "EE"))
nf.parse("-0,08")
```
This code throws an exception. Because in Estonian locale a different minus sign is used
We need to improve UX in this part. Suggested workaround: use `ParserOptions(Locale.FRANCE)`

koperagen:
Note for future: let's check how it's handled in R dplyr and maybe use it as a reference. I think they might be good at interpretation of all kinds of data

Jolanrensen:
> Ok, this is interesting. Here's the culprit:
> 
> ```
> val nf = NumberFormat.getInstance(Locale("et", "EE"))
> nf.parse("-0,08")
> ```
> 
> This code throws an exception. Because in Estonian locale a different minus sign is used. So DataFrame tries to parse this column using locale specific rules and fails. Next step of parsing is locale agnostic and it creates wrong values. We need to improve UX in this part. The fact that locale parameter defines not only `,` and `.` is surprising. Maybe it shouldn't.. Suggested workaround: use `ParserOptions(Locale.FRANCE)`

Actually it makes sense that the locale parameter also looks at minus signs etc. Just have a look at Java's [DecimalFormatSymbols](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/text/DecimalFormatSymbols.html), it includes the zero digit, grouping separator, decimal separator, per mille sign, percent char, NaN, minus sign, monetary decimal separator, exponent separator, etc... The decimal character is just the tip of the iceberg really.
So, if you read a csv with an estonian locale it will do so :).
[...]

PR #935, opened 2024-10-28
Fast double parser

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/607

Created under csv umbrella issue: https://github.com/Kotlin/dataframe/issues/827

Introduces [FastDoubleParser](https://github.com/wrandelshofer/FastDoubleParser) as optional alternative double parser. It can be enabled as experimental option (for now) in `ParserOptions`.

The [configurable version](https://github.com/wrandelshofer/FastDoubleParser/issues/82) of this double parser was created for us :) and the way I implemented it, it can solve some of the [locale-fallback-issues](https://github.com/Kotlin/dataframe/issues/607) we were facing before. 
Check the tests I created in this PR for more some examples.

It's also [faster](https://github.com/wrandelshofer/FastDoubleParser/issues/82#issuecomment-2425048098) than `NumberFormat`!

FastDoubleParser is also used by [Deephaven CSV](https://github.com/deephaven/deephaven-csv/tree/main), which helps with our [new csv implementation](https://github.com/Kotlin/dataframe/pull/903).


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/utils.kt
```diff
@@ -19,24 +19,31 @@ import kotlin.annotation.AnnotationTarget.VALUE_PARAMETER
  * {@include [Indent]}
  *
  */
+@ExcludeFromSources
```
Jolanrensen, line 22: These documentation interfaces can be [excluded](https://github.com/Kotlin/dataframe/blob/master/KDOC_PREPROCESSING.md#excludefromsources-annotation-excluding-code-content-from-sources) from the sources as they are only ever accessed from kdoc.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/FastDoubleParser.kt
```diff
@@ -0,0 +1,233 @@
...
+import java.nio.charset.Charset
+import java.text.DecimalFormatSymbols
+import java.text.NumberFormat
+import java.text.ParsePosition
+import java.util.Locale
+
+private val logger = KotlinLogging.logger {}
+
+// (lowercase) strings that are recognized to represent infinity and NaN in doubles in all locales
+private val INFINITIES = arrayOf("∞", "inf", "infinity", "infty")
```
zaleslaw, line 16: How did you collect this information? Are some standards somewhere available?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/FastDoubleParser.kt
```diff
@@ -0,0 +1,233 @@
...
+private val PLUS_INFINITIES = INFINITIES.map { "+$it" }
+private val MINUS_INFINITIES = INFINITIES.map { "-$it" }
+private val NANS = arrayOf("nan", "na", "n/a")
+
+/**
+ * Parses a [String]/[CharSequence], [CharArray], or [ByteArray] into a [Double].
+ *
+ * If [ParserOptions.useFastDoubleParser] is enabled, it will try to parse the input with an _EXPERIMENTAL_
+ * fast double parser, [FastDoubleParser](https://github.com/wrandelshofer/FastDoubleParser).
+ * If not, or if it fails, it will use [NumberFormat] to parse the input.
```
zaleslaw, line 26: Great approach!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/FastDoubleParser.kt
```diff
@@ -0,0 +1,233 @@
...
+                        add(lowercase)
+                    }
+                }
+            }
+
+        return NumberFormatSymbols.fromDecimalFormatSymbols(localDecimalFormatSymbols)
+            .withPlusSign(setOf('+'))
+            .withDecimalSeparator(DecimalFormatSymbols::getDecimalSeparator.fromLocalWithFallBack())
+            .withGroupingSeparator(DecimalFormatSymbols::getGroupingSeparator.fromLocalWithFallBack())
+            .withExponentSeparator(DecimalFormatSymbols::getExponentSeparator.fromLocalWithFallBack())
```
zaleslaw, line 134: What does it mean `set(get.fromLocalWithFallBack)` construction?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/FastDoubleParser.kt
```diff
@@ -0,0 +1,233 @@
...
+    private fun String.parseToDoubleOrNullFallback(): Double? =
+        when (lowercase()) {
+            in INFINITIES, in PLUS_INFINITIES -> Double.POSITIVE_INFINITY
+
+            in MINUS_INFINITIES -> Double.NEGATIVE_INFINITY
+
+            in NANS -> Double.NaN
+
+            else -> {
+                // not thread safe; must be created here
```
zaleslaw, line 150: why is thread safety important here? Or this is a reason to create an Instance of Number Format?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/FastDoubleParser.kt
```diff
@@ -0,0 +1,233 @@
...
+                val result = numberFormat.parse(this, parsePosition)?.toDouble()
+                if (parsePosition.index != this.length || parsePosition.errorIndex != -1) {
+                    null
+                } else {
+                    result
+                }
+            }
+        }.also {
+            if (it == null) {
+                logger.debug { "Could not parse '$this' as Double with NumberFormat with locale '$locale'." }
```
zaleslaw, line 162: Probably we could also add some logs about **experimental fast double parsing** like a marker for our logs

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/FastDoubleParserTests.kt
```diff
@@ -0,0 +1,163 @@
...
+    fun `can fast parse doubles`() {
+        val parser = FastDoubleParser(ParserOptions(locale = Locale.ROOT, useFastDoubleParser = true))
+
+        val numbers = listOf(
+            "+12.45",
+            "-13.35",
+            "100123.35",
+            "-204,235.23",
+            "1.234e3",
+            "3e-04", // failed with old double parser
```
zaleslaw, line 40: Will it fail, if we disable fast parsing? Or they aligned now in the terms of supported functionality, locales?
And one just is faster than another?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/FastDoubleParserTests.kt
```diff
@@ -0,0 +1,163 @@
...
+        // CharArray
+        numbers.map { parser.parseOrNull(it.toCharArray()) }.shouldContainInOrder(expectedDoubles)
+
+        // ByteArray
+        numbers.map { parser.parseOrNull(it.toByteArray()) }.shouldContainInOrder(expectedDoubles)
+    }
+
+    @Test
+    fun `can fast parse german locale`() {
+        val parser = FastDoubleParser(ParserOptions(locale = Locale.GERMANY, useFastDoubleParser = true))
```
zaleslaw, line 72: Vielen Dank im Voraus!

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/FastDoubleParserTests.kt
```diff
@@ -0,0 +1,163 @@
...
+        // CharArray
+        numbers.map { parser.parseOrNull(it.toCharArray()) }.shouldContainInOrder(expectedDoubles)
+
+        // ByteArray
+        numbers.map { parser.parseOrNull(it.toByteArray()) }.shouldContainInOrder(expectedDoubles)
+    }
+
+    @Test
+    fun `can fast parse french locale`() {
+        val parser = FastDoubleParser(ParserOptions(locale = Locale.FRANCE, useFastDoubleParser = true))
```
zaleslaw, line 102: I started to think: could we write a test, which just generates checks for all possible locales and constants are obtained from Double values -> String -> parseDouble again?

zaleslaw (changes requested):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/FastDoubleParser.kt
Jolanrensen: Deephaven holds a single instance of this FastDoubleParser class but can parse columns in parallel. Previously I put `numberFormat` outside the function, but this caused bugs, so I put it inside here and it works fine :). I put the warning here because you might assume you could save an object creation by putting `val numberFormat = NumberFormat.getInstance(locale)` inside the class and reusing it, but no, you cannot.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/FastDoubleParser.kt
Jolanrensen: you can read it as a sentence :) `::getDecimalSeparator.fromLocalWithFallBack()`. This means it calls `localDecimalFormatSymbols.getDecimalSeparator()` first and then adds fallback characters from `fallbackDecimalFormatSymbols.getDecimalSeparator()`.

Jolanrensen: I made it work on the function reference so that the function can be called from different `DecimalFormatSymbols` instances

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/FastDoubleParser.kt
Jolanrensen: We already had "inf" and "infinity". I added "infty" because it's used in the math world, like LaTeX, and finally, I supposed the infinity symbol itself should also be recognizable. Any others we should add?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/FastDoubleParserTests.kt
Jolanrensen: nope, it will fail in the old one. We cannot make the old one behave the same, as it's not supported by NumberFormat.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/FastDoubleParser.kt
Jolanrensen: KLogger already provides a marker. It recognizes the file/class it's called in. So this will look something like: `DEBUG org.jetbrains.kotlinx.dataframe.impl.io.FastDoubleParser - Could not parse 'hello' as Double with NumberFormat with locale 'en_US'.`

Jolanrensen: we can just search for "FastDoubleParser" :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/FastDoubleParserTests.kt
Jolanrensen: "all" locales? XD I can try, but that's like hundreds of locales and may be system-dependent.

Jolanrensen: Also, there are some locales like Arabic where NumberFormat itself fails to add "-" in front of negative numbers :/ It's fixed in later Java versions, but not 8 yet.

Jolanrensen: Okay this was actually a really good idea! It made me catch two separate bugs which I reported to [FastDoubleParser](https://github.com/wrandelshofer/FastDoubleParser/issues/82).
- Minus signs in RTL languages, like Arabic, are (in Java 8) marked with the [Arabic letter mark](https://en.wikipedia.org/wiki/Arabic_letter_mark) instead of, you know, a minus sign... FastDoubleParser cannot handle this, because the character is a format symbol, not a normal one. `NumberFormat` does handle it, so to align our behavior with theirs, I now manually check for format symbol minus signs and replace them with '-' before parsing.
- In Finnish and other langauges, they use `exponentSeparator='e'` and `nan="epäluku"`, which breaks as both contain the letter 'e'. I've reported this and luckily our fallback method catches it.

Jolanrensen:
Thanks for the review :) I'll merge it for now so I can continue with the csv parser

Issue #558, opened 2024-01-12
Describe breaks on `Number` column


Jolanrensen:
Describe breaks on `Number` columns. This happens because the `Iterable<Number>.std()` function accepts `Number` but doesn't convert them to `Double` (like `mean()` does).

There are a couple more missing actually:

 - [x]  `cumSum` 
   - Misses `Byte`, `Short`
   - ~Has `DataColumn` overloads but not `Iterable`/`Sequence`~
 - [x]  `mean` 
   - ~Has `Sequence<Double | Float>` but not for other `Number` types~
 - [x]  `median`
   - Misses `Float`, `Byte`, `Short`, `Number` (it only works on `Comparable`)
   - Needs to handle other types consistently
   - ~No `Sequence` overloads~
   - Cannot `skipNA` (if applicable)
 - [x]   `min` and `max`
    - internal `Iterable<T>.min` and `max` are not used and can be removed. Stdlib functions for Comparable sequences and iterables are used instead.
    - Misses `Number` (it only works on `Comparable`)
    - `Short` and `Byte` are converted to `Int` for some reason
 - [x]  `std`
   - Breaks if type is `Number`
   - `Short` and `Byte` are cast to `Int` which works but is a bit iffy
   - ~Iterable overloads missing for `Number`, `Short`, `Byte`~
   - ~Sequence overloads missing~
   - ~Nullable overloads missing for Iterable (and sequence)~
- [ ]  `varianceAndMean`
  - also provides `std(ddof: Int)` function without docs of what ddof even means, as well as `count`. Could have a better name. Also can produce nulls?? this screams for documentation.
  - variance functions are missing on DataColumns entirely (had to be added separately for Kandy)
  - Misses `Short`, `Byte`, `Number`, and nullable overloads
  - ~Misses Sequence overloads~
- [x]  `sum`
  - Has `TODO`s where types are amiss
  - Misses `Float`(!), `Short`, `Byte`, `Number` in various `Iterable` overloads.

- [x] All are also missing `BigInteger` as we're supporting `BigDecimal` too.
- [ ] There are plenty of public overloads on `Iterable` and `Sequence`. It's fine to have them internally, but I feel like we're clogging the public scope here. `mean`, for instance, is already covered in the stdlib.
- [ ] We need to honor some conversion table (see below)
[...]

Issue #352, opened 2023-04-12
Misleading exception message in "sum" aggregator (when aggregating no columns)


koperagen:
we have an example in readme with 
`pivot { destination }.sum { recentDelays.colsOf<Int?>() } into "total delays to"`
but changing the type to `colsOf<Int>` leads to 
```
Sum is not supported for kotlin.Number
java.lang.IllegalArgumentException: Sum is not supported for kotlin.Number
```

When in fact, all direct Number inheritors are supported, it's just nullable types are not. Maybe we can support them, maybe just change the error message

koperagen:
Actually, my explanation is wrong. `Int?` is supported by sum aggregation. Something is just wrong

Jolanrensen:
I suspect this is because recentDelays is a column group with only `Int?` columns, so you try to sum no columns. Sum can only accept Numbers, so it assumes you gave it 0 Number columns to sum, then the exception occurs... So yes, sum should support Number columns, but it should also give a proper error when no columns are selected to sum

Jolanrensen:
https://github.com/Kotlin/dataframe/pull/937 will remove the exception, but not solve the issue. The `sum { recentDelays.colsOf<Int>() }` will now be `0.0`.
There's still the odd switch to `Number` for providing an empty columnset to `sum` and https://github.com/Kotlin/dataframe/pull/937 makes `Double` the default for `Number` columns.

PR #937, opened 2024-10-29
Statistics fixes

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/558

WIP

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt
@@ -9,6 +9,7 @@ import org.jetbrains.kotlinx.dataframe.math.cumSum
 import org.jetbrains.kotlinx.dataframe.math.defaultCumSumSkipNA
 import org.jetbrains.kotlinx.dataframe.typeClass
 import java.math.BigDecimal
+import java.math.BigInteger
 import kotlin.reflect.KProperty
 import kotlin.reflect.typeOf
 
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/math/stdMean.kt
```diff
@@ -114,3 +115,20 @@ public fun Iterable<BigDecimal>.varianceAndMean(): BasicStats {
     }
     return BasicStats(count, mean.toDouble(), variance.toDouble())
 }
+
+@JvmName("bigIntegerVarianceAndMean")
+public fun Iterable<BigInteger>.varianceAndMean(): BasicStats {
+    var count = 0
+    var sum = BigInteger.ZERO
+    for (element in this) {
```
zaleslaw, line 123: Old good plain **java**

core/api/core.api
```diff
@@ -2719,7 +2719,7 @@ public final class org/jetbrains/kotlinx/dataframe/api/ColumnDescription_Extensi
 	public static final fun ColumnDescription_top (Lorg/jetbrains/kotlinx/dataframe/ColumnsContainer;)Lorg/jetbrains/kotlinx/dataframe/DataColumn;
 	public static final fun ColumnDescription_top (Lorg/jetbrains/kotlinx/dataframe/DataRow;)Ljava/lang/Object;
 	public static final fun ColumnDescription_type (Lorg/jetbrains/kotlinx/dataframe/ColumnsContainer;)Lorg/jetbrains/kotlinx/dataframe/DataColumn;
-	public static final fun ColumnDescription_type (Lorg/jetbrains/kotlinx/dataframe/DataRow;)Lkotlin/reflect/KType;
+	public static final fun ColumnDescription_type (Lorg/jetbrains/kotlinx/dataframe/DataRow;)Ljava/lang/String;
```
zaleslaw, line 2722: very useful to track changes!

zaleslaw (approved):

koperagen:
Could you please add a test for #352? Maybe even if something else covers it, good to have one that is closer to use case

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
```diff
@@ -44,13 +49,26 @@ public inline fun <reified T> AnyCol.isType(): Boolean = type() == typeOf<T>()
...
- * Returns `true` if [this] column is comparable, i.e. its type is a subtype of [Comparable] and its
- * type argument is not [Nothing].
+ * Returns `true` if [this] column is inter-comparable, i.e.
+ * its elements can be compared with each other.
+ *
+ * Technically, this means the elements' common type is a subtype of [Comparable] with
+ * the type argument not being [Nothing].
  */
-public fun AnyCol.isComparable(): Boolean =
+public fun AnyCol.isInterComparable(): Boolean =
```
koperagen, line 71: isInterComparable is a new / not widespread term, for me it raises some questions like what's the difference (if i were to see it for the first time)
For users isComparable can be described as "filter columns for which max(), min(), median() and other (we actually only have 3) order operations make sense"
You wouldn't expect to find max() value for mixed column of String and Double values, so it should make things clearer

Jolanrensen: Well, that's actually the reason I changed it, because for me `isComparable` can be interpreted in two ways:
- "is comparable": so can call max(), min(), etc., not true for mixed number columns
- "is `Comparable`"; implements the interface, which would be true for mixed number columns

you need KDocs to explain what it actually means. inter-comparable means "comparable inside" (or something, I came up with it too), which narrows down the meaning to the first. Of course, other name suggestions are very welcome if you can think of any :)

Jolanrensen:
> Could you please add a test for #352? Maybe even if something else covers it, good to have one that is closer to use case

Very good comment. After I added a test I noticed this PR didn't actually fully fix https://github.com/Kotlin/dataframe/issues/352. It just fixed the error message. The actual issue needs to be solved separately. I will reorganize things a bit

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
Jolanrensen: With suggestion of ChatGPT I'll rename it to `.valuesAreComparable()`. Would that be okay?

Issue #634, opened 2024-03-20
Add binary compatibility validator to dataframe-core


devcrocod:
Add a [Binary compatibility validator](https://github.com/Kotlin/binary-compatibility-validator) in dataframe-core to ensure backward compatibility across releases.

This tool will help for maintaining stable releases and aiding projects like Kandy in assessing compatibility risks with new versions.

PR #877, opened 2024-09-23
0.14 post release Part 2

Jolanrensen:
- bumps versions
- removes deprecations

PR #938, opened 2024-10-29
Adds the binary compatibility validator plugin

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/634 by adding https://github.com/Kotlin/binary-compatibility-validator

See the individual commits for all checks. I started by adding the `.api` files built using the 0.14.1 release branch, to see whether we already made changes to the API.

This correctly showed me:
- removed deprecated functions in https://github.com/Kotlin/dataframe/pull/877, which is fine of course :)
- removed `GenericColumn` from https://github.com/Kotlin/dataframe/pull/923, which probably won't cause issues as it was just used from the compiler plugin
- Split openApi module https://github.com/Kotlin/dataframe/pull/916, for which we need to warn users (but that was already obvious)

The plugin automatically runs when `check` or `build` runs, so that's nice :) I added it only to the public modules, skipping the examples, plugins, etc.

When updating the public API we'll need to try to keep binary compatibility as best as possible, resorting to deprecating old functions instead of just removing/changing them. And if we introduce a breaking change, we'll need to inform the users in the release notes and run `apiDump` to overwrite the .api files.

A non-breaking change would be a `+` in the API check, this means we simply add another function or overload. A breaking change would be a `-`. Those will cause `NoSuchMethodError`s when called from an older version: 

image

For more information, check https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html

Hint: Sometimes `DeprecationLevel.HIDDEN` can help to remove a function from the API while keeping it in the bytecode. The Api checker plugin still shows it as a `-`, but it can still be called from a compiled library, avoiding a `NoShuchMethodError`.


dataframe-jdbc/build.gradle.kts
```diff
@@ -5,6 +5,7 @@ plugins {
         alias(kover)
         alias(ktlint)
         alias(jupyter.api)
+        alias(binary.compatibility.validator)
```
zaleslaw, line 8: Let's disable for a while

zaleslaw: From other hand, it's just a check and could help to discover stuff, which should be documented!

zaleslaw (approved):

dataframe-jdbc/build.gradle.kts
Jolanrensen: It doesn't hurt to add it. It will just inform you when you make a binary breaking change, after which you can call `apiDump`, commit those and the error goes away. But most importantly it makes you aware of the changes that we'll need to inform our users about for the next release.

PR #942, opened 2024-11-06
[Compiler plugin] Propagate nullability in toDataFrame tree conversion

koperagen:
There are 3 cases where plugin assumed non-null type for column and it actually should be nullable. With this fix, compile schema matches runtime schema


Jolanrensen (approved):
Looks good :)

Jolanrensen:
I do have one potential bug: How does the compiler plugin receive type `T` from `toDataFrame`? Does it actually get all the type information like the function or does it assume that the type can be received from the first index of generic types? 

In other words, how does it handle something like this?
```kt
import org.jetbrains.kotlinx.dataframe.*
import org.jetbrains.kotlinx.dataframe.annotations.*
import org.jetbrains.kotlinx.dataframe.api.*
import org.jetbrains.kotlinx.dataframe.io.*

@DataSchema
data class D(
    val s: String
)

class Subtree(
    val p: Int,
    val l: List<Int>,
    val ld: List<D>,
)

class Root(val a: Subtree)

class MyList(val l: List<Root?>): List<Root?> by l

fun box(): String {
    val l = listOf(
        Root(Subtree(123, listOf(1), listOf(D("ff")))),
        null
    )
    val df = MyList(l).toDataFrame(maxDepth = 2)
    df.compareSchemas(strict = true)
    return "OK"
}

```

PR #944, opened 2024-11-07
Render FormattedFrame stored inside columns as HTML

koperagen:
With this change it will be possible to format structured data:
image

```
df.groupBy { receiverType }.toDataFrame("api")
    .convert { api }
    .with { it.format { all() }.where { refine || interpretable }.with { background(green) } }
```

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
@@ -5,6 +5,7 @@ import org.jetbrains.kotlinx.dataframe.AnyCol
 import org.jetbrains.kotlinx.dataframe.AnyFrame
 import org.jetbrains.kotlinx.dataframe.AnyRow
 import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.api.FormattedFrame
 import org.jetbrains.kotlinx.dataframe.api.FormattingDSL
 import org.jetbrains.kotlinx.dataframe.api.RowColFormatter
 import org.jetbrains.kotlinx.dataframe.api.asColumnGroup
@@ -138,13 +139,13 @@ internal var sessionId = (Random().nextInt() % 128) shl 24
 internal fun nextTableId() = sessionId + (tableInSessionId++)
 
 internal fun AnyFrame.toHtmlData(
-    configuration: DisplayConfiguration = DisplayConfiguration.DEFAULT,
+    defaultConfiguration: DisplayConfiguration = DisplayConfiguration.DEFAULT,
     cellRenderer: CellRenderer,
 ): DataFrameHtmlData {
     val scripts = mutableListOf<String>()
-    val queue = LinkedList<Pair<AnyFrame, Int>>()
+    val queue = LinkedList<RenderingQueueItem>()
 
-    fun AnyFrame.columnToJs(col: AnyCol, rowsLimit: Int?): ColumnDataForJs {
+    fun AnyFrame.columnToJs(col: AnyCol, rowsLimit: Int?, configuration: DisplayConfiguration): ColumnDataForJs {
         val values = if (rowsLimit != null) rows().take(rowsLimit) else rows()
         val scale = if (col.isNumber()) col.asNumbers().scale() else 1
         val format = if (scale > 0) {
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
```diff
@@ -196,6 +204,36 @@ internal fun AnyFrame.toHtmlData(
     return DataFrameHtmlData(style = "", body = body, script = script)
 }
 
+private interface DataFrameLike {
+    fun configuration(default: DisplayConfiguration): DisplayConfiguration
+
+    fun df(): AnyFrame
+}
+
+private fun dataFrameLikeOrNull(value: Any?): DataFrameLike? =
```
Jolanrensen, line 213: I'd probably make this an extension function, like `Any?.asDataFrameLikeOrNull()`, similar to `.asIntOrNull()` etc.

Jolanrensen:
So, now you can have different configurations for each different nested DataFrame? That's cool :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/rendering/RenderingTests.kt
```diff
@@ -196,4 +202,29 @@ class RenderingTests : TestBase() {
...
+
+        val formatted = dataFrameOf("a")(df.format { all() }.with { empty })
+        val nestedFrame = dataFrameOf("a")(df)
+        val configuration = DisplayConfiguration(enableFallbackStaticTables = false)
+        tableInSessionId = 0
+        val formattedHtml = formatted.toStandaloneHTML(configuration).toString()
+        tableInSessionId = 0
+        val regularHtml = nestedFrame.toStandaloneHTML(configuration).toString()
+
+        formattedHtml.replace("api.FormattedFrame", "DataFrame") shouldBe regularHtml
```
Jolanrensen, line 221: maybe also check whether "api.FormattedFrame" is actually present, otherwise this test would succeed if formatting did nothing.

koperagen:
> So, now you can have different configurations for each different nested DataFrame? That's cool :)

Yes. For example, i wanted to have green background for all dataframe API that are annotated with refine or interpretable. Later i decided that it will be more convenient to groupBy this dataframe by receiver type and have the same formatting. But i couldn't do it anymore, without this change

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/rendering/RenderingTests.kt
```diff
@@ -196,4 +202,29 @@ class RenderingTests : TestBase() {
...
+
+    @Test
+    fun `render nested FormattedFrame as DataFrame`() {
+        val empty = object : CellAttributes {
+            override fun attributes(): List<Pair<String, String>> = emptyList()
+        }
+        val df = dataFrameOf("b")(1)
+
+        val formatted = dataFrameOf("a")(df.format { all() }.with { empty })
+        val nestedFrame = dataFrameOf("a")(df)
```
Jolanrensen, line 214: It's probably okay, as formatting is usually the final step in the operation, but what would happen if you were to perform a format on a nested DF, then perform an operation on the frame column, like move, rename, or filter, and then convert it to html? Would it lose the formatting?

Jolanrensen (approved):
thanks :) looks good!

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/rendering/RenderingTests.kt
koperagen: Look at it this way: we had `api: DataColumn<DataFrame<Type>>`, and after this operation it becomes `api: DataColumn<FormattedFrame<Type>>` 
```
df.groupBy { receiverType }.toDataFrame("api")
    .convert { api }
    .with { it.format { all() }.where { refine || interpretable }.with { background(green) } }
```

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/rendering/RenderingTests.kt
koperagen: Second test checks actual formatting. First test is to emphasize that FormattedFrame should be treated same as DataFrame in toHTML

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/rendering/RenderingTests.kt
Jolanrensen: okay makes sense :)

PR #947, opened 2024-11-08
[Important Fix!] Parse started to removed unselected columns

Jolanrensen:
Fixed `parseImpl()` removing unselected columns by reverting implementation to use `convert` instead of `insert`.

We didn't have a test that ran `parse { someCol }`, but I found it running the example in README.md


koperagen:
Why is it different from original code? isFrameColumn changed 
```
internal fun <T> DataFrame<T>.parseImpl(options: ParserOptions?, columns: ColumnsSelector<T, Any?>) =
    convert(columns).to {
        when {
            it.isFrameColumn() -> it.cast<AnyFrame?>().parse(options)

            it.isColumnGroup() ->
                it.asColumnGroup()
                    .parse(options) { all() }
                    .asColumnGroup(it.name())
                    .asDataColumn()

            it.typeClass == String::class -> it.cast<String?>().tryParse(options)

            else -> it
        }
    }
```

koperagen (approved):

PR #948, opened 2024-11-08
Add castTo(Function) overload for workflows that use compiler plugin 

koperagen:


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cast.kt
```diff
@@ -42,6 +42,36 @@ public inline fun <reified T> AnyFrame.castTo(
     verify: Boolean = true,
 ): DataFrame<T> = cast<T>(verify = verify)
 
+/**
```
Jolanrensen, line 45: Very nice and good explanation!

Jolanrensen (approved):

PR #951, opened 2024-11-14
[Compiler plugin] Support valueCounts

koperagen:
Important part of this PR is first commit. Other two are rather trivial refactoring. 

Inline all trivial delegates:

It was supposed to be a source of truth to help choose correct arguments, but in reality it's more reliable to look at existing interpreters


plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/loadInterpreter.kt
```diff
@@ -248,6 +249,7 @@ internal inline fun <reified T> String.load(): T {
         "Update0" -> Update0()
         "Aggregate" -> Aggregate()
         "DataFrameOf3" -> DataFrameOf3()
+        "ValueCounts" -> ValueCounts()
```
zaleslaw, line 252: https://kotlin.github.io/dataframe/valuecounts.html - very strange GROUP BY + DISTINCT

zaleslaw (approved):

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/loadInterpreter.kt
koperagen: I agree :)

PR #957, opened 2024-11-19
Add `df.properties()` function to improve visibility of generated API for typed column access

koperagen:
It's not always easy to tell if code generation in notebooks or compiler plugin worked as expected or find columns you're interested in without printing schema. With this change, people will have an option to overview generated column properties

In notebooks:
```
val df = DataFrame.read("file.csv")
==== next code cell
df. // column properties are mixed together with methods, not easy to find unless you already know names
df.properties(). // easy to overview available columns
```
image

In compiler plugin:
```
val df = @Import DataFrame.read("file.csv")
df.properties().
```
image

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataFrameGet.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataFrameGet.kt
@@ -4,6 +4,7 @@ import org.jetbrains.kotlinx.dataframe.AnyCol
 import org.jetbrains.kotlinx.dataframe.AnyColumnReference
 import org.jetbrains.kotlinx.dataframe.ColumnSelector
 import org.jetbrains.kotlinx.dataframe.ColumnsContainer
+import org.jetbrains.kotlinx.dataframe.ColumnsScope
 import org.jetbrains.kotlinx.dataframe.ColumnsSelector
 import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.DataFrame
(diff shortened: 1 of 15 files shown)


Jolanrensen:
It will clash if you have `df.group { a and b }.into("columns")`, right? 
I suspect this is the reason we have ~`df.size`~ `df.size()`, ~`df.nrow`~ `df.rowsCount()`, etc in the public api.
Only functions, no properties. Now, `df.columns()` already exists, but maybe `df.columnScope()` could work?

koperagen:
Generated properties override "columns", so for users it means that they loose ability to overview columns if they have column with this name. It won't prevent them from accessing their "columns" column though. That's why i think it's a good trade off, no?
image

image

Jolanrensen:
Hmm that means the API would have certain parts be available depending on the data.
I think this is not a good idea. It may create unexpected behavior and copy/pasting of examples failing, etc. Plus, it's a slippery slope to adding more properties to DataFrame and more potential clashes... So no, I would say we maintain our convention of not having any public properties to DataFrame, just the generated ones.

Functions are fine of course :)

koperagen:
There's an important part: it's not available in selection dsl, so potential conflicts are limited to only `df.` column access. Generally, it's not even useful api - It's an utility to mitigate completion list clutter and the fact there's no good way to see generated code
We can simply agree to not add properties other than this one

Jolanrensen:
Yes, but we'd still have conflicts that don't need to be there. What's wrong with making a function instead?
I think `df.availableColumns()`, for instance, would be a nice alternative that can never clash :)

koperagen:
Because `df.columns` is a better name. We'd have conflicts - if someone wants to overview columns (which wasn't even possible before) and miraculously have "columns" in their dataframe

Jolanrensen:
it's also an option to deprecate the old `df.columns()` -> `df.getColumns()` and migrate the functionality to this. That would be fine by me, but a property, I can't agree with

zaleslaw:
I need a pause for a week to play with it

koperagen:
Alternative option: function named properties. example: `df.properties()`

zaleslaw:
I'd like properties more "https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/to-properties.html"
Maybe toProperties?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataFrameGet.kt
```diff
@@ -58,6 +59,36 @@ public fun <T> ColumnsContainer<T>.getFrameColumn(columnName: String): FrameColu
...
+ * ==== next code cell
+ * df. // column properties are mixed together with methods, not easy to find unless you already know names
+ * df.properties(). // easy to overview available columns
+ * ```
+ * In compiler plugin:
+ * ```
+ * val df = @Import DataFrame.read("file.csv")
+ * df.properties().
+ * ```
+ * By design, it can't be used in columns selection DSL. Because "properties()" is optional - for convenient overview.
```
Jolanrensen, line 81: hmm I'd leave out this bit. Users can figure out for themselves this doesn't work and all your examples are calling upon a `DataFrame` instance anyway

Jolanrensen (approved):
Useful! Thanks!

PR #959, opened 2024-11-20
Prepare compilation pipeline that hides API overloads - KProperty and ColumnAccessor, for compiler plugin workflow 

koperagen:
With DataFrame public API now, for most operations there're 4 overloads. 
KProperty and ColumnAccessor overloads were needed to support more convenient, robust safe access to columns. With compiler plugin, there's an idea these overloads won't be useful and only clutter code completion popup and make on-boarding harder. 
This PR adds new tiny compiler plugin to compile the same sources where all these overloads are public, and make them internal in the process. 
Publication of a new artifact  will be in a separate PR because it's trickier that i thought

experimental:
image

core:
image


Jolanrensen:
Maybe the name `-experimental` could be a bit more descriptive. It's probably temporary, but still, we might want to publish another experiment in the future and name that `dataframe-experimental`. So I'd go with something like `dataframe-core-minimal`, or `dataframe-core-for-compiler-plugin`, or something like that.

core/build.gradle.kts
```diff
@@ -295,6 +294,40 @@ tasks.withType<org.jetbrains.dokka.gradle.AbstractDokkaLeafTask> {
...
 // endregion
 
+// region experimental artifact
+val experimentalJar by tasks.registering(Jar::class) {
+    archiveBaseName.set("experimental")
+    from(experimental.output)
+}
+
+val experimentalJarSources by tasks.registering(Jar::class) {
+    dependsOn("kspKotlin")
```
Jolanrensen, line 304: If it's dependent on the generated sources, it should probably also depend on the task `processKDocsMain`, right?

core/build.gradle.kts
```diff
@@ -408,6 +441,11 @@ kotlinPublications {
         description = "Dataframe core API"
         packageName = artifactId
     }
+    publication {
+        publicationName = "dataframe-experimental"
+        description = "Excel support for Kotlin Dataframe"
+        composeOfTaskOutputs(listOf(experimentalJar, experimentalJarSources))
```
Jolanrensen, line 447: omg, I'm gonna have to look into doing this for the publishing of generated sources too. That's way easier than temporarily modifying the kotlin main sourceset before publishing each time.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/Plugin.kt
```diff
@@ -43,3 +43,6 @@ public annotation class ScopeProperty
 
 @Target(AnnotationTarget.FUNCTION)
 internal annotation class Check
+
+@Target(AnnotationTarget.FUNCTION)
+internal annotation class ApiOverload
```
Jolanrensen, line 48: Please add a small KDoc how it should be used and what it does.

Also, if we would go with `dataframe-core-minimized`, we could make this class be named `@ExtendedApi` or something. Indicating that this API will be larger.

Makes it also easier to recommend to users in the future: "If you don't have or you cannot run the compiler plugin, you can use the extended API to make use of the old column accessors or KProperties"

plugins/public-api-modifier/src/org/jetbrains/kotlinx/dataframe/PublicApiModifierRegistrar.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.jetbrains.kotlin.fir.declarations.FirSimpleFunction
+import org.jetbrains.kotlin.fir.extensions.FirExtensionRegistrar
+import org.jetbrains.kotlin.fir.extensions.FirExtensionRegistrarAdapter
+import org.jetbrains.kotlin.fir.extensions.FirStatusTransformerExtension
+import org.jetbrains.kotlin.fir.resolve.fqName
+import org.jetbrains.kotlin.fir.symbols.impl.FirClassLikeSymbol
+import org.jetbrains.kotlin.name.FqName
+
+@OptIn(ExperimentalCompilerApi::class)
+class PublicApiModifierRegistrar : CompilerPluginRegistrar() {
```
Jolanrensen, line 20: this is a good template compiler plugin haha. I don't think many are as easy as this one :)

Jolanrensen (changes requested):
This would have taken a lot of time, wouldn't it? haha. Nice job thinking about the user experience in combination with the compiler plugin :)

I just have some suggestions regarding naming to prevent future headaches :)

(the module name is perfect btw, very descriptive :) )

koperagen:
I was thinking `dataframe-core-compact-api` or just `dataframe-core-compact`. ExtendedApi annotation seems like an option too, but it kind of swaps default, right? Like, this way we need dataframe-core by default and dataframe-core-extended :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/Plugin.kt
koperagen: I'll provide explanation here

Jolanrensen:
> I was thinking `dataframe-core-compact-api` or just `dataframe-core-compact`. 

Maybe... Although "compact" suggests everything is still included, just more compressed somehow, which is not the case, we actually take stuff out.

AI suggestions:
- dataframe-core-lite (good option!)
- dataframe-core-core (LOL no XD)
- dataframe-core-basic (meh)
- dataframe-core-mini (maybe)
- dataframe-core-reduced (too long imo)
- dataframe-core-slim (I guess...)
- dataframe-core-minimal (could be a good option)

>ExtendedApi annotation seems like an option too, but it kind of swaps default, right? Like, this way we need dataframe-core by default and dataframe-core-extended :)

Yeah I see what you mean.. Anyway, we can always rename the annotation later :) `@FullApi` could be an option?

AI suggestions:
- `@NotInLiteVersion` (lol)
- `@FullCoreOnly` (mmm)
- `@ExcludedFromLite` (is clear tho)

koperagen:
@Jolanrensen I decided to split this PR. I'll setup publishing later, it's trickier than i thought :(

Jolanrensen (approved):

PR #965, opened 2024-11-25
Clarify the situation when generated extension property causes ClassCast or NPE exceptions

koperagen:
Now if compiler plugin or jupyter integration has a bug and generated property returns type that mismatches runtime, we get vague ClassCast or NPE. This PR tries to give a little bit more info about it and propose a workaround

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -9918,6 +9918,11 @@ public final class org/jetbrains/kotlinx/dataframe/exceptions/ColumnNotFoundExce
 	public fun getMessage ()Ljava/lang/String;
 }
 
+public final class org/jetbrains/kotlinx/dataframe/exceptions/ColumnTypeMismatchesColumnValuesException : java/lang/RuntimeException {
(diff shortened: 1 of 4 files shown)


koperagen:
For `convert.with`
Yes, there are other cases, but for example this one is a bit better. Because this["a"] as String is checked cast, so it will fail right there and with sensible stack trace. 
```
val DataRow<Marker>.a get() = this["a"] as String
df.filter { a != "something" } // throws ClassCastException if `a` does not contain Strings
```
When DataColumn accessors are called, it's unchecked cast and value can travel quite a bit before library "extract" value from the column. So stack trace and exception message are less helpful - i think convert with is popular example of such operation

Jolanrensen:
@koperagen I see :) Well, then `convert.with` throwing the new exception is a good start :) We'll likely need to add it in other places as well, but maybe that can be done later after we discover where else it pops up.

Jolanrensen (approved):

PR #989, opened 2024-12-05
Miscellaneous 0.15 fixes

Jolanrensen:
all need to be cherry-picked to the 0.15.0 release branch

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
@@ -71,6 +71,7 @@ internal class Integration(private val notebook: Notebook, private val options: 
 
     // TODO temporary settings while these experimental modules are being developed
     private val enableExperimentalCsv = options["enableExperimentalCsv"]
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
```diff
@@ -152,6 +156,15 @@ internal class Integration(private val notebook: Notebook, private val options:
 
     override fun Builder.onLoaded() {
         if (version != null) {
+            if (enableExperimentalCsv?.toBoolean() == true) {
+                println("Enabling experimental CSV module: dataframe-csv")
```
koperagen, line 160: Do you want to keep it? It can't be disabled on user end

Jolanrensen: I think it's good to have it, a good check for the user that the dataframe integration received the special command correctly and activated the dataframe-csv module. 
It's also good to see it in the output of the notebook, to see something experimental is going on, like a small warning.
When the module is included by default I'll remove the message of course

PR #990, opened 2024-12-06
Sort df.compileTimeSchema() columns according to df.schema() so they're easier to compare

koperagen:
This is quality of life improvement. I print both schemas when i test implementation of join operation. I used to get compileTimeSchema in somewhat random order, it makes diff useless. Now when columns are present in both schemas, they will go in a definite order

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/schema.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/schema.kt
@@ -1,0 +1,55 @@
+package org.jetbrains.kotlinx.dataframe.impl.api
+
+import org.jetbrains.kotlinx.dataframe.columns.ColumnPath
+import org.jetbrains.kotlinx.dataframe.impl.schema.DataFrameSchemaImpl
+import org.jetbrains.kotlinx.dataframe.impl.schema.getSchema
+import org.jetbrains.kotlinx.dataframe.schema.ColumnSchema
+import org.jetbrains.kotlinx.dataframe.schema.DataFrameSchema
+import kotlin.reflect.KClass
+
+@PublishedApi
+internal fun compileTimeSchemaImpl(runtimeSchema: DataFrameSchema, klass: KClass<*>): DataFrameSchema {
+    val compileSchema = getSchema(klass)
+    val root = ColumnPath(emptyList())
+    val order = buildMap {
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/schema.kt
```diff
@@ -0,0 +1,54 @@
...
+@PublishedApi
+internal fun compileTimeSchemaImpl(runtimeSchema: DataFrameSchema, klass: KClass<*>): DataFrameSchema {
+    val compileSchema = getSchema(klass)
+    val root = ColumnPath(emptyList())
+    val order = mutableMapOf<ColumnPath, Int>()
+    runtimeSchema.putColumnsOrder(order, path = root)
+    return compileSchema.sorted(order, path = root)
+}
+
+internal fun DataFrameSchema.putColumnsOrder(order: MutableMap<ColumnPath, Int>, path: ColumnPath) {
```
Jolanrensen, line 19: kinda strange to see `DataFrameSchema` be the receiver. I'd expect the mutable map to be the receiver, since that's what you're modifying. That could also enable the notation:
```kt
val order = buildMap<ColumnPath, Int> {
    putColumnsInOrderOf(runtimeSchema, path)
}
```

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/schema.kt
```diff
@@ -0,0 +1,54 @@
...
+            }
+
+            is ColumnSchema.Group -> {
+                column.schema.putColumnsOrder(order, columnPath)
+            }
+        }
+    }
+}
+
+internal fun DataFrameSchema.sorted(order: Map<ColumnPath, Int>, path: ColumnPath): DataFrameSchema {
```
Jolanrensen, line 35: here it's okay to have it as a receiver :) I'd only name it "sortedBy(order)" to be more expressive.

Jolanrensen (changes requested):
just some small syntactical tips, otherwise lgtm :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/schema.kt
koperagen: true, i changed it

Issue #899, opened 2024-10-01
Move :generator into :plugins


Jolanrensen:
The generator module exists to generate forbidden kotlin keywords and is used by the :core module. Since it's a gradle plugin, however, we should likely move it under :plugins, like the other plugins we have.

Issue #995, opened 2024-12-10
Update to Kotlin 2.1+


Jolanrensen:
Bumping this requires some fixes in the :generator plugin and potentially some other places too

Omico:
Gradle 8.11.1 has been released. Don't forget to run the `gradle wrapper` task to update the wrapper jar & scripts.
![image](https://github.com/user-attachments/assets/bf0c4db4-4764-4e35-afb6-27298159cc47)

Jolanrensen:
After trying to update to Kotlin 2.1 in https://github.com/Kotlin/dataframe/pull/997, I bumped into more and more issues, especially regarding our infrastructure.

I would seriously consider dropping our dogfooding approach with the gradle plugin, especially since we're gonna drop it in favor of the compiler plugin. Not only is the KSP version causing issues, but the actual Kotlin Gradle plugin is clashing because it has multiple mismatching versions on the classpath.

PR #996, opened 2024-12-10
Keywords generator plugin: moved and fixed for Kotlin 2.1

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/899

This PR is required for updating to Kotlin 2.1 (https://github.com/Kotlin/dataframe/issues/995)

- Moved `:generator` plugin to worker-based `:plugins:keywords-generator` plugin. This is required for Kotlin 2.1 to avoid tons of warnings. Explained here: https://kotlinlang.org/docs/whatsnew21.html#compiler-symbols-hidden-from-the-kotlin-gradle-plugin-api
- Bumped KotlinPoet version to 2.0.0

We should be able to merge this even before bumping to Kotlin 2.1


plugins/keywords-generator/src/main/kotlin/org/jetbrains/dataframe/keywords/KeywordsGeneratorPlugin.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.gradle.kotlin.dsl.get
+import org.gradle.kotlin.dsl.register
+import org.jetbrains.kotlinx.dataframe.BuildConfig
+import java.io.File
+
+abstract class KeywordsGeneratorPlugin : Plugin<Project> {
+
+    override fun apply(target: Project): Unit = with(target) {
+        // from https://kotlinlang.org/docs/whatsnew21.html#compiler-symbols-hidden-from-the-kotlin-gradle-plugin-api
+        val taskDependencyScope = configurations.create("keywordsGeneratorScope")
```
koperagen, line 16: For me "scope" here was misleading, took me longer to realize it simply is a new configuration

plugins/keywords-generator/src/main/kotlin/org/jetbrains/dataframe/keywords/KeywordsGeneratorPlugin.kt
```diff
@@ -0,0 +1,44 @@
...
+        dependencies.add(taskDependencyScope.name, "$KOTLIN_COMPILER_EMBEDDABLE:$KOTLIN_COMPILER_VERSION")
+        val taskResolvableConfiguration = configurations.create("keywordGeneratorResolvable") {
```
koperagen, line 18: What is the purpose? I know there are resolvable configurations in gradle, but this one seems to duplicate taskDependencyScope without modifications?

plugins/keywords-generator/src/main/kotlin/org/jetbrains/dataframe/keywords/KeywordsGeneratorTask.kt
```diff
@@ -0,0 +1,41 @@
...
+    abstract val executor: WorkerExecutor
+
+    @get:Classpath
+    abstract val kotlinCompiler: ConfigurableFileCollection
+
+    @OutputDirectory
+    lateinit var srcDir: File
+
+    @Input
+    override fun getGroup() = "codegen"
```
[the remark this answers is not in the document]
Jolanrensen: I just copied this over from the old implementation. Shall I still remove it?

plugins/keywords-generator/src/main/kotlin/org/jetbrains/dataframe/keywords/KeywordsGeneratorPlugin.kt
Jolanrensen: just took it from https://kotlinlang.org/docs/whatsnew21.html#compiler-symbols-hidden-from-the-kotlin-gradle-plugin-api

plugins/keywords-generator/src/main/kotlin/org/jetbrains/dataframe/keywords/KeywordsGeneratorPlugin.kt
Jolanrensen: also not sure, but they do it in the example https://kotlinlang.org/docs/whatsnew21.html#compiler-symbols-hidden-from-the-kotlin-gradle-plugin-api so there must be a reason for it I guess

zaleslaw (approved):

plugins/keywords-generator/src/main/kotlin/org/jetbrains/dataframe/keywords/KeywordsGeneratorPlugin.kt
Jolanrensen: It represents a `DependencyScopeConfiguration`, will fail in gradle 9.0 if we don't split it

plugins/keywords-generator/src/main/kotlin/org/jetbrains/dataframe/keywords/KeywordsGeneratorPlugin.kt
Jolanrensen: It represents a `ResolvableConfiguration`, will fail in gradle 9.0 if we don't split it

plugins/keywords-generator/src/main/kotlin/org/jetbrains/dataframe/keywords/KeywordsGeneratorTask.kt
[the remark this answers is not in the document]
Jolanrensen: it will be put in "other" if we don't supply a group (which is ginormous). Maybe a separate group for codegen stuff is not such a bad idea.

koperagen (approved):
Thank you!

Issue #998, opened 2024-12-11
Failed to convert single character to Enum using parser


hantsy:
After upgrading to v0.15, a new issue is occurred in our application. 

There is a column that accepts a single character in the CSV, and we use the following parser to convert it from a string directly.

```kotlin 
 df.
...
.convertTo<xxx>{
 parser { ACHType.fromSymbol(it) }

}
```

which worked well with the former version, but failed now with the following information.

```bash
Type converter from kotlin.Char to xxx.ACHType? is not found for column 'ACH Type'
```

Jolanrensen:
Ah it's because we added Char recognition to our parser system.

It should work if you write 
`convert<Char>.with { ACHType.fromSymbol(it.toString()) }`

I'll have a look whether we can pass Chars (as String) through the parser function as well, that way we can keep the previous notation working. Also, I'll add some tests to catch it in the future. Thanks!

hantsy:
When using 

```kotlin
.convertTo<DataSchemaType>{
    parser { ACHType.fromSymbol(it) }
    parser { other parsing from it}
}
```

Here, `it` is always a string.

hantsy:
In the original post, I think the problem is the `parser` is not called.

It did not work as other parsers.

Jolanrensen:
@hantsy yes that's true. Could you share `df.schema()` in your example?

hantsy:
I create an example project to produce the issue, https://github.com/hantsy/dataframe-sandbox/blob/master/src/main/kotlin/Main.kt

When the mainKt, it will display the following info.

```bash
12:58:11.221 [pool-1-thread-4] DEBUG org.jetbrains.kotlinx.dataframe.impl.io.FastDoubleParser -- Could not parse 'N' as Double with NumberFormat with locale 'en_US'.
Amount: Double
Last 4: Int
ACH Type: Char
Exception in thread "main" org.jetbrains.kotlinx.dataframe.exceptions.TypeConverterNotFoundException: Type converter from kotlin.Char to org.example.ACHType? is not found for column 'ACH Type'
	at org.jetbrains.kotlinx.dataframe.impl.api.ConvertKt.convertToTypeImpl$convertPerCell(convert.kt:189)
	at org.jetbrains.kotlinx.dataframe.impl.api.ConvertKt.convertToTypeImpl(convert.kt:209)
	at org.jetbrains.kotlinx.dataframe.api.ConvertKt.convertTo(convert.kt:137)
[...]

PR #999, opened 2024-12-11
Handling Chars as Strings implicitly

Jolanrensen:
Related to https://github.com/Kotlin/dataframe/issues/998

This PR can best be reviewed per-commit. It contains 3 parts where we can enhance Char support:

The [first commit](https://github.com/Kotlin/dataframe/pull/999/commits/f86a6ccea2da632fda2bf7f2563e42bb8c72b86a) adds String-fallback for Char columns to `df.convert()...` and `charCol.convertTo<X>()`. This means you can now do things like

```kt
enum class EnumClass { A, B }

columnOf('A', 'B').convertTo<EnumClass>()
// previously you could only do this
columnOf('A', 'B').convertToString().convertTo<EnumClass>() 
```

Converting Char -> Int remains unchanged, it takes the char code instead of the value, similar to casting a char to int in java.

The [second commit](https://github.com/Kotlin/dataframe/pull/999/commits/5c54f58224497ecb6200f3796152b21419a766e0) adds "treating chars as strings" behavior to the `df.convertTo<Schema> {}` DSL. This makes the behavior mentioned in https://github.com/Kotlin/dataframe/issues/998 possible again:

```kt
df.convertTo<Schema>() {
    // can now be used for both String and chars
    parser { /* it: String -> */ MyCustomClass(it) }

    // will be used instead of parser {} if present
    convert<Char>().with { /* it: Char -> */ ... }
}
```
This we could change to `charParser {}` and `parser {}` separately if we wish so. It's still up for debate.

The [third commit](https://github.com/Kotlin/dataframe/pull/999/commits/36d6acc104dc0c9c3b1a8e4db3b8b3a8b9bd1bcb) introduces parsing of `Char` columns, similar to String columns.
This means you can now do:
```kt
columnOf('1', '2', '3').parse() // results in DataColumn<Int>

columnOf('a', 'b', 'c').parse() // results in DataColumn<String>
```
and `Char` columns will also be considered when calling `DataFrame.parse()`.

I still need to update the docs, but first I want to be sure there are not other places that could benefit from this Char-as-String treatment.

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -6501,8 +6501,12 @@ public final class org/jetbrains/kotlinx/dataframe/api/ParseKt {
 	public static synthetic fun parse$default (Lorg/jetbrains/kotlinx/dataframe/DataFrame;[Lorg/jetbrains/kotlinx/dataframe/columns/ColumnReference;Lorg/jetbrains/kotlinx/dataframe/api/ParserOptions;ILjava/lang/Object;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun parseAnyFrameNullable (Lorg/jetbrains/kotlinx/dataframe/DataColumn;Lorg/jetbrains/kotlinx/dataframe/api/ParserOptions;)Lorg/jetbrains/kotlinx/dataframe/DataColumn;
 	public static synthetic fun parseAnyFrameNullable$default (Lorg/jetbrains/kotlinx/dataframe/DataColumn;Lorg/jetbrains/kotlinx/dataframe/api/ParserOptions;ILjava/lang/Object;)Lorg/jetbrains/kotlinx/dataframe/DataColumn;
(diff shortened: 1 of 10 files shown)


hantsy:
> columnOf('a', 'b', 'c').parse() // results in DataColumn<String>

I remember any chars in the ascii table can be converted into Int automatically in Java.

Jolanrensen:
> > columnOf('a', 'b', 'c').parse() // results in DataColumn
> 
> I remember any chars in the ascii table can be converted into Int automatically in Java.

yes indeed! However, in java you need to be careful when you want to convert to/from ascii codes or the readable values. In Kotlin, this has been made clearer:

```kt
// get ascii code from char
'1'.code == 49

// get readable value
'1'.digitToInt() == 1

// and back, from ascii code
49.toChar == '1'

// back to readable char
1.digitToChar()  == '1'
```

In DataFrame, I'd like `convert` to use the ascii conversion methods, while `parse` should provide the readable value conversion alternative.

Jolanrensen:
Surpassed by https://github.com/Kotlin/dataframe/pull/1420

Issue #890, opened 2024-09-26
Cannot call `.map {}` on `ColumnGroup`


Jolanrensen:
Annoying overload resolution ambiguity bug

![image](https://github.com/user-attachments/assets/a15f9f0b-6596-4433-b240-043ec94c77a3)

PR #1007, opened 2024-12-19
#890 - Cannot call .map {} on ColumnGroup

ysknsid25:
Fix issue https://github.com/Kotlin/dataframe/issues/890

If you want to use `DataFrame<T>.map()`, you can usually just call it with the default arguments. If you have something else you want to `transform`, I don't mind returning the result of passing it to you.

# before 

image

# after

image

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/map.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/map.kt
@@ -65,7 +65,7 @@ public fun <T, R> DataColumn<T>.mapIndexed(
 
 // region DataFrame
 
-public fun <T, R> DataFrame<T>.map(transform: RowExpression<T, R>): List<R> = rows().map { transform(it, it) }
+public fun <T, R> DataFrame<T>.map(transform: RowExpression<T, R>, rows: Iterable<DataRow<T>> = rows()): List<R> = rows.map { transform(it, it) }
 
 public inline fun <T, reified R> ColumnsContainer<T>.mapToColumn(
     name: String,


Jolanrensen:
Thank you very much!

It does, however, seem that your fix breaks the notation: 
```kt
df.map { ... }
```
as seen in the test here: https://github.com/ysknsid25/dataframe/blob/b67c665cb10375ac13f2dd9bca8aa838af522082/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Modify.kt#L1130

I think it may be easier to just provide another `map` function but defined on `ColumnGroup` instead of `DataFrame`. I suspect this will work, because the compiler always chooses the function that matches the provided type the closest. Could you try?

ysknsid25:
I defined the following in `map.kt`, but it did not resolve the ambiguous overload 🫠

```
// region ColumnGroup

public fun <T, R> ColumnGroup<T>.map(transform: (AnyCol) -> R): List<R> = columns().map { transform(it) }

// endregion
```

image

Jolanrensen:
I see :( that's a shame.

Then the only solution may be to add
```kt
public fun <R> map(transform: RowExpression<T, R>): List<R> = asDataFrame().map(transform)
```
directly inside the `ColumnGroup` interface. Member functions do always have priority over extension functions, so it would probably work fine. (just need to check if it doesn't break anything else :) )

Jolanrensen:
> I see :( that's a shame.
> 
> Then the only solution may be to add
> 
> ```kotlin
> public fun <R> map(transform: RowExpression<T, R>): List<R> = asDataFrame().map(transform)
> ```
> 
> directly inside the `ColumnGroup` interface. Member functions do always have priority over extension functions, so it would probably work fine. (just need to check if it doesn't break anything else :) )

Another would be to use `@HidesMember` from https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/src/kotlin/internal/Annotations.kt

PR #1015, opened 2025-01-12
Documentation for date-time pattern

DestBro:
This commit adds new documentation detailing how to parse date-time columns from CSV files using a specific format pattern. It explains two approaches: providing the pattern as a raw string (e.g., "dd/MMM/yy h:mm a") and supplying a DateTimeFormatter instance (e.g., DateTimeFormatter.ofPattern("dd/MMM/yy h:mm a")). These options ensure that columns are correctly recognized and parsed as date-time rather than strings.

### DIFF
--- a/docs/StardustDocs/topics/read.md
+++ b/docs/StardustDocs/topics/read.md
@@ -172,7 +172,8 @@ val df = DataFrame.readCSV(
 
 ### Work with specific date-time formats
 
-When parsing date or date-time columns, you might encounter formats different from the default ISO_LOCAL_DATE_TIME.
+When parsing date or date-time columns, you might encounter formats different from the default `ISO_LOCAL_DATE_TIME`.
+Like:
 
 <table>
 <tr><th>date</th></tr>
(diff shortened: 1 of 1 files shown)


DestBro:
Thanks for the feedback! Honestly, it was a bit tricky to search for the relevant docs at first, which is why I ended up here with the keywords I used. I didn’t see any problem adding the documentation in this place, especially since there’s already a similar block of code above referencing **ParserOptions**. In my opinion, more documentation never hurts! :)

### Work with locale-specific numbers
Sometimes columns in your CSV can be interpreted differently depending on your system locale.
<table>
<tr><th>numbers</th></tr>
<tr><td>12,123</td></tr>
<tr><td>41,111</td></tr>
</table>
Here a comma can be decimal or thousands separator, thus different values.
You can deal with it in two ways:

1) Provide locale as a parser option

```
  val df = DataFrame.readCSV(
      file,
      parserOptions = ParserOptions(locale = Locale.UK),
  )
```

2) Disable type inference for a specific column and convert it yourself

```
  val df = DataFrame.readCSV(
      file,
      colTypes = mapOf("colName" to ColType.String),
  )
```

Jolanrensen:
@DestBro Great job! I'll do some small formatting and merge it into our docs :) thanks!

PR #1022, opened 2025-01-21
`remove` docs and deprecating `df - cols`

### LINKED ISSUE #819: Add KDocs for `remove`

The [remove operation](https://kotlin.github.io/dataframe/remove.html) contains just a few overloads which differ just by how the columns should be selected (we have templates for this concept), so it should be relatively simple to get started getting used to the KDoc preprocessor with this function.

What is important to tell users that removing nested columns in column groups is supported!

A grammar DSL could be possible, showing both the `-` overload and the normal name

AndreiKingsley:
* Fixes #819


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/remove.kt
```diff
@@ -6,38 +6,93 @@ import org.jetbrains.kotlinx.dataframe.DataFrame
...
+ *
+ * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
+ *
+ * See [Selecting Columns][Select.SelectSelectingOptions].
+ *
+ * For more information: {@include [DocumentationUrls.Remove]}
+ */
+internal interface Remove
+
+/** {@set [SelectingColumns.OperationArg] [remove][remove]} */
```
Jolanrensen, line 37: OperationArg does not exist. OPERATION does

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt
```diff
@@ -79,6 +79,9 @@ internal const val IS_URL_IMPORT = "org.jetbrains.kotlinx.dataframe.io.isUrl"
 
 private const val MESSAGE_0_17 = "Will be ERROR in 0.17."
 
+internal const val MINUS = "This minus overload will be removed in favor of `remove`. $MESSAGE_0_17"
```
Jolanrensen, line 82: to speed up the cleanup, let's actually make this ERROR in 0.16 already :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/remove.kt
```diff
@@ -6,38 +6,93 @@ import org.jetbrains.kotlinx.dataframe.DataFrame
...
+/**
+ * ## The Remove Operation
+ *
+ * Removes the specified [columns] from the original [DataFrame] and returns a new [DataFrame] without them.
+ *
+ * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
+ *
+ * See [Selecting Columns][Select.SelectSelectingOptions].
+ *
+ * For more information: {@include [DocumentationUrls.Remove]}
```
Jolanrensen, line 33: We will probably move this to a `@see` tag later, but for now it looks good

Jolanrensen (changes requested):
just a tiny thing, otherwise, it's another operation kdoc'ed :D

PR #1024, opened 2025-01-21
Add an example notebook with analyses for the Top 12 German companies

devcrocod:


zaleslaw (changes requested):
What should be changed before to became an official example?

The next few points:

1. Add the Header and description to the Notebook, its main purpose and Summary for what goals are you going to achieve
2. Add some explanation as cells with texts to the given results/charts
3. Add a way, how you got the dataset, could it be repeated by the user
4. Rename a dataset to something shorter and in typical style
5. Rename a folder in notebooks
6. Also feel free to add a readme for that case
7. Fell free to enrich with links on Kandy documentation when you using or customizing some charts

devcrocod:
Added a title, description of steps, and general explanations.

Published the notebook on Datalore and included it in the README.

Some points I didn't:
3. There is data in the same folder as the notebook.
4. In my opinion, the title is appropriate, as it clearly conveys the purpose of the example. I'm not sure what style you mean.
5. The same applies to the folder name.
7. This example uses very simple charts that don't require additional explanations.

PR #1029, opened 2025-01-27
Rename to camelCase support in compiler plugin

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1025

Adds compiler plugin support for `DataFrame.renameToCamelCase()` and `DataFrame.rename {}.toCamelCase()`

Got some help from Junie :) The first function was almost entirely written by AI, the second one needed a more guiding hand from me.

See commit messages for more specifics

### DIFF
--- a/plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/rename.kt
+++ b/plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/rename.kt
@@ -1,6 +1,5 @@
 package org.jetbrains.kotlinx.dataframe.plugin.impl.api
 
-import org.jetbrains.kotlinx.dataframe.api.rename
 import org.jetbrains.kotlinx.dataframe.api.renameToCamelCase
 import org.jetbrains.kotlinx.dataframe.plugin.impl.AbstractInterpreter
 import org.jetbrains.kotlinx.dataframe.plugin.impl.AbstractSchemaModificationInterpreter
@@ -18,9 +17,8 @@ import org.jetbrains.kotlinx.dataframe.plugin.impl.toPluginDataFrameSchema
 class Rename : AbstractInterpreter<RenameClauseApproximation>() {
     private val Arguments.receiver by dataFrame()
     private val Arguments.columns: ColumnsResolver by arg()
-    override fun Arguments.interpret(): RenameClauseApproximation {
-        return RenameClauseApproximation(receiver, columns)
-    }
+
+    override fun Arguments.interpret(): RenameClauseApproximation = RenameClauseApproximation(receiver, columns)
 }
 
 class RenameClauseApproximation(val schema: PluginDataFrameSchema, val columns: ColumnsResolver)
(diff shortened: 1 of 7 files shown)


plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/rename.kt
```diff
@@ -28,55 +31,90 @@ class RenameInto : AbstractSchemaModificationInterpreter() {
...
+    val Arguments.receiver by dataFrame()
+
+    override fun Arguments.interpret(): PluginDataFrameSchema =
+        receiver.asDataFrame().renameToCamelCase().toPluginDataFrameSchema()
+}
+
+class RenameToCamelCaseClause : AbstractSchemaModificationInterpreter() {
+    val Arguments.receiver: RenameClauseApproximation by arg()
+
+    override fun Arguments.interpret(): PluginDataFrameSchema {
```
koperagen, line 111: i'd suggest convert selected columns to path and `receiver.df.asDataFrame().rename { paths }.toCamelCase().toPluginDataFrameSchema()`

plugins/kotlin-dataframe/testData/box/renameToCamelCase.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.jetbrains.kotlinx.dataframe.io.*
+
+fun box(): String {
+    // Test DataFrame.renameToCamelCase()
+    val df = dataFrameOf("first_name", "last_name", "user_age")(
+        "John", "Doe", 30
+    )
+    val renamed = df.renameToCamelCase()
+
+    // Verify extension properties
```
koperagen, line 13: this is redundant here, no point testing runtime behaviour

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/rename.kt
```diff
@@ -28,55 +31,90 @@ class RenameInto : AbstractSchemaModificationInterpreter() {
...
+        name = name,
+        columns = f(columns(), nextName, selected, path),
     )
-}
 
-internal fun SimpleFrameColumn.map(selected: ColumnsSet, path: List<String>, nextName: () -> String): SimpleFrameColumn {
-    return SimpleFrameColumn(
-        name,
-        f(columns(), nextName, selected, path)
+internal fun SimpleFrameColumn.map(
```
koperagen, line 91: Please, for easy and fast reviews let's hold changes related to linting and other non-functional changes

koperagen (changes requested):
as mentioned in the comments

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/rename.kt
Jolanrensen: I put it in a separate commit, so you can skip those if you wish

plugins/kotlin-dataframe/testData/box/renameToCamelCase.kt
Jolanrensen: not entirely, it's still worth checking that the generated property actually has the value we expect it to have. Could catch a potential bugs like types not being recognized or accessors mismatching the correct columns

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/rename.kt
Jolanrensen: and it promotes the idea that the original function solves the compiler plugin part :) which is great. I just need to become a bit more comfortable with the helper functions available to me in the compiler plugin module, but I'll get there :)

PR #1031, opened 2025-01-27
Add some AccessApiOverload annotations

koperagen:


zaleslaw (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -34,6 +34,7 @@ public interface SortDsl<out T> : ColumnsSelectionDsl<T> {
 
     public fun String.desc(): SingleColumn<Comparable<*>?> = invoke<Comparable<*>>().desc()
 
+    @AccessApiOverload
```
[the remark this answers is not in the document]
koperagen: At least those from different column selection DSLs

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataRow.kt
```diff
@@ -28,13 +28,16 @@ public interface DataRow<out T> {
 
     public operator fun <R> get(expression: RowExpression<T, R>): R = expression(this, this)
 
+    @AccessApiOverload
     public operator fun <R> get(column: ColumnReference<R>): R
```
Jolanrensen, line 32: hmm maybe not exclude these ones yet? Let's say we have a `DataRow<Any>`, calling `row["name"]` provides no way of specifying the type. You can however do `row[column<String>("name")]` to get the type. (potentially even `row["name"<String>()]`) It would also suffice if we could get a typed `row.get<String>("name")` function I believe.

Jolanrensen (changes requested):

koperagen:
> Do you actually also run a test somewhere to check whether the minimized dataframe-core still compiles? It could be that we use our public api somewhere internally
We use it, but it still compiles

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataRow.kt
koperagen: we have `typed.first().getValue<String>("name")`

Jolanrensen:
> > Do you actually also run a test somewhere to check whether the minimized dataframe-core still compiles? It could be that we use our public api somewhere internally
> 
> We use it, but it still compiles

oh right, you only change visibility, very cool :)

Jolanrensen (approved):

Issue #1028, opened 2025-01-24
Candidates for removal


koperagen:
We aim to stabilize API, but before that we need to remove a number of functions that do not bring clear value, don't seem useful. That includes shortcuts - things that can be easily replaced with other functions, or poorly designed API that can be re-introduced later, or things that don't have a use case

Here we'll keep a list of functions that can we consider for removal and want to hear from you. Please share your opinion if something should be kept as is

`DataRow.isEmpty` and `DataRow.isNotEmpty` - hard to imagine situation where it'd be used
`DataFrame.copy` is useless
`DataRow.getRows()` is a short cut that can be easily replaced with `df().get()`, and also hard to imagine a usecase

These two IMO should be renamed to `addAll` to give priority for `add("col") { }` which is a primary way to add columns
`public fun <T> DataFrame<T>.add(vararg columns: AnyBaseCol): DataFrame<T> = addAll(columns.asIterable())`
`public fun <T> DataFrame<T>.add(vararg dataFrames: AnyFrame): DataFrame<T> = addAll(dataFrames.asIterable())`

koperagen:
[org/jetbrains/kotlinx/dataframe/api/convert.kt:80](https://github.com/Kotlin/dataframe/blob/89e0d416b0127e0b49b09af5531333b774eb9731/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt#L80)

PR #1032, opened 2025-01-27
Add @CandidateForRemoval annotation

koperagen:
https://github.com/Kotlin/dataframe/issues/1028

For `DataRow.isEmpty` and `DataRow.isNotEmpty` is hard to imagine situation where it'd be used
`DataFrame.copy` is useless 
`DataRow.getRows()` is a short cut that can be easily replaced, and also hard to imagine a usecase

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/CandidateForRemoval.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/CandidateForRemoval.kt
@@ -1,0 +1,9 @@
+package org.jetbrains.kotlinx.dataframe.annotations
+
+/**
+ * Functions that can be replaced with other API, such as shortcuts without clean value,
+ * or something not properly designed, and so will be considered to be removed from API.
+ * If you see a function marked with it and think it should be kept, please let us know in the GitHub issue:
+ * https://github.com/Kotlin/dataframe/issues/1028
+ */
+internal annotation class CandidateForRemoval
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/CandidateForRemoval.kt
```diff
@@ -0,0 +1,9 @@
+package org.jetbrains.kotlinx.dataframe.annotations
+
+/**
+ * Functions that can be replaced with existing API such as shortcuts without clean value,
+ * or something not properly designed. Can be considered to be removed from API.
```
Jolanrensen, line 5: I think this should be one sentence, no?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataRowApi.kt
```diff
@@ -20,8 +21,10 @@ import kotlin.experimental.ExperimentalTypeInference
 import kotlin.reflect.KProperty
 import kotlin.reflect.KType
 
+@CandidateForRemoval
 public fun AnyRow.isEmpty(): Boolean = owner.columns().all { it[index] == null }
```
Jolanrensen, line 25: could be moved to internal maybe? it's used at least once

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataRowApi.kt
```diff
@@ -166,12 +169,16 @@ public fun AnyRow.columnNames(): List<String> = df().columnNames()
 
 public fun AnyRow.columnTypes(): List<KType> = df().columnTypes()
 
+@CandidateForRemoval
```
Jolanrensen, line 172: I think these ones should just be documented and better tested. It's in the same family of functions as `relative` which is powerful in the `add` DSL. And to populate a column with values using, for instance `getRow(0).someCol` as part of the calculation seems quite critical. Again, we just need more examples. Though, marking them as CandidateForRemoval is a good way for us to start looking into it again later :)

Jolanrensen (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/CandidateForRemoval.kt
koperagen: Rephrased it a little

PR #1034, opened 2025-01-28
Allow running Gradle tasks with -PskipKodex to publish locally faster

koperagen:
So by default we will run everything, but with this extra parameter build will ignore configuration that changes jar and triggers processing
image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -108,6 +108,10 @@ This library is built with Gradle.
   things up during development.
 * Make sure to pass the extra parameter `-Pkotlin.dataframe.debug=true` to enable debug mode. This flag will
   make sure some extra checks are run, which are important but too heavy for production.
+* The parameter `-PskipKodex` allows you to skip [kdoc processing](KDOC_PREPROCESSING.md),
+  making local publishing faster: `./gradlew publishToMavenLocal -PskipKodex`.
+  This, however, publishes the library with "broken" KDocs, 
+  so it's only meant for faster iterations during development.
 
 You can import this project into IDEA, but you have to delegate the build actions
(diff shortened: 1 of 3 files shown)


KDOC_PREPROCESSING.md
```diff
@@ -57,6 +57,8 @@ This can be seen in action in the `core:processKDocsMain` and `core:changeJarTas
 `processKDocsMain` task is executed first, which processes the KDocs in the source files and writes them to the
 `generated-sources` folder. The `changeJarTask` task then makes sure that any `Jar` task in the `core` module uses the
 `generated-sources` folder as the source directory instead of the normal `src` folder.
+It's possible to optionally skip this step, for example, when you publish the library locally during development, 
+by providing `-PskipKodex` project property: `./gradlew publishToMavenLocal -PskipKodex`
```
Jolanrensen, line 61: providing *the

CONTRIBUTING.md
```diff
@@ -108,6 +108,8 @@ This library is built with Gradle.
   things up during development.
 * Make sure to pass the extra parameter `-Pkotlin.dataframe.debug=true` to enable debug mode. This flag will
   make sure some extra checks are run, which are important but too heavy for production.
+* Parameter `-PskipKodex` allows you to skip [kdoc processing](KDOC_PREPROCESSING.md), making local publishing faster
```
Jolanrensen, line 111: *The parameter

Jolanrensen: I miss a small part here or in the other place maybe that you generally don't want to use this setting as it produces a library with "broken" KDocs; it's only for faster development if you don't care about the KDocs

Jolanrensen (approved):

Issue #761, opened 2024-07-01
`NullPointerException` using `except` to exclude double nested column


Jolanrensen:
To reproduce:

```kt
val df = dataFrameOf("a.b", "a.c.d", "d.e", "d.f")(1, 3.0, 2, "b")
    .move { all() }.into { it.name.split(".").toPath() }
```
| a | d |
| :--- | :--- |
| {"b": 1, "c": {"d": 3}} | {"e": 2, "f": "b"} |

This works:
```kt
df.select { cols(a) except a.b }
```

This breaks:
```kt
df.select { cols(a) except a.c.d }
```
```
java.lang.NullPointerException
	at org.jetbrains.kotlinx.dataframe.impl.columns.UtilsKt$allColumnsExceptKeepingStructure$2.invoke(Utils.kt:454)
	at org.jetbrains.kotlinx.dataframe.impl.columns.UtilsKt$allColumnsExceptKeepingStructure$2.invoke(Utils.kt:454)
	at org.jetbrains.kotlinx.dataframe.api.ReplaceKt.with(replace.kt:53)
	at org.jetbrains.kotlinx.dataframe.impl.columns.UtilsKt.allColumnsExceptKeepingStructure(Utils.kt:454)
	at org.jetbrains.kotlinx.dataframe.impl.columns.UtilsKt.allColumnsExceptKeepingStructure$default(Utils.kt:412)
	at org.jetbrains.kotlinx.dataframe.api.AllExceptKt$exceptInternal$1.invoke(allExcept.kt:1193)
	at org.jetbrains.kotlinx.dataframe.api.AllExceptKt$exceptInternal$1.invoke(allExcept.kt:1190)
	at org.jetbrains.kotlinx.dataframe.impl.columns.ConstructorsKt$createColumnSet$1.resolve(constructors.kt:154)
	at org.jetbrains.kotlinx.dataframe.impl.columns.UtilsKt.resolve(Utils.kt:480)
	at org.jetbrains.kotlinx.dataframe.impl.columns.ConstructorsKt$toColumnSet$1.invoke(constructors.kt:178)
	at org.jetbrains.kotlinx.dataframe.impl.columns.ConstructorsKt$toColumnSet$1.invoke(constructors.kt:175)
	at org.jetbrains.kotlinx.dataframe.impl.columns.ConstructorsKt$createColumnSet$1.resolve(constructors.kt:154)
[...]

PR #1036, opened 2025-01-28
All except simplification: Option 1

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/761

This PR rewrites `except` to disallow pointing to nested columns. The columns selection DSL should only be allowed to "select" columns, not restructure it; we have `move`, `remove`, etc. for that.

When a user now calls, say 
```kt
df.select { all() except userData.age }
```
they will receive an `IllegalArgumentException`: 
`` Cannot exclude the nested columns '[userData/age]' from the column set '[userData/{age, name}, otherCol1, otherCol2]' with the except-functions. Use the `DataFrame<*>.remove { }` operation instead. ``
because `userData.age` is not directly selected by `all()`, only its parent, `userdata`, is.

TODO: 
- [ ] Documentation website

----

Some internal struggle: It seems like `except` was used internally in a couple places and it relied on the old behavior.

[Rewriting this](https://github.com/Kotlin/dataframe/pull/1036/commits/21c6a370318081a1cc1475432bc5713836da6474) now pushes me to this horrible notation:

Before:
```kt
df.groupBy { allExcept(columnSetWithNestedCols) }...
```

After:
```kt
df.groupBy {
    (this as DataFrame<*>)
        .remove(columnSetWithNestedCols)
        .columns()
        .toColumnSet()
}...
```

(something like `df.remove {}.groupBy {}` wouldn't not work, because you still want the columns inside the group)

Having to write something like this makes me rethink what we should do about this. Clearly there's a need for _selecting the top-level columns of a version of the current dataframe where a (set of) column(s) and their parents are removed._ Though, it seems out of the scope for `except` (and the 0.15 version of `except` even [breaks](https://github.com/Kotlin/dataframe/issues/761) in some cases). We could:
- Fix the bugs in `except` and keep it as is 
  - (there are likely quite a few more we don't know about yet)
  - The ones we know can be fixed: https://github.com/Kotlin/dataframe/pull/1038
- introduce a new function that has the old behavior with a good name
- Create a way to access the original DF in the selection dsl and promote the idea that you can use functions like `remove` on it. Probably also needs a `df.toColumnSet()` function.
- Merge this as is and just block except with nested columns
  - (We only use the concept 3 times after all)
- Something else

@zaleslaw @koperagen @nikitinas @belovrv WDYT?

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -9410,6 +9410,9 @@ public abstract interface class org/jetbrains/kotlinx/dataframe/documentation/Ds
 public abstract interface class org/jetbrains/kotlinx/dataframe/documentation/DslGrammarTemplateColumnsSelectionDsl$DslGrammarTemplate$COLUMN_SET_PART {
 }
 
+public abstract interface class org/jetbrains/kotlinx/dataframe/documentation/DslGrammarTemplateColumnsSelectionDsl$DslGrammarTemplate$ColumNameDef {
+}
+
 public abstract interface class org/jetbrains/kotlinx/dataframe/documentation/DslGrammarTemplateColumnsSelectionDsl$DslGrammarTemplate$ColumnDef {
 }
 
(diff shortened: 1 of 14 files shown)


zaleslaw:
I believe that we should split our internallyUsedExcept and newOneExcept in our publicAPI

Jolanrensen:
As discussed on Slack, we might actually need the ability to select column groups with some column removed, especially in `groupBy`, so it may be best to revert this, fix the NPE, and revert/stabilize https://github.com/Kotlin/dataframe/pull/1030. Let's revisit it later

Jolanrensen:
Let's go with [Option 2](https://github.com/Kotlin/dataframe/pull/1038)

PR #1038, opened 2025-01-29
All except fix: Option 2

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/761

- Fixes the two bugs of https://github.com/Kotlin/dataframe/issues/761
- Reverts https://github.com/Kotlin/dataframe/pull/1030
- Removes `@ExperimentalExceptCsDsl` and renames `exceptNew` to `except`
- Removes the infix of those functions for clarity and simplicity
- Adds kdocs/grammar for new notation
- Adds `@AccessApiOverload` for KProperties and `@ExcludeFromSources` for doc interfaces
- Updates docs on the website


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/allExcept.kt
```diff
@@ -282,4 +311,79 @@ class AllExceptTests : ColumnsSelectionDslTests() {
...
+        listOf(
+            df.select { name }.remove { name.firstName }.alsoDebug(),
+            df.select { cols(name) except name.firstName },
+            df.select { name.except { cols { "first" in it.name } } },
+            df.select { name.except { cols { "first" in it.name } and cols { "first" in it.name } } },
+            df.select { name.except { firstName } },
+            df.select { name.except { firstNameAccessor } },
+            df.select { name.except { firstName and firstName } },
+            df.select { name.except { firstNameAccessor and firstNameAccessor } },
+//            df.select { name.except(name.firstName and name.firstName) }, // not allowed
```
zaleslaw, line 327: Looks like we need a test for this, but it's not possible in that manner and should be kind of tests for compiler plugin, when it's a String which should be compiled or not

zaleslaw: Or could we make assert on that?

Jolanrensen: yes indeed, that's a good idea for later :)

Issue #1039, opened 2025-01-29
Stabilize fast double parser


Jolanrensen:
This will be done in two steps:
- First we set the default for `useFastDoubleParser` in all parser options to `true`
- Later if we know it's stable in all cases, we can remove this parser option altogether

Jolanrensen:
I moved it, part 1 is done for beta1, we can do part 2 when we're sure it's stable enough. Doesn't even need to be beta 2 per se

Jolanrensen:
maybe we can call it stable in beta5

Jolanrensen:
Actually, I don't think it would hurt keeping the parser option. It could help DataFrame's flexibility to allow switching between different Double parsers. Let's consider it stable for now and keep the option

PR #1040, opened 2025-01-29
Stabilize FastDoubleParser part 1

Jolanrensen:
Relates to https://github.com/Kotlin/dataframe/issues/1039

- Sets `useFastDoubleParser` parser option to true by default.
- Improved `FastDoubleParser` fallback mechanism to use all possible safe alternative decimal symbols for a given locale.
- Fixed breaking tests and added some new ones
- Updated `parse` documentation on the website with the behavior of the fallback mechanism etc.

### DIFF
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ParserTests.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ParserTests.kt
@@ -1,5 +1,6 @@
 package org.jetbrains.kotlinx.dataframe.io
 
+import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.matchers.shouldBe
 import kotlinx.datetime.LocalDateTime
 import kotlinx.datetime.LocalTime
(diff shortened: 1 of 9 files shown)


docs/StardustDocs/topics/parse.md
```diff
@@ -54,8 +74,13 @@ df.parse(options = ParserOptions(locale = Locale.CHINA, dateTimeFormatter = Date
 <dataFrame src="org.jetbrains.kotlinx.dataframe.samples.api.Modify.parseWithOptions.html"/>
 <!---END-->
 
+### Global Parser Options
+
 You can also set global parser options that will be used by default in [`read`](read.md), [`convert`](convert.md),
-and `parse` operations:
+and other `parse` operations.
+These can be seen as a global fallback for the `parserOptions` argument.
```
zaleslaw, line 81: I suppose we need to be simpler with "fallback" definition and give a short explanation what is this, how it influences on something or add some references who are not familiar with that

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ParserTests.kt
```diff
@@ -145,9 +145,9 @@ class ParserTests {
         parsed.toList() shouldBe listOf(1, 2, null, 3, null, null, 4.0, 5.0)
     }
 
-    @Test // This does not yet use fastDoubleParser!
+    @Test
     fun `converting String to Double in different locales`() {
-        val currentLocale = Locale.getDefault()
+        val systemLocale = Locale.getDefault()
```
zaleslaw, line 150: is it machine dependent test?

Jolanrensen: no, this backs up the system locale and restores it at the end of the test. During the test we set `Locale.setDefault()` to be machine independant

docs/StardustDocs/topics/parse.md
Jolanrensen: alright, expanded a bit on it

Issue #1045, opened 2025-01-31
Provide toSequenceOf<SomeDataClass> similar to toListOf


martinitus:
In our application, we use the dataframe library to parse CSV and preprocess it before we transform the rows into our internal domain objects. This works really well and feels quite natural (coming from spark, pandas, polars, ...).

However, for large datasets, working with eagerly created collections sometimes seems wasteful, as we sometimes immediately chain other mappings or filters - i.e. we discard the List that gets returned from DataFrame.toListOf<K>.

AFAIK it would be way more memory efficient to use a asSequenceOf<K> in those scenarios.

Looking at the implementation of `toListOf()`:

```Kotlin
// org/jetbrains/kotlinx/dataframe/api/toList.kt
public inline fun <reified T> AnyFrame.toListOf(): List<T> = toListImpl(typeOf<T>()) as List<T>
```

and 

```Kotlin
// org/jetbrains/kotlinx/dataframe/impl/api/toList.kt
 internal fun AnyFrame.toListImpl(type: KType): List<Any> {
    ....
       return rows().map { row ->
        val parameters = convertedColumns
            .map { row[it] }
            .toTypedArray()
        constructor.call(*parameters)
    }
}
```
 
and 

```Kotlin
// org/jetbrains/kotlinx/dataframe/api/DataFrameGet.kt
public fun <T> DataFrame<T>.rows(): Iterable<DataRow<T>> = ...
```

It looks like a `asSequenceOf()` method could trivially be added?

martinitus:
I would be willing to contribute this - given my above assessment is correct and such a contribution is welcome!

PR #1046, opened 2025-01-31
generalize toListImpl to support conversions into lists and sequences

martinitus:
The previous implementation already was based on iterating over rows, with this generalization, the user can now decide whether to eagerly convert a Dataframe into a list, or use lazy transformations via a sequence.

Resolves: #1045

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -9808,8 +9808,8 @@ public final class org/jetbrains/kotlinx/dataframe/impl/api/ToDataFrameKt {
 	public static final fun createDataFrameImpl (Ljava/lang/Iterable;Lkotlin/reflect/KClass;Lkotlin/jvm/functions/Function1;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 }
 
-public final class org/jetbrains/kotlinx/dataframe/impl/api/ToListKt {
-	public static final fun toListImpl (Lorg/jetbrains/kotlinx/dataframe/DataFrame;Lkotlin/reflect/KType;)Ljava/util/List;
+public final class org/jetbrains/kotlinx/dataframe/impl/api/ToSequenceKt {
+	public static final fun toSequenceImpl (Lorg/jetbrains/kotlinx/dataframe/DataFrame;Lkotlin/reflect/KType;)Ljava/lang/Iterable;
 }
 
 public final class org/jetbrains/kotlinx/dataframe/impl/api/UpdateKt {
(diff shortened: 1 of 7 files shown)


martinitus:
I am note entirely sure, if I correctly followed the pattern of separation between interface and implementation - Not that familiar with Kotlin. Let me know if you want some stuff changed / reorganized!

koperagen:
Hi! It's a nice idea and contributions are welcome, so thank you for the interesting proposal :) I looked at the implementation and noticed that data is still converted eagerly for two reasons:
`convertedColumns` does part of the work, specifically conversion of nested lists or objects. Making it eager would be more difficult task
`AnyFrame.toSequenceImpl(type: KType): Iterable<Any>` is still not lazy, i think you'd need `return rows().asSequence().map {` to avoid converting all rows

martinitus:
Nice to hear that :-)

Oh yes - I just looked up the difference between Kotlin Iterable and Sequence -  I thought they were both lazy 🤷🏼 - Now I understand ! Added to `asSequence` where you suggested.

W.r.t. to making the inner / nested objects lazy: I don't think that would give the user a large benefit - at least not when converting into data classes which probably usually hold lists anyway: When the lazy outer conversion reaches a nested List object it has to convert that one eagerly anyway? 

That said: I could have a look into implementing that - but I am not sure if I would succeed 😁

koperagen:
> W.r.t. to making the inner / nested objects lazy: I don't think that would give the user a large benefit - at least not when converting into data classes which probably usually hold lists anyway: When the lazy outer conversion reaches a nested List object it has to convert that one eagerly anyway?
> 
> That said: I could have a look into implementing that - but I am not sure if I would succeed 😁

In fact i also suggest to maybe not improve this further for now. If your dataframes are flat, without ColumnGroup or FrameColumn, it will work as you expect with Sequence. Here's what will happen if hierarchical data is involved:

```
data class Nested(val a: String, val b: Int)

data class MySchema(val i: Int, val d: Double, val nested: Nested)

df.toSequenceOf<MySchema>().take(10).toList()
```

Entire `nested` column is recursively converted from ColumnGroup to DataColumn<Nested>, and then lazy sequence of `MySchema` is cut off at 10 items. 

Example with FrameColumn:

```
data class Nested(val a: String, val b: Int)

data class MySchema(val i: Int, val d: Double, val nested: List<Nested>)

df.toSequenceOf<MySchema>().take(10).toList()
```

Entire `nested` column is recursively converted from FrameColumn to DataColumn<List<Nested>>, and then lazy sequence of `MySchema` is cut off at 10 items.

martinitus:
Are any other changes needed? Seems like the pipeline suceeded:-) would be awesome for our project to get this into 0.16;-)

koperagen:
> Are any other changes needed? Seems like the pipeline suceeded:-) would be awesome for our project to get this into 0.16;-)

No, everything is in check. Thank you

PR #1055, opened 2025-02-10
[Compiler plugin] Support more GroupBy shortcuts

koperagen:
Now these are supported:
```
GroupBy.[minOf | maxOf]
GroupBy.[first | last | maxBy | minBy].into(columName) 
GroupBy.count 
```

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/concat.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/concat.kt
@@ -42,6 +42,15 @@ public fun <T, G> GroupBy<T, G>.concat(): DataFrame<G> = groups.concat()
 
 // endregion
 
+// region ReducedGroupBy
+
+public fun <T, G> ReducedGroupBy<T, G>.concat(): DataFrame<G> =
+    groupBy.groups.values()
+        .map { reducer(it, it) }
+        .concat()
+
+// endregion
+
 // region Iterable
 
 public fun <T> Iterable<DataFrame<T>>.concat(): DataFrame<T> = concatImpl(asList())
--- a/plugins/kotlin-dataframe/testData/box/groupBy_count.kt
+++ b/plugins/kotlin-dataframe/testData/box/groupBy_count.kt
@@ -1,0 +1,16 @@
+import org.jetbrains.kotlinx.dataframe.*
+import org.jetbrains.kotlinx.dataframe.annotations.*
+import org.jetbrains.kotlinx.dataframe.api.*
+import org.jetbrains.kotlinx.dataframe.io.*
+
+fun box(): String {
+    val df = dataFrameOf("a")(1, 1, 2, 3, 3).groupBy { a }.count()
+    val i: Int = df.count[0]
+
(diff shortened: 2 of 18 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/concat.kt
```diff
@@ -42,6 +42,15 @@ public fun <T, G> GroupBy<T, G>.concat(): DataFrame<G> = groups.concat()
 
 // endregion
 
+// region ReducedGroupBy
+
+public fun <T, G> ReducedGroupBy<T, G>.concat(): DataFrame<G> =
```
Jolanrensen, line 47: is this used anywhere? If I look for usages I cannot find any

Jolanrensen: maybe we just need a junit test, we can do it later, just make sure it's in our excel sheet in the right place after moving it here :)

plugins/kotlin-dataframe/testData/box/groupBy_count.kt
```diff
@@ -0,0 +1,13 @@
+import org.jetbrains.kotlinx.dataframe.*
+import org.jetbrains.kotlinx.dataframe.annotations.*
+import org.jetbrains.kotlinx.dataframe.api.*
+import org.jetbrains.kotlinx.dataframe.io.*
+
+fun box(): String {
+    val df = dataFrameOf("a")(1, 1, 2, 3, 3).groupBy { a }.count()
```
Jolanrensen, line 7: can you add a test for count("otherName") as well? :)

plugins/kotlin-dataframe/testData/box/groupBy_maxOfMinOf.kt
```diff
@@ -0,0 +1,16 @@
+import org.jetbrains.kotlinx.dataframe.*
+import org.jetbrains.kotlinx.dataframe.annotations.*
+import org.jetbrains.kotlinx.dataframe.api.*
+import org.jetbrains.kotlinx.dataframe.io.*
+
+fun box(): String {
+    val df = dataFrameOf("a")(1, 1, 2, 3, 3).groupBy { a }.add("id") { index() }.maxOf { 123 }
```
Jolanrensen, line 7: same here, maybe test a custom name as well, or do you think that would be overkill?

Jolanrensen (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/concat.kt
koperagen: i think it was overlooked because it was not in a good place! I will likely revisit it, because sometimes it can produce empty dataframe, need to check that with compiler plugin we'd have a reasonable exception in this case

plugins/kotlin-dataframe/testData/box/groupBy_count.kt
koperagen: sure

Issue #961, opened 2024-11-21
☂ Statistics streamlining


Jolanrensen:
Continuation of https://github.com/Kotlin/dataframe/issues/558 which fixed the most annoying bugs related to `describe`.

See https://github.com/Kotlin/dataframe/issues/558  for more information.

Our statistics functions need some more love. We used to have many missing types (mostly fixed by https://github.com/Kotlin/dataframe/pull/937), but there are yet some more inconsistencies to be solved:

> As mentioned here https://github.com/Kotlin/dataframe/issues/543, some functions like median(ints) might result in an unexpectedly rounded Int in return. It might be better to let all functions return `Double` and then handle `BigInteger` / `BigDecimal` separately for now, as they're java-specific [for now](https://youtrack.jetbrains.com/issue/KT-20912).

> ~There are plenty of public overloads on `Iterable` and `Sequence`. It's fine to have them internally, but I feel like we're clogging the public scope here. mean, for instance, is already covered in the stdlib.~

> ~We'll need to hide public functions that are not on DataColumn as @AndreiKingsley will probably make a statistics library for that anyway.~

>  We need to honor some conversion table (see below)

We won't support `UByte`, `UShort`, `UInt`, and `ULong` since they don't inherit `Number`.

We also drop support for `BigNumber` and `BigDecimal` as this makes generic typing and conversion very difficult and unpredictable.

Progress:
- [x] underlying fixes https://github.com/Kotlin/dataframe/pull/1078
- [x] mean https://github.com/Kotlin/dataframe/pull/1091
- [x] sum https://github.com/Kotlin/dataframe/pull/1103
- [x] min https://github.com/Kotlin/dataframe/pull/1108
- [x] max https://github.com/Kotlin/dataframe/pull/1108
- [x] std https://github.com/Kotlin/dataframe/pull/1119
- [x] median https://github.com/Kotlin/dataframe/pull/1122
- [x] percentile https://github.com/Kotlin/dataframe/pull/1149
- [x] cumSum https://github.com/Kotlin/dataframe/pull/1152

| Function                  | Conversion                                             | extra information                                                                               | nulls in input                                                                   |
|---------------------------|--------------------------------------------------------|-------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------|
| **mean**                  | Int -> Double                                          | **For all: Double.NaN if no elements**                                                          | **All nulls are filtered out**                                                   |
[...]

PR #1060, opened 2025-02-13
add percentile and p25 and p75 to describe

### LINKED ISSUE #1054: Add 25 and 75 percentiles in `describe`
- [ ] Unit tests
- [ ] KDocs for describe where it missed
- [ ] Update the site documentation

AndreiKingsley:
* closes #1054.
* add `percentile` function (similar to `median`)
* add `percentile` site docs
* add simple percentile unit tests
* add p25 and p75 statistics to `describe`
* add `describe` KDocs
* update `describe` site docs
* update `describe` tests


desc.ipynb
Jolanrensen, line 1: this should not be here, right?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/describe.kt
```diff
@@ -25,35 +28,132 @@ public interface ColumnDescription {
...
+ * - **`p25`** — The 25th percentile value (first quartile).
+ * - **`median`** — The median value (50th percentile / second quartile).
+ * - **`p75`** — The 75th percentile value (third quartile).
+ * - **`max`** — The maximum value in the column.
+ *
+ * For non-numeric columns, statistical metrics
+ * such as `mean` and `std` will return `null`. If a column is not [Comparable],
+ * percentile values (`min`, `p25`, `median`, `p75`, `max`) will also return `null`.
+ */
+internal interface SummaryMetrics
```
Jolanrensen, line 60: Nice :D

Jolanrensen: You can mark this documentation interface as `@ExcludeFromSources`, just like the `interface Describe` below, since their contents are only included in other places.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/describe.kt
```diff
@@ -25,35 +28,132 @@ public interface ColumnDescription {
...
+
+/**
+ * ## The Describe Operation
+ *
+ * Computes descriptive statistics for {@get COLUMNS all} columns in a given [DataFrame], including nested columns,
+ * returning a [DataFrame] with key summary metrics for each column.
+ *
+ * This function provides a statistical summary for each column, including its type, count, uniqueness,
+ * missing values, most frequent values, and statistical measures if applicable.
+ * It automatically traverses nested column groups to include all non-grouped columns in the summary.
```
Jolanrensen, line 70: I don't understand this sentence

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/describe.kt
```diff
@@ -25,35 +28,132 @@ public interface ColumnDescription {
...
+ */
 public fun <T> DataColumn<T>.describe(): DataFrame<ColumnDescription> = describeImpl(listOf(this))
 
 // endregion
 
 // region DataFrame
 
+/**
+ * {@include [Describe]}
+ * (
```
Jolanrensen, line 114: typo

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/describe.kt
```diff
@@ -25,35 +28,132 @@ public interface ColumnDescription {
...
+ * @return A [DataFrame] where each row represents the descriptive statistics of a single column in the input DataFrame.
+ *         The output contains one row per described column with the summary metrics as columns.
+ */
 public fun <T> DataFrame<T>.describe(): DataFrame<ColumnDescription> =
     describe {
         colsAtAnyDepth { !it.isColumnGroup() }
     }
 
+/**
```
Jolanrensen, line 123: very nice usage of the template :D

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/describe.kt
Jolanrensen: oh btw, probably link to `ColumnDescription` so users can click on it and explore the DataSchema result.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/percentile.kt
```diff
@@ -0,0 +1,294 @@
...
+import org.jetbrains.kotlinx.dataframe.impl.aggregation.modes.aggregateOf
+import org.jetbrains.kotlinx.dataframe.impl.aggregation.modes.of
+import org.jetbrains.kotlinx.dataframe.impl.columns.toComparableColumns
+import org.jetbrains.kotlinx.dataframe.impl.suggestIfNull
+import org.jetbrains.kotlinx.dataframe.math.percentileOrNull
+import kotlin.reflect.KProperty
+
+// region DataColumn
+
+public fun <T : Comparable<T>> DataColumn<T?>.percentile(percentile: Double): T =
```
Jolanrensen, line 27: it's okay to do the KDocs later, but don't forget to add these to the excel sheet

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/percentile.kt
```diff
@@ -0,0 +1,294 @@
...
+    noinline expression: (T) -> R?,
+): R = percentileOfOrNull(percentile, expression).suggestIfNull("percentileOf")
+
+// endregion
+
+// region DataRow
+
+public fun AnyRow.rowPercentileOrNull(percentile: Double): Any? =
+    Aggregators.percentile(percentile).aggregate(
+        values().filterIsInstance<Comparable<Any?>>().toValueColumn(),
```
Jolanrensen, line 49: This will break when you have a `String` and `Number` column for instance. While they both are `Comparable`, they are not comparable to each other. I see we make the same mistake in `rowMin` etc. (it's [noted](https://github.com/Kotlin/dataframe/issues/961#issuecomment-2659202472)). It's probably best to filter for `Number` values, like `rowMean`.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/math/median.kt
```diff
@@ -13,77 +10,4 @@ public inline fun <reified T : Comparable<T>> Iterable<T>.median(): T = medianOr
...
-                list = greater
-                greater = temp
-                temp = list
-                k -= less.size + equal
-            }
-        }
-    }
-    return list[0]
-}
+internal inline fun <reified T : Comparable<T>> Iterable<T?>.median(type: KType): T? = percentile(50.0, type)
```
Jolanrensen, line 13: nice

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/math/percentile.kt
```diff
@@ -0,0 +1,138 @@
+package org.jetbrains.kotlinx.dataframe.math
+
+import org.jetbrains.kotlinx.dataframe.impl.asList
+import java.math.BigDecimal
+import java.math.BigInteger
+import kotlin.reflect.KType
+import kotlin.reflect.typeOf
+
+public inline fun <reified T : Comparable<T>> Iterable<T>.percentileOrNull(percentile: Double): T? =
```
Jolanrensen, line 9: I don't think DataFrame should have these functions public. It pollutes the scope on Iterables and it's outside the scope of DataFrame to offer these to users. They should be part of a statistics library if you want them publicly accessible :). (mentioned here https://github.com/Kotlin/dataframe/issues/961)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/math/percentile.kt
```diff
@@ -0,0 +1,138 @@
...
+public inline fun <reified T : Comparable<T>> Iterable<T>.q3(): T = q3OrNull()!!
+
+@PublishedApi
+internal inline fun <reified T : Comparable<T>> Iterable<T?>.q1(type: KType): T? = percentile(25.0, type)
+
+@PublishedApi
+internal inline fun <reified T : Comparable<T>> Iterable<T?>.q3(type: KType): T? = percentile(75.0, type)
+
+@PublishedApi
+internal inline fun <reified T : Comparable<T>> Iterable<T?>.percentile(percentile: Double, type: KType): T? {
```
Jolanrensen, line 30: It's fine for now to return the same type it receives, as that's what median did too. However, I'll likely change it pretty soon according to this table: https://github.com/Kotlin/dataframe/issues/961

docs/StardustDocs/topics/describe.md
```diff
@@ -10,21 +10,27 @@ describe [ columns ]
...
+- **`mean`** — The arithmetic mean (only for numeric columns).
+- **`std`** — The standard deviation (only for numeric columns).
+- **`min`** — The minimum value in the column.
+- **`p25`** — The 25th percentile value (first quartile).
+- **`median`** — The median value (50th percentile / second quartile).
+- **`p75`** — The 75th percentile value (third quartile).
+- **`max`** — The maximum value in the column.
+
+For non-numeric columns, statistical metrics
+such as `mean` and `std` will return `null`. If a column is not `Comparable`,
```
Jolanrensen, line 32: maybe specify it like "if the values in a column are incomparable" which is closer to the truth. A column can be `Comparable<Nothing>` but then these statistics will fail.

Jolanrensen (changes requested):
Thanks a lot! good addition :)

desc.ipynb
AndreiKingsley: yep)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/describe.kt
AndreiKingsley: where?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/percentile.kt
AndreiKingsley: Let's add an unified solution for all these functions

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/math/percentile.kt
AndreiKingsley: yeah, that's true. Let's get rid of all of them In the other PR?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/math/percentile.kt
AndreiKingsley: Yes, we could reconsider this logic, I add all of this in accordance with  existing  `median`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/percentile.kt
Jolanrensen: sure! I'd probably go with a route like this, similar to `describeImpl` with `convertToComparableOrNull`:
- First filter for all `Comparable<*>` values in the row
- If all these values are comparable to each other, run the statistic function on it
- If not, filter for `Number` values. These are not directly comparable to each other but can be made comparable by converting them to Double/BigDecimal, and thus all statistics functions will work. Though, then we skip any other potentially comparable values, like `String`.
- Else return `null`

I think this would produce the most expected behavior. wdyt?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/statistics/percentile.kt
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    @Test
+    fun `row percentile`() {
+        val df = dataFrameOf("a", "b", "c")(
+            1, 3, 3,
+            2, 4, 10,
+            7, 7, 1,
+        )
+        df.mapToColumn("", Infer.Type) { it.rowPercentile(25.0) } shouldBe columnOf(1, 2, 1)
```
Jolanrensen, line 32: Btw. Do you think it would be a good idea to offer an enum? Like `Percentile.Q1`, `Percentile.Q2`, etc.

Jolanrensen: or probably an `object` with `val Q1 = 0.25`, etc.

AndreiKingsley: Well, it could be a good idea, but for statistics library :). I believe simple percentile is enough for now.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/percentile.kt
AndreiKingsley: Looks nice! But let's do it in a new PR for all statistical functions.

desc.ipynb
AndreiKingsley: done

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/describe.kt
AndreiKingsley: done

docs/StardustDocs/topics/percentile.md
```diff
@@ -0,0 +1,32 @@
...
+df.percentileOf(25.0) { (weight ?: 0) / age } // percentile of expression evaluated for every row
+```
+
+<!---END-->
+
+<!---FUN percentileAggregations-->
+
+```kotlin
+df.percentile(25.0)
+df.age.percentile(25.0)
```
zaleslaw, line 24: is it possible to embed a table here, like in valueColumns https://kotlin.github.io/dataframe/valuecounts.html

Other statistics are poorly covered with examples, should be improved

zaleslaw (approved):

docs/StardustDocs/topics/percentile.md
AndreiKingsley: We could review all statistical functions documentation.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/describe.kt
Jolanrensen: there's a "(" that doesn't do anything

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/percentile.kt
Jolanrensen: sure! Then I'd make this `Number` for now until we fix it for all

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/statistics/percentile.kt
Jolanrensen: true

Jolanrensen (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/percentile.kt
AndreiKingsley: done

PR #1063, opened 2025-02-17
Added KDocs for cumSum operation with Kodex

### LINKED ISSUE #1062: Add KDocs for `cumSum` with KoDEx

zaleslaw:
Documentation and overloads are added for DataColumn, DataFrame, and GroupBy, ensuring consistency across API components.
Fixes #1062

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt
@@ -5,7 +5,10 @@ import org.jetbrains.kotlinx.dataframe.ColumnsSelector
 import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.DataFrame
 import org.jetbrains.kotlinx.dataframe.annotations.AccessApiOverload
+import org.jetbrains.kotlinx.dataframe.api.Select.SelectSelectingOptions
 import org.jetbrains.kotlinx.dataframe.columns.toColumnSet
+import org.jetbrains.kotlinx.dataframe.documentation.DocumentationUrls
+import org.jetbrains.kotlinx.dataframe.documentation.ExcludeFromSources
 import org.jetbrains.kotlinx.dataframe.impl.nothingType
 import org.jetbrains.kotlinx.dataframe.impl.nullableNothingType
 import org.jetbrains.kotlinx.dataframe.math.cumSum
@@ -18,6 +21,39 @@ import kotlin.reflect.typeOf
 
 // region DataColumn
 
+/**
+ * ## The CumSum Operation
+ *
+ * Computes the cumulative sums of the values in the each column from the {@get [CumSumDocs.DATA_TYPE]}
+ * from the first row to the last row.
+ *
+ * __NOTE:__ If the column contains nullable values and `skipNA` is set to `true`,
+ * null values are skipped when computing the cumulative sum.
+ * Otherwise, any null value encountered will propagate null values in the output from that point onward.
+ *
+ * {@get [CumSumDocs.CUMSUM_PARAM] @param [columns\]
+ * The names of the columns to apply cumSum operation.}
+ *
+ * @param [skipNA\] Whether to skip null values (default: `true`).
+ *
+ * @return A new {@get [CumSumDocs.DATA_TYPE]} of the same type with the cumulative sums.
+ *
+ * {@get [CumSumDocs.CUMSUM_PARAM] @see [Selecting Columns][SelectSelectingOptions].}
+ * @see {@include [DocumentationUrls.CumSum]}
+ */
+@ExcludeFromSources
+@Suppress("ClassName")
+private interface CumSumDocs {
+    interface CUMSUM_PARAM
+
+    interface DATA_TYPE
+}
+
+/**
+ * {@include [CumSumDocs]}
+ * {@set [CumSumDocs.DATA_TYPE] [DataColumn]}
+ * {@set [CumSumDocs.CUMSUM_PARAM]}
+ */
 public fun <T : Number?> DataColumn<T>.cumSum(skipNA: Boolean = defaultCumSumSkipNA): DataColumn<T> =
     when (type()) {
         typeOf<Double>() -> cast<Double>().cumSum(skipNA).cast()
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt
```diff
@@ -18,6 +21,36 @@ import kotlin.reflect.typeOf
...
+ *
+ * @return A new {@get DATA_TYPE} of the same type with the cumulative sum of the values.
+ *
+ * {@get [CUMSUM_PARAM] @see [Selecting Columns][SelectSelectingOptions].}
+ * @see {@include [DocumentationUrls.CumSum]}
+ */
+@ExcludeFromSources
+@Suppress("ClassName")
+private interface CumSumDocs {
+    interface CUMSUM_PARAM
```
Jolanrensen, line 46: It's a bit odd you created an argument interface for `CUMSUM_PARAM` but not for `DATA_TYPE`. Best to probably also add an interface for that for better refactorability :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt
```diff
@@ -18,6 +21,36 @@ import kotlin.reflect.typeOf
...
+}
+
+/**
+ * {@include [CumSumDocs]}
+ * {@set DATA_TYPE [DataColumn]}.
```
Jolanrensen, line 51: Why is there a period after the `{@set ...}`?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt
```diff
@@ -18,6 +21,36 @@ import kotlin.reflect.typeOf
...
+ * Otherwise, any null value encountered will propagate null values in the output from that point onward.
+ *
+ * {@get [CUMSUM_PARAM] @param [columns]
+ * The names of the columns to apply cumSum operation.}
+ *
+ * @param [skipNA] Whether to skip null values (default: `true`).
+ *
+ * @return A new {@get DATA_TYPE} of the same type with the cumulative sum of the values.
+ *
+ * {@get [CUMSUM_PARAM] @see [Selecting Columns][SelectSelectingOptions].}
```
Jolanrensen, line 40: I see you used `CUMSUM_PARAM` as a 2-piece template you can "turn off" by setting it to nothing later. Interesting! I used this strategy in the grammar DSLs before. There I named these interfaces something like `SELECTING_COLUMNS_PART` and I added some `// comment` to the interface about how to treat this argument. Maybe it's useful to keep using this naming convention.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt
```diff
@@ -18,6 +21,36 @@ import kotlin.reflect.typeOf
 
 // region DataColumn
 
+/**
+ * ## The CumSum Operation
+ *
+ * Computes the cumulative sum of the values in the {@get DATA_TYPE}.
```
Jolanrensen, line 27: The way it's written now suggests we get a single value result, while we actually we get a new column containing all cumulative sum values (or a df with new columns containing...)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt
```diff
@@ -18,6 +21,36 @@ import kotlin.reflect.typeOf
...
+ *
+ * __NOTE:__ If the column contains nullable values and `skipNA` is set to `true`,
```
Jolanrensen, line 29: I'd write "`` `skipNA` ``" like "`[skipNA\]`" so users can click on it in the kdocs :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt
```diff
@@ -18,6 +21,36 @@ import kotlin.reflect.typeOf
...
+ * skips null values when computing the cumulative sum.
```
Jolanrensen, line 30: *null values are skipped

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt
```diff
@@ -79,25 +112,46 @@ private val supportedClasses = setOf(
 
 // region DataFrame
 
+/**
+ * {@include [CumSumDocs]}
+ * {@set DATA_TYPE [DataFrame]}.
```
Jolanrensen, line 117: Unclear what "the values in the DataFrame" means. It's run on each column individually, and we get a new df with the new cumulative sum columns

Jolanrensen (changes requested):
see comments :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt
```diff
@@ -18,6 +21,39 @@ import kotlin.reflect.typeOf
 
 // region DataColumn
 
+/**
+ * ## The CumSum Operation
+ *
+ * Computes the cumulative sums of the values in the each column from the {@get [CumSumDocs.DATA_TYPE]}
+ * from the first row to the last row.
```
Jolanrensen, line 28: @zaleslaw when adding this doc to `DataColumn.cumSum` you're now talking about rows inside a data column. This is not entirely wrong, I suppose, but it looks a bit weird. Usually we're talking about rows (across multiple columns) and columns (across multiple rows). A single value in a column is either a "cell" or a "value", not a "row". right?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt
```diff
@@ -18,6 +21,39 @@ import kotlin.reflect.typeOf
 
 // region DataColumn
 
+/**
+ * ## The CumSum Operation
+ *
+ * Computes the cumulative sums of the values in the each column from the {@get [CumSumDocs.DATA_TYPE]}
```
Jolanrensen, line 27: @zaleslaw "in the each column" is not correct english

Issue #1068, opened 2025-02-19
`rowSum()` breaks for `Int` + `Float`


Jolanrensen:
```kt
dataFrameOf("a", "b")(1, 2f)[0].rowSum()
```
results in `java.lang.IllegalStateException: Can not find common number type for class kotlin.Int and class kotlin.Float`

should be considered in https://github.com/Kotlin/dataframe/issues/961

Issue #557, opened 2024-01-12
Handling of `Number` types can be unexpected


Jolanrensen:
Currently, numbers are inferred similar to type inference in Kotlin.

For instance:
```kt
DataFrame.readJsonStr("""[ { "a": 1 }, { "a": 1.0 } ]""")
```
results in a column: `a: Number`. This can cause issues with statistics like `std`, as they don't support `Number`s as of now: https://github.com/Kotlin/dataframe/issues/558
It might be easier to work with if numbers would converge in the closest common ancestor in terms of information. For instance: `Int + Double + Float => Double`, since `Double`s can store all information a `Double` and `Int` holds.

This would need to be carefully considered, however, as it's inconsistent with Kotlin's type inference.

Jolanrensen:
A system like https://github.com/Kotlin/dataframe/pull/1070 should help

PR #1070, opened 2025-02-19
Unified number types

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1068

Fixed `getCommonNumberType` and `commonNumberClass` functions that are used solely by `sum` at the moment.

I introduced a future-proof rewrite of the function and added support for unsigned- and big numbers. Moved it to a separate file and added tests. We will reuse this logic in more places later. I created a small DAG implementation for this, as it's smaller than yet another dependency.

```
           BigDecimal
           /       \
     BigInteger     \
       /    \        \
   ULong   Long    Double
..   |    /   |   /   |  \..
  \  |   /    |  /    |
    UInt     Int    Float
..   |    /   |   /      \..
  \  |   /    |  /
   UShort   Short
     |    /   |
     |   /    |
   UByte     Byte
```
The idea of this is that numbers can be converted lossless to a higher number type, so providing, say `UInt` and `Float` can be auto-converted to `Double` at runtime safely. The only place it's currently done is when collecting numbers across multiple columns and summing them, but I intend to reuse this logic in other statistical functions when needed, in parsing, or [in JSON](https://github.com/Kotlin/dataframe/issues/557).

`NumbersAggregator` now converts numbers in its input to a common number type first, before aggregating, not relying on smart-casts anymore. 
To avoid heavy reflection calls, types can be supplied to aggregateMixed() if you're aware of them.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/NumberTypeUtils.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/NumberTypeUtils.kt
@@ -1,0 +1,143 @@
+package org.jetbrains.kotlinx.dataframe.impl
+
+import org.jetbrains.kotlinx.dataframe.impl.api.createConverter
+import org.jetbrains.kotlinx.dataframe.impl.commonNumberType
+import java.math.BigDecimal
+import java.math.BigInteger
+import kotlin.reflect.KClass
+import kotlin.reflect.KType
+import kotlin.reflect.full.withNullability
+import kotlin.reflect.typeOf
+
+/**
+ * Number type graph, structured in terms of number complexity.
+ * A number can always be expressed lossless by a number of a more complex type (any of its parents).
+ *
+ * ```
+ *         BigDecimal
+ *         /      \\
+ *   BigInteger    |
+ *         |       |
+ *       ULong     |
+ *         |       |
+ *       Long    Double
+ *          \\   /    |
+ *           UInt    |
+ *            |      |
+ *           Int   Float
+ *            |   /
+ *           UShort
+ *             |
+ *           Short
+ *             |
+ *           UByte
+ *            |
+ *           Byte
+ * ```
+ *
+ * For any two numbers, we can find the nearest common ancestor in this graph
+ * by calling [DirectedAcyclicGraph.findNearestCommonVertex].
+ * @see getCommonNumberClass
+ * @see commonNumberClass
+ */
+internal val numberTypeGraph: DirectedAcyclicGraph<KType> by lazy {
+    dagOf(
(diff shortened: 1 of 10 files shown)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/types/UtilTests.kt
```diff
@@ -414,4 +417,44 @@ class UtilTests {
             typeOf<Set<Nothing>?>(),
         ).commonTypeListifyValues() shouldBe typeOf<Collection<out Nothing?>?>()
     }
+
+    @Test
+    fun `common number types`() {
+        // Same type
+        getCommonNumberClass(Int::class, Int::class) shouldBe Int::class
```
zaleslaw, line 424: These tests are great, could you please add a diagramm from your PR here to extract the context about the DAG of type conversion

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/statistics/sum.kt
```diff
@@ -61,4 +63,24 @@ class SumTests {
         df.sum { value2 } shouldBe expected2
         df.sum { value3 } shouldBe expected3
     }
+
+    // Issue #1068
```
zaleslaw, line 67: add direct link please

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/NumberTypeUtils.kt
```diff
@@ -0,0 +1,143 @@
...
+import kotlin.reflect.KType
+import kotlin.reflect.full.withNullability
+import kotlin.reflect.typeOf
+
+/**
+ * Number type graph, structured in terms of number complexity.
+ * A number can always be expressed lossless by a number of a more complex type (any of its parents).
+ *
+ * ```
+ *         BigDecimal
```
zaleslaw, line 17: Ok, found the diagramm here, probably refer to this or just copy-paste to the tests

zaleslaw (approved):
Really intersting innovation and extensible, thanks for graph theory puzzle here!

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/types/UtilTests.kt
Jolanrensen: I'll put it somewhere in /documentation so we can include it with KoDEx :)

Jolanrensen:
I'll rename common number types -> unified number type before merging, because I think that better reflects the idea. A "common type" of `Int` and `Double` could be interpreted as using Kotlin's type system, so then it'd be `Number`... However, a "unified type" sounds like a new concept, which is what we do here

PR #1073, opened 2025-02-21
JSON reading: unified numbers

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/557

todo.


Jolanrensen:
> @Jolanrensen may be add `unifyNumbers` to `readJson`? I had an experience with APIs that returns JSON with different numeric types inside one column 😒 (and it was important).

definitely, that sounds like a good idea :)

AndreiKingsley (approved):

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
```diff
@@ -383,7 +410,7 @@ class JsonTests {
         ).alsoDebug("df:")
 
         val res = DataFrame.readJsonStr(df.toJson()).alsoDebug("res:")
-        res shouldBe df
+        res shouldBe df.convert { colsOf<Double?>() }.toFloat()
```
zaleslaw, line 413: I see that we are reducing Double -> Float what is possible, but why not Float -> Double as a most common type?

zaleslaw (approved):

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
Jolanrensen: There's no number unification happening in this test at all; in fact, we're not reducing Double to Float on purpose, however because we're writing a `Double` "1.0" and "3.0" to JSON it can be read back from JSON as `Float`. If any of the numbers were too large to fit in a `Float`,  or if the column contained both floats and integers, the result would have been `Double`.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/readJson.kt
```diff
@@ -229,18 +239,18 @@ internal fun fromJsonListAnyColumns(
 
                             v.longOrNull != null -> collector.add(v.long)
 
-                            v.doubleOrNull != null -> collector.add(v.double)
-
                             v.floatOrNull != null -> collector.add(v.float)
 
+                            v.doubleOrNull != null -> collector.add(v.double)
```
Jolanrensen, line 244: @zaleslaw It's because of this change that we can now get floats out of json too instead of just doubles.

PR #1077, opened 2025-02-25
Add Compiler Plugin support for statistics on GroupBy

zaleslaw:
PR contains:

1. test scenarious for sum*/min/max/std/mean/median family on GroupBy
2. ported test scenarious for Compiler Plugin
3. Changes in Compiler Plugin

Tests for mean/std with results in Double doesn't work well (regarding incorrect resulting types), but please review the logic and  keep test muted for a while


plugins/kotlin-dataframe/testData/box/groupBy_sum.kt
```diff
@@ -0,0 +1,54 @@
...
+    )
+
+    // scenario #0: all numerical columns
+    val res0 = personsDf.groupBy { city }.sum()
+    val sum01: Int? = res0.age[0]
+    // TODO: Compilation error - actual type it(kotlin.Number & kotlin.Comparable<*>)
+    // `val sum02: Double? = res0.weight[0]
+
+    // scenario #1: particular column
+    val res1 = personsDf.groupBy { city }.sumFor { age }
```
koperagen, line 28: sumFor, maxFor need support too because now "age" column comes from T
```
public fun <T, C : Number> Grouped<T>.sumFor(columns: ColumnsForAggregateSelector<T, C?>): DataFrame<T> =
    Aggregators.sum.aggregateFor(this, columns)
```
While there's `age`, problem is that there's also column `height` in compile time schema, but there's no such column in runtime. In tests i have this utility to compare if compile time schema is compatible with runtime:
"res1.compareSchemas()"

plugins/kotlin-dataframe/testData/box/groupBy_sum.kt
```diff
@@ -0,0 +1,75 @@
...
+        "Hank", 60, "Paris", 80.2, "1.75", 5,
+        "Isla", 22, "London", 75.1, "1.85", 43,
+    )
+
+    // scenario #0: all numerical columns
+    val res0 = personsDf.groupBy { city }.sum()
+    val sum01: Int? = res0.age[0]
+    // TODO: Compilation error - actual type it(kotlin.Number & kotlin.Comparable<*>)
+    // `val sum02: Double? = res0.weight[0]
+    res0.compareSchemas()
```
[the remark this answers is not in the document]
koperagen: yes, it will check that compile time schema is compatible with runtime schema, so 
1. there are no columns in compile time that don't exist in runtime
2. all column types are either equal to runtime, or their subtype. Do compile `col: Any` when runtime is `col: Int` is ok. compile `col: Any` when runtime `col: Any?` is not ok

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/groupBy.kt
```diff
@@ -199,4 +205,195 @@ abstract class GroupByAggregator(val defaultName: String) : AbstractSchemaModifi
...
+) : FirImplicitBuiltinTypeRef(source, StandardClassIds.Comparable) */
+
+abstract class GroupByAggregatorComparable() : AbstractSchemaModificationInterpreter() {
+    val Arguments.receiver by groupBy()
+
+    override fun Arguments.interpret(): PluginDataFrameSchema {
+        val resolvedColumns = receiver.groups.columns()
+            .filter {
+                it is SimpleDataColumn
+                    && it.type.type.isSubtypeOfComparable(session)
```
zaleslaw, line 345: @koperagen here, I am trying to check if types if comparable to emulate the interComparable() internal function behaviour for median/min/max functions

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/groupBy.kt
```diff
@@ -199,4 +205,195 @@ abstract class GroupByAggregator(val defaultName: String) : AbstractSchemaModifi
...
+        if (name == null) { // TODO: add an example, should be double type
+            val resolvedColumns = columns?.resolve(receiver.keys)?.map { it.column }!!.toList()
+            return PluginDataFrameSchema(receiver.keys.columns() + resolvedColumns)
+        } else {
+            val aggregated =
+                makeNullable(
+                    simpleColumnOf(
+                        name ?: defaultName,
+                        session.builtinTypes.doubleType.type
+                    )
```
zaleslaw, line 367: here, I need to create it with type Comparable<Any?>, I don't know is it a good idea or not, but it could be solved only if we updated a sealed hiearchy in Kotlin with `FirImplicitComparableTypeRef` with wrap for class ```
FirImplicitComparableTypeRef(
    source: KtSourceElement?
) : FirImplicitBuiltinTypeRef(source, StandardClassIds.Comparable)
```

File: plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/groupBy.kt (hunk 3 above)
Comment (koperagen) [reply plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/groupBy.kt:350]: valuesAreComparable should be impemented like this, to follow `public fun <T : Comparable<T>> Iterable<T>.maxOrNull(): T?` signature. 
```
public fun AnyCol.valuesAreComparable(): Boolean {
    val other = Comparable::class.createType(listOf(KTypeProjection(KVariance.IN, type)))
    return type.isSubtypeOf(other)
}
```

Now it's possible to break it with this code:
```
val i = 1
    val i1: Comparable<Int> = object : Comparable<Int> {
        override fun compareTo(other: Int): Int {
            return other
        }
    }
    val col by columnOf(i, i1)
    dataFrameOf(col).max().print()
```
Because indeed, Comparable<Int> does not implement Comparable<Comparable<Int>>
It's not possible to call `listOf(i, i1).maxOrNull()`

Comment (koperagen) [reply plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/groupBy.kt:350]: So for plugin:
```
fun isInterComparable(col: SimpleDataColumn): Boolean {
      val comparable = StandardClassIds.Comparable.constructClassLikeType(arrayOf(col.type.type))
      return col.type.type.isSubtypeOf(comparable, session)
  }
```

**Comment by zaleslaw** [comment]
@Jolanrensen @koperagen please have a look at the supported scenarious, I want to merge it before https://github.com/Kotlin/dataframe/pull/1078 changes

File: plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/groupBy.kt (hunk 3 above)
Comment (Jolanrensen) [reply plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/groupBy.kt:405]: @koperagen I created a branch with this suggested fix, but it doesn't produce the expected results either https://github.com/Kotlin/dataframe/tree/values-are-comparable-fix

Comment (Jolanrensen) [reply plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/groupBy.kt:405]: @koperagen actually, looking better at it I think you're right and your fix is too. There's some tests that need to be rewritten.

Comment (Jolanrensen) [reply plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/groupBy.kt:405]: In theory, this should be the plugin version then:
```kt
fun isIntraComparable(col: SimpleDataColumn): Boolean {
    val comparable = StandardClassIds.Comparable.constructClassLikeType(
        typeArguments = arrayOf(col.type.type.withNullability(ConeNullability.NOT_NULL, session.typeContext)),
        isNullable = col.type.type.isNullable,
    )
    return col.type.type.isSubtypeOf(comparable, session)
}
```
(didn't test this yet)

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/groupBy.kt
```diff
@@ -199,4 +210,217 @@ abstract class GroupByAggregator(val defaultName: String) : AbstractSchemaModifi
...
+/** Implementation for `median` */
+class GroupByMax0 : GroupByAggregator3(defaultName = "max")
+
+abstract class GroupByAggregatorMean(val defaultName: String) : AbstractSchemaModificationInterpreter() {
+    val Arguments.receiver by groupBy()
+    val Arguments.name: String? by arg(defaultValue = Present(null))
+    val Arguments.columns: ColumnsResolver? by arg()
+
+    override fun Arguments.interpret(): PluginDataFrameSchema {
+        if (name == null) {
```
koperagen, line 323: Missing a test case here:
`personsDf.groupBy { city }.mean { age and yearsToRetirement }.compareSchemas()`

Runtime:
city: String
mean: Double

Compile:
age: Double
yearsToRetirement: Double
city: String

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/groupBy.kt
```diff
@@ -199,4 +210,217 @@ abstract class GroupByAggregator(val defaultName: String) : AbstractSchemaModifi
...
+
+abstract class GroupByAggregatorExpressionComparable(val defaultName: String) :
+    AbstractSchemaModificationInterpreter() {
+    val Arguments.receiver by groupBy()
+    val Arguments.name: String? by arg(defaultValue = Present(null))
+    val Arguments.expression by type()
+
+    override fun Arguments.interpret(): PluginDataFrameSchema {
+
+        val aggregated = makeNullable(simpleColumnOf(name ?: defaultName, createComparableType(session)))
```
koperagen, line 239: Looks like it needs to be updated after Jolan's PR
Now runtime type is Comparable<Any?> which seems wrong, because all values are actually Int. We discussed what result type we want to median

Jolanrensen (changes requested):

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/statistics.kt
```diff
@@ -0,0 +1,445 @@
+package org.jetbrains.kotlinx.dataframe.api
+
+import io.kotest.matchers.shouldBe
+import org.junit.Test
+
+@Suppress("ktlint:standard:argument-list-wrapping")
+class StatisticsTests {
```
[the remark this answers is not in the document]
Jolanrensen: oh I see, it's reused in the compiler plugin

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/groupBy.kt
zaleslaw: It seems that we agreed that support of multiple columns is under the question?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/mean.kt
```diff
@@ -98,9 +100,12 @@ public inline fun <T, reified D : Number> DataFrame<T>.meanOf(
 // endregion
 
 // region GroupBy
-
+@Refine
+@Interpretable("GroupByMean1")
```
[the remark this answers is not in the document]
koperagen: I suggest `GroupByMeanDefault` for `GroupBy.mean()`, `GroupByMeanFor` for `GroupBy.meanFor` and `GroupByMean` for` mean`

koperagen (approved):

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/groupBy.kt
Jolanrensen: True, at least until after the statistics types are set in stone. That said, for `mean`, the result will always be `Double` :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/mean.kt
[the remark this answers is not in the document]
zaleslaw: Ok, agree, will do some renaming/refactoring in next PR in that area

Jolanrensen:
Merged a failing build

PR #1078, opened 2025-02-26
Aggregator implementation rework

Jolanrensen:
- Aggregator improvements
  - Refactoring, renaming, documenting
  - Number aggregator now unifies numbers properly
  - Added `UnifiedNumberTypeOptions` such that aggregators can choose to not work with big numbers
  - Fixed (potential) bugs
    - like nulls not always being filtered out of columns
    - ~`preservesTypes` not always being honored~
    - `preservesTypes` removed since it can use `calculateReturnTypeOrNull`
  - Performance improvements by reducing runtime reflection calls
    - This is done partly by the `calculateReturnTypeOrNull` function that needs to be supplied for each aggregator. Often it's as simple as `{ _, _ -> typeOf<Double>() }` but it can reduce runtime value checks equal to the number of columns. These functions might be reused by the compiler plugin btw ;)
    - The other function that's improved is ~`aggregateMixed`~ `aggregateCalculatingType`. Often used in row-functions. It's unnecessary to do runtime reflection since we know the type of each cell in the row after all.
  
- Renamed `interComparableColumns` to `intra-`, because that's what it actually does. ("inter" would mean "in between columns", "intra" means "inside a column")


dataframe-csv/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readTsv.kt
```diff
@@ -367,7 +369,8 @@ public fun DataFrame.Companion.readTsv(
  *
  * [DataFrame.readTsvStr][readTsvStr]`("a,b,c", delimiter = ",")`
  *
- * @param url The URL from which to fetch the data. Can also be compressed as `.gz` or `.zip`, see [Compression].
+ * @param url The URL from which to fetch the data.
```
zaleslaw, line 372: Could you please convert `@param url `to `@param [url]` ? it's just easier to read

dataframe-csv/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/DelimParams.kt
```diff
@@ -17,16 +17,28 @@ import org.jetbrains.kotlinx.dataframe.io.QuoteMode
...
+    /**
+     * @param file The file to read.
+     *   Can also be compressed as `.gz` or `.zip`, see [Compression][org.jetbrains.kotlinx.dataframe.io.Compression].
+     */
     interface FILE_READ
 
-    /** @param url The URL from which to fetch the data. Can also be compressed as `.gz` or `.zip`, see [Compression]. */
+    /**
+     * @param url The URL from which to fetch the data.
+     *   Can also be compressed as `.gz` or `.zip`, see [Compression][org.jetbrains.kotlinx.dataframe.io.Compression].
```
zaleslaw, line 34: A good candidate for https://kotlinlang.org/docs/kotlin-doc.html#see-identifier

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/statistics.kt
```diff
@@ -362,12 +364,12 @@ class StatisticsTests {
     fun `max on GroupBy`() {
         // scenario #0: all numerical columns
         val res0 = personsDf.groupBy("city").max()
-        res0.columnNames() shouldBe listOf("city", "name", "age", "height", "yearsToRetirement") // TODO: DOUBLE weight?
+        res0.columnNames() shouldBe listOf("city", "name", "age", "weight", "height", "yearsToRetirement")
```
zaleslaw, line 367: Thanks for fixing

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/math/sum.kt
```diff
@@ -95,6 +97,11 @@ internal fun <T : Number> Iterable<T?>.sum(type: KType): T =
         else -> throw IllegalArgumentException("sum is not supported for $type")
     }
 
+// T: Number? -> T
```
zaleslaw, line 100: `//` -> `/** */ `?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/math/std.kt
```diff
@@ -35,6 +37,11 @@ internal fun <T : Number> Iterable<T?>.std(
     }
 }
 
+// T: Number? -> Double
```
zaleslaw, line 40: `//` -> `/** */ `?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/math/mean.kt
```diff
@@ -43,6 +45,11 @@ internal fun <T : Number> Sequence<T>.mean(type: KType, skipNA: Boolean = skipNA
     }
 }
 
+// T: Number? -> Double
```
zaleslaw, line 48: `//` -> `/** */ `?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/modes/ofRowExpression.kt
```diff
@@ -75,7 +75,8 @@ internal inline fun <T, reified C, reified R> Grouped<T>.aggregateOf(
     val type = typeOf<R>()
     return aggregateInternal {
         val value = aggregator.aggregateOf(df, expression)
-        yield(path, value, type, null, false)
+        val inferType = !aggregator.preservesType
```
zaleslaw, line 78: sorry, could not parse the logic here, could you please explain or add a comment in the code

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/TwoStepNumbersAggregator.kt
```diff
@@ -0,0 +1,167 @@
...
+     *   If `null`, the types of [values] will be calculated at runtime (heavy!).
+     */
+    @Suppress("UNCHECKED_CAST")
+    override fun aggregateCalculatingType(values: Iterable<Number>, valueTypes: Set<KType>?): Return? {
+        val valueTypes = valueTypes ?: values.types()
+        val commonType = valueTypes
+            .unifiedNumberType(PRIMITIVES_ONLY)
+            .withNullability(false)
+
+        if (commonType == typeOf<Double>() && (typeOf<ULong>() in valueTypes || typeOf<Long>() in valueTypes)) {
```
zaleslaw, line 132: For Float and Long, the precision loss can still occur but to a lesser extent. 

Float has a lower precision compared to Double, and converting Long to Float can also result in precision loss, although less severe compared to Double.

Why do we avoid this in your solution for Floats?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/FlatteningAggregator.kt
```diff
@@ -0,0 +1,71 @@
...
+        val allValues = columns.asSequence().flatMap { it.values() }.filterNotNull()
+        return aggregate(allValues.asIterable(), commonType)
+    }
+
+    /**
+     * Creates [FlatteningAggregator].
+     *
+     * @param getReturnTypeOrNull Functional argument for the [calculateReturnTypeOrNull] function.
+     * @param aggregator Functional argument for the [aggregate] function.
+     * @param preservesType If `true`, [Value][Value]`  ==  `[Return][Return].
```
zaleslaw, line 57: Sorry, again could not understand  [Value][Value]`  ==  `[Return][Return].

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/Aggregators.kt
```diff
@@ -1,52 +1,136 @@
...
     }
 
-    val median by mergedValues<Comparable<Any?>, Comparable<Any?>> { median(it) }
+    // T: Comparable<T>? -> T
+    val median by flatteningPreservingTypes<Comparable<Any?>> { type ->
+        median(type)
+    }
 
-    val sum by extendsNumbers { sum(it) }
+    // T: Number -> T
```
zaleslaw, line 132: // -> /** */ ?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/Aggregators.kt
```diff
@@ -1,52 +1,136 @@
...
+    // T: Comparable<T>? -> T
+    val percentile by withOneOption { percentile: Double ->
+        flatteningPreservingTypes<Comparable<Any?>> { type ->
+            percentile(percentile, type)
+        }
     }
 
-    val median by mergedValues<Comparable<Any?>, Comparable<Any?>> { median(it) }
+    // T: Comparable<T>? -> T
+    val median by flatteningPreservingTypes<Comparable<Any?>> { type ->
```
zaleslaw, line 128: Great abstraction - I'd like it

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/AggregatorOptionSwitch.kt
```diff
@@ -1,33 +1,72 @@
...
+     * Creates [AggregatorOptionSwitch2].
+     *
+     * Used like:
+     * ```kt
+     * val myAggregator by AggregatorOptionSwitch2.Factory { param1: Param1, param2: Param2 ->
+     *   MyAggregator.Factory(param1, param2)
+     * }
+     * ```
+     */
+    class Factory<in Param1, in Param2, out AggregatorType : Aggregator<*, *>>(
```
zaleslaw, line 67: Very hard to read, could it be formatted in another way?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/Aggregator.kt
```diff
@@ -2,23 +2,91 @@ package org.jetbrains.kotlinx.dataframe.impl.aggregation.aggregators
...
+    fun calculateReturnTypeOrNull(type: KType, emptyInput: Boolean): KType?
 }
 
 @PublishedApi
-internal fun <T> Aggregator<*, *>.cast(): Aggregator<T, T> = this as Aggregator<T, T>
+internal fun <Type> Aggregator<*, *>.cast(): Aggregator<Type, Type> = this as Aggregator<Type, Type>
 
 @PublishedApi
-internal fun <T, P> Aggregator<*, *>.cast2(): Aggregator<T, P> = this as Aggregator<T, P>
+internal fun <Value, Return> Aggregator<*, *>.cast2(): Aggregator<Value, Return> = this as Aggregator<Value, Return>
```
zaleslaw, line 81: Nice naming for Generics

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/UnifyingNumbers.kt
```diff
@@ -12,16 +14,23 @@ package org.jetbrains.kotlinx.dataframe.documentation
...
+ * potentially losing a little precision, but a warning will be given.
+ *
+ * See [UnifiedNumberTypeOptions] for these settings.
  */
 internal interface UnifyingNumbers {
 
     /**
      * ```
-     *            BigDecimal
+     *           (BigDecimal)
```
zaleslaw, line 29: Will it be removed in the next PRs and will you keep the DAG after removal?

zaleslaw (approved):

dataframe-csv/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readTsv.kt
Jolanrensen: these are out of date generated sources. Not relevant to this pr

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/statistics.kt
Jolanrensen: didn't even do that on purpose XD

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/math/sum.kt
Jolanrensen: sure, it makes sense here to have them as KDoc

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/TwoStepNumbersAggregator.kt
Jolanrensen: Unification only happens upwards, so Floats become Doubles but not the other way around. Longs will never become Floats. With PRIMITIVES_ONLY, Longs can however become Double now with a slight loss in precision (we would use BigDecimal earlier), hence the warning.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/FlatteningAggregator.kt
Jolanrensen: Simply put, if you have an aggregation where the type of the values is the same as the return type, like `min()`, then you need to set `preservesType = true`.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/Aggregators.kt
Jolanrensen: I will keep them like `//` here because it's just an implementation hint, not relevant from other places.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/AggregatorOptionSwitch.kt
Jolanrensen: how do you suggest? it's the best I could do with 3 named generics

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/UnifyingNumbers.kt
Jolanrensen: JSON can still create Big numbers and maybe other places where number unification needs to happen can do it. It's only in statistics where we don't want big numbers, right?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/AggregatorOptionSwitch.kt
Jolanrensen: I could turn the AggregatorType bound to a where-clause, but it's not much better readable
```kt
class Factory<in Param1, in Param2, out AggregatorType>(
    val getAggregator: (param1: Param1, param2: Param2) -> AggregatorProvider<AggregatorType>,
) : Provider<AggregatorOptionSwitch2<Param1, Param2, AggregatorType>> by
    Provider({ name -> AggregatorOptionSwitch2(name, getAggregator) })
    where AggregatorType : Aggregator<*, *>
```

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/FlatteningAggregator.kt
Jolanrensen: I will see if I can remove `preservesType` btw. Since we can `calculateReturnTypeOrNull()` this seems unnecessary.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/modes/ofRowExpression.kt
Jolanrensen: Rewrote it now with `calculateReturnType`. 
But the gist of this specific overload `aggregateOf` is that it executes the calls like `meanOf`, `minOf`, etc, so the result of calling a `statisticOf { rowExpression }` function from within a `GroupBy`.
The result of calling the aggregator on the dataframe with the rowExpression here is a single `value: R`, which will be "yielded" with a certain path and type. This makes it so that `GroupBy` can, say, store it in a new `mean` column :). It's very high level, but hopefully the idea is a bit clearer now.

PR #1079, opened 2025-02-27
Extend CS DSL support in compiler plugin

koperagen:
I made adapter layer, so most implementations in compiler plugin just delegate to core. So most of this PR are tests for all different overloads, making sure they work

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/columns/ColumnWithPath.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/columns/ColumnWithPath.kt
@@ -3,9 +3,6 @@ package org.jetbrains.kotlinx.dataframe.columns
 import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.api.asColumnGroup
 import org.jetbrains.kotlinx.dataframe.api.isColumnGroup
-import org.jetbrains.kotlinx.dataframe.impl.columns.ColumnGroupWithPathImpl
-import org.jetbrains.kotlinx.dataframe.impl.columns.FrameColumnWithPathImpl
-import org.jetbrains.kotlinx.dataframe.impl.columns.ValueColumnWithPathImpl
 import org.jetbrains.kotlinx.dataframe.impl.columns.addParentPath
 import org.jetbrains.kotlinx.dataframe.impl.columns.addPath
 import org.jetbrains.kotlinx.dataframe.impl.columns.depth
@@ -65,19 +62,4 @@ public interface ColumnWithPath<out T> : DataColumn<T> {
 
 public val <T> ColumnWithPath<T>.depth: Int get() = path.depth()
 
-public fun ColumnWithPath(column: DataColumn<*>, path: ColumnPath): ColumnWithPath<*> =
-    when (column) {
-        is FrameColumn<*> -> {
-            FrameColumnWithPathImpl(column, path)
-        }
-
-        is ColumnGroup<*> -> {
-            ColumnGroupWithPathImpl(column, path)
-        }
-
-        is ValueColumn<*> -> {
-            ValueColumnWithPathImpl(column, path)
-        }
-
-        else -> error("Can't add path to ${column.javaClass}")
-    }
+public fun ColumnWithPath(column: DataColumn<*>, path: ColumnPath): ColumnWithPath<*> = column.addPath(path)
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/columns/ColumnWithPath.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/columns/ColumnWithPath.kt
@@ -3,6 +3,9 @@ package org.jetbrains.kotlinx.dataframe.columns
 import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.api.asColumnGroup
 import org.jetbrains.kotlinx.dataframe.api.isColumnGroup
+import org.jetbrains.kotlinx.dataframe.impl.columns.ColumnGroupWithPathImpl
+import org.jetbrains.kotlinx.dataframe.impl.columns.FrameColumnWithPathImpl
+import org.jetbrains.kotlinx.dataframe.impl.columns.ValueColumnWithPathImpl
 import org.jetbrains.kotlinx.dataframe.impl.columns.addParentPath
 import org.jetbrains.kotlinx.dataframe.impl.columns.addPath
 import org.jetbrains.kotlinx.dataframe.impl.columns.depth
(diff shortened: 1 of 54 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/columns/ColumnWithPath.kt
```diff
@@ -61,3 +64,20 @@ public interface ColumnWithPath<out T> : DataColumn<T> {
 }
 
 public val <T> ColumnWithPath<T>.depth: Int get() = path.depth()
+
+public fun ColumnWithPath(column: DataColumn<*>, path: ColumnPath): ColumnWithPath<*> =
```
Jolanrensen, line 68: Duplicate logic to `BaseColumn<T>.addPath(path: ColumnPath)`. They should probably call each other if we want to keep both

Jolanrensen:
Wow really cool! :D That's so many supported functions :)

Does the compiler plugin need to understand the entire expression or can it also just understand a part of it? Like:
```kt
df.select { name and cols { false } }.name
```

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/columnsSelectionDsl.kt
```diff
@@ -19,12 +31,49 @@ internal class And10 : AbstractInterpreter<ColumnsResolver>() {
...
+        return listOf(resolveSingle(context))
+    }
+
+    override fun resolveSingle(context: ColumnResolutionContext): ColumnWithPath<Any?> {
+        return ColumnWithPath(col.column.asDataColumn(), col.path)
+    }
 }
 
-interface ColumnsResolver {
+interface ColumnsResolver : org.jetbrains.kotlinx.dataframe.columns.ColumnSet<Any?> {
```
Jolanrensen, line 57: Why does this not inherit `org.jetbrains.kotlinx.dataframe.columns.ColumnsResolver<Any?>` like the name suggests? I suspect this is gonna be confusing in the future (oh it's because I sealed it... oops... oh well, just have to be careful with instance checks for when some columnsResolver is a singlecolumn or columnset)

koperagen:
> Does the compiler plugin need to understand the entire expression or can it also just understand a part of it? Like:

There's no such partial evaluation in place now

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/columns/ColumnWithPath.kt
koperagen: That's a good point, thank you. I'll call addPath from ColumnWithPath

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/columnsSelectionDsl.kt
Jolanrensen: sounds great :)

plugins/kotlin-dataframe/testData/selectionDslTestUtils.kt
```diff
@@ -0,0 +1,60 @@
...
+}
+
+@DataSchema
+interface Name2 {
+    val firstName: DataRow<FirstNames>
+    val lastName: String
+}
+
+@DataSchema
+interface Person2 {
```
zaleslaw, line 36: sometimes I suppose it's bad that some numerical types as Long/Double are not presented in this dataset

Jolanrensen: true, I suppose weight could be a double actually, that's pretty realistic

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/select.kt
```diff
@@ -55,19 +63,134 @@ internal class And0 : AbstractInterpreter<ColumnsResolver>() {
...
+
+internal class AllAfter1 : AbstractInterpreter<ColumnsResolver>() {
+    val Arguments.receiver by ignore()
+    val Arguments.column: SingleColumnApproximation by arg()
+    override fun Arguments.interpret(): ColumnsResolver {
+        return columnsResolver { allAfter(column) }
+    }
+}
+
+internal class AllAfter2 : AbstractInterpreter<ColumnsResolver>() {
```
zaleslaw, line 104: What's the difference between AllAfter1/AllAfter2

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/select.kt
```diff
@@ -55,19 +63,134 @@ internal class And0 : AbstractInterpreter<ColumnsResolver>() {
...
+            receiver.asColumnGroup().allColsAfter(column)
+        }
+    }
+}
+
+internal class AllBefore0 : AbstractInterpreter<ColumnsResolver>() {
+    val Arguments.receiver: ColumnsResolver by arg()
+    val Arguments.column: SingleColumnApproximation by arg()
+    override fun Arguments.interpret(): ColumnsResolver {
+        return columnsResolver { receiver.allBefore(column) }
```
zaleslaw, line 126: Seems like it could be one class with the reference on function in some cases

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/DataFrameAdapter.kt
```diff
@@ -35,19 +35,20 @@ fun SimpleCol.asDataColumn(): DataColumn<*> {
...
-            is FrameColumn<*> -> {
-                SimpleFrameColumn(it.name(), it[0].columns().mapBack())
-            }
-
-            else -> {
-                SimpleDataColumn(it.name(), it[0] as TypeApproximation)
-            }
+private fun List<AnyCol>.mapBack(): List<SimpleCol> = map { it.asSimpleColumn() }
+
+fun AnyCol.asSimpleColumn(): SimpleCol {
```
zaleslaw, line 40: could you add some comment, when to use this function

zaleslaw (approved):
Please answer my comments, before or after merging

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/DataFrameAdapter.kt
koperagen: sure

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/select.kt
koperagen: CS DSL functions differ a little, but plugin implementation happened to be the same. You're right, i'll update it to use abstract class

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/select.kt
koperagen: Would be cool! I tried to get the reference to allBefore, but it's strange.. Hard to get the reference to an extension function inside interface, plus we have ambiguity. So this doesn't compile

  public companion object {
        
        internal val AllColumnsSelectionDsl<*>.a: ColumnSelectionDsl<*>.(ColumnSelector<*, *>) -> ColumnSet<*> get() = ColumnsSelectionDsl<*>::allBefore
    }

PR #1081, opened 2025-02-28
To dataframe improvements

### LINKED ISSUE #676: Missing overloads for `Iterable<ArrayLike>.toDataFrame()`
Currently, running
```kt
listOf(intArrayOf(1, 2, 3), intArrayOf(4, 5, 6)).toDataFrame()
```
results in:
```
No accessor found for property val kotlin.IntArray.size: kotlin.Int
kotlin.reflect.jvm.internal.KotlinReflectionInternalError: No accessor found for property val kotlin.IntArray.size: kotlin.Int
```
This is because `toDataFrame()` defaults to `toDataFrame { properties() }`, so it tries to dissect the properties of the `IntArray` class in this instance.

Instead, I believe this should create a `DataFrame<ValueProperty<IntArray>>`, like for the other primitives.

There is a challenge to adding these though. I found adding the overload `Iterable<BooleanArray>.toDataFrame()` breaks calls to `Iterable<AnyBaseCol>.toDataFrame()` for some reason. And adding them like `<B : BooleanArray?> Iterable<B>.toDataFrame()` is impossible, because arrays are illegal supertypes...

AndreiKingsley:
* Fixes #676

Fixed behavior for primitive arrays (**due to compiler limitations, it is impossible to make extensions for primitive arrays**), added extensions for common value types.


plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/utils/Names.kt
```diff
@@ -65,6 +65,9 @@ object Names {
     val LOCAL_DATE_CLASS_ID = kotlinx.datetime.LocalDate::class.classId()
     val LOCAL_DATE_TIME_CLASS_ID = kotlinx.datetime.LocalDateTime::class.classId()
     val INSTANT_CLASS_ID = kotlinx.datetime.Instant::class.classId()
+    val DATE_TIME_PERIOD_CLASS_ID = kotlinx.datetime.DateTimePeriod::class.classId()
```
koperagen, line 68: I'll later remove dependency on kotlinx.datetime, please call ClassId constructor with String :)

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/toDataFrame.kt
```diff
@@ -190,34 +190,56 @@ class Exclude1 : AbstractInterpreter<Unit>() {
...
             this.isSubtypeOf(
                 ConeClassLikeTypeImpl(
                     ConeClassLikeLookupTagImpl(
-                        ClassId(FqName("java.time.temporal"), Name.identifier("Temporal"))
+                        ClassId(FqName("java.time.temporal"), Name.identifier("TemporalAccessor"))
+                    ), arrayOf(), isNullable = false
+                ), session
+            ) ||
+            this.isSubtypeOf(
+                ConeClassLikeTypeImpl(
```
koperagen, line 230: let's refactor it a little, extract class ids in Names object and then call Names.TEMPORAL.constructClassLikeType instead of ConeClassLikeTypeImpl...

koperagen (changes requested):
Add a test for compiler plugin part covering the new case

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/toDataFrame.kt
AndreiKingsley: Ok, btw why `isNullable=false` here?

koperagen: Good catch! For number type i have two subtypeOf, one with nullable type and other with non-nullable. But 1 check with nullable = true will be enough and it's more correct. Test case that works incorrectly now:
```
class Test(val temporal: Temporal?)

listOf(Test(LocalDateTime.parse("01/01/08 00:02:04", DateTimeFormatter.ofPattern("MM/dd/yy HH:mm:ss"))), Test(null)).toDataFrame(maxDepth = 2)   
```

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
```diff
@@ -38,20 +39,37 @@ import kotlin.reflect.jvm.javaField
 import kotlin.reflect.typeOf
 
 private val valueTypes = setOf(
+    Any::class,
```
Jolanrensen, line 42: Uhm, `Any` doesn't cause issues with other classes? I guess this doesn't mean "a subtype of", rather the exact type? Maybe mention this in kdoc for future us

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
```diff
@@ -38,20 +39,37 @@ import kotlin.reflect.jvm.javaField
...
+    kotlinx.datetime.DateTimePeriod::class,
+    kotlinx.datetime.DateTimeUnit::class,
 )
 
+/**
+ * Should be aligned with `ConeKotlinType.isValueType()` in
+ * plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/toDataFrame.kt
+ */
+@PublishedApi
 internal val KClass<*>.isValueType: Boolean
```
Jolanrensen, line 65: yess! This is great, a stable definition of what we consider a value type :D love it

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -26,7 +28,14 @@ import kotlin.reflect.KProperty
 @Interpretable("toDataFrameDefault")
 public inline fun <reified T> Iterable<T>.toDataFrame(): DataFrame<T> =
     toDataFrame {
-        properties()
+        // check if type is value: primitives, primitive arrays, datetime types etc.
+        if (T::class.isValueType) {
```
Jolanrensen, line 32: neat!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -327,6 +243,7 @@ public fun Map<String, Iterable<Any?>>.toDataFrame(): AnyFrame =
 @JvmName("toDataFrameColumnPathAnyNullable")
 public fun Map<ColumnPath, Iterable<Any?>>.toDataFrame(): AnyFrame =
     map {
+        DatePeriod
```
Jolanrensen, line 246: typo?

Jolanrensen (approved):
very nice stability and consistency improvements!

AndreiKingsley:
> sad we cannot have this for arrays 

But as @koperagen said we don't need it since we have plugin !

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
AndreiKingsley: User can provide `Iterable<Any?>`.

AndreiKingsley: And yeah, these are exact types.

Jolanrensen: so now if a user has something like `listOf(cat, dog, house, table).toDataFrame()` they will get a `value` column, right?

koperagen (approved):

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
```diff
@@ -74,6 +74,13 @@ internal val KClass<*>.isValueType: Boolean
             this.isSubclassOf(TemporalAmount::class) ||
             this.isArray
 
+/**
+ * Checks if `KClass` has properties.
+ */
+@PublishedApi
+internal val KClass<*>.hasProperties: Boolean
+    get() = this.memberProperties.isNotEmpty()
```
Jolanrensen, line 82: This might cause issues with POJO's, java objects with getters+setters instead of memberProperties

AndreiKingsley: added check

Jolanrensen (approved):

PR #1082, opened 2025-03-04
DynamicDataFrameBuilder﻿ improvements

### LINKED ISSUE #715: "Smart" column add in `DynamicDataFrameBuilder﻿`
I can't rewrite this code with `DynamicDataFrameBuilder﻿`:
https://github.com/Kotlin/kandy/blob/2764bce7e9eec4888fdce71b306631cd4e0a208c/kandy-api/src/main/kotlin/org/jetbrains/kotlinx/kandy/dsl/internal/DatasetHandler.kt#L116
Here I want to do the following - if the column I want to add is already in the builder (with the same name and contains the same elements), it shouldn't be added again.

AndreiKingsley:
Closes #715.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
@@ -413,29 +413,78 @@ public class DataFrameBuilder(private val header: List<String>) {
 }
 
 /**
- * Helper class for implementing operations when column names can be potentially duplicated.
- * For example, operations involving multiple dataframes, computed columns or parsing some third-party data
+ * A builder class for dynamically constructing a DataFrame with provided columns.
+ * Allows adding columns manually while automatically handling duplicate column names by assigning unique names.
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
```diff
@@ -413,29 +413,78 @@ public class DataFrameBuilder(private val header: List<String>) {
...
+    /**
+     * Retrieves a column from the builder by its name.
+     *
+     * @param column The name of the column to retrieve.
+     * @return The column corresponding to the specified name, or `null` if no such column exists.
+     */
+    public fun get(column: String): AnyCol? = cols[column]
+
+    /**
+     * Converts the current `DynamicDataFrameBuilder` instance into a `DataFrame`.
```
Jolanrensen, line 482: using `[]` works better than just backticks, it's clickable and refactor safe

Jolanrensen (approved):

PR #1084, opened 2025-03-04
Prepare plugin to be moved to Kotlin repository

koperagen:
1. Removing all I/O, Import annotation
2. Publishing fatjar minimized to only whatever is needed for compiler plugin to work, as much as possible. New artifact will be `org.jetbrains.kotlinx:dataframe-core-all` (it's a variant of core)
3. Switching to -Xjvm-default=all
Now default methods in interfaces will be generated as java default methods in bytecode. This change removes 4 mb of duplicated code on DefaultImpls static classes in our core jar. Massive change in cope.api comes from this


Jolanrensen (approved):
Great job! :) Could I suggest you add a line to https://github.com/Kotlin/dataframe/blob/master/CONTRIBUTING.md#building telling about how to run a shadow build and what that does? Then I think it's complete :)

Jolanrensen:
> > Great job! :) Could I suggest you add a line to https://github.com/Kotlin/dataframe/blob/master/CONTRIBUTING.md#building telling about how to run a shadow build and what that does? Then I think it's complete :)
> 
> It's nothing special, think of it as "-sources.jar", "-javadoc.jar". It's published along with core by `publishToMavenLocal` or `publish`

then a small comment next to the shadow jar configuration could also do. Especially since the compiler plugin will be moved out of the repo, it's important to mark things relevant to the compiler plugin so that we don't forget that it's relevant in the future.

PR #699, opened 2024-05-16
importDataSchema tests 

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/696
by a fix done in the kotlin-jupyter kernel. It requires kotlin-jupyter 0.12.0.225 or up to work.

Issue #696, opened 2024-05-15
Jupyter `importDataSchema()` does not generate column accessors


Jolanrensen:
Jupyter `importDataSchema()` does not generate column accessors. This is likely due to nested `@DataSchema` interfaces being skipped:

```kt
%trackExecution
@DataSchema
interface A {
    val a: Int
}

interface B {
    @DataSchema
    interface C {
        val c: Int
    }
}
```
Executing:
```kt
val ColumnsContainer<Line_22_jupyter.A>.a: DataColumn<Int> @JvmName("A_a") get() = this["a"] as DataColumn<Int>
val DataRow<Line_22_jupyter.A>.a: Int @JvmName("A_a") get() = this["a"] as Int
val ColumnsContainer<Line_22_jupyter.A?>.a: DataColumn<Int?> @JvmName("NullableA_a") get() = this["a"] as DataColumn<Int?>
val DataRow<Line_22_jupyter.A?>.a: Int? @JvmName("NullableA_a") get() = this["a"] as Int?
 ```

Jolanrensen:
Related: https://youtrack.jetbrains.com/issue/KTNB-680/ClassAnnotationHandler-does-not-handle-nested-interfaces-classes

Issue #595, opened 2024-02-20
Update kotest-assertions to 5.8.0


Jolanrensen:
`io.kotest:kotest-assertions-core [5.5.4 -> 5.8.0]`, we cannot update atm thanks to kotlin-jupyter: https://github.com/Kotlin/kotlin-jupyter/issues/452

Kantis:
Upgrading jupyter to [0.12.0.217](https://github.com/Kotlin/kotlin-jupyter/releases/tag/0.12.0.217) should unblock upgrading kotest-assertions now. https://github.com/Kotlin/kotlin-jupyter/issues/452 was fixed in https://github.com/Kotlin/kotlin-jupyter/commit/e6a2dc14b8565a6ba2c6ce17ee0b0eb378251481 which is tagged with that version.

Jolanrensen:
Thanks, unfortunately, we're blocked from updating Jupyter thanks to https://github.com/Kotlin/dataframe/issues/700 :/

Issue #771, opened 2024-07-09
Integration with ktor-client


koperagen:
We should provide a function to get DataFrame from NotebookHttpResponse
I tried to add a compileOnly dependency on ktor-client, declared a function in dataframe-jupyter module. Due to classloading conflicts its work is unstable. 
If you execute cell with `%use dataframe` first, then `%use ktor-client`, there's this exception:
```
Line_12.jupyter.kts (1:15 - 18) Cannot access class 'org.jetbrains.kotlinx.jupyter.ktor.client.core.NotebookHttpResponse'. Check your module classpath for missing or conflicting dependencies
```
It happens because `NotebookHttpResponse.bodyAsDataFrame` is loaded with classloader that doesn't know anything about NotebookHttpResponse. 
So instead of declaring this function in a library, we can combine afterCellExecution + executeCode to declare it when dataframe is added and ktor-client is already on the classpath or when ktor-client is added to the classpath. This way the function will be executed in a cell with both ktor-client and dataframe classloaders available

koperagen:
New API that should help: https://github.com/Kotlin/kotlin-jupyter/blob/099c0a49907dbd0a06348aa57c741cf9d943012a/jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/Notebook.kt#L33

Jolanrensen:
This is not blocked anymore after https://github.com/Kotlin/dataframe/pull/1087

AndreiKingsley:
for some inspiration:
https://blog.jetbrains.com/kotlin/2024/08/track-and-analyze-github-star-growth-with-kandy-and-kotlin-dataframe

zaleslaw:
Very good candidate feature for 1.1

PR #1087, opened 2025-03-05
Post Java 11 version bumps

Jolanrensen:
Follow up for: https://github.com/Kotlin/dataframe/issues/700

Fixes https://github.com/Kotlin/dataframe/issues/696
Closes https://github.com/Kotlin/dataframe/pull/699
Fixes https://github.com/Kotlin/dataframe/issues/595
Unblocks https://github.com/Kotlin/dataframe/issues/771

Bumps various versions that were previously blocked, like Jupyter, Kotest, etc.

Bumps java toolchain version to 21 to be able to update simple-git. I also bumped TC to 21


core/api/core.api
```diff
@@ -4742,6 +4742,10 @@ public final class org/jetbrains/kotlinx/dataframe/columns/ColumnPath : java/uti
 	public fun add (Ljava/lang/String;)Z
 	public fun addAll (ILjava/util/Collection;)Z
 	public fun addAll (Ljava/util/Collection;)Z
+	public synthetic fun addFirst (Ljava/lang/Object;)V
```
zaleslaw, line 4745: Interesting, why it was added

zaleslaw (approved):

core/api/core.api
Jolanrensen: I'm not sure either, probably because it extends List

koperagen: Hmm.. It seems that target jdk is now Java 21 too: https://openjdk.org/jeps/431 and not 11 as we intend. Try Xjdk-release=11 in compileKotlin freeCompilerArgs

Jolanrensen: made an issue for it https://github.com/Kotlin/binary-compatibility-validator/issues/290

koperagen: https://kotlinlang.org/docs/compiler-reference.html#jvm-target-version
JVM target only seems to affect bytecode. jdk release limits API too

Jolanrensen: Ah I see, but if the target jvm is 11, then it's okay to compile it against a newer version, right? Having the target at 11 allows our users to set their projects to 11 and DF still works

Jolanrensen:
okay seems Kotlin 2.1 is sneaking in from some dependency, need to figure out which, because it makes our jupyter tests fail

core/api/core.api
koperagen: I believe we need to set jdk release to 11 and toolchain to 21

Jolanrensen: I think release and target overlap. I'll set it, but I think setting the target and sourceCompatibility does the same

PR #1091, opened 2025-03-10
Mean statistics fixes

Jolanrensen:
Helps and follows https://github.com/Kotlin/dataframe/issues/961

~To be merged after https://github.com/Kotlin/dataframe/pull/1078~

- Redid implementation of `mean`. It's now based on `TwoStepNumbersAggregator`, such that mixed numbers are unified first.
- Big number support is dropped.
- Created generic `rowX()` function `aggregateOfRow()`, used by `rowMean()` and `rowMeanOf<T>()`
- Removed big numbers from `describe()`
- https://github.com/Kotlin/dataframe/pull/1091#issuecomment-2729355565


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
```diff
@@ -52,6 +53,8 @@ public fun AnyCol.isNumber(): Boolean = isSubtypeOf<Number?>()
 
 public fun AnyCol.isBigNumber(): Boolean = isSubtypeOf<BigInteger?>() || isSubtypeOf<BigDecimal?>()
 
+public fun AnyCol.isPrimitiveNumber(): Boolean = type().withNullability(false) in primitiveNumberTypes
```
AndreiKingsley, line 56: Should this function be public? I mean, it seems like users will almost never call it, and it pollutes the scope of column extensions. If it does, please add a KDoc.

AndreiKingsley (approved):
Please add some notes in comments for public `mean` about new behavior, so that we don't forget to mention it there when writing KDoc in the future.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
Jolanrensen: hmm, maybe not. On the other hand, we are only allowing statistics on columns with primitive number types, so it might be helpful to actually have a simple check or ability to filter columns for primitive number types. That said, to be helpful with the compiler plugin, we'd likely need a specific `primitiveNumberCols()` function in the selection DSL... I think I'll keep it for now and I'll remove `isBigNumber`

Jolanrensen: I see I forgot some overloads; they are still filtering for number columns instead of primitive number columns, oops, fixing them too

AndreiKingsley (approved):
Great! However, I wonder - are there any aggregators that take null into account?

PR #1092, opened 2025-03-11
Move/Insert After fix

### LINKED ISSUE #1056: Move/Insert after nested column breaks
To reproduce:
```kt
val df = dataFrameOf("a", "b", "c")(1, 2, 3, 4, 5, 6)
val grouped = df.group { a and b and c }.into("g")

grouped.insert("d") { g.b  * g.c }.after { g.b }
// or
grouped.move { g.a }.after { g.b }
```
Image

The problem lies with `move`, as `insert` calls `move` as part of its implementation.

We should either prohibit this behavior with a helpful exception or fix the behavior.

AndreiKingsley:
* Closes #1056.
* Adds `insert after` plugin support

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
@@ -11,6 +11,7 @@ import org.jetbrains.kotlinx.dataframe.api.getColumn
 import org.jetbrains.kotlinx.dataframe.api.getColumnGroup
 import org.jetbrains.kotlinx.dataframe.api.getColumnWithPath
 import org.jetbrains.kotlinx.dataframe.api.toDataFrame
+import org.jetbrains.kotlinx.dataframe.columns.ColumnPath
 import org.jetbrains.kotlinx.dataframe.columns.ColumnWithPath
 import org.jetbrains.kotlinx.dataframe.columns.UnresolvedColumnsPolicy
 import org.jetbrains.kotlinx.dataframe.impl.DataFrameReceiver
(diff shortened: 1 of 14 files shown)


koperagen (approved):
Don't forget to set milestone, assignee and labels for generated change log

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
```diff
@@ -23,20 +24,70 @@ internal fun <T, C> MoveClause<T, C>.afterOrBefore(column: ColumnSelector<T, *>,
...
+    }
+
     val removeRoot = removeResult.removedColumns.first().getRoot()
 
     val refNode = removeRoot.getOrPut(targetPath) {
-        val parentPath = it.dropLast(1)
-        val parent = if (parentPath.isEmpty()) df else df.getColumnGroup(parentPath)
-        val index = parent.getColumnIndex(it.last())
-        val col = df.getColumn(index)
+        val path = it.toList()
```
Jolanrensen, line 46: can be cast to List as well

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
```diff
@@ -23,20 +24,70 @@ internal fun <T, C> MoveClause<T, C>.afterOrBefore(column: ColumnSelector<T, *>,
...
+
+        // Find the group reference (if any) and adjust the path
+        val groupRefIndex = path.indexOfFirst { it.isEmpty() }
```
Jolanrensen, line 49: What is happening here? How does the first empty path piece correspond to a group node? And at which indices can this empty path piece occur?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
```diff
@@ -23,20 +24,70 @@ internal fun <T, C> MoveClause<T, C>.afterOrBefore(column: ColumnSelector<T, *>,
...
+        val effectivePath = if (groupRefIndex >= 0) {
+            // Nested column reference
+            path.take(groupRefIndex)
+        } else {
+            path.dropLast(1)
+        }
+
+        val columnName = if (groupRefIndex >= 0) {
+            // Use the next segment after group reference, or previous if no next segment
+            path.getOrNull(groupRefIndex + 1) ?: path[groupRefIndex - 1]
```
Jolanrensen, line 61: can this break if groupRefIndex == 0?

plugins/kotlin-dataframe/testData/box/insertAfter.kt
```diff
@@ -0,0 +1,12 @@
+import org.jetbrains.kotlinx.dataframe.*
+import org.jetbrains.kotlinx.dataframe.api.*
+
+fun box(): String {
+    val df = dataFrameOf("a", "b", "c")(1, 2, 3, 4, 5, 6)
+
+    val dfWithD = df.insert("d") { b  * c }.after { a }
```
Jolanrensen, line 7: maybe lint these as well? XD

Jolanrensen (approved):
Good job :) just have some questions

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
```diff
@@ -23,20 +24,70 @@ internal fun <T, C> MoveClause<T, C>.afterOrBefore(column: ColumnSelector<T, *>,
...
+    // Check if any source path is a prefix of the target path
+    sourcePaths.forEach { sourcePath ->
+        val sourceSegments = sourcePath.toList()
+        val targetSegments = targetPath.toList()
+
+        if (sourceSegments.size <= targetSegments.size &&
+            sourceSegments.indices.all { targetSegments[it] == sourceSegments[it] }
+        ) {
+            throw IllegalArgumentException(
+                "Cannot move column '${sourcePath.joinToString()}' after its child column '${targetPath.joinToString()}'",
```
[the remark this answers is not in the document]
koperagen: maybe "its own child"? is it a proper english? makes it clearer?

PR #1095, opened 2025-03-13
Splitting off Jupyter yet again

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/775

- moves jupyter integration to `:dataframe-jupyter`
- Refactors `org.jetbrains.dataframe` -> `org.jetbrains.kotlinx.dataframe` in codegen parts
- jupyter integration becomes a friend module. 
  - ~Trying out [Restrikt](https://github.com/ZwenDo/Restrikt) to do this responsibly.~ Restrikt seems too unstable for DF. Often getting compile errors in FIR or NoSuchElementExceptions in tests.
  - Until https://youtrack.jetbrains.com/issue/KTIJ-31881/Support-Xfriend-path-compiler-argument-in-the-Kotlin-plugin is fixed, I suggest collecting all suppressions in one region per file. That seems the safest option to me and easy to revert once intellij understands friend modules or we find another solution.
- enables typesafe project accessors
- Fixes and enables example notebook tests only in DEBUG mode
- Updated library descriptor to handle both the old and new df versions: https://github.com/Kotlin/kotlin-jupyter-libraries/pull/470. This will work whenever we release a version 1.x.x.


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/CodeWithConverter.kt
```diff
@@ -28,3 +25,6 @@ public data class CodeWithConverter(val declarations: Code, val converter: (Vari
             else -> declarations + "\n" + converter(name)
         }
 }
+
+public typealias Code = String
```
koperagen, line 29: Is it an option to not have these in public api?

dataframe-geo/build.gradle.kts
```diff
@@ -26,7 +26,8 @@ repositories {
 fun ExternalModuleDependency.excludeJaiCore() = exclude("javax.media", "jai_core")
 
 dependencies {
-    api(project(":core"))
+    api(projects.core)
+    implementation(projects.dataframeJupyter)
```
koperagen, line 30: Why do we need geo to depend on impl jupyter?

dataframe-jupyter/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+tasks.test {
+    maxHeapSize = "2048m"
+}
+
+kotlinPublications {
+    publication {
+        publicationName.set("dataframe-jupyter")
+        artifactId.set("dataframe-jupyter")
+        description.set("Kotlin DataFrame integration with Kotlin Jupyter")
+        packageName.set("dataframe-jupyter")
```
koperagen, line 56: Did you check coordinates of the artifact? Can be seen in `~/.m2/repository/` after you publishToMavenLocal. Should be under `org.jetbrains.kotlinx.dataframe-jupyter` i suppose

dataframe-jupyter/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/JupyterHtmlRenderer.kt
```diff
@@ -81,10 +74,10 @@ internal inline fun <reified T : Any> JupyterHtmlRenderer.render(
         val jsonEncodedDf = when {
             !ideBuildNumber.supportsDynamicNestedTables() -> {
                 buildJsonObject {
-                    put(NROW, df.size.nrow)
-                    put(NCOL, df.size.ncol)
-                    putJsonArray(COLUMNS) { addAll(df.columnNames()) }
-                    put(KOTLIN_DATAFRAME, encodeFrame(df.take(limit)))
+                    put(SerializationKeys.NROW, df.rowsCount())
```
koperagen, line 77: Looks like these few internal declarations can be hidden behing single `buildJsonObject(df, limit)`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/CodeWithConverter.kt
Jolanrensen: I think they're part of our codegen api. CodeWithConverter is public too after all

dataframe-geo/build.gradle.kts
Jolanrensen: Actually, yes, it uses `useSchema<>()` in its jupyter integration

dataframe-jupyter/build.gradle.kts
Jolanrensen: check :)

dataframe-geo/build.gradle.kts
koperagen: If we expect to use geo Integration to work only when jupyter Integration is configured (which should be), dependency can be `compileOnly` here

Jolanrensen: I'm not 100% sure, @AndreiKingsley ?

Jolanrensen: Let's fix this after merging, so we can revert back to java 8 sooner

PR #1096, opened 2025-03-13
Publish compiler plugin :core subset from another module

koperagen:


Jolanrensen (approved):
Very clean solution! It makes it clear that we have a compiler plugin living somewhere else, and what we do for it on our side :)

PR #1103, opened 2025-03-18
Sum statistics and aggregator improvements

Jolanrensen:

- Simplified `sum` implementation, based on the table at https://github.com/Kotlin/dataframe/issues/961
- Fixed public `sum` overloads so they always return the correct type based on the table
- tiny fixes for `aggregateOfRow`
- Selecting "no" columns in `mean` and `sum` will now select primitive- or mixed numbers only. This automatically excludes columns with unsupported numbers and in the slight case unsupported numbers end up in a mixed column, we provide a helpful exception.


AndreiKingsley (approved):
Nice!

PR #1106, opened 2025-03-24
Set version to 1.0.0

Jolanrensen:
https://github.com/Kotlin/dataframe/pull/1095 + https://github.com/Kotlin/kotlin-jupyter-libraries/pull/470 only works when our build version starts with a "1", so it makes sense to bump our dev version to our first 1.X beta release :). We can follow the naming standard of Kotlin itself.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 projectName=dataframe
-version=0.16.0
+version=1.0.0-Beta1
 jupyterApiTCRepo=
 kotlin.jupyter.add.scanner=false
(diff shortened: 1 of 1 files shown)


koperagen:
We'll have `1.0.0-Beta1-dev-*` dev releases given this version number

Jolanrensen:
> We'll have `1.0.0-Beta1-dev-*` dev releases given this version number

Ah that's right, for RCs we avoided this so we should probably continue to do so

PR #1110, opened 2025-03-26
Ensure a predictable order of columns after aggregation of `GroupBy`

koperagen:
We need to match the order of columns in the runtime and in compiler plugin
GroupBy with the same schema should give the same result after aggregation, no matter how it was created
We cannot track position of group in original df, so we align `asGroupBy` with `groupBy` and move `group` column to end

### DIFF
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/PlaylistJsonTest.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/PlaylistJsonTest.kt
@@ -1,5 +1,6 @@
 package org.jetbrains.kotlinx.dataframe.io
 
+import io.kotest.assertions.withClue
 import io.kotest.matchers.shouldBe
 import org.jetbrains.kotlinx.dataframe.DataFrame
 import org.jetbrains.kotlinx.dataframe.DataRow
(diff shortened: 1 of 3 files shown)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/PlaylistJsonTest.kt
```diff
@@ -220,8 +221,10 @@ class PlaylistJsonTest {
             minBy { snippet.publishedAt }.snippet into "earliest"
         }
 
-        res.columnsCount() shouldBe typed.columnsCount() + 1
-        res.getColumnIndex("earliest") shouldBe typed.getColumnIndex("items") + 1
+        withClue(res.columnNames() to typed.columnNames()) {
```
Jolanrensen, line 224: wow what is this magic :o

Jolanrensen (approved):

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/PlaylistJsonTest.kt
koperagen: It's a kotest utility. When assertion fails, it'll also print a "clue". In this case pair of lists

PR #1111, opened 2025-03-27
Functions inline

### LINKED ISSUE #1026: Functions that cán be inlined should be inlined
`df.forEach {}` is inlined, which is good, it allows non-local returns etc. Other functions that cán be inlined, like `df.map {}` should also be inlined.

AndreiKingsley:
Closes #1026


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataFrame.kt
```diff
@@ -135,6 +135,8 @@ public operator fun <T> DataFrame<T>.get(columnRange: ClosedRange<String>): Data
 
 internal val ColumnsContainer<*>.ncol get() = columnsCount()
 internal val AnyFrame.nrow get() = rowsCount()
+
+@PublishedApi
 internal val AnyFrame.indices get() = indices()
```
koperagen, line 140: inline functions that use this internal properties can use already public functions instead

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/all.kt
```diff
@@ -1239,6 +1242,7 @@ internal fun ColumnsResolver<*>.allFromInternal(colByPredicate: ColumnFilter<*>)
  * @param colByPredicate the predicate function used to determine if a column should be included in the returned ColumnSet
  * @return a new ColumnSet containing all columns that come before the first column that satisfies the given predicate
  */
+@PublishedApi
 internal fun ColumnsResolver<*>.allBeforeInternal(colByPredicate: ColumnFilter<*>): ColumnSet<*> {
```
Jolanrensen, line 1246: missed an inline?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/any.kt
```diff
@@ -14,6 +14,6 @@ public fun <T> DataColumn<T>.any(predicate: Predicate<T>): Boolean = values.any(
 
 // region DataFrame
 
-public fun <T> DataFrame<T>.any(predicate: RowFilter<T>): Boolean = rows().any { predicate(it, it) }
+public inline fun <T> DataFrame<T>.any(crossinline predicate: RowFilter<T>): Boolean = rows().any { predicate(it, it) }
```
Jolanrensen, line 17: crossinline unneeded

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/associate.kt
```diff
@@ -6,10 +6,10 @@ import org.jetbrains.kotlinx.dataframe.RowExpression
 
 // region DataFrame
```
Jolanrensen, line 7: crossinline unneeded for both

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
```diff
@@ -25,15 +25,15 @@ public fun <T> DataColumn<T>.count(predicate: Predicate<T>? = null): Int =
 
 public fun AnyRow.count(): Int = columnsCount()
 
-public fun AnyRow.count(predicate: Predicate<Any?>): Int = values().count(predicate)
+public inline fun AnyRow.count(crossinline predicate: Predicate<Any?>): Int = values().count(predicate)
```
Jolanrensen, line 28: crossinline unneeded

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
```diff
@@ -25,15 +25,15 @@ public fun <T> DataColumn<T>.count(predicate: Predicate<T>? = null): Int =
...
 
 // endregion
 
 // region DataFrame
 
 public fun <T> DataFrame<T>.count(): Int = rowsCount()
 
-public fun <T> DataFrame<T>.count(predicate: RowFilter<T>): Int = rows().count { predicate(it, it) }
+public inline fun <T> DataFrame<T>.count(crossinline predicate: RowFilter<T>): Int = rows().count { predicate(it, it) }
```
Jolanrensen, line 36: crossinline unneeded

Jolanrensen (changes requested):
Could you do another run over the changes to see if crossinline is needed? I encountered a lot of cases already and crossinline blocks non-local returns, while allowing those is the main goal of the PR. crossinline is only needed for lambdas that need to be executed in a non-inlined scope

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/joinWith.kt
```diff
@@ -25,7 +25,7 @@ internal class JoinedDataRowImpl<A, B>(
     override val right: DataRow<B> = DataRowImpl(index1, rightOwner)
 }
 
-internal fun <A, B> DataFrame<A>.joinWithImpl(
+internal inline fun <A, B> DataFrame<A>.joinWithImpl(
```
Jolanrensen, line 28: hmm I'm doubting the benefits outweigh the downsides for inlining a function as large as this. Remember inlining copies over everthing in this function to the execution site increasing code size and compile time for the user. @nikita, what do you think?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/describe.kt
```diff
@@ -117,12 +117,11 @@ private fun DataColumn<Any?>.convertToComparableOrNull(): DataColumn<Comparable<
         valuesAreComparable() -> asComparable()
 
         // Found incomparable number types, convert all to Double first
-        isNumber() -> cast<Number?>().map {
-            if (it?.isPrimitiveNumber() == false) {
-                // Cannot calculate statistics of a non-primitive number type
+        isNumber() -> with(cast<Number?>()) {
```
Jolanrensen, line 120: I'd revert this back to `map` with non-local break. Then all values are visited just once, now they are visited twice.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/Utils.kt
```diff
@@ -189,8 +189,9 @@ internal fun <A, B> SingleColumn<A>.transformSingle(
  * or it can be used as a [TransformableColumnSet]<[B]>, where a [ColumnsResolverTransformer] can be injected before
  * the [converter] is applied.
  */
-internal fun <A, B> ColumnsResolver<A>.transform(
-    converter: (List<ColumnWithPath<A>>) -> List<ColumnWithPath<B>>,
+@PublishedApi
+internal inline fun <A, B> ColumnsResolver<A>.transform(
+    crossinline converter: (List<ColumnWithPath<A>>) -> List<ColumnWithPath<B>>,
 ): TransformableColumnSet<B> =
     object : TransformableColumnSet<B> {
```
Jolanrensen, line 196: remember me someday to remove this TransformableColumnSet stuff. It was the result of `.recursively()`. I have a hunch we can inline more if `transformResolve` doesn't exist anymore :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/take.kt
```diff
@@ -60,8 +60,8 @@ public fun <T> DataFrame<T>.takeLast(n: Int = 1): DataFrame<T> {
 /**
  * Returns a DataFrame containing first rows that satisfy the given [predicate].
  */
-public fun <T> DataFrame<T>.takeWhile(predicate: RowFilter<T>): DataFrame<T> =
-    firstOrNull { !predicate(it, it) }?.let { take(it.index) } ?: this
+public inline fun <T> DataFrame<T>.takeWhile(crossinline predicate: RowFilter<T>): DataFrame<T> =
```
Jolanrensen, line 63: crossinline unneeded

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/duplicate.kt
```diff
@@ -11,7 +11,7 @@ public fun <T> DataFrame<T>.duplicate(n: Int): FrameColumn<T> = List(n) { this }
 
 public fun <T> DataFrame<T>.duplicateRows(n: Int): DataFrame<T> = duplicateRowsImpl(n)
 
-public fun <T> DataFrame<T>.duplicateRows(n: Int, filter: RowFilter<T>): DataFrame<T> =
+public inline fun <T> DataFrame<T>.duplicateRows(n: Int, crossinline filter: RowFilter<T>): DataFrame<T> =
```
Jolanrensen, line 14: crossinline unneeded

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/colsOfKind.kt
```diff
@@ -199,9 +199,9 @@ public interface ColsOfKindColumnsSelectionDsl {
  * @param filter The filter function to apply on each column. Must accept a ColumnWithPath object and return a Boolean.
  * @return A [TransformableColumnSet] containing the columns of given kinds that satisfy the filter.
  */
-internal fun ColumnsResolver<*>.columnsOfKindInternal(
+internal inline fun ColumnsResolver<*>.columnsOfKindInternal(
```
Jolanrensen, line 202: currently there's no benefit for having these CS DSL internal functions inlined, because they are only called from non-inlined functions. However, I think they can stay inlined, because it reminds us we can move the CS DSL to a fully inlined context-parameter based API in the future :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/colsOf.kt
```diff
@@ -262,7 +262,10 @@ public inline fun <reified C> SingleColumn<DataRow<*>>.colsOf(
  * match the given [filter] and are the given [type].
  */
 @Suppress("UNCHECKED_CAST")
-internal fun <C> ColumnsResolver<*>.colsOfInternal(type: KType, filter: ColumnFilter<C>): TransformableColumnSet<C> =
+internal inline fun <C> ColumnsResolver<*>.colsOfInternal(
```
Jolanrensen, line 265: there are a couple more functions that be inline in this file

Jolanrensen (changes requested):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/joinWith.kt
koperagen: I agree, let's keep it a regular fun

koperagen:
Looking at changes that make some internal functions inline, can you please verify size of our core jar in core/build/libs before and after this change? Let's make sure we don't accidentally affect it too much

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/colsOf.kt
AndreiKingsley: Which ones?

PR #1118, opened 2025-04-04
Support convert shortcuts

koperagen:
I tried to minimize repetition and come up with somewhat generic solution. However the way framework works, we still must have delegated properties matching function arguments. I decided to not provide a way for interpreter to ignore any unknown arguments in this PR

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
@@ -17,6 +17,7 @@ import org.jetbrains.kotlinx.dataframe.DataRow
 import org.jetbrains.kotlinx.dataframe.RowColumnExpression
 import org.jetbrains.kotlinx.dataframe.RowValueExpression
 import org.jetbrains.kotlinx.dataframe.annotations.AccessApiOverload
+import org.jetbrains.kotlinx.dataframe.annotations.Converter
 import org.jetbrains.kotlinx.dataframe.annotations.HasSchema
 import org.jetbrains.kotlinx.dataframe.annotations.Interpretable
 import org.jetbrains.kotlinx.dataframe.annotations.Refine
@@ -284,13 +285,38 @@ public fun <T : Any> DataColumn<T?>.convertToBoolean(): DataColumn<Boolean?> = c
 
 // region convert URL
 
-public fun <T, R : URL?> Convert<T, R>.toIFrame(
+@JvmName("toIframeFromUrlNullable")
+@Refine
+@Converter<IFRAME>(IFRAME::class, nullable = true)
+@Interpretable("ToSpecificType")
+public fun <T> Convert<T, URL?>.toIFrame(
+    border: Boolean = false,
+    width: Int? = null,
+    height: Int? = null,
+): DataFrame<T> = to { it.map { url -> url?.let { IFRAME(url.toString(), border, width, height) } } }
+
+@JvmName("toIframeFromUrl")
+@Refine
+@Converter<IFRAME>(IFRAME::class, nullable = false)
+@Interpretable("ToSpecificType")
+public fun <T> Convert<T, URL>.toIFrame(
     border: Boolean = false,
     width: Int? = null,
     height: Int? = null,
 ): DataFrame<T> = to { it.map { IFRAME(it.toString(), border, width, height) } }
 
-public fun <T, R : URL?> Convert<T, R>.toImg(width: Int? = null, height: Int? = null): DataFrame<T> =
+@JvmName("toImgFromUrlNullable")
+@Refine
+@Converter<IMG>(IMG::class, nullable = true)
+@Interpretable("ToSpecificType")
+public fun <T, R : URL?> Convert<T, URL?>.toImg(width: Int? = null, height: Int? = null): DataFrame<T> =
+    to { it.map { url -> url?.let { IMG(url.toString(), width, height) } } }
+
+@JvmName("toImgFromUrl")
+@Refine
+@Converter<IMG>(IMG::class, nullable = false)
+@Interpretable("ToSpecificType")
+public fun <T, R : URL?> Convert<T, URL>.toImg(width: Int? = null, height: Int? = null): DataFrame<T> =
     to { it.map { IMG(it.toString(), width, height) } }
 
 // endregion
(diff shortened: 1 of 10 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
```diff
@@ -284,13 +285,38 @@ public fun <T : Any> DataColumn<T?>.convertToBoolean(): DataColumn<Boolean?> = c
 
 // region convert URL
 
-public fun <T, R : URL?> Convert<T, R>.toIFrame(
+@JvmName("toIframeFromUrlNullable")
+@Refine
+@Converter<IFRAME>(IFRAME::class, nullable = true)
```
Jolanrensen, line 290: The compiler can infer the `<>` part if you pass a `KClass` right? :)

plugins/kotlin-dataframe/testData/box/convertToShortcuts.kt
```diff
@@ -0,0 +1,21 @@
+import org.jetbrains.kotlinx.dataframe.*
+import org.jetbrains.kotlinx.dataframe.annotations.*
+import org.jetbrains.kotlinx.dataframe.api.*
+import org.jetbrains.kotlinx.dataframe.io.*
+import kotlinx.datetime.*
+
+fun box(): String {
```
Jolanrensen, line 7: woow cool :D

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
koperagen: Surprisingly, no. If i remove it, compiler complains about wrong number of type arguments

Jolanrensen: but do you use the type argument? Or could it work if you accept `KClass<*>`?

koperagen: Type argument is not needed. Initially i wanted to only use type argument, but compiler doesn't see it - only KClass get class expression. Get class works as expected without type argument

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/convert.kt
```diff
@@ -140,3 +152,31 @@ internal class To0 : AbstractInterpreter<PluginDataFrameSchema>() {
...
+
+internal abstract class AbstractToSpecificType : AbstractInterpreter<PluginDataFrameSchema>() {
+    val Arguments.functionCall: FirFunctionCall by arg(lens = Interpreter.Id)
+    val Arguments.receiver: ConvertApproximation by arg()
+
+    override fun Arguments.interpret(): PluginDataFrameSchema {
+        val converterAnnotation = functionCall.calleeReference.toResolvedFunctionSymbol()?.getAnnotationByClassId(Names.CONVERTER_ANNOTATION, session)
+        val to = converterAnnotation?.getKClassArgument(Name.identifier("klass"), session)
+        val nullable = converterAnnotation?.getBooleanArgument(Name.identifier("nullable"), session)
+        return if (to != null && nullable != null) {
```
zaleslaw, line 164: this branch is only for proper nullability binding, isn't it?

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/convert.kt
```diff
@@ -140,3 +152,31 @@ internal class To0 : AbstractInterpreter<PluginDataFrameSchema>() {
...
     }
 }
+
+internal abstract class AbstractToSpecificType : AbstractInterpreter<PluginDataFrameSchema>() {
+    val Arguments.functionCall: FirFunctionCall by arg(lens = Interpreter.Id)
+    val Arguments.receiver: ConvertApproximation by arg()
+
+    override fun Arguments.interpret(): PluginDataFrameSchema {
+        val converterAnnotation = functionCall.calleeReference.toResolvedFunctionSymbol()?.getAnnotationByClassId(Names.CONVERTER_ANNOTATION, session)
+        val to = converterAnnotation?.getKClassArgument(Name.identifier("klass"), session)
```
zaleslaw, line 162: is this an only place where we could extract info from annotations?

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/interpret.kt
```diff
@@ -139,6 +139,7 @@ fun <T> KotlinTypeFacade.interpret(
     val arguments = mutableMapOf<String, Interpreter.Success<Any?>>()
     arguments += additionalArguments
     arguments += typeArguments
+    arguments[THIS_CALL] = Interpreter.Success(functionCall)
```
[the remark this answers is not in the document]
koperagen: I want to access FirFunctionCall from this new interpreter. Normally interpreters only look at provided arguments, but in this case we want to access `@Converter` annotation on function declaration

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
```diff
@@ -402,17 +482,51 @@ public fun DataColumn<String?>.convertToLocalTime(
     return map { it?.let { converter(it.trim()) ?: error("Can't convert `$it` to LocalTime") } }
 }
 
+@JvmName("toLocalTimeFromTLongNullable")
+@Refine
+@Converter(LocalTime::class, nullable = true)
```
zaleslaw, line 487: It's kind of pattern to define type info for outputs, isn't it? a-la @ReturnType as a more common approach

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/convert.kt
koperagen: if i understand correctly, yes

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/convert.kt
koperagen: for now it's the first interpreter that relies on annotations - to reduce number of classes. imagine 20 interpreters for each to* function instead of these 3

PR #1122, opened 2025-04-09
Median overhaul

Jolanrensen:

WIP

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
@@ -59,10 +59,10 @@ public inline fun <T, reified R : Comparable<R & Any>?> DataColumn<T>.maxOfOrNul
 // region DataRow
 
 @Deprecated(ROW_MAX_OR_NULL, level = DeprecationLevel.ERROR)
-public fun AnyRow.rowMaxOrNull(): Any? = error(ROW_MAX_OR_NULL)
+public fun AnyRow.rowMaxOrNull(): Nothing? = error(ROW_MAX_OR_NULL)
 
 @Deprecated(ROW_MAX, level = DeprecationLevel.ERROR)
-public fun AnyRow.rowMax(): Any = error(ROW_MAX)
+public fun AnyRow.rowMax(): Nothing = error(ROW_MAX)
 
 public inline fun <reified T : Comparable<T>> AnyRow.rowMaxOfOrNull(skipNaN: Boolean = skipNaN_default): T? =
     Aggregators.max<T>(skipNaN).aggregateOfRow(this) { colsOf<T?>() }
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/median.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/median.kt
@@ -34,12 +34,12 @@ public fun <T : Comparable<T & Any>?> DataColumn<T>.medianOrNull(): T? =
     Aggregators.median<T>().aggregateSingleColumn(this)
 
 public fun <T> DataColumn<T>.median(
-    skipNaN: Boolean = skipNaN_default,
+    skipNaN: Boolean = skipNaNDefault,
 ): Double
     where T : Comparable<T & Any>?, T : Number? = medianOrNull(skipNaN = skipNaN).suggestIfNull("median")
 
 public fun <T> DataColumn<T>.medianOrNull(
-    skipNaN: Boolean = skipNaN_default,
+    skipNaN: Boolean = skipNaNDefault,
 ): Double?
     where T : Comparable<T & Any>?, T : Number? = Aggregators.median<T>(skipNaN).aggregateSingleColumn(this)
 
@@ -52,14 +52,14 @@ public inline fun <T, reified R : Comparable<R & Any>?> DataColumn<T>.medianOfOr
 ): R? = Aggregators.median<R>().aggregateOf(this, expression)
 
 public inline fun <T, reified R> DataColumn<T>.medianOf(
-    skipNaN: Boolean = skipNaN_default,
+    skipNaN: Boolean = skipNaNDefault,
     crossinline expression: (T) -> R,
 ): Double
     where R : Comparable<R & Any>?, R : Number? =
     medianOfOrNull(skipNaN, expression).suggestIfNull("medianOf")
 
 public inline fun <T, reified R> DataColumn<T>.medianOfOrNull(
-    skipNaN: Boolean = skipNaN_default,
+    skipNaN: Boolean = skipNaNDefault,
     crossinline expression: (T) -> R,
 ): Double?
     where R : Comparable<R & Any>?, R : Number? =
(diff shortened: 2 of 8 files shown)


Jolanrensen:
https://youtrack.jetbrains.com/issue/KT-76683 is holding me back

Jolanrensen:
@zaleslaw https://github.com/Kotlin/dataframe/blob/median/plugins/kotlin-dataframe/testData/box/groupBy_median.kt now fails because the return type of median changed to:
```
comparable -> itself or null
primitive number -> double or null
```
Shall I ignore the test for now so you can cover it after it's merged?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/statistics.kt
```diff
@@ -153,52 +153,52 @@ class StatisticsTests {
...
+        median211 shouldBe 22.0
 
         // scenario #2.2: two columns with new name - schema changes but via columnSelector
         val res22 = personsDf.groupBy("city").median(name = "newAge") { "age"<Int>() and "yearsToRetirement"<Int>() }
         res22.columnNames() shouldBe listOf("city", "newAge")
 
-        val median221 = res22["newAge"][0] as Int
-        median221 shouldBe 32
+        val median221 = res22["newAge"][0] as Double
+        median221 shouldBe 32.5
```
zaleslaw, line 194: Cool!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/math/median.kt
```diff
@@ -1,10 +1,169 @@
...
+
+        type == typeOf<Long>() ->
+            logger.warn { "Converting Longs to Doubles to calculate the median, loss of precision may occur." }
+
+        // this means the sequence is empty
+        type == nothingType -> return null
+    }
+
+    // propagate NaN to return if they are not to be skipped
+    if (type.canBeNaN && !skipNaN && any { it.isNaN }) return Double.NaN
```
zaleslaw, line 58: Logic here and next 3 rows is really hard to understand and follow, are you sure, that it covers all the required mixed situation Double.NaN + null, could it be here?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/median.kt
```diff
@@ -13,231 +15,529 @@ import org.jetbrains.kotlinx.dataframe.annotations.Refine
...
+@OverloadResolutionByLambdaReturnType
+@Suppress("UNCHECKED_CAST")
+public fun <T, C> DataFrame<T>.medianOrNull(
+    skipNaN: Boolean = skipNaNDefault,
+    columns: ColumnsSelector<T, C>,
+): Double?
+    where C : Comparable<C & Any>?, C : Number? =
+    Aggregators.medianNumbers<C>(skipNaN).aggregateAll(this, columns)
+
+public fun <T> DataFrame<T>.median(vararg columns: String, skipNaN: Boolean = skipNaNDefault): Any =
```
zaleslaw, line 182: Having Any as a return type defeats the purpose of strong typing, do we really need this?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/math/median.kt
```diff
@@ -1,10 +1,169 @@
...
+                }. Only primitive numbers or self-comparables are supported.",
+            )
+
+        type == typeOf<BigDecimal>() || type == typeOf<BigInteger>() ->
+            throw IllegalArgumentException(
+                "Cannot calculate the median for big numbers in DataFrame. Only primitive numbers are supported.",
+            )
+
+        type == typeOf<Long>() ->
+            logger.warn { "Converting Longs to Doubles to calculate the median, loss of precision may occur." }
```
zaleslaw, line 51: probably debug mode?

zaleslaw (approved):

zaleslaw:
I have a strong feeling, that we need less overloads/median* methods

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/median.kt
AndrewKis: Agree, the should be `Number` at least, no?

Jolanrensen: Median works for any comparable, also Strings, datetimes etc. So `Number` would be wrong.

The only correct return type would be `R` where `R : Comparable<R>`, but since we don't know `R`, `Any` comes closest to the actual return type for the string-api. We use the same for the column accessors in `describe()`.

We also cannot return `Comparable<Any?>` because that would imply you could call `df.median("someCol").compareTo(anything)` and that will likely break. We also cannot give `Comparable<*>/Comparable<Nothing>` because a) we cannot return `Nothing` and b) `df.median("someCol").compareTo(anything)` will then also break.

We could, however, force users to give a specific type `R`, how do you like that idea?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/math/median.kt
Jolanrensen: It's just a warning logged right? users can disable it by changing the log level for `org.jetbrains.kotlinx.dataframe.math.MedianKt` on their end or by converting their columns to `Double` manually first. It's not a heavy check, so I think debug mode is not needed.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/math/median.kt
Jolanrensen: `null` cannot occur in this function because of the type `T : Comparable<T>`. The aggregator integration already filters them out. So yes, it covers all the cases. I think it reads like a sentence: if the type of the values cán be nan and nans are not skipped, then if any is nan, return nan. In all other cases we continue and ignore nans. How would you rewrite it?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
```diff
@@ -36,12 +36,12 @@ public fun <T : Comparable<T>> DataColumn<T?>.maxOrNull(skipNaN: Boolean = skipN
 
 public inline fun <T, reified R : Comparable<R & Any>?> DataColumn<T>.maxBy(
     skipNaN: Boolean = skipNaNDefault,
-    noinline selector: (T) -> R,
+    crossinline selector: (T) -> R,
```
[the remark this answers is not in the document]
Jolanrensen: `crossline` is more efficient than `noinline` and unfortunately we cannot do it without, so yes.

Jolanrensen:
> I have a strong feeling, that we need less overloads/median* methods

Please enlighten me if you know how :)
We have the following restrictions:
- When we encounter specific number columns, the return type should be Double
- When we encounter other comparables, the return type should be the same as the input
- For float/double columns, we need a skipNaN argument.
- Just like other statistics we have all the X, of/by/for, row, and -orNull overloads for columns, dataframes, pivots, groupBy, and pivotGroupBy. Nothing really groundbreaking here

I already tried to combine overloads where possible, so if the return type of the function does not change depending on the input type (like for `DataFrame.medianFor {} -> DataRow`) I just use one overload that accepts any comparable, number or not, and has the `skipNaN` argument. In other cases, where the return type does change, I made two variants; one that returns Double and another that returns T.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/median.kt
AndrewKis: > We could, however, force users to give a specific type R, how do you like that idea?
Why not, but please right an example first, not sure how it looks.

Jolanrensen: sure, something like:
```kt
val medianAge: Int = df.median<Int>("age")
val medianLapTime: Double = df.median<Double>("lap1", "lap2", skipNaN = true)
```

PR #1123, opened 2025-04-09
Functions inline

### LINKED ISSUE #820: Add KDocs for `group`/`ungroup`
Related to #811

The [`group`/`ungroup`](https://kotlin.github.io/dataframe/groupungroupflatten.html) operations might be fun to document too.  Aside from having some text about selecting columns (for which we have a kdoc template), they should also link to each other. Group should probably also link to `groupBy` since they have a similar name.

Grammar DSL could be possible, especially for `group`

### LINKED ISSUE #1026: Functions that cán be inlined should be inlined

AndrewKis:
* Closes #1026
* Closes #820.


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/group.kt
```diff
@@ -7,17 +7,102 @@ import org.jetbrains.kotlinx.dataframe.DataFrame
...
+import org.jetbrains.kotlinx.dataframe.documentation.Indent
+import org.jetbrains.kotlinx.dataframe.documentation.LineBreak
+import org.jetbrains.kotlinx.dataframe.documentation.SelectingColumns
 import org.jetbrains.kotlinx.dataframe.impl.columnName
 import kotlin.experimental.ExperimentalTypeInference
 import kotlin.reflect.KProperty
 
 // region DataFrame
 
+/**
```
Jolanrensen, line 25: are these supposed to be part of the PR? I mean, kdocs are great in any case

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/group.kt
```diff
@@ -7,17 +7,102 @@ import org.jetbrains.kotlinx.dataframe.DataFrame
...
+ */
+internal interface GroupDocs {
+
+    /**
+     * {@comment Version of [SelectingColumns] with correctly filled in examples}
+     * @include [SelectingColumns] {@include [SetGroupOperationArg]}
+     */
+    interface GroupSelectingOptions
+
+    /**
```
Jolanrensen, line 58: This seems incomplete. it only shows the `into` part of the grammar (with some odd indents). The `group` part seems to be missing

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/group.kt
```diff
@@ -31,23 +116,110 @@ public fun <T> DataFrame<T>.group(vararg columns: KProperty<*>): GroupClause<T,
...
+ *
+ * For more information: {@include [DocumentationUrls.Group]}
+ *
+ * @include [SelectingColumns.ColumnNames]
+ *
+ * ### Examples:
+ * ```kotlin
+ * // Group selected columns into an existing column group:
+ * df.group("age", "weight").into { info }
+ * // Group selected columns into a nested column group using a path that may
```
Jolanrensen, line 186: may look better with empty newlines in between the expressions

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/ungroup.kt
```diff
@@ -6,19 +6,65 @@ import org.jetbrains.kotlinx.dataframe.DataFrame
...
+import org.jetbrains.kotlinx.dataframe.documentation.DocumentationUrls
+import org.jetbrains.kotlinx.dataframe.documentation.ExcludeFromSources
+import org.jetbrains.kotlinx.dataframe.documentation.SelectingColumns
 import org.jetbrains.kotlinx.dataframe.impl.columns.toColumnSet
 import org.jetbrains.kotlinx.dataframe.impl.removeAt
 import kotlin.reflect.KProperty
 
 // region DataFrame
 
+/**
```
Jolanrensen, line 20: perhaps also link to `group`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/describe.kt
```diff
@@ -117,12 +117,11 @@ private fun DataColumn<Any?>.convertToComparableOrNull(): DataColumn<Comparable<
         valuesAreComparable() -> asComparable()
 
         // Found incomparable number types, convert all to Double first
-        isNumber() -> cast<Number?>().map {
-            if (it?.isPrimitiveNumber() == false) {
-                // Cannot calculate statistics of a non-primitive number type
-                return@convertToComparableOrNull null
+        isNumber() -> with(cast<Number?>()) {
```
Jolanrensen, line 120: what's the benefit of this change?

AndrewKis:
So, now also closes #820.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/group.kt
AndrewKis: done

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/describe.kt
AndrewKis: reverted

Jolanrensen (approved):

Issue #1124, opened 2025-04-11
Digitize function?


Jolanrensen:
Numpy has a digitize function: https://numpy.org/doc/stable/reference/generated/numpy.digitize.html It's used to bin values from a column into given bins.

DataFrame has had [a version of digitize](https://github.com/Kotlin/dataframe/blob/0.15.0/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/digitize.kt) for a while, but it's not well tested, or used, or optimized with binary search.

It will be [removed](https://github.com/Kotlin/dataframe/pull/1125) in its current state.

Please use this issue to let us know whether you need or use something like `digitize` and how you use it so we can decide together whether DataFrame needs a function like this.

AndrewKis:
Kandy statistics contains experimental `binBy`:

<img width="826" alt="Image" src="https://github.com/user-attachments/assets/6bfb9c29-2eaa-4b15-9df1-ee480dc3c70b" />

zaleslaw:
oh, sorry

PR #1125, opened 2025-04-11
Removed digitize functionality

Jolanrensen:
Removed [our version of digitize](https://github.com/Kotlin/dataframe/blob/0.15.0/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/digitize.kt), as it's not well tested, or used, or optimized.

If you need digitize, let us know at: https://github.com/Kotlin/dataframe/issues/1124


zaleslaw:
Bye-Bye-Bye!

zaleslaw (approved):

PR #1127, opened 2025-04-14
[Compiler plugin] Support `unfold`

koperagen:


plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/unfold.kt
```diff
@@ -0,0 +1,38 @@
...
+
+class DataFrameUnfold : AbstractSchemaModificationInterpreter() {
+    val Arguments.receiver: PluginDataFrameSchema by dataFrame()
+    val Arguments.properties by ignore()
+    val Arguments.maxDepth: Int by arg(defaultValue = Present(0))
+    val Arguments.columns: ColumnsResolver by arg()
+
+    override fun Arguments.interpret(): PluginDataFrameSchema {
+        return receiver.asDataFrame().replace { columns }.with {
+            val d = it.asSimpleColumn() as? SimpleDataColumn
```
zaleslaw, line 26: d?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/unfold.kt
```diff
@@ -6,28 +6,23 @@ import org.jetbrains.kotlinx.dataframe.ColumnsSelector
...
-import org.jetbrains.kotlinx.dataframe.impl.api.createDataFrameImpl
-import org.jetbrains.kotlinx.dataframe.typeClass
+import org.jetbrains.kotlinx.dataframe.impl.api.unfoldImpl
+import kotlin.reflect.KCallable
 import kotlin.reflect.KProperty
 
-public inline fun <reified T> DataColumn<T>.unfold(): AnyCol =
-    when (kind()) {
-        ColumnKind.Group, ColumnKind.Frame -> this
+public inline fun <reified T> DataColumn<T>.unfold(vararg roots: KCallable<*>, maxDepth: Int = 0): AnyCol =
```
zaleslaw, line 16: Do we have JUnit tests for this function?

zaleslaw (approved):

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/unfold.kt
koperagen: :) fixed it

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/unfold.kt
koperagen: I added a few now

PR #1139, opened 2025-04-22
[Compiler plugin] join operations support

koperagen:
In first commit i adapted `ColumnMatch` and `ColumnList` to embed them into compiler plugin column resolving mechanism
New util functions designed to make sure all column names generated by compiler plugin in join are exactly as in runtime, and make sure there are no missing columns. They can however sometimes have nullability where runtime narrows the type to non-nullable

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ColumnsList.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ColumnsList.kt
@@ -4,7 +4,13 @@ import org.jetbrains.kotlinx.dataframe.columns.ColumnResolutionContext
 import org.jetbrains.kotlinx.dataframe.columns.ColumnSet
 import org.jetbrains.kotlinx.dataframe.columns.ColumnsResolver
 
-internal class ColumnsList<C>(val columns: List<ColumnsResolver<C>>) : ColumnSet<C> {
+public interface ColumnsList<C> : ColumnSet<C> {
+    public val columns: List<ColumnsResolver<C>>
+}
+
+internal class ColumnListImpl<C>(override val columns: List<ColumnsResolver<C>>) :
+    ColumnSet<C>,
+    ColumnsList<C> {
     constructor(vararg columns: ColumnsResolver<C>) : this(columns.toList())
 
     override fun resolve(context: ColumnResolutionContext) = columns.flatMap { it.resolve(context) }
--- a/plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/joinDsl.kt
+++ b/plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/joinDsl.kt
@@ -1,13 +1,29 @@
 package org.jetbrains.kotlinx.dataframe.plugin.impl.api
 
+import org.jetbrains.kotlinx.dataframe.api.ColumnMatch
+import org.jetbrains.kotlinx.dataframe.columns.ColumnResolutionContext
+import org.jetbrains.kotlinx.dataframe.columns.ColumnWithPath
 import org.jetbrains.kotlinx.dataframe.plugin.impl.AbstractInterpreter
 import org.jetbrains.kotlinx.dataframe.plugin.impl.Arguments
+import org.jetbrains.kotlinx.dataframe.plugin.impl.PluginDataFrameSchema
+import org.jetbrains.kotlinx.dataframe.plugin.impl.data.ColumnWithPathApproximation
 
-internal data class ColumnMatchApproximation(val left: ColumnsResolver, val right: ColumnsResolver)
+internal data class ColumnMatchApproximation(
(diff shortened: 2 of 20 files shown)


Jolanrensen (approved):
nice :)

plugins/kotlin-dataframe/testData/testUtils.kt
```diff
@@ -59,3 +62,53 @@ fun compare(runtime: DataFrameSchema, schemas: List<DataFrameSchema>, strict: Bo
     }
 }
 
+// Usual DataFrameSchema.compare is either strict comparison where both set of columns and their type must be the same
+// or subtype relation where subset of columns can vary.
+// This checks that schemas have same set of columns, but compile time columns can be nullable where runtime is narrowed to non-nullable
+
+sealed interface Mismatch
+data class AcceptableMismatch(val path: ColumnPath, val compile: KType, val runtime: KType) : Mismatch
```
zaleslaw, line 70: Probably the nullability fact could be somehow reflected in naming, like NullabilityMismatch or something like this

plugins/kotlin-dataframe/testData/box/joinWithKinds.kt
```diff
@@ -0,0 +1,44 @@
...
+        "Alice", 15, "London", 54,
+        "Bob", 45, "Dubai", 87,
+        "Charlie", 20, "Moscow", null,
+        "Charlie", 40, "Milan", null,
+        "Bob", 30, "Tokyo", 68,
+        "Alice", 20, null, 55,
+        "Charlie", 30, "Moscow", 90,
+    )
+
+    val typed2 = dataFrameOf("name", "origin", "grade", "age")(
```
zaleslaw, line 17: Probably we need here one test for nested tables and their joins

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/joinDsl.kt
```diff
@@ -1,13 +1,29 @@
...
+import org.jetbrains.kotlinx.dataframe.api.ColumnMatch
+import org.jetbrains.kotlinx.dataframe.columns.ColumnResolutionContext
+import org.jetbrains.kotlinx.dataframe.columns.ColumnWithPath
 import org.jetbrains.kotlinx.dataframe.plugin.impl.AbstractInterpreter
 import org.jetbrains.kotlinx.dataframe.plugin.impl.Arguments
+import org.jetbrains.kotlinx.dataframe.plugin.impl.PluginDataFrameSchema
+import org.jetbrains.kotlinx.dataframe.plugin.impl.data.ColumnWithPathApproximation
 
-internal data class ColumnMatchApproximation(val left: ColumnsResolver, val right: ColumnsResolver)
+internal data class ColumnMatchApproximation(
```
zaleslaw, line 11: Could you give me a hint, what does Approximation mean. I see a lot in naming, but could not understand the idea

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ColumnsList.kt
```diff
@@ -4,7 +4,13 @@ import org.jetbrains.kotlinx.dataframe.columns.ColumnResolutionContext
 import org.jetbrains.kotlinx.dataframe.columns.ColumnSet
 import org.jetbrains.kotlinx.dataframe.columns.ColumnsResolver
 
-internal class ColumnsList<C>(val columns: List<ColumnsResolver<C>>) : ColumnSet<C> {
+public interface ColumnsList<C> : ColumnSet<C> {
+    public val columns: List<ColumnsResolver<C>>
+}
+
+internal class ColumnListImpl<C>(override val columns: List<ColumnsResolver<C>>) :
+    ColumnSet<C>,
```
zaleslaw, line 12: wow, formatting looks weird to me

zaleslaw (changes requested):
The implementation is clear for me, but I'm not sure that all required test paths are covered, could you please answer

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/joinDsl.kt
koperagen: In library we have ColumnMatch, and operations work with this class. We implement the same operation in compiler plugin, but we cannot cover 100% of usages, for example if some variable is used instead of constant. So it wouldn't be fair to just name it "ColumnMatch". Approximation indicates that it's "compile-time ColumnMatch". Something like this

plugins/kotlin-dataframe/testData/box/joinWithKinds.kt
koperagen: +

plugins/kotlin-dataframe/testData/testUtils.kt
koperagen: renamed

PR #1143, opened 2025-04-22
Add convert asColumn operation as compiler plugin friendly variant oа replace with

koperagen:
I think it's more or less common to perform column-wide operations in async / parallel context, so having such compiler plugin friendly operation is useful
Another use case, although not as handy, is creating a ColumnGroup like 
`df.convert { col }.asColumn { dataFrameOf("a" to listOf(123), "b" to listOf(321).asColumnGroup() }`


docs/StardustDocs/topics/convert.md
```diff
@@ -36,6 +36,21 @@ df.convert { name }.asFrame { it.add("fullName") { "$firstName $lastName" } }
...
 <!---END-->
 
+Similar to `replace with` operation, 
+columns can be converted in a compiler plugin-friendly fashion whenever you need operation on the entire column.
+
+<!---FUN convertAsColumn-->
+
+```kotlin
+df.convert { name }.asColumn { col ->
+    col.toList().parallelStream().map { it.toString() }.collect(Collectors.toList()).toColumn()
```
zaleslaw, line 46: Parallel streams should be avoided in **single-threaded libraries** because they can introduce race conditions, synchronization issues, and unnecessary overhead. 

_However,_ in this case, the use of` parallelStream` is localized and safe, as it only transforms column values without affecting global state or column names.

But! Even though this use of parallelStream is logically safe, it can unexpectedly increase CPU load, especially on weaker machines. Parallel streams use the shared ForkJoinPool, which may cause performance issues if the system has limited resources or is already running other parallel tasks. This can lead to slowdowns or contention for threads, impacting the overall responsiveness of the application.

When running in **Kotlin Notebooks**, parallel streams can compete with notebook execution and UI rendering for limited CPU resources. This may cause lags or freezes, especially in constrained environments like containers or shared servers. Therefore, parallelism in notebooks should be used carefully to avoid degrading the interactive experience.

koperagen: Usually i wouldn't recommend using parallel streams for any trivial operations, but there can be a situation when it's needed. For example, one time i used a library that performs IO, parses file into kind of AST. So
```
df.add("data") { 
  Library.parse(file)
}
```
In my case single threaded execution took literally minutes of real time because CPU was loaded 5%. Opting to use parallel made it 20 times faster. https://github.com/Kotlin/dataframe/issues/723. But in any case, outside parallel operations we should minimize number of operations that plugin can't understand. replace with is one of them, convert asColumn would be an alternative should users need it

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
```diff
@@ -129,6 +133,21 @@ public fun <T, C, R> Convert<T, DataRow<C>>.asFrame(
...
 ): DataFrame<T> = to { body(this, it.asColumnGroup()).asColumnGroup(it.name()) }
 
+/**
+ * Compiler plugin-friendly variant of [ReplaceClause.with]
+ * [ReplaceClause.with] allows to change both column types and names.
+ * Tracking of column name changes in arbitrary lambda expression is unreliable and generally impossible
+ * to do statically.
+ * This function ensures that all column names remain as is and only their type changes to [R]
+ * Example:
+ * df.convert { colsOf<MyClass>() }.asColumn { it.asList().parallelStream().map { }.toList().toColumn() }
```
Jolanrensen, line 143: please put the example in a code block

Jolanrensen:
How is `asColumn` different from `public fun <T, C> Convert<T, C>.to(columnConverter: DataFrame<T>.(DataColumn<C>) -> AnyBaseCol): DataFrame<T>` we already have? Just an extra type `R` right?

Then I'd either add `R` to `convert.to { }` or deprecate `to {}` in favor of the better named `asColumn {}`. wdyt?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/replace.kt
```diff
@@ -53,6 +53,10 @@ public fun <T, C> ReplaceClause<T, C>.with(newColumns: List<AnyCol>): DataFrame<
 }
 
 // TODO: Issue #418: breaks if running on ColumnGroup and its child
+
+/**
+ * @see [Convert.asColumn] for an alternative supported in the compiler plugin
```
Jolanrensen, line 58: using `@see` only renders the reference, the text after it is invisible in the IDE :)

Jolanrensen:
Actually, why don't we just add the type `R` to `ReplaceClause.with` if you want compiler plugin support for it?

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/convert.kt
```diff
@@ -151,6 +158,20 @@ internal class To0 : AbstractInterpreter<PluginDataFrameSchema>() {
...
 
+internal class ConvertAsColumn : AbstractSchemaModificationInterpreter() {
+    val Arguments.receiver: ConvertApproximation by arg()
+    val Arguments.typeArg0: TypeApproximation by arg()
+    val Arguments.typeArg2: TypeApproximation by arg()
+    val Arguments.type: TypeApproximation by type(name("columnConverter"))
+
+    override fun Arguments.interpret(): PluginDataFrameSchema {
+        return receiver.schema.asDataFrame()
+            .replace { receiver.columns.map { it.toPath() }.toColumnSet() }
```
Jolanrensen, line 169: it cannot call convert {}?

Jolanrensen (changes requested):

koperagen:
> Then I'd either add R to convert.to { } or deprecate to {} in favor of the better named asColumn {}. wdyt?
The major difference is ignoring name changes. So, let's say `df.replace { col }.with { produceRandomColumnWithRandomName(it) }` is a valid code - but impossible for the plugin to interpret. So convert asColumn renames all new columns to original name

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
koperagen: sure!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/replace.kt
koperagen: fixed

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/convert.kt
koperagen: Right, it can call convert asColumn directly. Why not :) Changed it

koperagen:
@Jolanrensen I think because `convert { }.to { }` works differently, let's postpone decision whether it should be removed or not

Jolanrensen:
@koperagen ooh right I see. Actually I don't think many people would mind if we restricted that names cannot change for the entire `convert` operation. For that they'd need `replace`. That way we could make `convert {}.to {}` and `convert {}.asColumn {}` the same and `replace {}.with {}` different.

Issue #353, opened 2023-04-12
Add a way to create hierarchical dataframes without variables


koperagen:
So this is supported
```
// DataFrame with 2 columns and 3 rows
val df = dataFrameOf(
    "name" to listOf("Alice", "Bob", "Charlie"),
    "age" to listOf(15, 20, 100)
)
```

But if you want to create columnGroup, you'll have to completely rewrite your code:
```
val name by columnOf(...) or val name = listOf().toColumn()
val age = ...
val group by columnOf(name, age)
val df = dataFrameof(group)
```

zaleslaw:
We add you @AndreiKingsley to be a watcher here, not the implementer

koperagen:
I have two options in mind
```
dataFrameOf {
  "name" from listOf("Alice", "Bob", "Charlie")
  "age" from listOf(15, 20, 100)
  "group" { 
     "name" from listOf("Alice", "Bob", "Charlie")
      "age" from listOf(15, 20, 100)
  }
}
```

Or
```
val df = dataFrameOf(
    "name" to listOf("Alice", "Bob", "Charlie"),
    "age" to listOf(15, 20, 100),
    "group" to columnGroupOf(
          "name" to listOf("Alice", "Bob", "Charlie"),
          "age" to listOf(15, 20, 100),
    )
)
```
But biggest disadvantage, columnGroupOf must return List<AnyRow> (which contradicts its name). I like second option more because it's simple.

AndreiKingsley:
@koperagen what's problem with
```
val df = dataFrameOf(
    "name" to columnOf("Alice", "Bob", "Charlie"),
    "age" to columnOf(15, 20, 100),
    "group" to columnGroupOf(
          "name" to listOf("Alice", "Bob", "Charlie"),
          "age" to listOf(15, 20, 100),
    )
)
```

koperagen:
> @koperagen what's problem with
> 
> ```
> val df = dataFrameOf(
>     "name" to columnOf("Alice", "Bob", "Charlie"),
>     "age" to columnOf(15, 20, 100),
>     "group" to columnGroupOf(
>           "name" to listOf("Alice", "Bob", "Charlie"),
>           "age" to listOf(15, 20, 100),
>     )
> )
> ```

Nice idea

koperagen:
more consistent api:
```
val df = dataFrameOf(
    "name" to columnOf("Alice", "Bob", "Charlie"),
    "age" to columnOf(15, 20, 100),
[...]

PR #1144, opened 2025-04-22
Add a constructor to create a nested dataframe from columns inplace

koperagen:
I believe it will be a very useful addition for samples, simplifies many of our existing examples that use dataFrameOf(vararg columns: AnyBaseCol) + variables (see the issue) and more robust compiler plugin support
https://github.com/Kotlin/dataframe/issues/353. We need it to migrate our docs before 1.0

Another example here: 
https://github.com/Kotlin/dataframe/blob/a0c1a185c43c95f648367b24a90c4e2c10492b3f/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/TestBase.kt#L57

### DIFF
--- a/docs/StardustDocs/topics/createDataFrame.md
+++ b/docs/StardustDocs/topics/createDataFrame.md
@@ -44,6 +44,23 @@ val df = dataFrameOf(
 
 <!---END-->
 
+Create DataFrame with nested columns inplace:
+
+<!---FUN createNestedDataFrameInplace-->
+
+```kotlin
+// DataFrame with 2 columns and 3 rows
+val df = dataFrameOf(
+    "name" to columnOf(
+        "firstName" to columnOf("Alice", "Bob", "Charlie"),
+        "lastName" to columnOf("Cooper", "Dylan", "Daniels"),
+    ),
+    "age" to columnOf(15, 20, 100),
+)
+```
+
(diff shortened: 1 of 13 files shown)


zaleslaw:
Create a tickets for docs for every moment to be reflected and make it higher probability to be not abandoned
Or do it in one PR (this PR)

docs/StardustDocs/topics/createDataFrame.md
```diff
@@ -44,6 +44,23 @@ val df = dataFrameOf(
...
 
+Create DataFrame with nested columns inplace:
+
+<!---FUN createNestedDataFrameInplace-->
+
+```kotlin
+// DataFrame with 2 columns and 3 rows
+val df = dataFrameOf(
+    "name" to columnOf(
+        "firstName" to columnOf("Alice", "Bob", "Charlie"),
```
zaleslaw, line 55: Cool nested tree, I'd like this, seems like DSL could be used here also

koperagen: We could try to design such DSL after 1.0, as a better alternative to existing DataFrameBuilder functions like fill, withValues, etc, etc

AndrewKis (commented):
Nice!

PR #1148, opened 2025-04-23
Add support for DataFrame `sum` operation with tests

zaleslaw:


plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/statistics.kt
```diff
@@ -0,0 +1,124 @@
...
+        primitiveTypeMap[coneType.lookupTag.classId.asString()]
+            ?.createType(nullable = nullable)
+    }
+}
+
+fun KType.toConeKotlinType(): ConeKotlinType? {
+    val kClass = this.classifier as? KClass<*> ?: return null
+    val classId = kClass.toClassId() ?: return null
+
+    return ConeClassLikeTypeImpl(
```
koperagen, line 63: There's FirSession available inside Arguments scope
session.builtinTypes.doubleType.coneType

For class id, there's StandardClassIds
And, if needed, there's ClassId.constructClassLikeType instead of calling ConeClassLikeTypeImpl directly
Please use these APIs

plugins/kotlin-dataframe/tests-gen/org/jetbrains/kotlin/fir/dataframe/DataFrameBlackBoxCodegenTestGenerated.java
```diff
@@ -233,6 +233,12 @@ public void testFlexibleReturnType() {
     runTest("testData/box/flexibleReturnType.kt");
   }
 
+  @Test
+  @TestMetadata("sum.kt")
```
koperagen, line 237: After you create new file in testData, please use this task to update
image

plugins/kotlin-dataframe/testData/box/sum.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    // scenario #1: particular column
+    val res1 = personsDf.sumFor { age }
+    val sum11: Int? = res1.age
+
+    // scenario #1.1: particular column with converted type
+    val res11 = personsDf.sumFor { dependentsCount }
+    val sum111: Int? = res11.dependentsCount
+
+    // scenario #2: sum of all values in two columns of Int hierarchy of types
```
koperagen, line 49: Let's remove scenario 2 from plugin tests for DataFrame.sum?  It returns Number so no schema changes

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/Aggregator.kt
```diff
@@ -37,12 +37,11 @@ import kotlin.reflect.full.withNullability
  * @param Return The type of the resulting value. Can optionally be nullable.
  * @see [invoke]
  */
-@PublishedApi
-internal class Aggregator<in Value : Any, out Return : Any?>(
-    val aggregationHandler: AggregatorAggregationHandler<Value, Return>,
-    val inputHandler: AggregatorInputHandler<Value, Return>,
-    val multipleColumnsHandler: AggregatorMultipleColumnsHandler<Value, Return>,
-    val name: String,
+public class Aggregator<in Value : Any, Return : Any?>(
```
Jolanrensen, line 40: why'd you remove `out`? it should only appear in the out-position of the aggregation, at least, that's the idea

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/multipleColumnsHandlers/TwoStepMultipleColumnsHandler.kt
```diff
@@ -31,7 +31,7 @@ import kotlin.reflect.KType
  *   If not supplied, the handler of the first step is reused.
  * @see [FlatteningMultipleColumnsHandler]
  */
-internal class TwoStepMultipleColumnsHandler<in Value : Any, out Return : Any?>(
+internal class TwoStepMultipleColumnsHandler<in Value : Any, Return : Any?>(
```
Jolanrensen, line 34: same here with `out`, though, I could understand the removal here because `Return` is also used as input for stepTwo

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/AggregatorOptionSwitch.kt
```diff
@@ -31,7 +30,7 @@ internal class AggregatorOptionSwitch1<in Param1, in Value : Any, out Return : A
          *   MyAggregator.Factory(param1)
          * }
          */
-        fun <Param1, Value : Any, Return : Any?> Factory(
+        internal fun <Param1, Value : Any, Return : Any?> Factory(
```
Jolanrensen, line 33: if we make everything public, the factory can also be public imo

plugins/kotlin-dataframe/testData/box/sum.kt
```diff
@@ -0,0 +1,74 @@
...
+        "Frank", 55, "Dubai", 78.9, "1.35", 10, 35.toShort(), 2.toByte(), 145000L,
+        "Grace", 29, "Moscow", 67.8, "1.65", 36, 5.toShort(), 1.toByte(), 70000L,
+        "Hank", 60, "Paris", 80.22, "1.75", 5, 40.toShort(), 4.toByte(), 200000L,
+        "Isla", 22, "London", 75.1, "1.85", 43, 1.toShort(), 0.toByte(), 30000L,
+    )
+
+    // scenario #0: all numerical columns
+    val res0 = personsDf.sum()
+
+    val sum01: Int? = res0.age
```
Jolanrensen, line 34: do all accessors become nullable? hmm, that doesn't seem ideal. @koperagen, do we have some compile-time

koperagen: no, but i think it can't be helped

koperagen: alternatively, do we really return nullable int for an empty dataframe? maybe we should throw an exception instead?
but we should be careful about GroupBy.sum

Jolanrensen: See https://github.com/Kotlin/dataframe/issues/961

Jolanrensen: For min/max/median/percentile, the result is nullable however, but only null if the input is empty.

@koperagen, is there a compile-time check for when a DF is empty or not? Otherwise we could generate the result-type more tightly.

koperagen: There's no such check

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/statistics.kt
```diff
@@ -0,0 +1,124 @@
...
+    return inputColumns.map { col -> createUpdatedColumn(col, statisticAggregator) }
+}
+
+private fun Arguments.createUpdatedColumn(
+    column: SimpleDataColumn,
+    statisticAggregator: Aggregator<Number, Number>
+): SimpleCol {
+    val originalType = column.type.type
+    val inputKType = originalType.toKType()
+    val resultKType = inputKType?.let { statisticAggregator.calculateReturnType(it, emptyInput = false) }
```
Jolanrensen, line 83: might cause issues for min/max/median/percentile to set `emptyInput = false`. I'd set it to `true`. The name `emptyInput` is also kind of a misnomer, it should be "inputPotentiallyEmpty" or something.
It changes `intCol.min()` from `Int` (when it's set to false) to `Int?` (when it's set to true).
Feel free to rename the argument everywhere it's used :)

plugins/kotlin-dataframe/testData/box/sum.kt
Jolanrensen: > alternatively, do we really return nullable int for an empty dataframe? maybe we should throw an exception instead? but we should be careful about how it will affect GroupBy.sum

No we don't, the result of sum is never null, the sum of an empty Int column is 0. Even if we have a column of type `Nothing?` we have a non-null return type, defaulting to 0.0.

Jolanrensen: > There's no such check

Okay then for those we can follow what I wrote here https://github.com/Kotlin/dataframe/pull/1148#discussion_r2058186955

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/Aggregator.kt
zaleslaw: Thanks, I will check twice, could I return the previous version or not, there were some problems with out types when I made it public

plugins/kotlin-dataframe/tests-gen/org/jetbrains/kotlin/fir/dataframe/DataFrameBlackBoxCodegenTestGenerated.java
zaleslaw: Missed it, thanks

plugins/kotlin-dataframe/testData/box/sum.kt
zaleslaw: Agree

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/AggregatorOptionSwitch.kt
zaleslaw: I will check this place twice, it was a tricky thing to make public not all the stuff

Issue #1121, opened 2025-04-09
Percentile/quantile estimation types


Jolanrensen:
Helps https://github.com/Kotlin/dataframe/issues/961
Follow up for https://github.com/Kotlin/dataframe/issues/543 and https://github.com/Kotlin/dataframe/pull/1060

Our current percentile implementation is based on the [Nearest Rank](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method) with rounding down for non-numbers and a form of [linear interpolation](https://en.wikipedia.org/wiki/Percentile#The_linear_interpolation_between_closest_ranks_method) for number types.
However, this implementation makes some bold assumptions that can influence the expected outcome, especially for smaller data sizes.

Small background into the topic:
If I mention [quantiles](https://en.wikipedia.org/wiki/Quantile); they are the generic form of cutting up a range into continuous intervals with equal probabilities (sub-range size). A [percentile](https://en.wikipedia.org/wiki/Percentile) is thus a 100-quantile; a subdivision of a range into a hundred pieces of equal size. A [quartile](https://en.wikipedia.org/wiki/Quantile#Examples) is like a 4-quantile. A [median](https://en.wikipedia.org/wiki/Median) can thus be seen as 2-quantile or the 2nd quartile, or 50th percentile, etc.

[...]

PR #1149, opened 2025-04-25
Percentile

Jolanrensen:

- rewrites percentile implementation using quantile implementation (R3 and R8) (https://github.com/Kotlin/dataframe/issues/1121)
- adds quantile implementation
  - based on https://en.wikipedia.org/wiki/Quantile
  - and Hyndman and Fan (1996) Sample Quantiles in Statistical Packages. The American Statistician, 50, 361-365. [doi.org/10.2307/2684934](https://www.jstor.org/stable/2684934)
  - Comparing output in tests with Apache Commons Statistics
  - Adds and tests R1, R3, R7, and R8. Will add others later when needed. Currently we only use R3 and R8. R3 for selecting, R8 for linear interpolation (as recommended by the H&F paper). Other statistics libraries often use R7 as default, so we can switch if needed.
  - Rewrote `median` to use R3 and R8 as well

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -3486,63 +3486,108 @@ public final class org/jetbrains/kotlinx/dataframe/api/ParserOptions {
 
 public final class org/jetbrains/kotlinx/dataframe/api/PercentileKt {
 	public static final fun percentile (Lorg/jetbrains/kotlinx/dataframe/DataColumn;D)Ljava/lang/Comparable;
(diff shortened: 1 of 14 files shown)


Jolanrensen:
Might conflict with https://github.com/Kotlin/dataframe/pull/1148 but should be solvable

zaleslaw (approved):

PR #1153, opened 2025-04-28
Add statistics on DataFrame support

zaleslaw:
- [ ] std
- [ ] median
- [ ] cumSum
- [ ] percentile
- [ ] min
- [ ] max
- [ ] mean

### DIFF
--- a/plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/statistics.kt
+++ b/plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/statistics.kt
@@ -1,25 +1,23 @@
 package org.jetbrains.kotlinx.dataframe.plugin.impl.api
 
-import org.jetbrains.kotlin.fir.types.ConeClassLikeErrorLookupTag
+import org.jetbrains.kotlin.fir.types.ConeClassLikeType
+import org.jetbrains.kotlin.fir.types.ConeKotlinType
+import org.jetbrains.kotlin.fir.types.ConeNullability
+import org.jetbrains.kotlin.fir.types.constructClassLikeType
 import org.jetbrains.kotlin.fir.types.isSubtypeOf
+import org.jetbrains.kotlin.name.ClassId
+import org.jetbrains.kotlinx.dataframe.impl.aggregation.aggregators.Aggregator
 import org.jetbrains.kotlinx.dataframe.impl.aggregation.aggregators.Aggregators
 import org.jetbrains.kotlinx.dataframe.plugin.impl.AbstractSchemaModificationInterpreter
 import org.jetbrains.kotlinx.dataframe.plugin.impl.Arguments
 import org.jetbrains.kotlinx.dataframe.plugin.impl.PluginDataFrameSchema
+import org.jetbrains.kotlinx.dataframe.plugin.impl.SimpleCol
 import org.jetbrains.kotlinx.dataframe.plugin.impl.SimpleDataColumn
 import org.jetbrains.kotlinx.dataframe.plugin.impl.dataFrame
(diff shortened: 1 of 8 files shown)


plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/statistics.kt
```diff
@@ -86,39 +85,88 @@ private fun Arguments.createUpdatedColumn(
...
+    private val Arguments.receiver: PluginDataFrameSchema by dataFrame()
+
+    override fun Arguments.interpret(): PluginDataFrameSchema {
+        val resolvedColumns = receiver.columns()
+            .filterIsInstance<SimpleDataColumn>()
+            .filter { isIntraComparable(it, session) }
+
+        val newColumns = generateStatisticResultColumns(aggregator, resolvedColumns)
+
+        return PluginDataFrameSchema(receiver.columns() + newColumns)
```
koperagen, line 130: I think it should be `PluginDataFrameSchema(newColumns)`
image

image

df.compareSchemas(strict = true) can help detect if compile schema has columns that don't exist in runtime for such cases

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/statistics.kt
```diff
@@ -86,39 +85,88 @@ private fun Arguments.createUpdatedColumn(
...
     private val Arguments.receiver: PluginDataFrameSchema by dataFrame()
     private val Arguments.columns: ColumnsResolver by arg()
 
     override fun Arguments.interpret(): PluginDataFrameSchema {
         val resolvedColumns = columns.resolve(receiver).map { it.column }.filterIsInstance<SimpleDataColumn>().toList()
 
-        val newColumns = generateStatisticResultColumns(sum, resolvedColumns)
+        val newColumns = generateStatisticResultColumns(aggregator, resolvedColumns)
 
         return PluginDataFrameSchema(receiver.columns() + newColumns)
```
koperagen, line 154: Same as for Aggregator0

plugins/kotlin-dataframe/src/org/jetbrains/kotlinx/dataframe/plugin/impl/api/groupBy.kt
```diff
@@ -321,129 +313,96 @@ class GroupBySumOf : GroupByAggregatorSumOf(defaultName = "sum")
...
+abstract class GroupByAggregator0(val defaultName: String, val aggregator: Aggregator<*, *>) : AbstractSchemaModificationInterpreter() {
     val Arguments.receiver by groupBy()
     val Arguments.name: String? by arg(defaultValue = Present(null))
     val Arguments.columns: ColumnsResolver? by arg()
 
     override fun Arguments.interpret(): PluginDataFrameSchema {
         if (name == null) {
-            val resolvedColumns = columns?.resolve(receiver.keys)?.map { it.column }!!.toList()
-            return PluginDataFrameSchema(receiver.keys.columns() + resolvedColumns)
+            val resolvedColumns = columns?.resolve(receiver.keys)?.map { (it.column as SimpleDataColumn) }!!.toList()
```
koperagen, line 323: why columns is nullable? i think you can `ColumnsResolver?` => `ColumnsResolver` and remove !!

koperagen (approved):
Very cool! Please check my comment, and let's wait for the build to finish. Otherwise looks good

koperagen:
```
val df = dataFrameOf("a")(1, 2, 3)
df.groupBy { a named "b" }.sum { a }
```
Do you have a test for this?
it'll be columns `b` and `a` - surprisingly when only one column is selected in sum(ColumnsSelector), aggregated column will have its name

Issue #1156, opened 2025-04-28
Publish core without AccessOverloadApi annotated functions by default and additional core-compact artifact


koperagen:

koperagen:
Some of these could be somewhat blocked because proper replacement would require compiler plugin to be available in project / notebook. But i recommend to just remove / replace with String API whenever possible, and discuss when in doubts

Jolanrensen:
> Some of these could be somewhat blocked because proper replacement would require compiler plugin to be available in project / notebook. But i recommend to just remove / replace with String API whenever possible, and discuss when in doubts

During KotlinConf and my demos/the workshop I indeed noticed a lot of deprecations now appear in notebooks. It's not ideal to have the safest possible path be deprecated while we're waiting for the compiler plugin.

As for our DataFrame project itself, I'd advise not adding the compiler plugin. The bootstrapping of the Gradle plugin has caused many issues (and it's still causing issues atm). Instead, we can generate extensions we need manually or rely on strings indeed :).

zaleslaw:
@koperagen what steps do we need to count this issue closed?

PR #1162, opened 2025-04-29
Migrate usages of AccessOverload APIs in inline functions to public alternatives

koperagen:
Related to https://github.com/Kotlin/dataframe/issues/1156
With Public API modifier enabled, such calls to AccessOverload APIs won't compile. We cannot make them PublishedAPI


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/testSets/person/DataFrameTests.kt
```diff
@@ -1177,8 +1177,9 @@ class DataFrameTests : BaseTest() {
             .split { others }.intoRows()
             .add(sum) { name.length + other().length }
 
-        val matrix = src.pivot { other }.groupBy { name }.with { sum }
+        val matrix = src.pivot { other }.groupBy { name }.with { sum() }
```
Jolanrensen, line 1180: why are these brackets now needed?

Issue #1151, opened 2025-04-26
Docs after json extraction


Jolanrensen:
Website needs to be updated for custom gradle setup after:
https://github.com/Kotlin/dataframe/pull/1147

PR #1164, opened 2025-04-30
json module docs

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1151

### DIFF
--- a/docs/StardustDocs/topics/gettingStartedGradleAdvanced.md
+++ b/docs/StardustDocs/topics/gettingStartedGradleAdvanced.md
@@ -124,6 +124,7 @@ dependencies {
     // Artifact containing all APIs and implementations
     implementation("org.jetbrains.kotlinx:dataframe-core:%dataFrameVersion%")
     // Optional formats support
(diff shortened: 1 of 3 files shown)


docs/StardustDocs/topics/gettingStartedGradleAdvanced.md
```diff
@@ -168,6 +171,16 @@ dependencies {
 
 </tabs>
 
+<warning>
```
koperagen, line 174: I remember recommendation to not use warnings in similar cases

Jolanrensen: which tags are available to use instead?

koperagen: I think plain text is good

Jolanrensen: There's also `note`: https://www.jetbrains.com.cn/en-us/help/writerside/semantic-markup-reference.html#note

PR #1172, opened 2025-05-05
Deprecate `@AccessApiOverload` functions

koperagen:


Jolanrensen:
@koperagen After experiencing the result of this change in the current state of notebooks (without compiler plugin) I did notice I was getting a lot of deprecation warnings :/. While REPL K2 + plugin is not available yet, this might give the wrong message to our users; that they're doing something wrong, which column accessors or KProperties are the safest way, currently, in notebooks if you don't want to split cells...

Maybe we would need to publish two versions, one for gradle with deprecations, one for notebooks without deprecations. At least until the compiler plugin works there as well.

PR #1186, opened 2025-05-11
Fixed issue on Top 12 German Companies sample

kikoso:
This PR modifies the handling of the ROA and ROE columns in the 12 German Companies example to address formatting issues. The values were initially strings with periods (e.g., 12.5), which caused incorrect calculations. This PR removes the periods and converts the values to Double to ensure accurate computation of the mean and standard deviation.

### DIFF
--- a/examples/notebooks/top_12_german_companies/top_12_german_companies.ipynb
+++ b/examples/notebooks/top_12_german_companies/top_12_german_companies.ipynb
@@ -6552,10 +6552,10 @@
    "source": [
     "// Group data by sector to compute average and standard deviations of ROA and ROE\n",
     "val roeAndRoaDf = companiesDf.groupBy { sector }.aggregate {\n",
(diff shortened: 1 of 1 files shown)


Jolanrensen:
In the cell you modified:
```kts
// Group data by sector to compute average and standard deviations of ROA and ROE
val roeAndRoaDf = companiesDf.groupBy { sector }.aggregate {
    ROA.mean() into "Avg ROA"
    ROA.std() into "Std ROA"
    ROE.mean() into "Avg ROE"
    ROE.std() into "Std ROE"
}

roeAndRoaDf
```
the columns `ROA` and `ROE` are both of type `Double`, not `String`.
They are `Double` already from the 4th cell and the line: `.convert { ROA and ROE }.with { it.replace(".", "").toDouble() }`, so the calculation is done correctly.

PR #1197, opened 2025-05-15
Prepare Split for compiler plugin support

koperagen:


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/split.kt
```diff
@@ -62,22 +66,27 @@ public typealias ColumnNamesGenerator<C> = ColumnWithPath<C>.(extraColumnIndex:
 
 // region default
 
+@Interpretable("SplitDefault")
 public inline fun <T, C : Iterable<R>, reified R> Split<T, C>.default(value: R?): SplitWithTransform<T, C, R> =
     by { it }.default(value)
 
+@Deprecated(SPLIT_STR, ReplaceWith("""by(",").default(value)"""))
```
Jolanrensen, line 73: `ReplaceWith` works better with explicit `this.` in it. Also, The text inside it can be put in the deprecations.kt file as well if you like.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt
```diff
@@ -160,6 +160,8 @@ internal const val TO_CSV = "toCsv() is deprecated in favor of toCsvStr() in dat
 internal const val TO_CSV_IMPORT = "org.jetbrains.kotlinx.dataframe.io.toCsvStr"
 internal const val TO_CSV_REPLACE = "this.toCsvStr()"
 
+internal const val SPLIT_STR = "Shortcut can be trivially replaced"
```
Jolanrensen, line 163: This isn't very descriptive. As a user I'd expect reading something like "This shortcut is deprecated in favor of `.by(",").into {}`". That makes it immediately clear what to do about it :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/split.kt
Jolanrensen: Also, it used to be done with `splitDefault()`, which does a trim as well: `split(",").map { it.trim() }`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt
koperagen: I don't like existing message that much, but i also don't think it's worth to repeat built-in deprecation actions:
image

image

koperagen: do you have another idea for the message?

Jolanrensen: that looks different than I remember... hmmm

koperagen: so is it ok?

Jolanrensen: "Please explicitly specify how the `String` should be split. This shortcut will be removed in version xxx" (where xxx specifies in which version it will be ERROR)

Issue #1222, opened 2025-05-27
Jupyter: Data schema generation bug with mixed nullability of similarly named column


Jolanrensen:
Can be reproduced on 1.0.0-Beta2:

```kts
val df1 = dataFrameOf(
    "group" to columnOf(
        "a" to columnOf(1, null, 3),
    )
)
val df2 = dataFrameOf(
    "group" to columnOf(
        "a" to columnOf(1, 2, 3),
    )
)
```
produces:
```kt
// for df1

@DataSchema(isOpen = false)
interface _DataFrameType1 {
    val a: Int?
}

@DataSchema
interface _DataFrameType {
    val group: _DataFrameType1
}

// for df2

@DataSchema(isOpen = false)
interface _DataFrameType3 {
    val a: Int
}

@DataSchema
interface _DataFrameType2 : _DataFrameType {
    override val group: _DataFrameType3 // Type of 'group' is not a subtype of overridden property 'val group: _DataFrameType1' defined in '_DataFrameType'
}
```

What I suspect was meant to be generated is something like this:
```kt
@DataSchema(isOpen = false)
interface _DataFrameType1 {
    val a: Int?
}

@DataSchema
interface _DataFrameType {
    val group: _DataFrameType1
}

// for df2

@DataSchema(isOpen = false)
interface _DataFrameType3 : _DataFrameType1 { // now the non-nullable variant extends the nullable variant
    override val a: Int // requires override
}

@DataSchema
interface _DataFrameType2 : _DataFrameType {
    override val group: _DataFrameType3
}
```

PR #1223, opened 2025-05-27
Potential bug fix for nested types codegen jupyter

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1222

Made it so that nested types in jupyter codegen follow their parent's openness. This makes it so that:
```kt
val df1 = dataFrameOf(
    "group" to columnOf(
        "a" to columnOf(1, null, 3),
    )
)
val df2 = dataFrameOf(
    "group" to columnOf(
        "a" to columnOf(1, 2, 3),
    )
)
```
produces something like:
```kt
// for df1

@DataSchema(isOpen = true)
interface _DataFrameType1 {
    val a: Int?
}

@DataSchema
interface _DataFrameType {
    val group: _DataFrameType1
}

// for df2

@DataSchema(isOpen = true)
interface _DataFrameType3 : _DataFrameType1 { // now the non-nullable variant extends the nullable variant
    override val a: Int // requires override
}

@DataSchema
interface _DataFrameType2 : _DataFrameType {
    override val group: _DataFrameType3
}
```

However, the old behavior (whenever a schema is generated, nested types, such as inside column groups and frame cols are generated with isOpen = false, even if the data schema itself has isOpen = true) has been like this since at least [2021](https://github.com/Kotlin/dataframe/commit/56155264e0454e9b47245193e774fd87b322d718#diff-e8077b565373f8f859df71a7675fdeb42c0c152878a21140923c3d9e10bd697bR65).

This change makes [this test](https://github.com/Kotlin/dataframe/blob/5f5f6a1c253e609666bcb1f0665342d7fa132f08/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/ReplCodeGenTests.kt#L270) fail, which indicates to me that it might have been done intentionally. @koperagen could you maybe explain a bit more about this behavior now that we know it can cause bugs like these?

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/SchemaProcessorImpl.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/SchemaProcessorImpl.kt
@@ -58,6 +58,7 @@ internal class SchemaProcessorImpl(
     private fun generateFields(
         schema: DataFrameSchema,
         visibility: MarkerVisibility,
+        openNestedDataSchemas: Boolean,
         requiredSuperMarkers: List<Marker> = emptyList(),
     ): List<GeneratedField> {
         val usedFieldNames =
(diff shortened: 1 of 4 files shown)


Jolanrensen:
>This change makes [this test](https://github.com/Kotlin/dataframe/blob/5f5f6a1c253e609666bcb1f0665342d7fa132f08/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/ReplCodeGenTests.kt#L270) fail, which indicates to me that it might have been done intentionally. @koperagen could you maybe explain a bit more about this behavior now that we know it can cause bugs like these?

Probably this was done to limit the number of inter-schema comparisons. It may be a better solution to add isOpen=false to the parent data schemas if nested schemas don't match exactly

PR #1233, opened 2025-06-03
fix kodex docs iframes

AndreiKingsley:


Jolanrensen:
Let's do this with a KoDEx version bump after I merge the fix there

Jolanrensen:
okay, if you bump KoDEx to 0.4.4, the fix should be applied

PR #1237, opened 2025-06-04
fix escaping in iframes

AndreiKingsley:


Jolanrensen (approved):
cool! thanks, please fix the linting, then you can merge

PR #1238, opened 2025-06-05
various docs fixes to lower the number of errors

Jolanrensen:


docs/StardustDocs/topics/ColumnSelectors.md
```diff
@@ -51,15 +51,15 @@ or `ColumnSet` that adheres to the optional given condition. If no column adhere
 ##### Col {collapsible="true"}
 `col(name)`, `col(5)`, `this[5]`
 
-Creates a [ColumnAccessor](DataColumn.md#column-accessors) (or `SingleColumn`) for a column with the given
```
AndreiKingsley, line 54: Should we add a short explanation about `ColumnAccessor` and its difference with `DataColumn`?

Jolanrensen: Maybe... But I'm not sure where. Should it be in DataColumn or here in the selectors?

Jolanrensen: then again... it's not really an accessor like we made before. It's just a resolver, maybe we could better rephrase it in the entire file

Jolanrensen: https://github.com/Kotlin/dataframe/issues/1240

PR #1247, opened 2025-06-11
add nulls parse option in readExcel

### LINKED ISSUE #1166: Incorrect XLSX parsing
[pjTDYzyiTmEte21D_oIlIY6HL8t_5FH7f-Global Library Data.xlsx](https://github.com/user-attachments/files/19978238/pjTDYzyiTmEte21D_oIlIY6HL8t_5FH7f-Global.Library.Data.xlsx)

After read has incorrect type in several columns - doesn't recognize empty cells as nulls but as `String`:

Image

Image

AndreiKingsley:
Fixes #1166.


dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -255,6 +271,8 @@ public fun DataFrame.Companion.readExcel(
  * when set to false, it operates as [NameRepairStrategy.MAKE_UNIQUE],
  * ensuring unique column names will make the columns be named according to excel columns, like "A", "B", "C" etc.
  * for unstructured data.
+ * @param parseEmptyAsNull when set to true, empty strings in cells are parsed as null (default true).
```
Jolanrensen, line 342: There may be more representations in strings that can refer to `null`. maybe it might be better to follow readCsv and other parsing options for this. It will be overkill to add the full ParserOptions as argument (since datetime and locales are already handled by excel), but csv, for instance uses `nullStrings: Set<String>= setOf("", "NA", "N/A", "null", "NULL", "None", "none", "NIL", "nil")`, so it might make sense to add a `nullStrings` argument here as well, probably with fewer defaults

AndreiKingsley: Created an issue:
https://github.com/Kotlin/dataframe/issues/1248

AndreiKingsley: I don't think is a good idea to introduce a new non-trivial logic just before release, need more time for implementation and testing.

Jolanrensen: okay sure, makes sense

Jolanrensen: speaking off, this is a binary-breaking change. Since we released a beta already, we cannot break the API like that without a fallback mechanism. Could you add the old function signatures back with `@Deprecated()` at level HIDDEN? That way, we won't break libraries calling the old function :)

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -59,6 +59,9 @@ public class Excel : SupportedDataFrameFormat {
         DefaultReadExcelMethod(pathRepresentation)
 }
 
+private const val MESSAGE_1_1 = "Will be ERROR in 1.1."
```
Jolanrensen: HIDDEN >= ERROR already. You can remove this message or say that they are removed in 1.1

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -265,11 +351,21 @@ public fun DataFrame.Companion.readExcel(
...
-    return readExcel(sheet, columns, formattingOptions, skipRows, rowsCount, nameRepairStrategy, firstRowIsHeader)
+    return readExcel(
+        sheet,
+        columns,
+        formattingOptions,
+        skipRows,
+        rowsCount,
+        nameRepairStrategy,
+        firstRowIsHeader,
+        parseEmptyAsNull,
```
koperagen, line 367: My 5 cents: how about enum parameter "emptyStringValue" with NULL, KEEP_AS_STRING options?

koperagen: or, if we're looking to extend this with other null strings https://github.com/Kotlin/dataframe/issues/1248, it could be some sort of sealed interface (String) -> String? with companion object { val NULL = ...; val KEEP_AS_STRING ... }

Jolanrensen: I'd like something like this too, but it would deviate even more for how we handle parsing `String -> null` with `parse()`, `ParserOptions`, and `readCsv()`. They now take a simple set of null-representing strings.
Maybe we could change it all to the same system at some point, but until then I'd keep this one as simple as possible.

koperagen: what about enum then?

Jolanrensen: an enum could work too indeed, as long as we can make a simple `@Deprecated` `ReplaceWith` eventually, I'm fine with it :)

PR #1252, opened 2025-06-13
Reusage schemas fix

### LINKED ISSUE #1222: Jupyter: Data schema generation bug with mixed nullability of similarly named column

Jolanrensen:
Fixes #1222 by replacing the `strictlyEqualNestedSchemas` parameter by a more explicit `ComparisonMode`.

Comparing two schemas in `LENIENT` mode can return `IsSuper`, `IsDerived`, `IsEqual`, or `None`.
Compating in `STRICT` mode can only return `IsEqual` or `None`, because the schema's need to exactly match.
`STRICT_FOR_NESTED_SCHEMAS` works in `LENIENT` mode for the top-level, but `STRICT` for nested schemas. This is often used in Jupyter notebooks, to prevent nested types from extending each other and thus avoid a potential comparison explosion. (There could be a lot of nested types)

Also, added documentation everywhere

Requires tiny patch in the compiler plugin

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/schema/ColumnSchema.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/schema/ColumnSchema.kt
@@ -6,7 +6,6 @@ import org.jetbrains.kotlinx.dataframe.DataFrame
 import org.jetbrains.kotlinx.dataframe.DataRow
 import org.jetbrains.kotlinx.dataframe.columns.ColumnKind
 import org.jetbrains.kotlinx.dataframe.schema.ComparisonMode.LENIENT
-import org.jetbrains.kotlinx.dataframe.schema.ComparisonMode.STRICT
 import org.jetbrains.kotlinx.dataframe.schema.ComparisonMode.STRICT_FOR_NESTED_SCHEMAS
 import kotlin.reflect.KType
 import kotlin.reflect.full.isSubtypeOf
@@ -58,7 +57,7 @@ public sealed class ColumnSchema {
         override val nullable: Boolean = false
         override val type: KType get() = typeOf<AnyRow>()
 
-        public fun compare(other: Group, comparisonMode: ComparisonMode = LENIENT): CompareResult =
+        public fun compare(other: Group, comparisonMode: ComparisonMode = STRICT_FOR_NESTED_SCHEMAS): CompareResult =
             schema.compare(
                 other = other.schema,
                 comparisonMode = comparisonMode,
@@ -92,7 +91,7 @@ public sealed class ColumnSchema {
         }
     }
 
-    public fun compare(other: ColumnSchema, comparisonMode: ComparisonMode = LENIENT): CompareResult {
+    public fun compare(other: ColumnSchema, comparisonMode: ComparisonMode = STRICT_FOR_NESTED_SCHEMAS): CompareResult {
         if (kind != other.kind) return CompareResult.None
         if (this === other) return CompareResult.Equals
         return when (this) {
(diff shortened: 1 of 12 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/schema/DataFrameSchema.kt
```diff
@@ -1,12 +1,17 @@
...
 
     /**
      * By default generated markers for leafs aren't used as supertypes: @DataSchema(isOpen = false)
-     * strictlyEqualNestedSchemas = true takes this into account for internal codegen logic
+     * [ComparisonMode.STRICT_FOR_NESTED_SCHEMAS] takes this into account for internal codegen logic
      */
-    public fun compare(other: DataFrameSchema, strictlyEqualNestedSchemas: Boolean = false): CompareResult
+    public fun compare(
+        other: DataFrameSchema,
+        comparisonMode: ComparisonMode = STRICT_FOR_NESTED_SCHEMAS,
```
koperagen: i think LENIENT should be default for more visibility - easier to see where our "special" codegen mode STRICT_FOR_NESTED_SCHEMAS handling is used

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/schema/ColumnSchema.kt
```diff
@@ -88,37 +88,27 @@ public abstract class ColumnSchema {
...
-    public fun compare(other: ColumnSchema): CompareResult = compare(other, false)
-
-    internal fun compareStrictlyEqualNestedSchemas(other: ColumnSchema): CompareResult = compare(other, true)
-
-    private fun compare(other: ColumnSchema, strictlyEqualNestedSchemas: Boolean): CompareResult {
+    public fun compare(other: ColumnSchema, comparisonMode: ComparisonMode = STRICT_FOR_NESTED_SCHEMAS): CompareResult {
         if (kind != other.kind) return CompareResult.None
         if (this === other) return CompareResult.Equals
         return when (this) {
             is Value -> compare(other as Value)
```
koperagen: Seems odd that comparison mode is not used here. How you tell the difference between nullable and non-nullable column?

Jolanrensen:
> Seems good, but it's quite hard to review with many stylistic / non-functional changes :(

Sorry, I just had no clue what was going on before refactoring, let alone debug how it should behave. Hopefully the new approach expresses the intention behind the code better :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/schema/DataFrameSchema.kt
Jolanrensen: I agree :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/schema/ColumnSchema.kt
Jolanrensen: That's true, I had it at one point, but it wasn't in the implementation before... Let's see what breaks if I add it back

Jolanrensen: Ah, the same behavior was achieved by `if (comparison != Equals && comparisonMode == STRICT) None else comparison` in the other file, but I improved it now so `ColumnSchema.Value` now also has a `comparisonMode` argument and the "strictness increase" is better explained.

Issue #1260, opened 2025-06-19
Improve AddDsl


AndreiKingsley:
Now API is very confusing.
Proposal:
```
df.add { df ->
    col("year of birth") { 2021 - age }
    col(df.age > 18 named "isAdult")
    colGroup("details") {
        col(df.name.lastName.length() named "last name length")
        col("full name") { name.firstName + " " + name.lastName }
    }
    col(df.colA)
}
```

PR #1261, opened 2025-06-19
`add` kdocs

### LINKED ISSUE #816: Add KDocs for `add`

The [add operation](https://kotlin.github.io/dataframe/add.html) already has some KDocs here and there, however, it does not yet use the preprocessor, it has no DSL grammar yet (which it could really use), and it's one of the most used functions in the library (AFAIK). It's a bit more difficult task though.

AndreiKingsley:
* Closes #816 
(except AddDsl #1260)

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
@@ -150,6 +150,8 @@ public typealias AddExpression<T, R> = Selector<AddDataRow<T>, R>
  * }
  * ```
  *
+ * For more information: {@include [DocumentationUrls.Add]}.
+ *
  * @param name name for a new column.
  * If it is empty, a unique column name will be generated.
  * Otherwise, it should be unique for original [DataFrame].
@@ -192,6 +194,8 @@ public inline fun <reified R, T> DataFrame<T>.add(
  * Also, you can use other methods such as [prev] and [next] to access other rows,
  * including [newValue][AddDataRow.newValue] to retrieve already computed values of this column
  * in previous rows.
+ *
+ * For more information: {@include [DocumentationUrls.Add]}.
  *
  * Returns a new [DataFrame] with the new column inserted at the given [path].
  * {@include [org.jetbrains.kotlinx.dataframe.documentation.ColumnPathCreation]}
@@ -351,6 +355,8 @@ public class AddDsl<T>(
  * }
  * ```
  *
+ * For more information: {@include [DocumentationUrls.Add]}.
+ *
  * @param body An [AddDsl] expression used to define new columns and column groups.
  * @return A new [DataFrame] with the added columns.
  */
@@ -390,6 +396,8 @@ public fun <T> DataFrame<T>.add(body: AddDsl<T>.() -> Unit): DataFrame<T> {
  *     else prev()!!.newValue<Int>() + prev()!!.prev()!!.newValue<Int>()
  * }
  * ```
+ *
+ * For more information: {@include [DocumentationUrls.Add]}.
  *
  * @param name name for a new column.
  * If it is empty, a unique column name will be generated.
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
@@ -21,6 +21,7 @@ import org.jetbrains.kotlinx.dataframe.columns.ColumnAccessor
 import org.jetbrains.kotlinx.dataframe.columns.ColumnPath
 import org.jetbrains.kotlinx.dataframe.columns.ColumnReference
 import org.jetbrains.kotlinx.dataframe.documentation.DocumentationUrls
+import org.jetbrains.kotlinx.dataframe.documentation.ExcludeFromSources
 import org.jetbrains.kotlinx.dataframe.exceptions.DuplicateColumnNamesException
 import org.jetbrains.kotlinx.dataframe.exceptions.UnequalColumnSizesException
 import org.jetbrains.kotlinx.dataframe.impl.api.insertImpl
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
```diff
@@ -15,69 +15,77 @@ import org.jetbrains.kotlinx.dataframe.Selector
...
 import org.jetbrains.kotlinx.dataframe.impl.api.insertImpl
 import org.jetbrains.kotlinx.dataframe.impl.columns.resolveSingle
 import org.jetbrains.kotlinx.dataframe.util.DEPRECATED_ACCESS_API
 import kotlin.reflect.KProperty
 
-/*
- * `add` operation adds new columns to DataFrame.
- */
-
 // region Add existing columns
```
Jolanrensen, line 32: A lot of repetitions here, but probably okay if we don't have to change it often in the future

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
```diff
@@ -115,14 +123,43 @@ public interface AddDataRow<out T> : DataRow<T> {
 public typealias AddExpression<T, R> = Selector<AddDataRow<T>, R>
 
 /**
- * Creates new column using row [expression] and adds it to the end of [DataFrame]
+ * Creates a new column using [AddExpression] and
```
Jolanrensen: either "using an [AddExpression]" or "using [expression]"

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
```diff
@@ -115,14 +123,43 @@ public interface AddDataRow<out T> : DataRow<T> {
...
+ * adds a new column to the end of this [DataFrame] (at the top level).
+ *
+ * An [AddExpression] allows to compute a value for each row in the new column
```
Jolanrensen: sounds a bit unnatural. 

With [expression] you can define the value each row in the new column needs to have. This can be based on the values from that row in the original dataframe.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
```diff
@@ -115,14 +123,43 @@ public interface AddDataRow<out T> : DataRow<T> {
...
+ * based on the values from that row in the original [DataFrame].
+ * Also, you can use other methods such as [prev] and [next] to access other rows,
+ * including [newValue][AddDataRow.newValue] to retrieve already computed values of this column
```
Jolanrensen: Combine it with `newValue()` to retrieve...

Best to tell them you need to do prev().newValue() instead of just newValue()

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
```diff
@@ -149,6 +186,38 @@ public inline fun <reified R, T> DataFrame<T>.add(
     noinline expression: AddExpression<T, R>,
 ): DataFrame<T> = add(column.path(), infer, expression)
 
+/**
```
Jolanrensen, line 197: okay I'd put this together with the other in a documentation interface. Such a large piece of text will be hard to maintain in multiple places

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
```diff
@@ -163,6 +232,10 @@ public inline fun <reified R, T> DataFrame<T>.add(
 
 // region Create and add several columns
 
+/**
+ * Receiver that is used by the [add] and [mapToFrame]
+ * for adding new columns and column groups based on [DataFrame] columns and row values.
+ */
 public class AddDsl<T>(
```
Jolanrensen, line 243: I'd probably still try to write some docs for the current state of the AddDsl. We already released beta2, so it will be part of 1.0 in this state. Could be a separate PR of course

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
```diff
@@ -253,6 +326,40 @@ public class AddDsl<T>(
     public infix fun AddGroup<T>.into(column: AnyColumnGroupAccessor): Unit = into(column.name())
 }
 
+/**
```
Jolanrensen, line 333: oh you already did! Okay, forget my note, it's fine to have the kdocs only here and leave out specific docs for the 1.0 release for each function :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
```diff
@@ -253,6 +326,40 @@ public class AddDsl<T>(
...
+ * ## Example
+ *
+ * ```kotlin
+ * val dfWithAdded = df.add {
+ *     // Add new column "yearOfBirth" computed as 2021 minus value in "age" column
+ *     "yearOfBirth" from { 2021 - age }
+ *
+ *     // Add column "is adult" with result of age > 18
+ *     age > 18 into "is adult"
+ *
```
Jolanrensen, line 350: maybe one more example using `expr {} into ""`?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
```diff
@@ -261,6 +368,47 @@ public fun <T> DataFrame<T>.add(body: AddDsl<T>.() -> Unit): DataFrame<T> {
     return dataFrameOf(this@add.columns() + dsl.columns).cast()
 }
 
+/**
+ * Creates a new column using [AddExpression] and
+ * adds a new column to the end of each group (i.e., [DataFrame]s) of this [GroupBy] (at the top level).
+ *
+ * An [AddExpression] allows to compute a value for each row in the new column
```
Jolanrensen: same unnatural sentence

Jolanrensen (changes requested):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
AndreiKingsley: IMHO it's hard to find exact repetitions that can be replaces with kodex templates 😢.

AndreiKingsley: done for add expression description

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
```diff
@@ -115,14 +124,50 @@ public interface AddDataRow<out T> : DataRow<T> {
...
 
 /**
- * Creates new column using row [expression] and adds it to the end of [DataFrame]
+ * With an [AddExpression], you define the value that each row in the new column should have.
+ * This can be based on values from the same row in the original [DataFrame].
  *
- * Original [DataFrame] is not modified.
+ * You can also use functions like [prev] and [next] to access other rows, and combine them with
+ * [newValue][AddDataRow.newValue] to reference values already computed in the new column.
+ * For example, use `prev().newValue()` to access the new column value from the previous row.
```
Jolanrensen, line 132: nice! very clear

Jolanrensen (approved):

Issue #1221, opened 2025-05-27
Missing jupyter codegen for `GroupBy` type


zaleslaw:
![image-4.png](https://github.com/user-attachments/assets/ce021c54-b0ab-41f9-a28d-afe7457fb48f)

No fields found if we mix operation + group by, but if we split in different cells operation and Group by, fields will be available

Jolanrensen:
Which versions of DataFrame (and Kandy) is this on?

zaleslaw:
It were default versions or kc@25

zaleslaw:
@belovrv could share more, also it was latest EAP

Jolanrensen:
![Image](https://github.com/user-attachments/assets/5d3f5e55-0742-449d-927e-2a67ee3da77f)

Looks like column accessor generation is not triggered at all when there's a variable with type `GroupBy`/`Grouped`, only when something is a column, row, or dataframe:

![Image](https://github.com/user-attachments/assets/90290bdb-ea19-4e2c-b152-a67392e2a91a)

koperagen:
https://github.com/Kotlin/dataframe/pull/663

Jolanrensen:
> [#663](https://github.com/Kotlin/dataframe/pull/663)

wow that's indeed exactly the same thing, just for a different reason. My attempt was a bit simpler but failed at the renaming columns part, which your solution handles better. I'll continue with yours :)

PR #1263, opened 2025-06-19
GroupBy Jupyter codegen

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1221

Uses this PR as its base: https://github.com/Kotlin/dataframe/pull/663:
> This was a long standing minor inconvenience. Having this variable in a cell1 you couldn't access `k` in keys in the cell2 cell1:
> 
> ```kotlin
> val groupBy = dataFrameOf("a")("1", "11", "2", "22")
>     .groupBy { expr { "a"<String>().length } named "k" }
> ```
> 
> cell2:
> 
> ```kotlin
> groupBy.keys.k
> ```
> 
> Other use case is adding a new column to `GroupBy`
> 
> ```kotlin
> val groupBy = dataFrameOf("a")("1", "11", "2", "22")
>     .groupBy { expr { "a"<String>().length } named "k" }
>     .add("newCol") { 42 }
> ```
> 
> ```kotlin
> groupBy.aggregate { newCol into "newCol" }
> ```

But it's now up-to-date and contains some refactoring: 
- Refactor of `CodeWithConverter` -> `CodeWithTypeCastGenerator`. It was never really a "converter", it can just generate a casting expression for an expression, so...
- I created the `TypeCastGenerator` interface (the name can still change of course). It has a specific implementation that uses the `cast<>()` functions from the DataFrame API (instead of having string literals in different places) which also supports multiple type-arguments (for GroupBy). 
- I did some additional renaming like `markerName` -> `targetTypeName` since it makes it easier to understand without a grasp of the larger context.


dataframe-jupyter/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
```diff
@@ -282,13 +293,16 @@ internal class Integration(private val notebook: Notebook, private val options:
...
+
             override val execution = FieldHandlerFactory.createUpdateExecution<Any> { instance, property ->
+                // TODO check property type first, then instance, Issue #1245
                 when (instance) {
                     is AnyCol -> updateAnyColVariable(instance, property, codeGen)
                     is ColumnGroup<*> -> updateColumnGroupVariable(instance, property, codeGen)
                     is AnyRow -> updateAnyRowVariable(instance, property, codeGen)
                     is AnyFrame -> updateAnyFrameVariable(instance, property, codeGen)
                     is ImportDataSchema -> updateImportDataSchemaVariable(instance, property)
+                    is GroupBy<*, *> -> updateGroupByVariable(instance, property, codeGen)
```
zaleslaw, line 305: I have one question here: do we need the same support for Pivot as for GroupBy? At least create a ticket and make a separate fix? or validate with test

Jolanrensen: Actually yes!
We can try to test this by writing `val pivot dataFrameOf(...).pivot { ... }` and trying to access columns in the cell below.

https://github.com/Kotlin/dataframe/issues/1268

koperagen (approved):

PR #1270, opened 2025-06-23
Removed types = ["TABLE"] argument from readAllSqlTables

Jolanrensen:
In other databases like DuckDB, this type is actually "BASE TABLE" instead of "TABLE", making the function return 0 results. Setting it to `null` makes the integration do its default behavior, which usually is no filtering at all.


Jolanrensen:
Okay this breaks on PostgreSQL. We need the types argument, but it should be customizable at least

Jolanrensen:
Adjusted it to include the `tableTypes` parameter instead. Hopefully it's okay to break the api dump for jdbc? as it's not considered stable yet

Jolanrensen:
@zaleslaw 
On second thought, if we only want `readAllSqlTables` to include "normal" tables, we could let `DbType` provide a `val tableType: String`. `readAllSqlTables` would then simply filter for that type.

Alternatively, we could allow customizing tableType filters, but then the default would be `[dbType.tableType]`. Just an idea.

Jolanrensen:
See https://github.com/Kotlin/dataframe/pull/1283 for an alternative approach

PR #1272, opened 2025-06-24
Gather kdocs

### LINKED ISSUE #815: Add KDocs for `gather`

The [gather operation](https://kotlin.github.io/dataframe/gather.html) is also similar to update in terms of how to add KDocs to it; users select a bunch of columns (we have templates for this concept), there's a lot of overloads with the same name, and there's a nice DSL grammar opportunity.

AndreiKingsley:
Close #815.


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/gather.kt
```diff
@@ -18,6 +25,115 @@ import kotlin.reflect.typeOf
...
+ * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
+ *
+ * For more information: {@include [DocumentationUrls.Gather]}
+ *
+ * See [Grammar].
+ */
+internal interface GatherDocs {
+
+    /**
+     * ## {@get [GATHER_OPERATION]} Operation Grammar
```
Jolanrensen: You never set this argument to something else than "[**`gather`**][gather]", do you?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/gather.kt
```diff
@@ -18,6 +25,115 @@ import kotlin.reflect.typeOf
...
+     * {@include [LineBreak]}
+     * {@include [DslGrammarLink]}
+     * {@include [LineBreak]}
+     *
+     * **[`gather`][gather]****`  {  `**`columnsSelector: `[`ColumnsSelector`][ColumnsSelector]**` }`**
```
Jolanrensen: markdown doesn't always like `****`, best put the first ones inside the reference or change them to `__`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/gather.kt
```diff
@@ -18,6 +25,115 @@ import kotlin.reflect.typeOf
 
 // region gather
 
+/**
+ * Reshapes the [DataFrame] by gathering specified [\columns] into two columns: keys and values
```
Jolanrensen, line 29: no (small) header? :/ okay we don't necessarily need them

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/gather.kt
```diff
@@ -18,6 +25,115 @@ import kotlin.reflect.typeOf
...
+ *
+ * This operation is the reverse of [pivot].
+ *
+ * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
+ *
+ * For more information: {@include [DocumentationUrls.Gather]}
+ *
+ * See [Grammar].
+ */
+internal interface GatherDocs {
```
Jolanrensen, line 61: very nice and clear!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/gather.kt
Jolanrensen: set/get only works when you `@include` a doc in another place, not when you link to it and users need to click to get there

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/gather.kt
```diff
@@ -44,6 +169,34 @@ public fun <T, C> DataFrame<T>.gather(vararg columns: KProperty<C>): Gather<T, C
 
 // endregion
 
+/**
+ * Filter values in columns previously selected by [gather] using a [filter][RowValueFilter].
+ *
+ * [RowValueFilter] provides each value as a lambda argument, allowing you
+ * to filter rows using a Boolean condition.
```
Jolanrensen: link to Boolean?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/gather.kt
```diff
@@ -56,9 +209,65 @@ public fun <T, C, K, R> Gather<T, C, K, R>.where(filter: RowValueFilter<T, C>):
...
+ *   .mapValues { (it + 0.5).toFloat() }
+ *   .into("series", "value")
+ * ```
+ * @return A new [Gather] instance with only non-null values retained.
+ */
 @Interpretable("GatherChangeType")
 public fun <T, C, K, R> Gather<T, C?, K, R>.notNull(): Gather<T, C, K, R> = where { it != null } as Gather<T, C, K, R>
 
+/**
+ * Explodes [List] values in the columns previously selected by [gather].
```
Jolanrensen: this doesn't really describe what "exploding" means. It would be best if we could link to the `explode` operation for this, but that doesn't have any kdocs yet. Maybe add a small description of what exploding means?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/gather.kt
```diff
@@ -84,6 +318,27 @@ public fun <T, C, K, R> Gather<T, List<C>, K, R>.explodeLists(): Gather<T, C, K,
         explode = true,
     ) as Gather<T, C, K, R>
 
+/**
+ * Applies a [transform] to the gathering keys —
```
Jolanrensen: ~~a~~

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/gather.kt
```diff
@@ -98,6 +353,27 @@ public inline fun <T, C, reified K, R> Gather<T, C, *, R>.mapKeys(
         explode = explode,
     )
 
+/**
+ * Applies a [transform] to the values from the columns previously selected by [gather].
```
Jolanrensen: ~~a~~

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/gather.kt
```diff
@@ -136,6 +464,35 @@ public class Gather<T, C, K, R>(
...
+/**
+ * Reshapes the columns previously selected by [gather] into two new columns:
+ * [keyColumn], containing the original column names, and [valueColumn], containing the corresponding cell values.
+ *
+ * Returns a new [DataFrame] with the reshaped structure.
+ * The original gathered columns are removed from the result,
+ * while all other columns remain unchanged —
+ * except that their values are duplicated for each generated key-value pair.
+ *
+ * Resulting key and value values can be adjusted using [mapKeys] and [mapValues], respectively.
```
Jolanrensen: If you phrase it like this, I'd read "Resulting" as "after this function", but you mean "before calling this function", right? Maybe just leave out "Resulting", I think it's clear enough without it. Same for the functions below.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/gather.kt
```diff
@@ -157,6 +514,32 @@ public fun <T, C, K, R> Gather<T, C, K, R>.into(keyColumn: KProperty<K>, valueCo
...
+ * ### Example
+ * ```kotlin
+ * // Gather names of all columns containing "series" in their name
+ * // into a single "seriesType" column, omitting the value column.
+ * df.gather { cols { it.name().contains("series") } }
+ *   .keysInto("seriesType")
+ * ```
+ *
+ * @param keyColumn The name of the column to store keys (original column names by default).
+ * @return A new [DataFrame] with reshaped columns.
```
Jolanrensen, line 534: maybe a @see to into and valuesInto?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/gather.kt
```diff
@@ -175,6 +558,32 @@ public fun <T, C, K, R> Gather<T, C, K, R>.keysInto(keyColumn: KProperty<K>): Da
...
+ * ### Example
+ * ```kotlin
+ * // Gather values of all `String` columns
+ * // into a single "tag" column, omitting the key column.
+ * df.gather { colsOf<String>() }
+ *   .valuesInto("tag")
+ * ```
+ *
+ * @param valueColumn The name of the column to store gathered values.
+ * @return A new [DataFrame] with reshaped columns.
```
Jolanrensen, line 579: maybe a @see to into and keysInto?

Jolanrensen:
really good! Just some small notes

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/gather.kt
AndreiKingsley: Haven't we agreed that we don't place header in kdocs?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/gather.kt
```diff
@@ -18,6 +25,114 @@ import kotlin.reflect.typeOf
...
+     */
+    interface Grammar
+
+    /*
+     * This argument providing the (clickable) name of the gather-like function.
+     * Note: If clickable, make sure to [alias][your type].
+     */
+    @Suppress("ClassName")
+    @ExcludeFromSources
+    interface GATHER_OPERATION
```
Jolanrensen: This can be removed

PR #1286, opened 2025-06-27
Kotlin 2.2

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/995

- Bumps Kotlin and KSP to 2.2
- Bumps kotlinpoet and dokka as well
- Moving to dev.zacsweers.kctfork for build testing with ksp
- Fixes for all plugins
- Disabled all tests for the compiler plugin, because #1290 
- Bumps jupyter to 0.13.0-481-1 as it's the latest version that does not rely on a kotlin dev version atm

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,6 +1,6 @@
 [versions]
 ksp = "2.2.0-2.0.2"
-kotlinJupyter = "0.12.0-383"
+kotlinJupyter = "0.13.0-481-1"
 
 ktlint = "12.3.0"
 


plugins/keywords-generator/src/main/kotlin/org/jetbrains/dataframe/keywords/KeywordsGeneratorAction.kt
```diff
@@ -21,6 +22,7 @@ abstract class KeywordsGeneratorAction : WorkAction<KeywordsGeneratorAction.Para
     private val taskPackageName = "org.jetbrains.kotlinx.dataframe.keywords"
 
     override fun execute() {
+        println("Generating keywords using Kotlin compiler: ${KotlinCompilerVersion.getVersion()}")
```
zaleslaw, line 25: println? Is it a debug stuff?

zaleslaw (approved):

plugins/dataframe-gradle-plugin/src/main/kotlin/org/jetbrains/dataframe/gradle/SchemaGeneratorPlugin.kt
```diff
@@ -43,10 +44,16 @@ class SchemaGeneratorPlugin : Plugin<Project> {
                 group = GROUP
                 dependsOn(*generationTasks.toTypedArray())
             }
-            tasks.withType(KspTaskJvm::class.java).configureEach {
+            tasks.withType(KspTask::class.java).configureEach {
                 dependsOn(generateAll)
+                dependsOn(*generationTasks.toTypedArray())
+            }
+            tasks.withType(KspAATask::class.java).configureEach {
+                error(
```
koperagen, line 52: Maybe log.error will be enough?

AndreiKingsley (approved):

plugins/keywords-generator/src/main/kotlin/org/jetbrains/dataframe/keywords/KeywordsGeneratorAction.kt
Jolanrensen: Yes, but I think it's helpful enough. It only runs during the build and it can help us detect whether we've updated our keywords generator version

plugins/dataframe-gradle-plugin/src/main/kotlin/org/jetbrains/dataframe/gradle/SchemaGeneratorPlugin.kt
Jolanrensen: Maybe, but in my experience, if `KspAATask` runs, you get very unhelpful class cast exceptions somewhere during the build process. It's hard to then scroll back to find this specific error log and figure out that is causing the class cast exception.

So, if `KspAATask` runs, the user cannot have `ksp.useKSP2=false`, and thus the dataframe plugin will fail at some point anyway. Better to let it fail early with a helpful error message :)

Issue #1006, opened 2024-12-19
Parse String to UUID


Jolanrensen:
UUIDs are now part of common Kotlin: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/.

We should be able to add this to our `.parse()` function if we can parse UUIDs like `UUID.parseOrNull()`.

EmmanuelBerkowicz:
Hey @Jolanrensen!
As mentioned, can you please assign this to me! I would love to work on this.
Emmanuel

Jolanrensen:
@AnosVoldygod assigned it to you :) Let me know if you need help with anything!

EmmanuelBerkowicz:
Hey @Jolanrensen ,
I've created a pull request!
Let me know if you have any feedback or corrections :)

PR #1287, opened 2025-06-29
Parse String to UUID #1006

EmmanuelBerkowicz:
## Summary
Implements UUID parsing support for DataColumn.parse() as per open issue #1006

## Changes
- Added UUID parser to `parsersOrder` list in `main/../parse.kt` using `java.util.UUID.fromString()` with exception handling
- Added test cases in `test/../parse.kt`:
  - Valid UUID strings are parsed to UUID objects
  - Invalid UUID strings remain String type

## Testing
- Both positive and negative test cases pass
- Unable to run full test suite due to Java version compatibility issue with simple-git plugin (unrelated to this change)

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/parse.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/parse.kt
@@ -49,7 +49,6 @@ import java.time.format.DateTimeFormatterBuilder
 import java.time.temporal.Temporal
 import java.time.temporal.TemporalQuery
 import java.util.Locale
-import java.util.UUID
 import kotlin.properties.Delegates
 import kotlin.reflect.KClass
 import kotlin.reflect.KType
@@ -57,6 +56,8 @@ import kotlin.reflect.full.withNullability
 import kotlin.reflect.jvm.jvmErasure
 import kotlin.reflect.typeOf
 import kotlin.time.Duration
+import kotlin.uuid.ExperimentalUuidApi
+import kotlin.uuid.Uuid
 import java.time.Duration as JavaDuration
 import java.time.Instant as JavaInstant
 import java.time.LocalDate as JavaLocalDate
@@ -427,6 +428,7 @@ internal object Parsers : GlobalParserOptions {
         }
     }
 
+    @OptIn(ExperimentalUuidApi::class)
     internal val parsersOrder = listOf(
         // Int
         stringParser<Int> { it.toIntOrNull() },
@@ -493,13 +495,13 @@ internal object Parsers : GlobalParserOptions {
         // Boolean
         stringParser<Boolean> { it.toBooleanOrNull() },
         // UUID
-        stringParser<UUID> { str ->
+        stringParser<Uuid> { str ->
 
             val uuidRegex = Regex("[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}")
 
             if (uuidRegex.matches(str)) {
                 try {
-                    UUID.fromString(str)
+                    Uuid.parse(str)
                 } catch (e: IllegalArgumentException) {
                     null
                 }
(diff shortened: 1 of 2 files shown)


EmmanuelBerkowicz:
Hey @Jolanrensen,
I'm still pretty new to open source, is there an issue?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/parse.kt
```diff
@@ -491,6 +494,15 @@ internal object Parsers : GlobalParserOptions {
         posixParserToDoubleWithOptions,
         // Boolean
         stringParser<Boolean> { it.toBooleanOrNull() },
+        //UUID
+        stringParser<UUID> {str ->
+            try{
+                UUID.fromString(str)
+            } catch(e: IllegalArgumentException){
```
Jolanrensen: While this approach works, the `parse()` function can make this piece of code be called very often (if you try to parse a column with 100 Strings, it gets called 100 times). Exceptions are meant for exceptional cases, not for default flows, as they are quite heavy to throw (a stacktrace needs to be built every time).
For this reason, we tried to avoid throwing exceptions for the other types we can parse, like dates etc. I think for UUIDs we can do the same thing :)

We could namely check the string by a Regex first, return null when the regex doesn't match, and call `UUID.fromString()` only when the regex does match. You can keep the `try {} catch()` around it for safety of course. This will save a lot of stack traces being created :)

I found this regex: "[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}" in several places online; I think it's safe to use it, but if you're unsure, please add some more tests.

Jolanrensen:
> Hey @Jolanrensen, I'm still pretty new to open source, is there an issue?

No issue :) I self-requested a review so I can comment on some particular parts of the PR. If you correct it and our test suite and the team is okay with it, we can merge it to the master

Jolanrensen:
Actually, as my collegue just pointed out to me, the idea was to use Kotlin's new experimental builtin Uuid class https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/ instead of the Java variant.

It works very similarly but is more Kotlinlike :). Feel free to OptIn to @ExperimentalUuidApi, I'll add a `ParserOption` for our users to let them decide whether they want to use the experimental UUID parser or not after you finish this part.

You can use the same Regex approach I mentioned, but as it's Kotlin itself, I'll also [ask](https://youtrack.jetbrains.com/issue/KT-78739/Add-kotlin.uuid.Uuid.parseOrNull) whether they can create a `parseOrNull()` function without exceptions :)

EmmanuelBerkowicz:
Thanks for the feedback @Jolanrensen !

I'll work on this when I get home and update!

Emmanuel

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/parse.kt
```diff
@@ -491,6 +492,21 @@ internal object Parsers : GlobalParserOptions {
         posixParserToDoubleWithOptions,
         // Boolean
         stringParser<Boolean> { it.toBooleanOrNull() },
+        // UUID
+        stringParser<UUID> { str ->
```
Jolanrensen: could you change it to `kotlin.uuid.Uuid`? :)

EmmanuelBerkowicz: I tried using kotlin.uuid.UUID but get 'Unresolved reference' errors. Should I update the Kotlin version in the project, or is there a specific setup needed for kotlin.uuid?"

EmmanuelBerkowicz: Oh! Uuid! Nevermind I'm figuring it out

EmmanuelBerkowicz: Hey @Jolanrensen ,
I've made the change and updated the test sections also.

Jolanrensen:
@EmmanuelBerkowicz Thanks! :) 

We will take it from here. I'd like to add a setting to `ParserOptions` to make sure people need to opt-in for this experimental Uuid type, but I won't bother you to have to deal with that

EmmanuelBerkowicz:
No worries @Jolanrensen ,
Thanks for the patience and letting letting me contribute!

PR #1288, opened 2025-06-30
filter kdocs

AndreiKingsley:


AndreiKingsley:
Deprecated `filterBy`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/filter.kt
```diff
@@ -34,21 +45,99 @@ public inline fun <T> DataColumn<T>.filter(predicate: Predicate<T>): DataColumn<
...
+ * A [RowFilter] provides each row as a lambda argument, allowing you to define filtering logic
+ * using a [Boolean] condition.
+ *
+ * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
+ *
+ * For more information, see: {@include [DocumentationUrls.Filter]}
+ *
+ * See also:
+ *  - [drop][DataFrame.drop], which drops rows based on values within the row.
+ *  - [distinct][DataFrame.distinct], which filters rows with duplicated values.
```
Jolanrensen: *filters out? If you say "filters", it could also mean that you only keep rows with duplicate values.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/filter.kt
```diff
@@ -34,21 +45,99 @@ public inline fun <T> DataColumn<T>.filter(predicate: Predicate<T>): DataColumn<
...
+ *
+ * Returns a new [DataFrame] containing only the rows where the value in the given [column] is `true`.
+ *
+ * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
+ *
+ * For more information, see: {@include [DocumentationUrls.Filter]}
+ *
+ * See also: [filter], which allows filtering rows based on values within the row.
+ *
+ * ### This Gather Overload
```
Jolanrensen: Gather overload? ;P

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt
```diff
@@ -110,6 +110,9 @@ internal const val CONVERT_TO_URL_REPLACE = "convertToUrl()"
 internal const val TO_URL = "This function is replaced by `toUrl()`. $MESSAGE_1_0"
 internal const val TO_URL_REPLACE = "toUrl()"
 
+internal const val FILTER_BY = "This function  is deprecated in favor of `filter { }`. $MESSAGE_1_0"
```
Jolanrensen: double spaces

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt
```diff
@@ -110,6 +110,9 @@ internal const val CONVERT_TO_URL_REPLACE = "convertToUrl()"
...
+internal const val FILTER_BY_REPLACE = "filter { column }"
```
Jolanrensen, line 114: I'm not sure whether the ReplaceWith can fully handle this transition... in `filterBy { column }`, column is a `ColumnAccessor<T>`, in `filter { column }`, it's a value `T`. But let's keep it like this, the user can probably figure it out.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/filter.kt
```diff
@@ -34,21 +45,99 @@ public inline fun <T> DataColumn<T>.filter(predicate: Predicate<T>): DataColumn<
...
+ */
+@ExcludeFromSources
+internal interface FilterByDocs
+
+/**
```
Jolanrensen: probably don't need kdocs for this if we're gonna deprecate it, right?

Jolanrensen (approved):

Issue #614, opened 2024-03-04
`prev()?.newValue()` unavailable for add-multiple DSL


Jolanrensen:
As is described [here](https://kotlin.github.io/dataframe/add.html#create-a-new-column-and-add-it-to-dataframe), adding a single column allows you to access the previous row with `prev()` and the new value of the previous row with `prev().newValue<T>()`.

This enables:
```kt
df.add("fibonacci") {
    if (index() < 2) 1
    else prev()!!.newValue<Int>() + prev()!!.prev()!!.newValue<Int>()
}
```

I'd expect this to be equivalent:

```kt
df.add {
    "fibonacci" from {
        if (index() < 2) 1
        else prev()!!.newValue<Int>() + prev()!!.prev()!!.newValue<Int>()
    }
}
```
but unfortunately, since `from {}` provides a `DataRow` as context instead of `AddDataRow`, `newValue()` is unavailable.
I don't know whether this was done due to a technical limitation, but if it was an oversight it surely must be added.

PR #1303, opened 2025-07-03
Makes `prev()?.newValue()` work in Add DSL

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/614

Changes RowExpression to AddExpression in AddDsl to allow `prev()?.newValue()` to be called.

@koperagen do you think this requires changes in the compiler plugin?


koperagen (approved):
Should be fine

Issue #1304, opened 2025-07-03
Make `kotlin.uuid.Uuid` parsing experimental via ParserOptions


Jolanrensen:
Follow up from https://github.com/Kotlin/dataframe/pull/1287

PR #1306, opened 2025-07-03
`parseExperimentalUuid` in ParserOptions

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1304

Added `parseExperimentalUuid` parser option for parsing to experimental `kotlin.uuid.Uuid`.
False by default until it is no longer experimental.
Docs are updated as well.

Follow up from https://github.com/Kotlin/dataframe/pull/1287


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/parse.kt
```diff
@@ -101,6 +107,8 @@ public interface GlobalParserOptions {
  * @param skipTypes a set of types that should be skipped during parsing. Parsing will be attempted for all other types.
  *   By default, it's an empty set. To skip all types except a specified one, use [convertTo] instead.
  * @param useFastDoubleParser whether to use [FastDoubleParser], defaults to `true`. Please report any issues you encounter.
+ * @param parseExperimentalUuid whether to allow parsing UUIDs to the experimental [kotlin.uuid.Uuid] type.
```
koperagen, line 116: And we now know that in notebooks dataframe properties with such type require additional opt in!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/parse.kt
```diff
@@ -73,6 +73,12 @@ public interface GlobalParserOptions {
     public val nulls: Set<String>
 
     public val skipTypes: Set<KType>
+
+    /**
+     * Whether to allow parsing UUIDs to the experimental [kotlin.uuid.Uuid] type.
+     * By default, this is false and UUIDs are not recognized.
+     */
+    public var parseExperimentalUuid: Boolean
```
koperagen, line 87: Btw, do you plan to later rename this property later? It will be an incompatible change.. Or drop it completely? Should we call it "parseUuid" and provide note about experimental status in kdocs?

Jolanrensen: When it's no longer experimental, this can be deprecated with Warning -> Error -> removed, right? No need to disable it if it's stable IMO

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/parse.kt
Jolanrensen: Let me test it quickly for this specific case, since it's never really surfacing the opt-in to the API. Otherwise I'll update the docs :)

Jolanrensen: Image

nope, no additional opt-in needed. Only when you're interacting with the Uuid afterwards. I'll mention that instead.

koperagen: Try to create df instance with such column to trigger extension properties generation

Jolanrensen: damn, you're right! Very nice error. I'll add it to the docs

Jolanrensen: https://github.com/Kotlin/dataframe/issues/1335

Issue #1281, opened 2025-06-26
Add Unit tests and KDocs for family of `generate<Something>` methods


zaleslaw:
We have no proper Unit tests and KDocs for this functionality which will be is highly useful in upcomping release
https://github.com/Kotlin/dataframe/blob/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/generateCode.kt

- generateCode
- generateInterfaces
- generateDataClasses

PR #1311, opened 2025-07-04
generateCode.kt refactor, KDocs, documentation fixes, and tests

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1281


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/CodeGenerationTests.kt
```diff
@@ -368,22 +394,360 @@ class CodeGenerationTests : BaseTest() {
...
+        val schema = typed.schema()
+        val code = schema.generateDataClasses(
+            markerName = "SchemaDataClass",
+            extensionProperties = true,
+            visibility = MarkerVisibility.EXPLICIT_PUBLIC,
+            useFqNames = false,
+        )
+
+        @Language("kotlin")
+        val expected =
```
koperagen: There are very similar tests on CodeGenerator itself. If you want to test generateDataClasses specifically, maybe you can compare results to results of CodeGenerator? 
like df.generateInterfaces should be CodeGenerator.generateCode(...)
So we have less code Strings to refactor should anything change.. or, what do you think?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/generateCode.kt
```diff
@@ -1,14 +1,89 @@
...
+ * This is useful when working with the compiler plugin in cases where the schema
+ * cannot be inferred automatically from the source.
+ *
+ * This function is a simplified wrapper for the more advanced and customizable
+ * [CodeGenerator] API.
+ * For more customizability, have a look at [CodeGenerator.create()][CodeGenerator.create].
+ *
+ * For more information: {@include [DocumentationUrls.DataSchemaGeneration]}
+ *
+ * @return [CodeString] – A value class wrapper for [String], containing
```
koperagen, line 38: and `CodeString.print()` utility :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/generateCode.kt
```diff
@@ -32,11 +117,39 @@ public fun <T> DataFrame<T>.generateCode(
...
+ * @include [CommonGenerateCodeDocs]
+ * {@set [CommonGenerateCodeDocs.RECEIVER] [DataFrame's][this] [schema][DataFrameSchema]}
+ * {@set [CommonGenerateCodeDocs.TYPES] interfaces}
+ *
+ * @include [Params.MarkerNameOptional]
+ */
+public fun <T> DataFrame<T>.generateInterfaces(markerName: String): CodeString =
+    schema().generateInterfaces(markerName = markerName)
+
+/**
```
koperagen: Please mention that data classes can be used with "toListOf"

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/generateCode.kt
```diff
@@ -1,14 +1,89 @@
...
+    interface ExtensionProperties
+
+    /** @param visibility Visibility modifier for the generated declarations (markers). */
+    interface Visibility
+
+    /** @param useFqNames If `true`, fully qualified type names will be used in generated code. */
+    interface UseFqNames
+
+    /**
+     * @param nameNormalizer Strategy for converting column names (with spaces, underscores, etc.) to valid Kotlin identifiers.
```
koperagen: I think it's better to use word "Kotlin-style identifiers" for `default` normalizer
And 
"Generated properties will still refer to columns by their actual name via @ColumnName annotation`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/generateCode.kt
```diff
@@ -1,14 +1,89 @@
...
+@ExcludeFromSources
+private interface Params {
+
+    /** @param markerName The base name to use for generated data schema declarations (markers). */
+    interface MarkerName
+
+    /** @include [MarkerName] If not specified, generates a name from type [T]. */
+    interface MarkerNameOptional
+
+    /** @param fields Whether to generate fields (`val ...:`) inside the generated data schema declarations (markers). */
```
koperagen: How to put it.. Technically field = false, extensions = true is enough to refer to columns in dataframe
But plugin won't be able to extract schema from such dataschema. At least we should state it, at best maybe even remove this option?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/generateCode.kt
```diff
@@ -1,14 +1,89 @@
...
+    interface Fields
+
+    /** @param extensionProperties Whether to generate extension properties in addition to data schema declarations (markers). */
```
koperagen: Here we should say that if either compiler plugin or ksp are used in the project, extension properties will conflict with generated ones. 
Soo, same as with fields, does it even make sense to have this option..

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/generateCode.kt
```diff
@@ -1,14 +1,89 @@
...
+    interface ExtensionProperties
+
+    /** @param visibility Visibility modifier for the generated declarations (markers). */
+    interface Visibility
```
koperagen, line 74: To add some context, this argument was added to generate valid code for https://kotlinlang.org/docs/api-guidelines-simplicity.html#use-explicit-api-mode
Maybe it's a useful info, maybe not :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/generateCode.kt
```diff
@@ -1,14 +1,89 @@
...
+// region Docs
+
+/**
+ * Generates a [CodeString] containing generated [@DataSchema][DataSchema] $[TYPES]
+ * for the given $[RECEIVER]
+ * (including all nested [frame columns][FrameColumn] and [column groups][ColumnGroup]).
+ *
+ * These generated declarations can also be called "markers".
+ *
+ * This is useful when working with the compiler plugin in cases where the schema
```
AndreiKingsley: compiler plugin link

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/generateCode.kt
AndreiKingsley: I'd even add more info about Data Classes uses from here:
 https://kotlin.github.io/dataframe/dataschema-data-classes-generation.html#generatedataclasses

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/generateCode.kt
```diff
@@ -19,6 +94,16 @@ public inline fun <reified T> DataFrame<T>.generateCode(
...
+ * @include [CommonGenerateCodeDocs]
+ * {@set [CommonGenerateCodeDocs.RECEIVER] [DataFrame's][this] [schema][DataFrameSchema]}
+ * {@set [CommonGenerateCodeDocs.TYPES] interfaces and extension properties}
+ *
+ * @include [Params.MarkerName]
+ * @include [Params.Fields] Default is `true`.
+ * @include [Params.ExtensionProperties] Default is `true`.
+ * @include [Params.Visibility] Default is [MarkerVisibility.IMPLICIT_PUBLIC].
+ */
 public fun <T> DataFrame<T>.generateCode(
```
AndreiKingsley: For `generateCode`, add more info about extension properties.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/generateCode.kt
Jolanrensen: I see :) I'm not sure whether we need to mention that specific case. If they are looking to change the visibility because they use explicit api mode, they'll probably find this setting anyway

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/generateCode.kt
Jolanrensen: I'm not sure, but it was `true` by default for `generateCode`. There was probably a reason for that, so I kept it.
We could keep it but default it to false?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/generateCode.kt
Jolanrensen: yes, it makes sense to remove it. It's a very advanced use case

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/generateCode.kt
Jolanrensen: I put it in the parameter description

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/generateCode.kt
koperagen: Let's keep it but switch default to false, i didn't know about it :) But for now we'll recommend using generateCode with plugin and properties will surely conflict

Jolanrensen: then it might make more sense to recommend generateInterfaces or generateDataClasses in combination with the plugin. Actually, generateCode is probably not even needed, is it?

Jolanrensen:
@AndreiKingsley @koperagen Alright, I decided to refactor the API as well:
- generateCode is deprecated in favor of generateInterfaces
- generateInterfaces has all arguments generateDataClasses has

Please check again as many things have changed :) I did take your feedback into account

docs/StardustDocs/topics/concepts/DataSchemaGenerationMethods.md
```diff
@@ -164,90 +99,93 @@ val customers: List<Customer> = df.cast<Customer>().toList()
...
+
+This is useful when working with the [compiler plugin](Compiler-Plugin.md)
+in cases where the schema cannot be inferred automatically from the source.
+
+### Arguments {id="generateInterfaces-arguments"}
+
+* `markerName`: `String?` — The base name to use for generated interfaces.  
+  If `null`, uses the `T` type argument of `DataFrame` simple name.  
+  Default: `null`.
+* `extensionProperties`: `Boolean` – Whether to generate extension properties in addition to `interface` declarations.
```
AndreiKingsley: "extension properties" link

AndreiKingsley (approved):
Nice!

Issue #1292, opened 2025-07-02
Flatten the hierarchy of `Modify` page to the one level, based on alphabetical index


zaleslaw:
- [ ] [Modify](https://kotlin.github.io/dataframe/modify.html) flatten this based on alphabetical order
 - [ ] Keep the current structure and grouping on the `Modify` page, but links should be correct

PR #1312, opened 2025-07-04
Flatten navigation of Modify operations

koperagen:
The core idea is to embed all meta-information about "related" or "opposite" operations on the pages, so navigation can be simple and sorted

https://github.com/Kotlin/dataframe/issues/1292

### DIFF
--- a/docs/StardustDocs/d.tree
+++ b/docs/StardustDocs/d.tree
@@ -91,73 +91,61 @@
             </toc-element>
         </toc-element>
         <toc-element topic="modify.md">
-            <toc-element topic="sliceRows.md">
-                <toc-element topic="xs.md"/>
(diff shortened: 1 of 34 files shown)


AndreiKingsley:
May be then also make a restructure of [operations](https://kotlin.github.io/dataframe/operations.html) page according to new structure of inner pages?
It could be a good operations overview but now it looks dirty and hard to navigate.

koperagen:
> May be then also make a restructure of [operations](https://kotlin.github.io/dataframe/operations.html) page according to new structure of inner pages? It could be a good operations overview but now it looks dirty and hard to navigate.

You mean - create headers for modify, access data, general info, etc on that page? If yes - i'll create an issue and do it. I like the idea

PR #1323, opened 2025-07-14
Docs small fixes

AndreiKingsley:


zaleslaw (changes requested):
Let's remove TBD topics from this PR, we are planning to deploy docs and it mislead

AndreiKingsley:
@zaleslaw these pages are hidden and can only be accessed by links in FAQ

zaleslaw:
Wow, 😮, ok, of they will be not displayed , I am fine with that

PR #1325, opened 2025-07-15
Added readOnly mode

### LINKED ISSUE #1308: How about readOnly for sql query?
For some case, I should check sql that is a select statement.  How about add readonly for query? 

```kotlin
public fun DataFrame.Companion.readSqlQuery(
    dbConfig: DbConnectionConfig,
    sqlQuery: String,
    limit: Int = DEFAULT_LIMIT,
    inferNullability: Boolean = true,
    dbType: DbType? = null,
    strictValidation: Boolean = true,
): AnyFrame {
    DriverManager.getConnection(dbConfig.url, dbConfig.user, dbConfig.password).use { connection ->
        //    conn.isReadOnly = true
        //    conn.autoCommit = false
        return readSqlQuery(connection, sqlQuery, limit, inferNullability, dbType, strictValidation)
    }
}
```

Or to add readonly for DbConnectionConfig?

zaleslaw:
Fixes #1308

### DIFF
--- a/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
+++ b/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
@@ -22,6 +22,7 @@ import java.sql.Ref
 import java.sql.ResultSet
 import java.sql.ResultSetMetaData
 import java.sql.RowId
+import java.sql.SQLException
 import java.sql.SQLXML
 import java.sql.Time
 import java.sql.Timestamp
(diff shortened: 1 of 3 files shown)


dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -104,13 +105,105 @@ public data class TableColumnMetadata(
...
+ * Users do not have access to the underlying `Connection` instance and cannot commit or close it manually.
+ *
+ * ### Examples:
+ *
+ * ```kotlin
+ * // Safe read-only connection (default)
+ * val config = DbConnectionConfig("jdbc:sqlite::memory:")
+ * val df = DataFrame.readSqlQuery(config, "SELECT * FROM books")
+ *
+ * // Explicitly allow JDBC defaults (still protected against mutations)
```
koperagen: "allow JDBC defaults" wasn't clear to me. Maybe "Use default JDBC connection settings"?

PR #1334, opened 2025-07-18
Add deprecated filterNotNull for discoverability of dropNulls

koperagen:
We discussed that `dropNulls` is not obvious for people familiar with `filterNotNull` from Kotlin Collections and decided to keep our convention but add a few pointers

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
@@ -383,6 +383,15 @@ public fun <T> DataFrame<T>.dropNulls(whereAllNull: Boolean = false, columns: Co
     }
 }
 
+@Refine
+@Interpretable("DropNulls0")
+@Deprecated(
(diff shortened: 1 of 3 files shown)


zaleslaw (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
```diff
@@ -383,6 +383,15 @@ public fun <T> DataFrame<T>.dropNulls(whereAllNull: Boolean = false, columns: Co
...
 }
 
+@Refine
+@Interpretable("DropNulls0")
+@Deprecated(
+    "DataFrame conventional name for filterNot* functions is drop*",
+    ReplaceWith("dropNulls(columns = columns)"),
+    DeprecationLevel.ERROR,
+)
+public fun <T> DataFrame<T>.filterNotNull(columns: ColumnsSelector<T, *>): DataFrame<T> = dropNulls(columns = columns)
```
Jolanrensen, line 393: If you annotate something with ERROR, there's no need to write an implementation for it; the function cannot be called.

Issue #1280, opened 2025-06-26
Add KDocs and Unit tests for function `format`


zaleslaw:
https://github.com/Kotlin/dataframe/blob/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/format.kt

We need minimal but clear KDocs and Unit tests that clarifies cases of usage and set up

Issue #251, opened 2023-02-03
textColor does not work for number colums


pacher:
#### To reproduce:
```kotlin
df.format().with { textColor(green) }
```
**expected** all cells are green
**actual** number columns are not
#### What is happening:
Resulting `td` html elements do have correct style  `<td style="color:#ff0000">`.
But for number columns a nested `<span class="numbers">` element is rendered inside `td` ([source](https://github.com/Kotlin/dataframe/blob/514d3dd99857c4bb10496539e84a894e10308b57/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt#L448))
and for css class `numbers` there is a css rule ([source](https://github.com/Kotlin/dataframe/blob/514d3dd99857c4bb10496539e84a894e10308b57/core/src/main/resources/table.css#L139)) which overrides text color set by `format` operator in outer `td` style.

If I remove this rule, I don't see a difference, so I am disabling this css rule for a quick-fix with additional css
```css
    table.dataframe .formatted .numbers {
        color: unset;
    } 
```
but there must be a reason why this rule exists, so something somewhere might break if we just delete it. Would be really great if somebody more versed in dataframe css could have a look at this.

It's also possible that there could be other issues like this where css overrides explicit formatting with `format` operator.

Or was it intentional? And text color was meant really only for text as opposed to numbers? Shall we add `numberColor` to `FormattingDSL` in this case?

Jolanrensen:
We probably need to add `!important` https://www.w3schools.com/css/css_important.asp to these manual color settings so that they will not be overridden

Issue #509, opened 2023-11-20
API names should follow CamelCase rules


Jolanrensen:
Currently, the API is inconsistent and thus unpredictable when and how it camelCases certain names. We should follow the same rules everywhere, proposing: https://developer.android.com/kotlin/style-guide#camel_case.

This means:
 - `DataFrame.readJson()` correct :)
 - `DataFrame.readCSV()` -> `DataFrame.readCsv()`
 - `writeCSV()` -> `writeCsv()`
 - `toCsv()` correct :)
 - `DataFrame.readTSV()` -> `DataFrame.readTsv()`
 - `DataFrame.readArrowIPC()` -> `DataFrame.readArrowIpc()`
 - `isURL()` -> `isUrl()`
 - `asURL()` -> `asUrl()`
 - `CSVType` -> `CsvType`
 - `toHtmlData()` correct :)
 
Many more

Jolanrensen:
Duplicate of https://github.com/Kotlin/dataframe/issues/383

PR #1346, opened 2025-07-24
`format` kdocs++

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1280
Fixes https://github.com/Kotlin/dataframe/issues/251 (I needed it in the example :) )

- Adds full KDocs and grammar for `format`
- **Breaking change!** Renamed `FormatDSL` to `FormatDsl`, and `RGBColor` to `RgbColor`, following https://github.com/Kotlin/dataframe/issues/509 before 1.0 is fully released, as it will be considerably more harmful to do later. Cannot go through deprecation cycle unfortunately.
- Adds `at()`, and `notNull()` to `format`, similar to `update`, because they seem useful.
- Fixed `where` filters to add to previous ones instead of overwriting the filter, similar to `update`
- Fixed KDocs for Update.notNull() because I completely misunderstood how it worked when I wrote the original KDocs
- Bumped Kandy version to include https://github.com/Kotlin/kandy/pull/477/commits/2fbddba135d4aaa4aeee7e66921752ce999ffd9f; support for `FormattedFrame` "sample dataframes" on the website
- Wrote documentation for format on the website including the full grammar and examples :)
- Adds JUnit tests for format

https://github.com/Kotlin/dataframe/issues/1356 and https://github.com/Kotlin/dataframe/issues/1354 were discovered but can be done later.


core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/testSets/person/FormattingTests.kt
```diff
@@ -33,7 +33,7 @@ class FormattingTests : BaseTest() {
 
         val formatter = formattedFrame.formatter!!
         for (row in 0 until typed.nrow) {
-            FormattingDSL.formatter(typed[row], typed.age)!!.attributes().size shouldBe
+            FormattingDsl.formatter(typed[row], typed.age)!!.attributes().size shouldBe
```
koperagen, line 36: Maybe add a deprecated type aliases for FormattingDSL and RGBColor?

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Modify.kt
```diff
@@ -1303,4 +1309,38 @@ class Modify : TestBase() {
...
+
+    @Suppress("UNCHECKED_CAST")
+    @Test
+    fun formatExample_strings() {
+        // SampleStart
+        df
+            .format().with { bold and textColor(black) }
+            .format("isHappy").with { background(if (it as Boolean) green else red) }
+            .format("weight").notNull().with { linearBg(it as Int, 50 to blue, 90 to red) }
+            .format("age").perRowCol { row, col ->
```
koperagen, line 1337: While we do have this function, using it is worrisome. We calculate min and max (which take full scan of the column) for each row, so it's n^2 complexity out of nowhere. 

```kt
val col = df["age"] as DataColumn<Int>
val min = col.min()
val max = col.max()


.format("age").with { 
    textColor(
        linear(value = it, from = min to blue, to = max to green)
    )
}
```

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/testSets/person/FormattingTests.kt
Jolanrensen: That could help with source-compatibility indeed :) Unfortunately binary compatibility will always break because we use `FormattingDsl` as receiver in a lamda.
I'll add them

core/generated-sources/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Modify.kt
Jolanrensen: that... is a very good point. Let me see if I can adjust the examples to avoid repetitive calculations

Jolanrensen: in practice it will probably not be that bad. I wonder how many people will try to format a table with hundreds of rows, but it's good to hint towards effecient behavior

Jolanrensen:
@AndreiKingsley sure :) just bump the kandy version when you're done. I think I could already merge this version because the tables look good enough, right?

PR #1349, opened 2025-07-25
android example

AndreiKingsley:


examples/android-example/app/keks.ipynb
Jolanrensen, line 1: test file? :P

examples/android-example/app/build.gradle.kts
```diff
@@ -0,0 +1,81 @@
...
+    implementation(libs.androidx.material3)
+    testImplementation(libs.junit)
+    androidTestImplementation(libs.androidx.junit)
+    androidTestImplementation(libs.androidx.espresso.core)
+    androidTestImplementation(platform(libs.androidx.compose.bom))
+    androidTestImplementation(libs.androidx.ui.test.junit4)
+    debugImplementation(libs.androidx.ui.tooling)
+    debugImplementation(libs.androidx.ui.test.manifest)
+
+    implementation("org.jetbrains.kotlinx:dataframe:1.0.0-dev-7831")
```
Jolanrensen: maybe some note about which version to use?

examples/android-example/app/src/main/java/com/example/myapplication/MainActivity.kt
```diff
@@ -0,0 +1,124 @@
...
+        setContent {
+            MaterialTheme {
+                Surface(modifier = Modifier.fillMaxSize()) {
+                    DataFrameScreen(df)
+                }
+            }
+        }
+    }
+}
+
```
Jolanrensen, line 56: It may be nice to add a preview too:

```kt
@Preview(showBackground = true)
@Composable
fun DefaultPreview() {
    val df = dataFrameOf(
        "name" to listOf("Andrei", "Nikita", "Jolan"),
        "age" to listOf(22, 16, 37)
    ).cast<Person>()
    DataFrameScreen(df)
}
```

That way people don't have to build+run the app on an emulator/device but they can preview what their dataframe looks like in Android Studio

examples/android-example/app/build.gradle.kts
AndreiKingsley: WDYM? I'd just like release version to be here (mow Beta-2 doesn't work, so I use dev).

koperagen: `implementation("org.jetbrains.kotlinx:dataframe:1.0.0-dev-7831")`
For Android maybe it's worth to use core + json etc. dataframe includes apache arrow that, as we now know, doesn't work on Android =(

PR #1359, opened 2025-08-04
Provide `filter` overload to replace `colGroups(predicate<ColumnGroup>)`

koperagen:
Before this PR this was possible:
`df.select { colGroups { it.containsColumn("name") } }`
but this wasn't:
`df.select { colGroups().filter { it.containsColumn("name") }`
Because DataColumn<DataRow<T>> is not a ColumnGroup =( So the cast was needed. 

With a little compromise now following will be possible:
`df.select { colGroups().filter { it.data.containsColumn("name") }`

I decided against making complex inheritance hierarchy, so unlike with normal filter, our new ColumnGroupWithPath is not a ColumnGroup/DataColumn itself. It only contains a data: ColumnGroup, same as ColumnWithPath. Inheritance required deep and imo too complex changes!

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/filter.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/filter.kt
@@ -1,5 +1,6 @@
 package org.jetbrains.kotlinx.dataframe.api
 
+import org.jetbrains.kotlinx.dataframe.AnyRow
 import org.jetbrains.kotlinx.dataframe.ColumnFilter
 import org.jetbrains.kotlinx.dataframe.ColumnSelector
 import org.jetbrains.kotlinx.dataframe.ColumnsSelector
@@ -8,6 +9,7 @@ import org.jetbrains.kotlinx.dataframe.DataFrame
 import org.jetbrains.kotlinx.dataframe.Predicate
 import org.jetbrains.kotlinx.dataframe.RowFilter
 import org.jetbrains.kotlinx.dataframe.annotations.AccessApiOverload
+import org.jetbrains.kotlinx.dataframe.columns.ColumnGroupWithPath
 import org.jetbrains.kotlinx.dataframe.columns.ColumnPath
 import org.jetbrains.kotlinx.dataframe.columns.ColumnReference
 import org.jetbrains.kotlinx.dataframe.columns.ColumnSet
@@ -159,6 +161,25 @@ public interface FilterColumnsSelectionDsl {
     @Suppress("UNCHECKED_CAST")
     public fun <C> ColumnSet<C>.filter(predicate: ColumnFilter<C>): ColumnSet<C> =
         colsInternal(predicate as ColumnFilter<*>).cast()
+
+    /**
+     * ## Filter [ColumnSet]
+     *
+     * #### For example:
+     * ```kotlin
+     * df.convert {
+     *  colsAtAnyDepth().colGroups()
+     *    .filter { it.data.containsColumn("myCol")
+     * }.with { ... }
+     * ```
+     */
+    @Suppress("INAPPLICABLE_JVM_NAME")
+    @JvmName("filterColumnGroups")
+    public fun ColumnSet<AnyRow>.filter(predicate: Predicate<ColumnGroupWithPath<*>>): ColumnSet<*> =
+        colsInternal { columnWithPath ->
+            columnWithPath.isColumnGroup() &&
+                predicate(ColumnGroupWithPath(columnWithPath, columnWithPath.path))
+        }
 }
 
 // endregion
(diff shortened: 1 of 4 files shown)


Jolanrensen:
What do you mean with
```kt
df.select { colGroups().filter { it.containsColumn("name") }
```
is impossible?

When I check it, `colGroups()` returns `ColumnSet<AnyRow>`, meaning `filter {}` works as expected

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/filter.kt
```diff
@@ -159,6 +161,25 @@ public interface FilterColumnsSelectionDsl {
...
+     * ```kotlin
+     * df.convert {
+     *  colsAtAnyDepth().colGroups()
+     *    .filter { it.data.containsColumn("myCol")
+     * }.with { ... }
+     * ```
+     */
+    @Suppress("INAPPLICABLE_JVM_NAME")
+    @JvmName("filterColumnGroups")
+    public fun ColumnSet<AnyRow>.filter(predicate: Predicate<ColumnGroupWithPath<*>>): ColumnSet<*> =
```
Jolanrensen, line 178: I think this intermediate type is not needed, right?
`colGroups {}` takes a `Predicate<ColumnGroup<*>>`; we can duplicate that here:

```kt
fun ColumnSet<AnyRow>.filter(predicate: Predicate<ColumnGroup<*>>): ColumnSet<*> =
    colsInternal { columnWithPath ->
        columnWithPath.isColumnGroup() && predicate(columnWithPath)
    }
```

Jolanrensen: `.isColumnGroup()` does an instance check and auto-cast so is safe :)

koperagen: Yes but it won't hurt to have `path` property, same as `ColumnSet<*>.filter` does

Jolanrensen: now I understand what you're trying to achieve :) 
Still, we'll have to take a slightly further look, because I see we have the `ColumnGroupWithPathImpl` class. Adding a completely separate `ColumnGroupWithPath` would be confusing at best

koperagen: As alternative we could rename and expose ColumnGroupWithPathImpl, but 
1. ColumnWithPath is used extensively across whole column selection and ColumnGroupWithPathImpl is important implementation detail
2. ColumnGroupWithPath is needed only for one filter overload, and is very simple itself - it's only a class with 2 properties.

Jolanrensen: What if we don't introduce a new class and just create:
```kt
@Suppress("UNCHECKED_CAST")
@get:JvmName("columnAsColumnGroup")
public val <T> ColumnWithPath<DataRow<T>>.column: ColumnGroup<T>
    get() = data as ColumnGroup<T>

@get:JvmName("columnAsFrameColumn")
public val <T> ColumnWithPath<DataFrame<T>>.column: FrameColumn<T>
    get() = data as FrameColumn<T>

public val <T> ColumnWithPath<T>.column: DataColumn<T>
    get() = data
```

I think the name "column" is easier to discover in the `ColumnWithPath` class than "data".

We could even try to find a way to hide `data` in `ColumnWithPath`, but that probably even won't be necessary.

(btw I tried calling them "data" with `@kotlin.internal.HidesMembers`, but unfortunately that doesn't seem to work :( )

koperagen: Good idea, i'll try

koperagen: In CS DSL scope for DataColumn<DataRow<*>> is very polluted though =( There are functions named column. IMO totally impossible to find anything just from looking at completion. 
image
image

But if person knows what they're looking for, they can use `asColumnGroup` even without `column`.

For comparison completion for ColumnGroupWithPath looks clearer:
image

I tried to limit visibility of functions from outer scope with DslMarker, but i'm afraid everything is still visible despite Dsl Scope violation errors. 

So i'd go with ColumnGroupWithPath for better discoverability

Jolanrensen: That's a good point. Just try to be as transparant as possible in KDoc if you decide to go this route. I want to prevent confusion between ColumnGroupWithPathImp and ColumnGroupWithPath, especially if they are separate things.

Also, how's the situation for FrameColumn?

PR #1361, opened 2025-08-04
Fix jupyter logger

### LINKED ISSUE #1155: Fix excel module logger
Image

AndreiKingsley:
Fixes #1155 + geodataframe jupyter fix + new geo-jupyter module


dataframe-geo-jupyter/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/IntegrationGeo.kt
```diff
@@ -67,7 +72,7 @@ internal class IntegrationGeo : JupyterIntegration() {
             val generatedDf = execute(
                 codeWithTypeCastGenerator = replCodeGeneratorImpl.process(geo.df, kProperty),
                 expression = "(${kProperty.name}.df as DataFrame<*>)",
-            )
+            ).let { "`" + it.toString() + "`" }
```
Jolanrensen: `` "`$it`" `` is a bit shorter ;P

But what does this do? It renders the DF as string and adds crs? Isn't it better to output the DF natively and output CRS as string?

dataframe-geo-jupyter/build.gradle.kts
Jolanrensen, line 1: I'm wondering, should this be a separate module? I think it can just be part of the dataframe-jupyter module, potentially with the `useExperimentalGeo` flag

dataframe-geo-jupyter/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/IntegrationGeo.kt
AndreiKingsley: What do you mean by "this"? 
`FieldHandlerFactory.createUpdateExecution<GeoDataFrame<*>>` adds inner df schema retrieving for jupuyter.
new `render` section adds a new render for geodf like that:
image

dataframe-geo-jupyter/build.gradle.kts
AndreiKingsley: dataframe-geo is really heavy and I wouldn't like it to be a part of general `dataframe` artifact, neither `%use dataframe` in notebooks. So I'd really like to keep them separate

examples/android-example/local.properties
```diff
@@ -0,0 +1,8 @@
+## This file must *NOT* be checked into Version Control Systems,
```
koperagen: Need to remove this file :D

koperagen:
What changes in terms of setup? I reckon previously we could `use dataframe(useExperimentalGeo = true)`, do you need to update dependencies there to include a new module?

AndreiKingsley:
@koperagen as I said above, I'd like to remove `useExperimentalGeo`.

koperagen (approved):

AndreiKingsley:
Removed with the [latest commit](https://github.com/Kotlin/dataframe/pull/1361/commits/ea7b9e78e611f03dda56a487762165384103a1ba)

dataframe-geo-jupyter/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/IntegrationGeo.kt
Jolanrensen: Nice :) 

Is it always static, though? or is that just your screenshot?

dataframe-jupyter/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
```diff
@@ -169,11 +168,6 @@ internal class Integration(private val notebook: Notebook, private val options:
             if (enableExperimentalCsv?.toBoolean() == true) {
                 println("CSV module is already enabled by default now.")
             }
-            if (enableExperimentalGeo?.toBoolean() == true) {
-                println("Enabling experimental Geo module: dataframe-geo")
```
Jolanrensen, line 173: instead of completely removing it, please provide a helpful message of what to do instead :)

We mentioned the `enableExperimentalGeo` in some release, so if it suddenly does nothing anymore, people will be confused

AndreiKingsley: Fixed

dataframe-geo-jupyter/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/IntegrationGeo.kt
```diff
@@ -54,6 +54,11 @@ internal class IntegrationGeo : JupyterIntegration() {
             useSchema<WithMultiLineStringGeometry>()
         }
 
+        render<GeoDataFrame<*>> {
+            println("GeoDataFrame with ${it.crs?.name?.code} CRS and inner dataframe:")
```
ileasile, line 58: It's not ideal, better use `notebook.display("text")` for such things. Reason: we can't guarantee standard stream sync in multithreading envs.

PR #1369, opened 2025-08-07
cols with default argument deprecation

### LINKED ISSUE #1284: Remove default predicate for `cols {}` in Selection DSL
In the selection dsl you can write:
`myGroup.cols()` if you want to refer to all columns nested inside the group.
This is allowed, because the default argument makes it `myGroup.cols { true }`. 

However, because we use a predicate, we cannot get compiler plugin support for that function.
It's better to let users use `all()` and `allCols()` instead. It's more expressive and can get compiler plugin backing :)

We need to simply create overloads for `cols()` without argument, [deprecate](https://github.com/Kotlin/dataframe/blob/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt) them and point people to the `all()` family.

AndreiKingsley:
closes #1284


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt
```diff
@@ -213,4 +213,10 @@ internal const val ALL_COLS_EXCEPT_REPLACE = "this.allColsExcept { other }"
 internal const val ALL_COLS_EXCEPT_REPLACE_VARARG = "this.allColsExcept { others.toColumnSet() }"
 internal const val EXCEPT_REPLACE = "this.except { other }"
 internal const val EXCEPT_REPLACE_VARARG = "this.except { others.toColumnSet() }"
+
+internal const val COLS_TO_ALL = "This `cols()` overload will be removed in favor of `all()`. $MESSAGE_1_0"
```
Jolanrensen: You added it in the `// region keep across releases` but with a message for 1.0. How do you want to deprecate this?

Jolanrensen (changes requested):
Looks like there are tests that don't compile anymore due to the introduced ERROR, otherwise, (and besides my note) looks fine :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt
AndreiKingsley: move to 1.0 ERROR section

PR #1370, opened 2025-08-07
skipKodex gradle parameter in dataframe-csv

koperagen:


dataframe-csv/build.gradle.kts
```diff
@@ -107,7 +107,7 @@ tasks.named("runKtlintCheckOverGeneratedSourcesSourceSet") {
 // the target of `processKdocMain`, and they are returned to normal afterward.
 // This is usually only done when publishing
 val changeJarTask by tasks.registering {
-    outputs.upToDateWhen { false }
+    outputs.upToDateWhen { project.hasProperty("skipKodex") }
```
Jolanrensen, line 110: also update:

```kts
tasks.configureEach {
    if (name.startsWith("publish")) {
        dependsOn(processKDocsMain, changeJarTask)
    }
}
```

just like in :core

Jolanrensen: otherwise, good addition :)

koperagen: good catch!

PR #1371, opened 2025-08-07
clarify how to create BaseColumn

koperagen:
It's not quite obvious how to call this function just from the signature: 
`columnOf(vararg columns: Pair<String, AnyBaseCol>)`
I added a small cue

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
@@ -261,6 +261,24 @@ public inline fun <reified T> columnOf(vararg values: T): DataColumn<T> =
 
 public fun columnOf(vararg values: AnyBaseCol): DataColumn<AnyRow> = columnOf(values.asIterable()).forceResolve()
 
+/**
+ * Example:
+ * ```
+ * val columnGroup = columnOf(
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
```diff
@@ -261,6 +261,24 @@ public inline fun <reified T> columnOf(vararg values: T): DataColumn<T> =
 
 public fun columnOf(vararg values: AnyBaseCol): DataColumn<AnyRow> = columnOf(values.asIterable()).forceResolve()
 
+/**
+ * Example:
+ * ```
```
Jolanrensen: add `kotlin` or `kt` to make the example highlight correctly :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
```diff
@@ -314,6 +323,18 @@ public fun dataFrameOf(columns: Iterable<AnyBaseCol>): DataFrame<*> {
     return DataFrameImpl<Unit>(cols, nrow)
 }
 
+/**
+ * Example:
+ * ```
```
Jolanrensen: same here

PR #1373, opened 2025-08-11
Migration to kotlin.time.Instant and kotlinx-datetime 0.7.1

lucgirardin:
It would be a good idea to migrate to kotlin.time.Instant and kotlin.time.Clock before releasing version 1.0!


Jolanrensen:
Already done by https://github.com/Kotlin/dataframe/pull/1368, but thanks anyway!

Issue #1356, opened 2025-07-29
`format { col.group }` impossible


Jolanrensen:
We cannot HTML `format` inside column groups. This seems to be a limit of the `format` implementation.

Issue #982, opened 2024-12-03
HTML formatting for a column affects other columns with same name


Jolanrensen:
Given the following example
```kt
val df = dataFrameOf("firstName", "lastName", "age", "city", "weight", "isHappy")(
    "Alice", "Cooper", 15, "London", 54, true,
    "Bob", "Dylan", 45, "Dubai", 87, true,
    "Charlie", "Daniels", 20, "Moscow", null, false,
    "Charlie", "Chaplin", 40, "Milan", null, true,
    "Bob", "Marley", 30, "Tokyo", 68, true,
    "Alice", "Wolf", 20, null, 55, false,
    "Charlie", "Byrd", 30, "Moscow", 90, true,
).group("firstName", "lastName").into("name")
    .groupBy("city").toDataFrame()
    .add("cityCopy") { "city"<String>() }
    .group("city").into("cityGroup")
    .rename("cityCopy").into("city")

// affects city, cityGroup.city, and group[*].city
df.format("city").with {
    bold and italic and textColor(green)
}.toStandaloneHTML().openInBrowser()
```
We'd expect only `city` to be affected by the format operation, but `cityGroup.city` and all `city` columns inside `group` are also affected.

![image](https://github.com/user-attachments/assets/6b64fd59-8dad-41a9-b96f-17469f1a70db)

Jolanrensen:
![image](https://github.com/user-attachments/assets/83e74475-e7a3-4b17-9df8-e7995dedcac5)

It also seems that frame columns and column groups are excluded? I'd expect the `group` column to get black backgrounds as well in this example. (https://github.com/Kotlin/dataframe/issues/1356)

Jolanrensen:
I will revisit the format implementation later.

We will need to make a recommended way to format nested dataframes https://github.com/Kotlin/dataframe/issues/1375

Jolanrensen:
Actually, currently the example is even more broken than it was before...

```
cityGroup:
    city: String?
group: *
    name:
        firstName: String
        lastName: String
    age: Int
    city: String?
    weight: Int?
    isHappy: Boolean
city: String?
```

this is when it's querying for `name.firstName`. It tries to fetch its parent from the outer dataframe, instead of the nested one... 
```
Column not found: '[name]'
java.lang.IllegalArgumentException: Column not found: '[name]'
	at org.jetbrains.kotlinx.dataframe.api.DataFrameGetKt.getColumn(DataFrameGet.kt:107)
[...]

PR #1374, opened 2025-08-11
Format colgroups

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1356

Now you can actually do this:
```kt
df
    .format().with { background(white) }
    .format { age }.with { background(blue) and textColor(white) }
    .format { name }.with { background(green) }
    .format { name.firstName }.with { italic and underline }
    .format { name.colsOf<String>() }.where { it.startsWith("C") }.with { background(red) }
```
image

Note, this does not solve https://github.com/Kotlin/dataframe/issues/982 yet


koperagen:
This now formats both firstName and lastName?
`df.format { name }.with { italic and underline }`
 If so, add "Columns inherit formatting from containing column group" or "Formatting a ColumnGroup applies to all its children (or nested columns)"

Jolanrensen:
> This now formats both firstName and lastName? `df.format { name }.with { italic and underline }` If so, add "Columns inherit formatting from containing column group" or "Formatting a ColumnGroup applies to all its children (or nested columns)"

I shall expand the kdocs a bit :)

Jolanrensen:
I'll also update the documentation site btw, I forgot about that, then I'll merge :)

PR #1380, opened 2025-08-14
Extension property name generation fix & docs & tests

### LINKED ISSUE #1193: Bug with line break in column name props generation + align `renameToCamelCase` and normalization behavior
Bug:

Image

- [x] In DataFrame `renameToCamelCase` API
- [x] in compiler plugin (should be 😄)
- [ ] in NameNormalizer.Companion.default
- [ ] in ValidFieldName.Companion#of
- [ ] jupyter codegen

Might be other places, need to inspect.

AndreiKingsley:
Fixes #1193

### DIFF
--- a/docs/StardustDocs/topics/extensionPropertiesApi.md
+++ b/docs/StardustDocs/topics/extensionPropertiesApi.md
@@ -163,3 +163,52 @@ See [Compiler Plugin Example](https://github.com/Kotlin/dataframe/tree/plugin_ex
 IDEA project with basic Extension Properties API examples.
 </tab>
 </tabs>
+
+## Properties name generation
+
+By default, each extension property is generated with a name equal to the original column name.
+
+```kotlin
+val df = dataFrameOf("size_in_inches" to listOf(..))
+df.size_in_inches
+```
+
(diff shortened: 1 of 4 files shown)


docs/StardustDocs/topics/extensionPropertiesApi.md
```diff
@@ -163,3 +163,52 @@ See [Compiler Plugin Example](https://github.com/Kotlin/dataframe/tree/plugin_ex
...
+If the original column name cannot be used as a property name (for example, if it contains spaces 
+or has a name equal to a keyword in Kotlin), 
+it will be enclosed in backticks.
+
+```kotlin
+val df = dataFrameOf("size in inches" to listOf(..))
+df.`size in inches`
+```
+
+However, sometimes the original column name contains special symbols
```
koperagen, line 185: Sentence is not clear to me, sorry. Why there's "and can still.."? Because these two parts don't seem to be very connected

Jolanrensen: +1

AndreiKingsley: oh, there should be simply "can't" 😃

PR #1383, opened 2025-08-14
Add support for read-only mode in DuckDB and SQLite

### LINKED ISSUE #1365: Default `readOnly = true` in `DbConnectionConfig` breaks SQLite
To reproduce:

```kt
    val connection = DbConnectionConfig("jdbc:sqlite:/path/to/my.db")
    val dfs = DataFrame.readAllSqlTables(connection)
```

causes:
```
Exception in thread "main" java.sql.SQLException: Cannot change read-only flag after establishing a connection. Use SQLiteConfig#setReadOnly and SQLiteConfig.createConnection().
	at org.sqlite.jdbc3.JDBC3Connection.setReadOnly(JDBC3Connection.java:146)
	at org.jetbrains.kotlinx.dataframe.io.ReadJdbcKt.readAllSqlTables(readJdbc.kt:1464)
	at org.jetbrains.kotlinx.dataframe.io.ReadJdbcKt.readAllSqlTables$default(readJdbc.kt:776)
	at example.Sqlite_testKt.main(sqlite test.kt:10)
	at example.Sqlite_testKt.main(sqlite test.kt)
```

It works fine when I set `readOnly = false`, but it seems like `true` is not a good default for all databases

zaleslaw:
This commit introduces `createConnection` methods for `DuckDb` and `Sqlite` to handle read-only mode during connection creation. It also updates test configurations and ensures proper handling of in-memory databases.

Fixes #1365

### DIFF
--- a/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DuckDb.kt
+++ b/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DuckDb.kt
@@ -256,5 +256,6 @@ public object DuckDb : DbType("duckdb") {
      * Checks if the DuckDB URL represents an in-memory database.
      * In-memory DuckDB URLs are either "jdbc:duckdb:" or "jdbc:duckdb:" followed only by whitespace.
      */
-    private fun String.isInMemoryDuckDb(): Boolean = this.trim() == "jdbc:duckdb:" || matches("jdbc:duckdb:\\s*$".toRegex())
+    private fun String.isInMemoryDuckDb(): Boolean =
+        this.trim() == "jdbc:duckdb:" || matches("jdbc:duckdb:\\s*$".toRegex())
 }


dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/local/duckDbTest.kt
```diff
@@ -624,12 +623,35 @@ class DuckDbTest {
...
+        // Test in-memory database (cannot be read-only)
+        val config = DbConnectionConfig("jdbc:duckdb:")
         val df = config.readDataFrame("SELECT 1, 2, 3")
         df.values().toList() shouldBe listOf(1, 2, 3)
     }
+
+    @Test
+    fun `change read mode with persistent database`() {
+        // Test read-only mode with a temporary file
+        val tempDir = kotlin.io.path.createTempDirectory("duckdb-test-")
```
Jolanrensen: can you not import this function?

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -75,4 +78,22 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
      * @return A new SQL query with the limit clause added.
      */
     public open fun sqlQueryLimit(sqlQuery: String, limit: Int = 1): String = "$sqlQuery LIMIT $limit"
+
+    /**
+     * Creates a database connection using the provided configuration.
+     * This method is only called when working with DbConnectionConfig (internally managed connections).
```
Jolanrensen: link to DbConnectionConfig

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -28,4 +31,22 @@ public object Sqlite : DbType("sqlite") {
         )
 
     override fun convertSqlTypeToKType(tableColumnMetadata: TableColumnMetadata): KType? = null
+
+    override fun createConnection(dbConfig: DbConnectionConfig): Connection =
+        if (dbConfig.readOnly) {
+            try {
```
Jolanrensen: I think if you include SQLite via `compileOnly()` you don't need to rely on reflection :) Users can still provide their own version of SQLite

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DuckDb.kt
```diff
@@ -216,4 +223,38 @@ public object DuckDb : DbType("duckdb") {
...
+            dbConfig.user.takeIf { it.isNotEmpty() }?.let { setProperty("user", it) }
+            dbConfig.password.takeIf { it.isNotEmpty() }?.let { setProperty("password", it) }
+
+            // Handle DuckDB limitation: in-memory databases cannot be opened in read-only mode
+            if (dbConfig.readOnly && !dbConfig.url.isInMemoryDuckDb()) {
+                setProperty("access_mode", "read_only")
+            } else if (dbConfig.readOnly) {
+                logger.warn {
+                    "Cannot create read-only in-memory DuckDB database (url=${dbConfig.url}). " +
+                        "In-memory databases require write access for initialization. Connection will be created without read-only mode."
```
Jolanrensen, line 247: "without read-only mode" is better written as "in read-write mode"

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DuckDb.kt
```diff
@@ -216,4 +223,38 @@ public object DuckDb : DbType("duckdb") {
...
+        }
+
+        return DriverManager.getConnection(dbConfig.url, properties)
+    }
+
+    /**
+     * Checks if the DuckDB URL represents an in-memory database.
+     * In-memory DuckDB URLs are either "jdbc:duckdb:" or "jdbc:duckdb:" followed only by whitespace.
+     */
+    private fun String.isInMemoryDuckDb(): Boolean = this == "jdbc:duckdb:" || matches("jdbc:duckdb:\\s*$".toRegex())
```
Jolanrensen: this.trim() == "jdbc:duckdb:"

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
zaleslaw: Never do in that manner, but thanks for the advice!

Issue #1129, opened 2025-04-16
`renderToString` extension method is `internal`, making it difficult to write a humanized string representation to non-stdout


henrylegrys:
[AnyFrame#renderToString](https://github.com/Kotlin/dataframe/blob/3e1cca9071d3aa1ccd3f3ee60fc9acb0a7607776/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/string.kt#L22) is internal and only exposed via the `print` extension method. This method appears to be the only way to customize the parameters of `renderToString` - in the other places it is exposed, for example `DataFrame#toString`, there is no way to pass custom arguments.

In my application I have a custom `Appendable` that I want to print the formatted representation to instead - currently this appears to be impossible.

Would it be possible to either:
- change `renderToString`'s visibility to public, allowing for applications to write it to other streams; or
- add another extension method similar to `print`, but which can specify an Appendable as the target?

I suppose there are also some open questions about terminal capabilities (in the case of outputting to a terminal) - for example, not all terminals have Unicode support, and therefore might not support the crop symbols used for table borders. Perhaps some kind of TTY-aware interface could be added?

zaleslaw:
Fixed in #1395

PR #1395, opened 2025-08-19
Make AnyFrame.renderToString public

bwjohnson92:
Fixes request in #1129 

Two updates regarding the current DataFrame print flow -  
Setting AnyFrame.renderToString to public, allowing ability to call it externally.  
Updating DataFrame.print to include the rowIndex parameter, as it was the only parameter missing, being passed in to renderToString.

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -3762,11 +3762,13 @@ public final class org/jetbrains/kotlinx/dataframe/api/PivotKt {
 
 public final class org/jetbrains/kotlinx/dataframe/api/PrintKt {
 	public static final fun print (Lorg/jetbrains/kotlinx/dataframe/DataColumn;)V
-	public static final fun print (Lorg/jetbrains/kotlinx/dataframe/DataFrame;IIZZZZ)V
+	public static final synthetic fun print (Lorg/jetbrains/kotlinx/dataframe/DataFrame;IIZZZZ)V
+	public static final fun print (Lorg/jetbrains/kotlinx/dataframe/DataFrame;IIZZZZZ)V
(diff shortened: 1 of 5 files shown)


Jolanrensen:
Thanks! This is very helpful :) It does need some KDocs, but we can add that later.

Since you changed the signature of `print()`, could you add a `@Deprecated` version of `print()` with level HIDDEN with the old signature that simply calls the new one? That way we remain binary compatibility between Beta2 and Beta3+ :) 
We usually put a string constant [here](https://github.com/Kotlin/dataframe/blob/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt) to keep track of deprecations across cycles. Since the old signature will be unavailable to call in 1.0, you can put it in the "region WARNING in 0.15, ERROR in 1.0".

Also run an `apiDump` from gradle to update our .api file :).

bwjohnson92:
Hey! Thanks for the feedback!

A deprecated function signature has been added, with the new message in deprecationMessages, and an apiDump has been run with the changed files pushed! Let me know if you need or want any other changes!

Jolanrensen:
I ran our CI street, and it looks like there are still some linting errors. This can probably be fixed by running a `ktLintFormat` from gradle. You can check the build yourself by running `build` from gradle: https://github.com/Kotlin/dataframe/blob/master/CONTRIBUTING.md#building

bwjohnson92:
Cheers Jolanrensen - there was indeed some linting errors. `ktLintFormat` has been run, and the build looks to be working locally. Hopefully that makes the CI happy!

PR #1396, opened 2025-08-19
Columngroup map fix

AndreiKingsley:
Fixes https://github.com/Kotlin/dataframe/issues/890


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/map.kt
```diff
@@ -25,7 +26,7 @@ import kotlin.reflect.typeOf
 
 @Deprecated(DEPRECATED_ACCESS_API)
 @AccessApiOverload
-public inline fun <C, reified R> ColumnReference<C>.map(
+internal inline fun <C, reified R> ColumnReference<C>.map(
```
Jolanrensen, line 29: does this affect our.api file? otherwise do an `apiDump`

AndreiKingsley: I did run, seems like it doesn't.

PR #1397, opened 2025-08-19
Calling `df.convert { }.with { }` from compiler plugin won't cause static initialization of TimeZone class

koperagen:
before, calling df.convert { }.with { } from compiler plugin implementation would cause `ClassNotFoundException: kotlinx.datetime.TimeZone`. As expected because we don't want to bundle kotlinx.datetime in the plugin. 
now with lives in its own class and can be safely called

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convertRowCellImpl.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convertRowCellImpl.kt
@@ -1,0 +1,27 @@
+package org.jetbrains.kotlinx.dataframe.impl.api
+
+import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.RowValueExpression
+import org.jetbrains.kotlinx.dataframe.api.Convert
+import org.jetbrains.kotlinx.dataframe.api.Infer
+import org.jetbrains.kotlinx.dataframe.api.asColumn
+import org.jetbrains.kotlinx.dataframe.api.name
+import org.jetbrains.kotlinx.dataframe.exceptions.ColumnTypeMismatchesColumnValuesException
+import org.jetbrains.kotlinx.dataframe.impl.columns.newColumn
+import kotlin.reflect.KType
+
+@PublishedApi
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convertRowCellImpl.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.jetbrains.kotlinx.dataframe.RowValueExpression
+import org.jetbrains.kotlinx.dataframe.api.Convert
+import org.jetbrains.kotlinx.dataframe.api.Infer
+import org.jetbrains.kotlinx.dataframe.api.asColumn
+import org.jetbrains.kotlinx.dataframe.api.name
+import org.jetbrains.kotlinx.dataframe.exceptions.ColumnTypeMismatchesColumnValuesException
+import org.jetbrains.kotlinx.dataframe.impl.columns.newColumn
+import kotlin.reflect.KType
+
+@PublishedApi
```
Jolanrensen: Could you maybe add a comment for this function (either as `//` or `/** @comment */` why this function needs to be in this file and not in `convert.kt`? I'm afraid we won't remember anymore why we moved it in a few years.

Jolanrensen: Also, if we follow our naming convention, the file should probably be named `withRowCell.kt`. Maybe if we do that, we're less inclined to move it in the future ;P

Jolanrensen (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convertRowCellImpl.kt
koperagen: i'd like to keep it close to original convert.kt, now they're close together
as for the comment, why not rely on my commit message?) i like to track such changes via git

Jolanrensen: Those help, but I don't consider them a main source-of-truth, as refactorings etc. can alter the git blame, so it takes longer to find

PR #1402, opened 2025-08-21
Rename kdocs

### LINKED ISSUE #1388: Add KDocs for `Rename` function
DSL is covered, so it's only other part
https://github.com/Kotlin/dataframe/blob/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/rename.kt

AndreiKingsley:
close #1388.


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/rename.kt
```diff
@@ -41,17 +139,47 @@ public fun <T, C> DataFrame<T>.rename(vararg cols: ColumnReference<C>): RenameCl
...
+ *
+ * Use the following methods to perform the conversion:
+ * - [into(name)][RenameClause.into] — renames selected columns to the specified names.
+ * - [into { nameExpression }][RenameClause.into] — renames selected columns using a custom expression,
+ *   which takes the column and its path and returns a new name.
+ * - [toCamelCase()][RenameClause.toCamelCase] — renames all selected columns to `camelCase`.
+ *
+ * See [Grammar][RenameDocs.Grammar] for more details.
+ */
 @HasSchema(schemaArg = 0)
```
zaleslaw, line 180: Looks like we need a catalogue / design doc with all annotations, invented this year @koperagen ?
Could you create a simple md doc with them?

zaleslaw (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/rename.kt
```diff
@@ -23,12 +29,104 @@ import kotlin.reflect.KProperty
...
+private interface CommonRenameDocs
+
+/**
+ * Renames columns in the [DataFrame].
+ *
+ * This function allows renaming multiple columns in a single call by supplying a list of name pairs.
+ * Each pair consists of the current column name and the desired new name.
+ *
+ * Example:
+ * ```
```
Jolanrensen: kt

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/rename.kt
```diff
@@ -23,12 +29,104 @@ import kotlin.reflect.KProperty
...
+ * df.rename("oldName1" to "newName1", "oldName2" to "newName2")
+ * ```
+ *
+ * @param mappings A vararg of pairs where each pair consists of the original column name (`first`)
+ * and the new column name (`second`).
+ * @return A new [DataFrame] with the renamed columns.
+ */
 @Refine
 @Interpretable("RenameMapping")
 public fun <T> DataFrame<T>.rename(vararg mappings: Pair<String, String>): DataFrame<T> =
```
Jolanrensen, line 114: should this one be mentioned in the grammar as well?

Jolanrensen: maybe also renameToCamelCase?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/rename.kt
```diff
@@ -23,12 +29,104 @@ import kotlin.reflect.KProperty
...
+     */
+    interface RenameSelectingOptions
+
+    /**
+     * ## Rename Operation Grammar
+     * {@include [LineBreak]}
+     * {@include [DslGrammarLink]}
+     * {@include [LineBreak]}
+     *
+     * **[`rename`][rename]**`  { columnsSelector: `[`ColumnsSelector`][ColumnsSelector]`  }`
```
Jolanrensen: Don't forget everything explicitly written in Kotlin code needs to be bold, so also the braces etc.

Jolanrensen: That allows you to see the difference between `[ optional contents ]` and **`this[]`** get calls, but also between `( grouped | elements)` and **`function(`**`calls`**`)`**

Jolanrensen: if you have the time, could you check previous PRs for this as well? I might have missed this

Jolanrensen: I did the same for format: https://github.com/Kotlin/dataframe/blob/b4a00369b1de8163f88b3dd593518499e8bb8fbd/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/format.kt#L144

(And we also describe it in the DslGrammar description)

Jolanrensen (changes requested):
as commented

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/rename.kt
AndreiKingsley: I believe they're out of grammar. I mean we should mention them, but not as part of operation grammar

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/rename.kt
Jolanrensen: I'd still mark examples with kotlin or kt so the highlighting is correct

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/rename.kt
koperagen: I recently updated the kdoc for HasSchema, should be clear now. They're all in the same file

PR #1412, opened 2025-08-26
Util functions fixes

### LINKED ISSUE #1389: Add KDocs for a set of tiny functions
- [ ] `any` https://github.com/Kotlin/dataframe/blob/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/any.kt
- [ ] `between` https://github.com/Kotlin/dataframe/blob/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/between.kt
- [ ] `associate` https://github.com/Kotlin/dataframe/blob/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/associate.kt
- [ ] `asSequence` https://github.com/Kotlin/dataframe/blob/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/asSequence.kt
- [ ] `asIterable` https://github.com/Kotlin/dataframe/blob/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/asIterable.kt

Highly likely, JUnit test doesn't exist, please create to cover the main scenario at least before writing Documentation

AndreiKingsley:
KDocs and tests for a set of small util functions + deprecate obsolete ones.
Closes #1389

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/any.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/any.kt
@@ -2,18 +2,46 @@ package org.jetbrains.kotlinx.dataframe.api
 
 import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.DataRow
 import org.jetbrains.kotlinx.dataframe.Predicate
 import org.jetbrains.kotlinx.dataframe.RowFilter
 import org.jetbrains.kotlinx.dataframe.columns.values
 
 // region DataColumn
 
+/**
+ * Returns `true` if at least one element in this [DataColumn] satisfies the given [predicate].
+ *
+ * This is a convenience alias that delegates to [Iterable.any] on the column's [values].
+ *
+ * @param predicate A lambda function that takes a value from the column
+ * and returns `true` if it matches the condition.
+ * @return `true` if at least one element matches the [predicate], `false` otherwise.
+ * @see [DataColumn.all]
+ * @see [DataColumn.filter]
+ * @see [DataColumn.count]
+ */
 public fun <T> DataColumn<T>.any(predicate: Predicate<T>): Boolean = values.any(predicate)
 
 // endregion
 
 // region DataFrame
 
+/**
+ * Returns `true` if at least one row in this [DataFrame] satisfies the given [predicate].
+ *
+ * {@include [org.jetbrains.kotlinx.dataframe.documentation.RowFilterDescription]}
+ *
+ * ### Example
+ * ```kotlin
+ * // Check if there is at least one row where "age" is greater than 18
+ * val hasAdults = df.any { age > 18 }
+ * ```
+ *
+ * @param predicate A [RowFilter] lambda that takes a [DataRow] (as both `this` and `it`)
+ * and returns `true` if the row should be considered a match.
+ * @return `true` if at least one row satisfies the [predicate], `false` otherwise.
+ */
 public inline fun <T> DataFrame<T>.any(predicate: RowFilter<T>): Boolean = rows().any { predicate(it, it) }
 
 // endregion
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/any.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/any.kt
@@ -41,6 +41,7 @@ public fun <T> DataColumn<T>.any(predicate: Predicate<T>): Boolean = values.any(
  * @param predicate A [RowFilter] lambda that takes a [DataRow] (as both `this` and `it`)
  * and returns `true` if the row should be considered a match.
  * @return `true` if at least one row satisfies the [predicate], `false` otherwise.
+ * @see [DataFrame.all]
  */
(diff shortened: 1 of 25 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/any.kt
```diff
@@ -2,18 +2,46 @@ package org.jetbrains.kotlinx.dataframe.api
...
+/**
+ * Returns `true` if at least one element in this [DataColumn] satisfies the given [predicate].
+ *
+ * This is a convenience alias that delegates to [Iterable.any] on the column's [values].
+ *
+ * @param predicate A lambda function that takes a value from the column
+ * and returns `true` if it matches the condition.
+ * @return `true` if at least one element matches the [predicate], `false` otherwise.
+ * @see [DataColumn.all]
+ * @see [DataColumn.filter]
```
Jolanrensen, line 21: I don't think filter and count are very relevant, only `all`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/any.kt
```diff
@@ -2,18 +2,46 @@ package org.jetbrains.kotlinx.dataframe.api
...
+ *
+ * ### Example
+ * ```kotlin
+ * // Check if there is at least one row where "age" is greater than 18
+ * val hasAdults = df.any { age > 18 }
+ * ```
+ *
+ * @param predicate A [RowFilter] lambda that takes a [DataRow] (as both `this` and `it`)
+ * and returns `true` if the row should be considered a match.
+ * @return `true` if at least one row satisfies the [predicate], `false` otherwise.
```
Jolanrensen, line 43: this should then also link to DataFrame.all

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/between.kt
```diff
@@ -5,6 +5,18 @@ import org.jetbrains.kotlinx.dataframe.impl.between
 
 // region DataColumn
 
+/**
+ * Returns a [DataColumn] of [Boolean] values indicating whether each element
```
Jolanrensen, line 9: each element in `[this]`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/tail.kt
```diff
@@ -1,9 +1,12 @@
 package org.jetbrains.kotlinx.dataframe.api
 
 import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.util.TAIL
+import org.jetbrains.kotlinx.dataframe.util.TAIL_REPLACE
 
 // region DataFrame
 
+@Deprecated(TAIL, ReplaceWith(TAIL_REPLACE), DeprecationLevel.ERROR)
```
Jolanrensen: I would keep `tail` as shortcut for `takeLast(5)`, since we also have `head` as a shortcut for `take(5)`. It's easy to get a quick glance at the beginning or end of a dataframe

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Create.kt
```diff
@@ -79,7 +78,8 @@ class Create : TestBase() {
         // SampleEnd
     }
 
-    @Test
+    // Not used anymore
```
Jolanrensen: please remove then :)

examples/kotlin-dataframe-plugin-example/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/plugin/Main.kt
```diff
@@ -6,9 +6,11 @@ import org.jetbrains.kotlinx.dataframe.api.add
 import org.jetbrains.kotlinx.dataframe.api.convert
 import org.jetbrains.kotlinx.dataframe.api.convertTo
 import org.jetbrains.kotlinx.dataframe.api.filter
+import org.jetbrains.kotlinx.dataframe.api.insert
```
Jolanrensen: unused imports

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/any.kt
AndreiKingsley: Why? IMHO these functions are really related.

AndreiKingsley: Since they all operate on column values with a Boolean condition.

zaleslaw: Agree with Andrew here

zaleslaw (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/any.kt
Jolanrensen: well so does `takeWhile`, `dropWhile` and I can think of some more.

I would only group `all {}` and `any {}` since they both reduce to a single boolean:
`Column<T>.function(predicate: (T) -> Boolean): Boolean`
 (`none {}` would be here too, if we had it... do we?)

Issue #1140, opened 2025-04-22
Notebook with project dependencies can end up without loaded Jupyter integration


koperagen:
Consider this setup:
Project with dataframe-core or dataframe `1.0.0-dev-6991` or later
Notebook with project dependencies 

DataFrame will be available, but no imports and no codegen. Because integration no longer comes with Jupyter Integration, so we need an extra dependency

We can mitigate this problem, add dependency on Jupyter to umbrella dataframe artifact. But proper fix should fix it for `dataframe-core` too. So probably will involve fixes on kernel side

Jolanrensen:
@ileasile there might be interest from other libraries as well for "hinting" to the jupyter kernel that a specific dependency is needed for a library to work. The [Kotlin Spark API](https://github.com/Kotlin/kotlin-spark-api) for instance has the same issue.

ileasile:
This exact issue could be fixable in the .json descriptor I believe

PR #1415, opened 2025-08-28
Load dataframe-jupyter from core in notebooks

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1140

Adds libraries.json which loads dataframe-jupyter from dataframe-core in notebooks. Requires https://github.com/Kotlin/kotlin-jupyter/pull/488.

TODO: Wait for the above PR to be merged and published. Adding this new argument will break existing kernels due to the `"descriptors"` argument being unknown.

### DIFF
--- a/core/src/main/resources/META-INF/kotlin-jupyter-libraries/libraries.json
+++ b/core/src/main/resources/META-INF/kotlin-jupyter-libraries/libraries.json
@@ -1,0 +1,9 @@
+{
+  "descriptors": [
+    {
+      "init": [
+        "@Suppress(\"INVISIBLE_MEMBER\", \"INVISIBLE_REFERENCE\") USE { dependencies(\"org.jetbrains.kotlinx:dataframe-jupyter:${org.jetbrains.kotlinx.dataframe.BuildConfig.VERSION}\") }"
+      ]
+    }
+  ]
+}


Jolanrensen:
https://github.com/Kotlin/kotlin-jupyter/pull/488 was merged, however, there's still the issue of providing "unsupported keys" in `libraries.json` for current kernels. Let's wait until Beta 4 to merge this PR, so people have a chance to bump kernel versions first.

zaleslaw (approved):

core/src/main/resources/META-INF/kotlin-jupyter-libraries/libraries.json
```diff
@@ -0,0 +1,9 @@
+{
+  "descriptors": [
+    {
+      "init": [
+        "@Suppress(\"INVISIBLE_MEMBER\", \"INVISIBLE_REFERENCE\") USE { dependencies(\"org.jetbrains.kotlinx:dataframe-jupyter:${org.jetbrains.kotlinx.dataframe.BuildConfig.VERSION}\") }"
```
koperagen: Maybe make it public or semi-hidden through some hacks? This suppress warning can break in future versions of kotlin

Jolanrensen: it's `internal`, so basically semi-hidden, right?

Jolanrensen: In my opinion we could also add it to a companion object of DataFrame, like `DataFrame.VERSION`; doesn't even need to be hidden

Jolanrensen:
wow that's a lot of failing tests all of a sudden, let me check..


...of course... the tests are trying to load the META-INF file from :core using a VERSION that doesn't exist. This file should be ignored by the tests

core/src/main/resources/META-INF/kotlin-jupyter-libraries/libraries.json
koperagen: semi hidden but also avoid suppressing invisible reference :) 
DataFrame.VERSION idk, not worth to pollute IO scope with it

Jolanrensen: I might not need it after all :) Currently, the tests in dataframe-jupyter are failing because the libraries.json is in the classpath. I'm trying to generate and include it on the fly upon publishing only

Jolanrensen:
Okay, I removed the libraries.json file from the resources. Now it's only added upon publishing, similar to running KoDEx only upon publishing. That way the tests that depend on :core don't have to deal with the file and potentially try to load a version of dataframe-jupyter that doesn't exit.

Jolanrensen:
"upon publishing" is not a good idea. It's still triggered from some integration tests that call `publishToMavenLocal`. This is likely why https://github.com/Kotlin/dataframe/pull/1034 was necessary.

Instead, I added a new gradle parameter `includeCoreLibrariesJson` similar to `skipKodex`. Our publisher calls it already:
https://jetbrains.team/p/kds/repositories/kotlin-ds-teamcity-config/revision/d170537ce93ed7ca9292efeecefb706e604c65fe

I ran a test-run of publish-dev and it seems to both build/test fine and publish with the json file included:
image

core/build.gradle.kts
```diff
@@ -251,6 +251,52 @@ val changeJarTask by tasks.registering {
     }
 }
 
+// generateLibrariesJson makes sure a META-INF/kotlin-jupyter-libraries/libraries.json file is generated
+// This file allows loading dataframe-jupyter when dataframe-core is present on its own in a Kotlin Notebook.
+val generatedJupyterResourcesDir = layout.buildDirectory.dir("generated/jupyter")
+val generateLibrariesJson by tasks.registering {
+    val outDir = generatedJupyterResourcesDir.get().asFile.resolve("META-INF/kotlin-jupyter-libraries")
```
koperagen, line 258: I recently fixed similar process resources task with dataframe version to set input/outputs for correct task invalidation somewhere in plugins. I think we need the same here

Jolanrensen: do you remember where? :)

Jolanrensen: something like `inputs.property("version", project.version)`?

PR #1420, opened 2025-09-01
Implicit char to string v2

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/998 and surpasses https://github.com/Kotlin/dataframe/pull/999.

(Can either be for ~Beta3 or~ Beta4, whatever suits us best)

We again have three parts, but slightly modified from #999:

### `String`-fallback for `Char` in `convert`
We get `String`-fallback for `Char` columns in `df.convert().to<>()`... and `col.convertTo<X>()`. This means you can now do things like:

```kt
enum class EnumClass { A, B }

columnOf('A', 'B').convertTo<EnumClass>()
// previously you could only do this
columnOf('A', 'B').convertToString().convertTo<EnumClass>()
```

Note: We do keep the old `convert` behavior of `Char` -> `Int` by ASCII code. This is in line with the JVM world. We have parsing now if you want the `Char` `'1'` to turn into the `Int` `1`.

Converting _from_ `String` columns can only result in enum instances (like above), value class instances, or call `parse()`, so we follow the same behavior for `Char` columns now.

### `Char` parsing

`df.parse()` now not only parses string columns, but also char ones (this can be changed if you don't agree, of course).
We also gain:
```kt
columnOf('1', '2', '3').parse() // results in DataColumn<Int>
columnOf('T', 'F', 'T').parse() // results in DataColumn<Boolean>
```
Here I deviate from #999, as
```kt
columnOf('a', 'b', 'c').parse() // throws IllegalStateException

// just like
columnOf("aa", "bb", "cc").parse() // throws IllegalStateException
```
because the parsing didn't do anything. However similar to `String` columns, there's `tryParse()` which can return the same type as the input.

### `convertTo<> { parser {} }` `Char` support

Fixes https://github.com/Kotlin/dataframe/issues/998 by trying out all provided `String` converters if `Char`s are encountered without a converter.

```kt
df.convertTo<Schema>() {
    // can now be used for both String and chars
    parser { /* it: String -> */ MyCustomClass(it) }

    // will be used instead of parser {} if present
    // convert<Char>().with { /* it: Char -> */ ... }
}
```

I also updated the docs and expanded upon `convertTo` a lot to prevent confusion, like what was experienced in #998.


docs/StardustDocs/topics/convert.md
```diff
@@ -71,6 +71,13 @@ df.convert { name }.asColumn { col ->
 * `LocalDate` (kotlinx.datetime and java.time)
 * `LocalTime` (kotlinx.datetime and java.time)
 * `Instant` (kotlinx.datetime, kotlin.time, and java.time)
+* `enum` classes (by name)
+
+Note that converting between `Char` and `Int` is done by ASCII character code.
```
zaleslaw: Let's make it as a Warning or NOTE - it's a very important fact

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ParserTests.kt
```diff
@@ -46,6 +46,12 @@ class ParserTests {
         DataFrame.parser.resetToDefault()
     }
 
+    @Test
+    fun `parse to Char`() {
```
zaleslaw, line 50: Should it be an opposite test, which has columnOf('a', 'b')
col.parse().type() shouldBe typeOf<**String**>()?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/parse.kt
```diff
@@ -323,7 +339,27 @@ public fun <T> DataFrame<T>.parse(options: ParserOptions? = null): DataFrame<T>
  * @return a new column with parsed values
  */
 public fun DataColumn<String?>.parse(options: ParserOptions? = null): DataColumn<*> =
-    tryParse(options).also { if (it.typeClass == String::class) error("Can't guess column type") }
+    tryParse(options).also { if (it.isSubtypeOf<String?>()) error("Can't guess column type") }
+
+/**
+ * Tries to parse a column of chars as strings into a column of a different type.
+ * Each parser in [Parsers] is run in order until a valid parser is found,
+ * a.k.a. that parser was able to parse all values in the column successfully. If a parser
```
zaleslaw, line 347: Great explanation!

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ParserTests.kt
Jolanrensen: No, that no longer happens. `columnOf('a', 'b').parse()` throws IllegalStateException because it cannot be parsed to any other type than `Char`. You could write `tryParse()` but then the result would still be of type `Char`

AndreiKingsley (approved):
Great, nothing to add!

As @zaleslaw mentioned, there should be more warnings (and I'd add it everywhere in docs) that now `Char` -> `Int` is converted in ASCII coding.

Issue #24, opened 2021-06-01
Support Multiplatform


ileasile:
Most of the library code is common, exceptions are IO parts and Jupyter integration. We may support KMP (at least K/JS) for this library

Jolanrensen:
There's probably too much JVM reflection going on for this to be easy, let alone viable :/

icecreamparlor:
Any Updates ?

Jolanrensen:
@icecreamparlor Nope. While it would be cool, there are just so many JVM dependencies in the project right now, so while in theory, it should be possible, it would be a huge undertaking. 

If performance would be the reason to go multiplatform, I think we still have a lot to gain when the [Vector API](https://openjdk.org/jeps/426) hits the JVM eventually, plus we have plans to convert our Lists to primitive arrays eventually https://github.com/Kotlin/dataframe/issues/30.

If, aside from performance, there are other needs for multiplatform support, I would be interested in seeing a proof of concept of (part of) the API, so we can then properly decide whether it would be worth the effort or not.

devcrocod:
* to support multiplatform, needed to switch to kotlinx.serialization - #312 
It would be nice if the part with the parser was separated into a separate module

* also appeared kotlinx-io - https://github.com/Kotlin/kotlinx-io

devcrocod:
Doubts arose about multiplatform support in dataframe, because the library uses a lot of reflections.

I've looked into this a bit, most of the reflection we use is in common. Therefore, if there are problems, then in isolated cases
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect/

Jolanrensen:
If you look at implementation files like this:
https://github.com/Kotlin/dataframe/blob/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt you can see we use `jvmErasure` a lot all over the place. I'm not sure if there's a common alternative for that. This needs to be checked.

devcrocod:
In fact, this is the only thing I noticed that is strongly tied to the platform.

In this case, for jvm everything will be the same, for native it will be possible to come up with a workaround, with wasmJs I’m not sure if it’s worth supporting at all

devcrocod:
I took a closer look at multiplatform support and conducted some experiments with it. Initially, I made some erroneous conclusions. 

Here are the issues I discovered:

- Support for generated code and proper source configuration
- DataFrame plugin - there is support for different targets in ksp.
[...]

PR #1421, opened 2025-09-01
Add API that debugger renderer can rely on to avoid breaking changes

koperagen:
This is needed so adding a new argument to `toJsonWithMetadata` won't break intellij plugin. Now, even with default value, this is a breaking change from POV of debugger caller.


dataframe-json/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/JsonFacadeForDebugger.java
```diff
@@ -0,0 +1,16 @@
+package org.jetbrains.kotlinx.dataframe.io;
+
+import org.jetbrains.kotlinx.dataframe.DataFrame;
+import java.util.Collections;
+
+class JsonFacadeForDebugger {
```
Jolanrensen, line 6: Any reason this is java? There's `@JvmStatic()` too if you need it :)

koperagen: More straightforward ABI: avoiding false assumption that parameter with a default value can be added, file name, static method

Jolanrensen: okay, but we got a failing javadoc task because we never used to have java XD

Jolanrensen: I think we don't add java source files from gradle

koperagen: pushed a fix disabling javadoc task. it was anyway an empty arfitact
image

Jolanrensen: Then the only issue I have left is https://github.com/Kotlin/dataframe/issues/24 but then again, when it comes around to that we can always move it to the jvm specific module

PR #1425, opened 2025-09-03
[Compiler plugin] Support more functions in toDataFrame DSL and AddDsl

koperagen:


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -169,6 +169,10 @@ public interface TraversePropertiesDsl {
     public fun preserve(vararg properties: KCallable<*>)
 }
 
+/**
+ * Store values of given type [T] in ValueColumns without transformation into ColumnGroups or FrameColumns.
+ */
+@Interpretable("PreserveT")
 public inline fun <reified T> TraversePropertiesDsl.preserve(): Unit = preserve(T::class)
```
Jolanrensen, line 176: never knew we had this function, thanks for the docs!

PR #1427, opened 2025-09-04
Prepare utility to assert compile time column order

koperagen:
Without this flag actual column order in the schema is unavailable, so i cannot check if my implementation of sorting in compiler toDataFrame counterpart does anything useful

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/schema.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/schema.kt
@@ -26,5 +26,9 @@ public fun GroupBy<*, *>.schema(): DataFrameSchema = toDataFrame().schema()
 
 // endregion
 
-public inline fun <reified T> DataFrame<T>.compileTimeSchema(): DataFrameSchema =
-    compileTimeSchemaImpl(schema(), T::class)
+/**
+ * [ordered] - if true, columns are ordered the same as in runtime schema for easier diff between the two.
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/schema.kt
```diff
@@ -26,5 +26,9 @@ public fun GroupBy<*, *>.schema(): DataFrameSchema = toDataFrame().schema()
 
 // endregion
 
-public inline fun <reified T> DataFrame<T>.compileTimeSchema(): DataFrameSchema =
-    compileTimeSchemaImpl(schema(), T::class)
+/**
+ * [ordered] - if true, columns are ordered the same as in runtime schema for easier diff between the two.
+ * if false, columns are ordered as they are represented in the compiler plugin
+ */
+public inline fun <reified T> DataFrame<T>.compileTimeSchema(ordered: Boolean = true): DataFrameSchema =
```
Jolanrensen, line 33: how's binary compatibility for this function? doesn't it break older versions of the compiler plugin?

if so, add a separate overload without argument :)

Jolanrensen (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/schema.kt
koperagen: No. Old versions of compiler plugin continue to use dataframe they've been published with, so it's ok

PR #1428, opened 2025-09-04
Update Gradle to 9.0

koperagen:
Our old Gradle plugin requires Kotlin 2.2.0 in Gradle runtime to work, because ParseKt refers to the new kotlin.time.Instant class and for 8.14.2 + 1.0.0-dev-8342 generate schema task fails ClassNotFound exception. (this problem itself bothers me a little, but ok)

Removing SimpleGit because it's for now incompatible with 9.0

### DIFF
--- a/examples/kotlin-dataframe-plugin-example/build.gradle.kts
+++ b/examples/kotlin-dataframe-plugin-example/build.gradle.kts
@@ -2,8 +2,8 @@ import org.jlleitschuh.gradle.ktlint.KtlintExtension
 
 plugins {
     id("org.jlleitschuh.gradle.ktlint") version "12.3.0"
-    kotlin("jvm") version "2.2.20-dev-3524"
-    kotlin("plugin.dataframe") version "2.2.20-dev-3524"
+    kotlin("jvm") version "2.2.20-Beta2"
+    kotlin("plugin.dataframe") version "2.2.20-Beta2"
 }
 
 group = "org.example"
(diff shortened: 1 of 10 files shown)


examples/kotlin-dataframe-plugin-example/build.gradle.kts
```diff
@@ -2,8 +2,8 @@ import org.jlleitschuh.gradle.ktlint.KtlintExtension
 
 plugins {
     id("org.jlleitschuh.gradle.ktlint") version "12.3.0"
-    kotlin("jvm") version "2.2.20-dev-3524"
-    kotlin("plugin.dataframe") version "2.2.20-dev-3524"
+    kotlin("jvm") version "2.2.20-Beta1"
```
Jolanrensen: Beta2 is also released already

gradle/libs.versions.toml
```diff
@@ -4,16 +4,16 @@ kotlinJupyter = "0.15.0-587"
 
 ktlint = "12.3.0"
 
-# make sure to sync both manually with :generator module
-kotlin = "2.2.0" # needs jupyter compatibility with Kotlin 2.1 to update
+kotlin = "2.2.20-Beta1"
```
Jolanrensen: Beta2?

gradle/libs.versions.toml
```diff
@@ -4,16 +4,16 @@ kotlinJupyter = "0.15.0-587"
 
 ktlint = "12.3.0"
 
-# make sure to sync both manually with :generator module
-kotlin = "2.2.0" # needs jupyter compatibility with Kotlin 2.1 to update
+kotlin = "2.2.20-Beta2"
```
Jolanrensen, line 7: please also bump ksp, I think 2.2.20-Beta2-2.0.2 is available.

(Can we substitute version in toml files? Like `ksp=$kotlin-2.0.2` or something? that would come in handy)

Issue #719, opened 2024-06-04
`Map<String, Any?>.toDataRow()` function?


Jolanrensen:
DataFrame has the function pair:
`Map<String, Iterable<Any?>>.toDataFrame()` and `AnyFrame.toMap(): Map<String, List<Any?>>`

But for `DataRow` there's just: `AnyRow.toMap(): Map<String, Any?>`.

Especially in notebooks it might be helpful to be able to convert a simple `Map` to a `DataRow` to get generated accessors.
It might also make notations like https://github.com/Kotlin/dataframe/issues/710 possible (so to construct a DataFrame from rows instead of from columns, albeit at a performance cost).

Jolanrensen:
Additionally, we should allow adding unfolding features like `maxDepth` to unpack properties. Plus, for large maps, it might be preferable to make a KeyValueDataFrame instead of a DataRow

CarloMariaProietti:
Hey, i would like to work on this issue

CarloMariaProietti:
Reading the issue, i understand that the returned DataRow should be part of a DataFrame which contains only that row. Is it correct?

Jolanrensen:
@CarloMariaProietti Hi! Thanks for picking this up :)

Yes, that's one way to tackle it, I'll check out the PR to see how it works

PR #1433, opened 2025-09-07
adding from_map_to_data_row function

CarloMariaProietti:
Solving issue #719

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/typeConversions.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/typeConversions.kt
@@ -406,7 +406,7 @@ public fun <T> DataRow<T>.toDataFrame(): DataFrame<T> = owner[index..index]
 
 public fun AnyRow.toMap(): Map<String, Any?> = df().columns().associate { it.name() to it[index] }
 
-public fun Map<String, Any?>.toDataRow() : DataRow<*> {
+public fun Map<String, Any?>.toDataRow(): DataRow<*> {
     val df = mapValues { listOf(it.value) }.toDataFrame()
     return DataRowImpl(0, df)
 }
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/typeConversions.kt
```diff
@@ -403,6 +406,11 @@ public fun <T> DataRow<T>.toDataFrame(): DataFrame<T> = owner[index..index]
 
 public fun AnyRow.toMap(): Map<String, Any?> = df().columns().associate { it.name() to it[index] }
 
+public fun Map<String, Any?>.toDataRow() : DataRow<*> {
```
Jolanrensen: this is not well linted, run `ktLintFormat` for automatic linting before committing :)

Jolanrensen:
Could you add a test for it? Maybe in `org.jetbrains.kotlinx.dataframe.testSets.person.DataRowTests`, that may be a good place. The implementation itself looks good :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/typeConversions.kt
CarloMariaProietti: Thank you for reviewing, i have run ktLintFormat. Am I ready to open a new pull request? I'll try to write tests as soon as i can.

Jolanrensen: no need to open a new pull request :) just push to the current branch

CarloMariaProietti:
I pushed a commit where i add test and fix the linting

Jolanrensen:
All looks well, the CI also passes. Thanks! Feel free to merge (are you allowed to do that btw? I'm not sure actually, otherwise I'll hit the button :) )

It will likely be part of Beta4 :)

CarloMariaProietti:
Thank you! I am not allowed to merge, it would be great if you could do it.

PR #1437, opened 2025-09-09
Join docs

### LINKED ISSUE #583: What is `filter` in join types?
https://kotlin.github.io/dataframe/join.html#join-types

AndreiKingsley:
Closes #583


docs/StardustDocs/topics/operations/multiple/join.md
```diff
@@ -0,0 +1,233 @@
+[//]: # (title: join)
+
+<!---IMPORT org.jetbrains.kotlinx.dataframe.samples.api.multiple.JoinSamples-->
+
+Joins two [`DataFrame`](DataFrame.md) object by join columns.
```
Jolanrensen: objects

Jolanrensen: ah you just copied this part from the original file

Jolanrensen:
While this looks much better already, I wonder if there's something we can do with colors, like we have for joinWith: https://kotlin.github.io/dataframe/joinwith.html

I think these could help a lot with the visualization of which values end up where. For instance, imagine a yellow dataframe and a light blue one being joined, and the resulting dataframe consisting of the yellow columns, the blue ones and green for the join column(s).
Something per-row, like in joinWith is also possible, of course. WDYT?

We could probably even use the `format {}` operation for this :) now that we have support for that

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/multiple/JoinSamples.kt
```diff
@@ -0,0 +1,271 @@
...
+
+    private fun nameToColor(name: String): RgbColor = when(name) {
+        "Alice"   -> RgbColor(189, 206, 233)
+        "Bob"     -> RgbColor(198, 224, 198)
+        "Charlie" -> RgbColor(219, 198, 230)
+        else -> RgbColor(255, 255, 255)
+    }
+
+    private fun nameAndCityToColor(name: String, city: String?): RgbColor = when(name to city) {
+        "Alice" to "London"  ->RgbColor(242, 210, 189)
```
Jolanrensen: linting

docs/StardustDocs/topics/operations/multiple/join.md
```diff
@@ -0,0 +1,362 @@
...
+// Merge a row when dfAges.firstName == dfCities.name.
+// With the given data all 3 names match → all rows merge.
+dfAges.join(dfCities) { firstName match right.name }
+```
+
+<!---END-->
+
+<inline-frame src="./resources/notebook_test_join_6.html" width="100%" height="500px"></inline-frame>
+
+If mapped columns have the same name, just select join columns from the left [`DataFrame`](DataFrame.md):
```
Jolanrensen: Hmm it's hard to see where the previous example ends and the new one begins. Maybe you could give them a small title?

Jolanrensen: you do exactly that below :) nice

Jolanrensen:
The colors help a lot! Thanks :) 
Though I still feel like some information is missing, as joins are column-centric. Maybe you could experiment with additional formatting? like text color/thickness per column and background colors for rows? just an idea

docs/StardustDocs/topics/operations/multiple/join.md
```diff
@@ -0,0 +1,393 @@
...
+<inline-frame src="./resources/notebook_test_join_11.html" width="100%" height="500px"></inline-frame>
+
+> In this example, the "city" columns from the left and right dataframes do not match to each other.
+> After joining, the "city" column from the right dataframe is included in the result dataframe
+> with the name **"city1"** to avoid a name conflict.  
+> { style = "note" }
+
+### Join with implicit keys (all columns with the same name) {collapsible="true"}
+
+If `joinColumns` is not specified, columns with the same name from both [`DataFrame`](DataFrame.md)
```
Jolanrensen, line 117: You say `joinColumns` is not specified, yet in the example, you show `{ name and city }`

docs/StardustDocs/topics/operations/multiple/join.md
```diff
@@ -0,0 +1,393 @@
...
+
+<inline-frame src="./resources/notebook_test_join_14.html" width="100%" height="500px"></inline-frame>
+
+
+<!---FUN notebook_test_join_17-->
+
+```kotlin
+// LEFT JOIN:
+// Keep ALL left rows. If (name, city) matches, attach right columns;
+// if not, right columns are null (e.g., Alice–London has no right match).
```
Jolanrensen: *columns from the right dataframe

Also, Alice-London does have a match, Bob-Dubai does not, so `isBusy == null`

docs/StardustDocs/topics/operations/multiple/join.md
```diff
@@ -0,0 +1,393 @@
...
+<!---END-->
+
+<inline-frame src="./resources/notebook_test_join_14.html" width="100%" height="500px"></inline-frame>
+
+<!---FUN notebook_test_join_15-->
+
+```kotlin
+// INNER JOIN:
+// Keep only rows where (name, city) match on both sides.
+// In this dataset both Charlies match twice (Moscow, Milan) → 2 merged rows.
```
Jolanrensen: what do you mean "both charlies"?, the result shows `Alice` and `Charlie`

docs/StardustDocs/topics/operations/multiple/join.md
```diff
@@ -0,0 +1,393 @@
...
+
+<inline-frame src="./resources/notebook_test_join_14.html" width="100%" height="500px"></inline-frame>
+
+
+<!---FUN notebook_test_join_19-->
+
+```kotlin
+// FULL JOIN:
+// Keep ALL rows from both sides. Where there's no match on (name, city),
+// the other side is filled with nulls.
```
Jolanrensen, line 354: could you make all `nulls` bold? so it's clearer these are new, added to fill in the gaps

Jolanrensen: `.format().with { if (it == null) bold else null }` will do

I still find the examples hard to follow, I'm wondering how we can make it as clear as possible :)

Jolanrensen (changes requested):

docs/StardustDocs/topics/operations/multiple/join.md
AndreiKingsley: Yep, I forgot to remove it!

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/multiple/JoinSamples.kt
```diff
@@ -281,7 +281,7 @@ class JoinSamples : DataFrameSampleHelper("join", "api") {
         dfLeft.rightJoin(dfRight) { name and city }
             // SampleEnd
             .colorized()
-            .format().with { if (it == null) bold else null }
+            .format { all() except (name and city) }.with { if (it == null) bold else null }
```
Jolanrensen, line 284: there's also `allExcept { name and city }` ;P but don't worry about it, this works fine :)

PR #1439, opened 2025-09-10
include JsonFacadeForDebugger in compilation

koperagen:
it wasn't in the jar because java didn't see the file in Kotlin source set; i changed java to point in the same directory because separate source sets for different languages is an outdated and bad practice :) 
image

### DIFF
--- a/dataframe-json/build.gradle.kts
+++ b/dataframe-json/build.gradle.kts
@@ -53,6 +53,12 @@ tasks.test {
     useJUnitPlatform()
 }
 
+sourceSets {
+    main {
+        java.srcDirs("src/main/kotlin")
+    }
+}
+
 val instrumentedJars: Configuration by configurations.creating {
     isCanBeConsumed = true
     isCanBeResolved = false


dataframe-json/build.gradle.kts
```diff
@@ -53,6 +53,12 @@ tasks.test {
     useJUnitPlatform()
 }
 
+sourceSets {
+    main {
+        java.srcDirs("src/main/kotlin")
```
Jolanrensen, line 58: wait, this works? In the past I could've sworn java had something hardcoded somewhere that .java files needed to be in a 'java' folder. For Kotlin this wasn't a constraint, so mixed projects would use the 'java' directory. This is still the case on Android.

koperagen: On android indeed :) But apparently it does, for example in Kotlin repo they don't even have src/main/kotlin, only src/ for main sources and test/ or testFixtures/ etc for tests, with both java and kotlin files

PR #1443, opened 2025-09-12
Independent framecol formatting

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/982

Removed the copying of `cellFormatter` from `DisplayConfiguration` to nested dataframe-like classes.
If the user wants to format them, manually make them `FormattedFrame`.

image

### DIFF
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/format.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/format.kt
@@ -333,11 +333,9 @@ class FormatTests : TestBase() {
             .rename("cityCopy").into("city")
 
         val formatted = df.format("city").with { bold and italic and textColor(green) }
-        val html =
-            formatted.toStandaloneHtml() + // expand the nested dataframes so we can see the difference
-                DataFrameHtmlData(script = "document.querySelectorAll('a.expander').forEach(a => a.click());")
+        val html = formatted.toHtml().toString()
 
-        html.toString().split("color:#00ff00").size - 1 shouldBe 12
-        html.toString().split("font-style:italic").size - 1 shouldBe 6
+        html.split("color:#00ff00").size - 1 shouldBe 12
+        html.split("font-style:italic").size - 1 shouldBe 6
     }
 }


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/format.kt
```diff
@@ -312,4 +313,31 @@ class FormatTests : TestBase() {
...
+    fun `formatting a column shouldn't affect nested columns with the same name`() {
+        val df = dataFrameOf("firstName", "lastName", "age", "city", "weight", "isHappy")(
+            "Alice", "Cooper", 15, "London", 54, true,
+            "Bob", "Dylan", 45, "Dubai", 87, true,
+            "Charlie", "Daniels", 20, "Moscow", null, false,
+            "Charlie", "Chaplin", 40, "Milan", null, true,
+            "Bob", "Marley", 30, "Tokyo", 68, true,
+            "Alice", "Wolf", 20, null, 55, false,
+            "Charlie", "Byrd", 30, "Moscow", 90, true,
+        ).group("firstName", "lastName").into("name")
```
koperagen, line 329: Ok that's funny

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/format.kt
```diff
@@ -312,4 +313,31 @@ class FormatTests : TestBase() {
...
+            .groupBy("city").toDataFrame()
+            .add("cityCopy") { "city"<String>() }
+            .group("city").into("cityGroup")
+            .rename("cityCopy").into("city")
+
+        val formatted = df.format("city").with { bold and italic and textColor(green) }
+        val html =
```
koperagen: Additional script doesn't seem to be needed for static String checks

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/format.kt
koperagen: I liked this pipeline and tried to port it to see the result 
image

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/format.kt
Jolanrensen: interesting, you're right indeed. But makes sense, all formatting should be there already, just collapsed

PR #1451, opened 2025-09-17
Add json api required for table widget in debugger to tests

koperagen:
With this we will be able to debug tests in :core with the table widget

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -71,6 +71,7 @@ dependencies {
     implementation(libs.sl4j)
     implementation(libs.kotlinLogging)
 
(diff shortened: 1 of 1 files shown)


core/build.gradle.kts
```diff
@@ -71,6 +71,7 @@ dependencies {
     implementation(libs.sl4j)
     implementation(libs.kotlinLogging)
 
+    testImplementation(projects.dataframeJson)
```
Jolanrensen, line 74: this line is already present below :p

koperagen: o-oh. But why didn't widget work on first attempt. Alright, i'll re-check on master, this PR is not needed

Issue #1432, opened 2025-09-05
Make compiler plugin work in samples (for docs)


Jolanrensen:
As a follow-up for https://github.com/Kotlin/dataframe/pull/1414 @AndreiKingsley tried to enable the compiler plugin in ~`:tests`~ `:samples`.
This fails currently because the compiler plugin expects the jar of `:core`/`:dataframe-compiler-plugin-core` on the classpath.

Currently, this is not the case, as we depend on the module, not the released version; the jar.

We could use a tactic I came up with for kotlin-spark:
- We create a [gradle task](https://github.com/Kotlin/kotlin-spark-api/blob/00d8b2696e89d11599889fe77e555ae0a5442646/compiler-plugin/build.gradle.kts#L116) that runs the jar-task on each build and copies the jar into a separate directory
- And depend on it from the examples project, like [here](https://github.com/Kotlin/kotlin-spark-api/blob/00d8b2696e89d11599889fe77e555ae0a5442646/build.gradle.kts#L19) and [here](https://github.com/Kotlin/kotlin-spark-api/blob/00d8b2696e89d11599889fe77e555ae0a5442646/build.gradle.kts#L131)

PR #1456, opened 2025-09-18
Enables Compiler plugin for :samples

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1432

- Enables the compiler plugin instead of the old gradle plugin for :samples
- Enables `instrumentedJars` configuration and artifact to each subproject with a `jar` task
- Makes sure `:samples` compilation depends on the jar task of each relevant module
- Replaces the `implementation(projects.dataframe)` in :samples with an explicit dependency on all `instrumentedJars` from all relevant modules
- Manually adds all transitive (`api()`) dependencies from relevant modules to :samples.
- Rewrites all samples to take advantage of the compiler plugin features :)


samples/build.gradle.kts
```diff
@@ -37,8 +35,44 @@ repositories {
...
+        .replace(File.separatorChar, '/')
+}
+
 dependencies {
-    implementation(projects.dataframe)
+    runtimeOnly(projects.dataframe) // Must depend on jars for the compiler plugin to work!
+    implementation(files(dependentProjectJarPaths))
+
+    // include api() dependencies from dependent projects, as they are not included in the jars
+    dependentProjects.forEach {
```
koperagen, line 70: Another way would be to have a shadowed jar in root project

Jolanrensen: yes, but I wasn't sure how the compiler plugin would react to a shadowed version of dataframe being on the classpath instead of the actual jar

Jolanrensen: and I have mixed experience with the shadowJar plugin

PR #1459, opened 2025-09-20
Format header

### LINKED ISSUE #1470: Add `formatHeader`
It would be useful to be able to customise not only the DF cells, but also the header with the names of the columns.

AndreiKingsley:
Closes #1470.


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/formatHeader.kt
```diff
@@ -0,0 +1,174 @@
...
+import org.jetbrains.kotlinx.dataframe.columns.toColumnSet
+import org.jetbrains.kotlinx.dataframe.impl.getColumnPaths
+
+// region docs
+
+/**
+ * A lambda used to format a column header (its displayed name) when rendering a dataframe to HTML.
+ *
+ * The lambda runs in the context of [FormattingDsl] and receives the [ColumnWithPath] of the header to format.
+ * Return a [CellAttributes] (or `null`) describing CSS you want to apply to the header cell.
```
Jolanrensen: *the CSS

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/formatHeader.kt
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * Examples:
+ * - Center a header: `attr("text-align", "center")`
```
Jolanrensen: *the header

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/formatHeader.kt
```diff
@@ -0,0 +1,174 @@
...
+ * - Make it bold: `bold`
+ * - Set custom color: `textColor(rgb(10, 10, 10))`
```
Jolanrensen: *a custom color

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/formatHeader.kt
```diff
@@ -0,0 +1,174 @@
...
+    override fun toString(): String =
+        "HeaderFormatClause(df=$df, columns=$columns, oldHeaderFormatter=$oldHeaderFormatter, oldCellFormatter=$oldCellFormatter)"
+}
+
+// endregion
+
+// region DataFrame.formatHeader
+
+/**
+ * **Experimental API. It may be changed in the future.**
```
Jolanrensen, line 48: Would it make sense to introduce an Experimental OptIn for this API?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/formatHeader.kt
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * Creates a new [FormattedFrame] that uses the specified [HeaderColFormatter] to format the selected headers.
+ *
+ * Header formatting is additive: attributes from already-applied header formatters are combined with the newly
+ * returned attributes using [CellAttributes.and]. If a parent column group is selected, its attributes are
+ * applied to its children unless explicitly overridden.
+ */
+@Suppress("UNCHECKED_CAST")
+public fun <T, C> HeaderFormatClause<T, C>.with(formatter: HeaderColFormatter<C>): FormattedFrame<T> {
+    val selectedPaths = df.getColumnPaths(UnresolvedColumnsPolicy.Skip, columns).toSet()
```
Jolanrensen: maybe move this to `formatHeaderImpl`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/formatHeader.kt
```diff
@@ -0,0 +1,174 @@
...
+    val oldHeader = oldHeaderFormatter
+
+    val composedHeader: HeaderColFormatter<Any?> = { col ->
+        val path = col.path
+        // Merge attributes from selected parents
+        val parentAttributes = if (path.size > 1) {
```
Jolanrensen: wait... so you're merging attributes from parents here, but also in html.kt? how does it work?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
```diff
@@ -234,6 +237,29 @@ internal fun AnyFrame.toHtmlData(
                 HtmlContent(html, style)
             }
         }
+        val headerStyle = run {
+            val hf = configuration.headerFormatter
+            if (hf == null) {
+                null
+            } else {
+                // collect attributes from parents
```
Jolanrensen, line 247: you also do this in formatHeader.kt?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
```diff
@@ -826,12 +853,15 @@ public class DataFrameHtmlData(
...
  * For example hosting *.ipynb files with outputs on GitHub
+ * @param cellFormatter Optional cell formatter applied to data cells during HTML rendering.
+ * @param headerFormatter Optional header formatter applied to column headers; supports inheritance for nested column groups.
  */
 public data class DisplayConfiguration(
     var rowsLimit: Int? = 20,
     var nestedRowsLimit: Int? = 5,
     var cellContentLimit: Int = 40,
     var cellFormatter: RowColFormatter<*, *>? = null,
+    var headerFormatter: HeaderColFormatter<*>? = null,
```
Jolanrensen, line 866: this is a public data class; you're breaking binary compatibility by adding this, especially halfway

Jolanrensen: We had the same issue with `ParserOptions` :/ unfortunately, using data classes in the public api is a bit of a nightmare. Adding a new argument breaks the constructor, `componentX()`, and `copy()` function. 
Check out `ParserOptions` for a workaround for this

Jolanrensen:
image

Very nice and very useful!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/formatHeader.kt
AndreiKingsley: I believe kdoc note is enough.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/formatHeader.kt
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// endregion
+
+// region DataFrame.formatHeader
+
+/**
+ * **Experimental API. It may be changed in the future.**
+ *
+ * Selects [columns] whose headers should be formatted.
```
koperagen: `If unspecified, all columns will be formatted.`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
```diff
@@ -843,6 +875,67 @@ public data class DisplayConfiguration(
         public val DEFAULT: DisplayConfiguration = DisplayConfiguration()
     }
 
+    /** For binary compatibility. */
+    @Deprecated(
+        message = DISPLAY_CONFIGURATION,
+        level = DeprecationLevel.HIDDEN,
+    )
+    public constructor(
```
Jolanrensen, line 883: while this fixes binary compatibility, it does not fix written compatibility. Let's just hope people used this class witt named parameters only.

Issue #1431, opened 2025-09-05
Add `none` function for DataFrame


AndreiKingsley:
Add `none` function by analogy with `all` and `any` (for consistency with stdlib).

PR #1462, opened 2025-09-25
adding `.none {}` to `DataFrame`

CarloMariaProietti:
About issue #1431

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/none.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/none.kt
@@ -7,7 +7,6 @@ import org.jetbrains.kotlinx.dataframe.Predicate
 import org.jetbrains.kotlinx.dataframe.RowFilter
 import org.jetbrains.kotlinx.dataframe.columns.ColumnSet
 import org.jetbrains.kotlinx.dataframe.columns.ColumnsResolver
-import org.jetbrains.kotlinx.dataframe.columns.size
 import org.jetbrains.kotlinx.dataframe.columns.values
 import org.jetbrains.kotlinx.dataframe.documentation.DslGrammarTemplateColumnsSelectionDsl.DslGrammarTemplate
 import org.jetbrains.kotlinx.dataframe.impl.columns.ColumnListImpl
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/none.kt
```diff
@@ -52,3 +54,25 @@ public interface NoneColumnsSelectionDsl {
 }
 
 // endregion
+
+// region DataFrame
```
Jolanrensen: For consistency with similar files, please put this function above the interface :)

Jolanrensen:
Great addition! Thanks for including KDocs and tests too!

I see `.all {}` also has an overload on `DataColumn<T>`, could you do the same for `.none {}` while you're at it? :) that will save us work in the future

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/none.kt
CarloMariaProietti: Thank you for reviewing! Pushed a commit trying to solve the problem.

CarloMariaProietti:
I'll try to do the same for .none {} . Should I open a new PR  for this?

Jolanrensen:
> I'll try to do the same for .none {} . Should I open a new PR for this?

I think it's fine if they're in the same PR as they're siblings anyway :) thanks!

Jolanrensen:
please run `ktlintFormat` to check everything is formatted right and `apiDump`, as you added a new public function :) Then I'll run our CI on the branch and we should be good to go :)

CarloMariaProietti:
I ran ktlintFormat but i don't know about apiDump

Jolanrensen:
> I ran ktlintFormat but i don't know about apiDump

alright, I'll do it for you this time. The idea is that you run the linter with `ktlintFormat` and commit the changes. It takes care of formatting and unused imports etc. `apiDump` updates the .api file of our module to check binary compatibility across versions (with `apiCheck`). This should also be run and committed for PRs :)

Jolanrensen:
The CI agrees :) I'll merge it, thanks!

Issue #1098, opened 2025-03-14
DataRow from Map creation API


AndreiKingsley:
After #1097, `Map` is handled as a value type in
```
Iterable<Map<*, *>>.toDataFrame()
```

 But it's also useful to have extensions that creates DataRow from map / DataFrame from list of maps
```
Map<String, *>.toDataRow()
Iterable<Map<String, *>>.toDataFrame()
```
as was previously discussed in #90 and #719.

Jolanrensen:
Since https://github.com/Kotlin/dataframe/issues/719 we now have `Map<String, *>.toDataRow()`

CarloMariaProietti:
Hey, I would like to work on this issue.

Jolanrensen:
@koperagen what's the status of your rework of toDataFrame regarding this?

CarloMariaProietti:
I wrote Iterable<Map<String, *>>.toDataFrame() and a test for it. If it is allowed, I'll send a PR.

Jolanrensen:
@CarloMariaProietti sure, go ahead :) Thanks! We might need to check if it works well in combination with our current changes of `toDataFrame()`, but that's okay

PR #1463, opened 2025-09-29
From rows to data frame

CarloMariaProietti:
Solving issue #1098

### DIFF
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
@@ -427,7 +427,7 @@ class CreateDataFrameTests {
 
     @Test
     fun `should convert iterables of maps representing rows to DataFrame with value columns`() {
-        val maps: Iterable<Map<String, *>> = listOf(mapOf("a" to 1, "b" to true), mapOf("c" to 2, "d" to false),)
+        val maps: Iterable<Map<String, *>> = listOf(mapOf("a" to 1, "b" to true), mapOf("c" to 2, "d" to false))
         val df = maps.toDataFrame()
         df["a"][0] shouldBe 1
         df["b"][0] shouldBe true


Jolanrensen:
@koperagen do you think it will cause issues in combination with https://github.com/Kotlin/dataframe/issues/1465 ? or should this be good to go?

koperagen:
> @koperagen do you think it will cause issues in combination with #1465 ? or should this be good to go?

No issues, but in Iterable<T>.toDataFrame we'll treat Map as value type and won't unfold it. I think it makes sense as a default

So results would be,
1. With this new function
`listOf(mapOf("a" to 123)).toDataFrame()` => 
```
a: Int
```

2. With generic toDataFrame and deep conversion
```
class MapContainer(val myMap: Map<String, *>)

listOf(MapContainer(mapOf("a" to 123))).toDataFrame(maxDepth = 2)
``` 
=>
```
myMap: Map<String, *>
```

3. Later Map can be converted 
df.convert { myMap }.with { it.toDataRow() }

Jolanrensen:
@koperagen and what about `listOf(myMap, 1, "").toDataFrame()`? This is what I'm most concerned about

koperagen:
> @koperagen and what about `listOf(myMap, 1, "").toDataFrame()`? This is what I'm most concerned about

Iterable<T>.toDataFrame relies on T to traverse properties. T of listOf(myMap, 1, "") is Any => 
image

Nothing changes here

Jolanrensen:
Okay, in that case we can probably add it safely. Could you update this branch from `master` and commit `apiDump` again? :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -425,6 +425,16 @@ class CreateDataFrameTests {
         df["value"].toList() shouldBe maps
     }
 
+    @Test
+    fun `should convert iterables of maps representing rows to DataFrame with value column`() {
```
Jolanrensen: *columns

CarloMariaProietti:
I fixed test name, pulled master, ran apiDump.

Jolanrensen:
> Solving issue #1098

Btw, if you write "Fixes #1098", GitHub automatically closes the issue when the PR is merged

Issue #1061, opened 2025-02-14
Add compiler plugin support for `cumSum`


zaleslaw:
https://github.com/Kotlin/dataframe/blob/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt

zaleslaw:
Closed in #1077

zaleslaw:
@Jolanrensen before I've started to implement this ticket, I've discovered, that cumSum is missing in Aggregators.kt and also there is no cumSumFor and  cumSumOf for GroupBy section...could you please add them, when you will finish the Statistics

Jolanrensen:
https://code.jetbrains.team/p/kt/repositories/kotlin/reviews/23858/timeline

PR #1467, opened 2025-09-30
Cum sum fixes

Jolanrensen:
some bug fixes and slight improvements required for https://github.com/Kotlin/dataframe/issues/1061.

Also expanded kdoc

### DIFF
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/statistics/cumsum.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/statistics/cumsum.kt
@@ -1,5 +1,6 @@
 package org.jetbrains.kotlinx.dataframe.statistics
 
+import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.matchers.shouldBe
 import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.api.columnOf
@@ -8,7 +9,10 @@ import org.jetbrains.kotlinx.dataframe.api.cumSum
 import org.jetbrains.kotlinx.dataframe.api.dataFrameOf
 import org.jetbrains.kotlinx.dataframe.api.groupBy
 import org.jetbrains.kotlinx.dataframe.api.map
+import org.jetbrains.kotlinx.dataframe.impl.nullableNothingType
+import org.jetbrains.kotlinx.dataframe.math.cumSumTypeConversion
 import org.junit.Test
+import kotlin.reflect.typeOf
 
 @Suppress("ktlint:standard:argument-list-wrapping")
 class CumsumTests {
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt
```diff
@@ -24,15 +24,19 @@ import kotlin.reflect.KProperty
...
+ * `null` and `NaN` values are skipped when computing the cumulative sum.
+ * When `false`, all values after the first `NA` will be `NaN` (for [Double] and [Float] columns)
+ * or `null` (for other columns).
  *
- * {@get [CumSumDocs.CUMSUM_PARAM] @param [columns\]
- * The names of the columns to apply cumSum operation.}
+ * `cumSum` only works on columns that contain solely primitive numbers.
  *
- * @param [skipNA\] Whether to skip null and NaN values (default: `true`).
+ * Similar to [sum][sum], [Byte][Byte]- and [Short][Short]-columns are converted to [Int][Int].
```
zaleslaw, line 33: Sum-sum! Bye-Bye-Bye!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt
```diff
@@ -212,10 +220,11 @@ public fun <T, G> GroupBy<T, G>.cumSum(
  * {@set [CumSumDocs.DATA_TYPE] [GroupBy]}
  * {@set [CumSumDocs.CUMSUM_PARAM]}
  */
+@Refine
+@Interpretable("GroupByCumSum0")
```
zaleslaw, line 224: If we are marking the functions for Compiler Plugin, should we add a test for this trivial case?
@koperagen

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt
Jolanrensen: Yep, unfortunately, there are KDoc rendering issues if you don't alias the types here

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/cumSum.kt
Jolanrensen: testing what? and how?

koperagen: No, usually i simply annotate functions and later adjust if anything is wrong (one time i forgot Refine annotation)

koperagen (changes requested):
Please add a test
And perform interactive rebase + force push before merge so fixup! commits don't end up in the master

PR #1468, opened 2025-10-01
Kotlin Contracts with generics!

Jolanrensen:
explored experimental generics in kotlin contracts and what they could bring to dataframe

Shall we adopt it? or wait until it's not experimental anymore?

https://youtrack.jetbrains.com/issue/KT-45683

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
@@ -1,8 +1,9 @@
-@file:OptIn(ExperimentalContracts::class)
+@file:OptIn(ExperimentalContracts::class, ExperimentalExtendedContracts::class)
 
 package org.jetbrains.kotlinx.dataframe.api
 
 import org.jetbrains.kotlinx.dataframe.AnyCol
+import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.columns.ColumnGroup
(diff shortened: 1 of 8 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
```diff
@@ -63,9 +80,15 @@ public fun AnyCol.isPrimitiveNumber(): Boolean = type().isPrimitiveNumber()
...
  */
-public fun AnyCol.isPrimitiveOrMixedNumber(): Boolean = type().isPrimitiveOrMixedNumber()
+public fun AnyCol.isPrimitiveOrMixedNumber(): Boolean {
+    contract { returns(true) implies (this@isPrimitiveOrMixedNumber is ValueColumn<Number?>) }
+    return type().isPrimitiveOrMixedNumber()
+}
 
-public fun AnyCol.isList(): Boolean = typeClass == List::class
+public fun AnyCol.isList(): Boolean {
+    contract { returns(true) implies (this@isList is ValueColumn<List<*>>) }
```
koperagen, line 89: Need to consider nullability: List<*>?
is it safe?

zaleslaw:
Fine to have an experinmental PR, but to early to use this feature in 1.0

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
Jolanrensen: oh you're right, should indeed be `List<*>?`

PR #1473, opened 2025-10-02
Candidates for removal deprecation

### LINKED ISSUE #1028: Candidates for removal

AndreiKingsley:
Fixes #1028.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/copy.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/copy.kt
@@ -1,10 +1,11 @@
 package org.jetbrains.kotlinx.dataframe.api
 
 import org.jetbrains.kotlinx.dataframe.DataFrame
-import org.jetbrains.kotlinx.dataframe.annotations.CandidateForRemoval
+import org.jetbrains.kotlinx.dataframe.util.MESSAGE_SHORTCUT_1_0
+
 // region DataFrame
 
-@CandidateForRemoval
+@Deprecated(MESSAGE_SHORTCUT_1_0, ReplaceWith("columns().toDataFrame().cast()"), DeprecationLevel.ERROR)
 public fun <T> DataFrame<T>.copy(): DataFrame<T> = columns().toDataFrame().cast()
 
 // endregion
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
@@ -6,7 +6,6 @@ import org.jetbrains.kotlinx.dataframe.ColumnSelector
 import org.jetbrains.kotlinx.dataframe.ColumnsSelector
 import org.jetbrains.kotlinx.dataframe.DataFrame
 import org.jetbrains.kotlinx.dataframe.annotations.AccessApiOverload
-import org.jetbrains.kotlinx.dataframe.annotations.CandidateForRemoval
 import org.jetbrains.kotlinx.dataframe.annotations.Interpretable
 import org.jetbrains.kotlinx.dataframe.annotations.Refine
 import org.jetbrains.kotlinx.dataframe.columns.ColumnGroup
@@ -24,6 +23,7 @@ import org.jetbrains.kotlinx.dataframe.impl.api.moveImpl
 import org.jetbrains.kotlinx.dataframe.impl.api.moveTo
 import org.jetbrains.kotlinx.dataframe.ncol
 import org.jetbrains.kotlinx.dataframe.util.DEPRECATED_ACCESS_API
+import org.jetbrains.kotlinx.dataframe.util.MESSAGE_SHORTCUT_1_0
 import org.jetbrains.kotlinx.dataframe.util.MOVE_TO_LEFT
 import org.jetbrains.kotlinx.dataframe.util.MOVE_TO_LEFT_REPLACE
 import org.jetbrains.kotlinx.dataframe.util.MOVE_TO_RIGHT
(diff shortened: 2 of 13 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
```diff
@@ -409,7 +409,7 @@ public fun <T, C> MoveClause<T, C>.into(
 /**
  * Move a single selected column to top level with a new name
  */
-@CandidateForRemoval
+@Deprecated(MESSAGE_SHORTCUT_1_0, ReplaceWith("into { pathOf(column) }"), DeprecationLevel.ERROR)
 @Refine
 @Interpretable("MoveInto0")
 public fun <T, C> MoveClause<T, C>.into(column: String): DataFrame<T> = pathOf(column).let { path -> into { path } }
```
koperagen, line 425: Do you remember why we want to remove it? It's supported by compiler plugin, seems useful

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/copy.kt
```diff
@@ -1,10 +1,11 @@
 package org.jetbrains.kotlinx.dataframe.api
 
 import org.jetbrains.kotlinx.dataframe.DataFrame
-import org.jetbrains.kotlinx.dataframe.annotations.CandidateForRemoval
+import org.jetbrains.kotlinx.dataframe.util.MESSAGE_SHORTCUT_1_0
+
 // region DataFrame
 
-@CandidateForRemoval
+@Deprecated(MESSAGE_SHORTCUT_1_0, ReplaceWith("columns().toDataFrame().cast()"), DeprecationLevel.ERROR)
```
Jolanrensen: we also usually put the replaceWith contents in the deprecations file. I'm impartial to it, as long as we have at least one constant String there to track the deprecations

Jolanrensen:
should we immediately go to ERROR? I don't think many of these overloads hurt. So while we will clean them up, I wouldn't say there's much hurry, so WARNING could suffice

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
```diff
@@ -87,6 +87,7 @@ class ConstructorsTests {
         col.type() shouldBe typeOf<AnyRow>()
         col.kind() shouldBe ColumnKind.Group
         col[0] shouldBe row
+        @Suppress("DEPRECATION_ERROR")
```
Jolanrensen, line 90: If we deprecate them, we should probably remove any internal uses

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
Jolanrensen: is it just `rename {}.into {}` maybe? or `move {}.under("")`?

koperagen: it's  `move { myCol }.toTop().rename { myCol }.into("newName")`

AndreiKingsley: Well, I really don't see a reason 😄 . Let's keep it.

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
AndreiKingsley: for `DataRow.isEmpty()`, I'd make it internal after deprecation and public removal for such tests.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/copy.kt
AndreiKingsley: I don't like it, because it's hard to validate when it's remote, especially in singular cases + losing IDE code analysis.
And btw we have a lot of them in the project:
image

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
Jolanrensen: > move { myCol }.toTop().rename { myCol }.into("newName")

or, you know `move { myCol }.into(pathOf("newName"))` ;P

Jolanrensen: but yes, let's keep it, as we treat column paths and the String API as the same thing :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/copy.kt
Jolanrensen: makes sense :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/constructors.kt
Jolanrensen: good idea! But we should put somewhere that we will make it internal maybe. Or should we have this deprecation suppression mechanism being seen as "we're planning not to remove it but to make it internal"?

Issue #1021, opened 2025-01-21
Write `move {}.before {}`


Jolanrensen:
We have `move after` but `move before` was never finished.

CarloMariaProietti:
Hey, I would like to work on this issue.

Jolanrensen:
@CarloMariaProietti go ahead and give it a try :) When I looked at it, it was a bit more complex than expected, so let me know if you need any help.

Feel free to ignore compiler plugin support for now (the `@Refine @Interpretable` stuff). If `move {}.before {}` works, we can add it later.

There's no need to create overloads for functions we annotated with `@Deprecated(DEPRECATED_ACCESS_API) @AccessApiOverload` as they are going to be removed in a future version of DataFrame.

CarloMariaProietti:
I have been trying for a few days  but for now I did not succed. It would be great if you could help!

Jolanrensen:
How far did you get? :) Did you find the [MoveClause.afterOrBefore()` implementation function](https://github.com/Kotlin/dataframe/blob/82bf7ba7705d575dab976e704d1390772a8a3e99/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt#L24)?

The way I think that function works is:
- it first removes all columns selected by the `move {}` operatation, 
- it does some checking to see whether the move is not done to a "deeper level" (aka we cannot move a column group into itself), 
- then it finds the position of the reference inside `after {}` or `before {}` as `refNode`, 
- and finally, all columns that were removed get turned into a `ColumnToInsert` with the given `refNode`/`referenceNode` and passed on to the `insert` implementation.

Apparently, `insertImpl` is the one that decides that these columns should be placed after the one referenced.

That seems to be done here: https://github.com/Kotlin/dataframe/blob/82bf7ba7705d575dab976e704d1390772a8a3e99/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/insert.kt#L83
There appears to be some logic in there that gives a new index for the columns to be inserted relative to the provided `referenceNode` from `ColumnToInsert`.

Maybe you could try adding an extra parameter to `ColumnToInsert`, like `insertAfterReferenceNode: Boolean = true` and adapt the behavior in `insertImpl` depening on this flag.

[...]

PR #1474, opened 2025-10-03
Add move before

CarloMariaProietti:
Solves issue #1021. It comes with a test case which is inspired by after's test case.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
@@ -621,7 +621,7 @@ public fun <T, C> MoveClause<T, C>.after(column: KProperty<*>): DataFrame<T> = a
 
 // endregion
 
-// region after
+// region before
 
 /**
  * Moves columns, previously selected with [move] to the position before the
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
```diff
@@ -621,13 +621,62 @@ public fun <T, C> MoveClause<T, C>.after(column: KProperty<*>): DataFrame<T> = a
...
+ * df.move("age", "weight").before("surname")
+ * ```
+ * @param [column] The [Column Name][String] specifying the column
+ * before which the selected columns will be placed.
+ */
+public fun <T, C> MoveClause<T, C>.before(column: String): DataFrame<T> = before { column.toColumnAccessor() }
+
+@Deprecated(DEPRECATED_ACCESS_API)
+@AccessApiOverload
+public fun <T, C> MoveClause<T, C>.before(column: AnyColumnReference): DataFrame<T> = before { column }
```
Jolanrensen: no need to add these deprecated overloads :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
```diff
@@ -78,7 +82,16 @@ internal fun <T, C> MoveClause<T, C>.afterOrBefore(column: ColumnSelector<T, *>,
         }
         ColumnToInsert(path, sourceCol.data, refNode)
     }
-    return removeResult.df.insertImpl(toInsert)
+    if (isAfter) {
+        return removeResult.df.insertImpl(toInsert)
+    }
+
+    // move target after last of toInsert
```
Jolanrensen: could you expand this doc a bit? I have a bit of a hard time understanding the logic below.

so... you insert the removed columns after the target column in the first place, and then you move the target column after the removed/inserted columns?

That's... really clever :D

But it took me a while to get it :) so please expand the doc a tiny bit :)

CarloMariaProietti: Thank you very much, I appreciate it. That's exactly the logic. I'll try to make the doc more undesrtandable.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
CarloMariaProietti: I'll remove them

CarloMariaProietti:
I implemented requested changes, In addition I noticed a bug when the function throws exception, so i fixed it.

Jolanrensen (approved):

Issue #1465, opened 2025-09-29
Fix all issues with toDataFrame in the library and compiler plugin


koperagen:
We got 
1. inconsistent behaviour for nested "value types" compared to iterable of value types. 
2. Exclude classes not working at all. 
3. empty nested interfaces / classes become empty columns groups
This all needs to be fixed both in library and compiler plugin

PR #1475, opened 2025-10-03
`.toDataFrame` fixes

koperagen:
Fixes https://github.com/Kotlin/dataframe/issues/1465
Btw i think strict `ktlint_standard_function-expression-body` rule is unnecessary, functions with return and with expression body can be a preference and coexist. At least for me this rule ALWAYS fails and forces me to rewrite

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
@@ -11,7 +11,6 @@ import org.jetbrains.kotlinx.dataframe.annotations.Interpretable
 import org.jetbrains.kotlinx.dataframe.annotations.Refine
 import org.jetbrains.kotlinx.dataframe.columns.ColumnPath
 import org.jetbrains.kotlinx.dataframe.impl.ColumnNameGenerator
-import org.jetbrains.kotlinx.dataframe.impl.api.canBeUnfolded
 import org.jetbrains.kotlinx.dataframe.impl.api.createDataFrameImpl
 import org.jetbrains.kotlinx.dataframe.impl.asList
 import org.jetbrains.kotlinx.dataframe.impl.columnName
@@ -21,6 +20,7 @@ import org.jetbrains.kotlinx.dataframe.util.DEPRECATED_ACCESS_API
 import kotlin.reflect.KCallable
 import kotlin.reflect.KClass
 import kotlin.reflect.KProperty
+import kotlin.reflect.typeOf
 
 // region read DataFrame from objects
 
@@ -28,21 +28,13 @@ import kotlin.reflect.KProperty
 @Interpretable("toDataFrameDefault")
 public inline fun <reified T> Iterable<T>.toDataFrame(): DataFrame<T> =
     toDataFrame {
-        // check if type is value: primitives, primitive arrays, datetime types etc.,
-        // or has no properties
-        if (!T::class.canBeUnfolded) {
-            // create a single `value` column
-            ValueProperty<T>::value.name from { it }
-        } else {
-            // otherwise creates columns based on properties
-            properties()
-        }
+        properties()
     }
 
 @Refine
 @Interpretable("toDataFrameDsl")
 public inline fun <reified T> Iterable<T>.toDataFrame(noinline body: CreateDataFrameDsl<T>.() -> Unit): DataFrame<T> =
-    createDataFrameImpl(T::class, body)
+    createDataFrameImpl(typeOf<T>(), body)
 
 @Refine
 @Interpretable("toDataFrame")
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
@@ -34,6 +34,7 @@ import kotlin.reflect.full.isSubclassOf
 import kotlin.reflect.full.memberFunctions
 import kotlin.reflect.full.memberProperties
 import kotlin.reflect.full.primaryConstructor
+import kotlin.reflect.full.starProjectedType
 import kotlin.reflect.full.withNullability
 import kotlin.reflect.jvm.isAccessible
 import kotlin.reflect.jvm.javaField
@@ -201,6 +202,16 @@ internal fun <T> Iterable<T>.createDataFrameImpl(
     body: CreateDataFrameDslImpl<T>.() -> Unit,
 ): DataFrame<T> {
     val builder = CreateDataFrameDslImpl(this, type)
+    builder.body()
+    return builder.columns.toDataFrameFromPairs()
+}
+
+@Deprecated("backward compatibility for Kandy", level = DeprecationLevel.ERROR)
+internal fun <T> Iterable<T>.createDataFrameImpl(
+    clazz: KClass<*>,
+    body: CreateDataFrameDslImpl<T>.() -> Unit,
+): DataFrame<T> {
+    val builder = CreateDataFrameDslImpl(this, clazz.starProjectedType)
     builder.body()
     return builder.columns.toDataFrameFromPairs()
 }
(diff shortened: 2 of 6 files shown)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -213,8 +234,7 @@ class CreateDataFrameTests {
         df.rowsCount() shouldBe 1
 
         val childCol = df[Entry::child]
-        childCol.kind() shouldBe ColumnKind.Group
-        childCol.asColumnGroup().columnsCount() shouldBe 0
+        childCol.kind() shouldBe ColumnKind.Value
```
AndreiKingsley, line 236: Oh, that was weird, didn't know!

AndreiKingsley (changes requested):
Great, this logic is now really clear!
But please add a KDoc - or at least a comment for now, - to clearly define the behavior of `toDataFrame`; it's not actually trivial.

Jolanrensen:
> Btw i think strict ktlint_standard_function-expression-body rule is unnecessary, functions with return and with expression body can be a preference and coexist. At least for me this rule ALWAYS fails and forces me to rewrite

I have to disagree. Adding ` { return  }`  for me feels like a waste of space, and mixing both expression functions and single-line-return functions in the same file feels messy.
You, for instance, wouldn't mix lambdas and normal function definitions in one place either, even though they do the same thing.
```kt
val myFunction: (String) -> Unit = { arg1 ->
    doSomething()
}

fun myFunction(arg1: String) {
    doSomething()
}
```
It increases cognitive load. I'd be fine if they were all single-line-returns too, but we're a JetBrains Kotlin library, so if we don't follow the [guidelines](https://kotlinlang.org/docs/coding-conventions.html#functions), then who does? :P

But I understand it can be annoying if it fails the build :) For that I can recommend installing the KtLint plugin in the IDE and binding it to the reformatter, Ctrl+Alt+L. It highlights these cases and can fix them with one push of a button, or even upon file-save. Personally, I use it with pleasure, and as a result, the ktLintCheck almost never complains even though I often forget commas, write too long lines etc. And yes, sometimes it's a bit of a back-and-forth between me and the linter, but in the end the code looks clean, consistent, and readable.

koperagen:
> It increases cognitive load. I'd be fine if they were all single-line-returns too, but we're a JetBrains Kotlin library, so if we don't follow the [guidelines](https://kotlinlang.org/docs/coding-conventions.html#functions), then who does? :P

I agree with the following

```
fun foo(): Int {     // bad
    return 1
}
```

```
fun foo() = 1        // good
```

However in Kotlin anything can be a single expression:

```
fun foo() = run { 
  // multiple statement go here!
}
```

Key point why i'd like that function to have a body: a lot of lines of code in the expression. We can keep this rule but split expressions into multiple statements, and technically linter won't complain
For trivial cases when there's only single line or otherwise not much code, always use it sure! Or if you want to rely on type inference, might as well abuse run { } - with Kotlin DataFrame compiler plugin for example.

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -156,8 +156,11 @@ class CreateDataFrameTests {
         df.a[0].v shouldBe 7
 
         val df2 = data.toDataFrame {
-            preserve(B::row)
-            properties { preserve(DataFrame::class) }
+            preserve(B::row) // this@toDataFrame: TraversePropertiesDsl - works
+            properties {
+                preserve(DataFrame::class) // this@properties: TraversePropertiesDsl - always works
+            }
+            preserve(B::row) // this@toDataFrame: TraversePropertiesDsl - doesn't
```
Jolanrensen: can we deprecate this function here? since it doesn't always work

.editorconfig
```diff
@@ -33,6 +33,7 @@ ktlint_standard_chain-method-continuation = disabled
 ktlint_ignore_back_ticked_identifier = true
 ktlint_standard_multiline-expression-wrapping = disabled
 ktlint_standard_when-entry-bracing = disabled
+ktlint_standard_function-expression-body = disabled
```
Jolanrensen: Turning this off project-wide disables the quick-fix too. I use this quite often :( I'd rather recommend suppressing it on a class/file/function basis if you really need it for clarity. The KtLint plugin has a quick-fix option for suppressing.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
```diff
@@ -403,3 +410,15 @@ internal fun convertToDataFrame(
         dataFrameOf(columns)
     }
 }
+
+private fun KType.classifierOrAny(): KClass<*> = classifier as? KClass<*> ?: Any::class
+
+private fun KClass<*>.properties(): List<KCallable<*>> {
```
Jolanrensen, line 427: I'd either rewrite it declaratively/functionally, like:
```kt
private fun KClass<*>.properties(): List<KCallable<*>> =
    memberProperties
        .filter { it.visibility == KVisibility.PUBLIC }
        // fall back to getter functions for pojo-like classes if no member properties were found
        .ifEmpty {
            memberFunctions
                .filter { it.visibility == KVisibility.PUBLIC && it.isGetterLike() }
        }
```
or imperatively:
```kt
private fun KClass<*>.properties(): List<KCallable<*>> {
    val publicMembers = memberProperties.filter { it.visibility == KVisibility.PUBLIC }
    return if (publicMembers.isEmpty()) {
        // fall back to getter functions for pojo-like classes if no member properties were found
        memberFunctions.filter { it.visibility == KVisibility.PUBLIC && it.isGetterLike() }
    } else {
        publicMembers
    }
}
```
(or a variant of either). In either case the linter won't complain while sticking to one code paradigm.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -21,28 +20,21 @@ import org.jetbrains.kotlinx.dataframe.util.DEPRECATED_ACCESS_API
...
-            properties()
-        }
+        properties()
     }
 
 @Refine
 @Interpretable("toDataFrameDsl")
 public inline fun <reified T> Iterable<T>.toDataFrame(noinline body: CreateDataFrameDsl<T>.() -> Unit): DataFrame<T> =
-    createDataFrameImpl(T::class, body)
+    createDataFrameImpl(typeOf<T>(), body)
```
Jolanrensen, line 37: nice! We should really do a sweep of `::class` usage in DF. `typeOf<T>()` is cheaper I believe

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -632,4 +652,43 @@ class CreateDataFrameTests {
         val df = files.toDataFrame(columnName = "files")
         df["files"][0] shouldBe File("data.csv")
     }
+
+    class MyEmptyDeclaration
+
+    class TestItem(val name: String, val containingDeclaration: MyEmptyDeclaration, val test: Int)
+
+    @Test
+    fun `preserve empty interface consistency`() {
```
Jolanrensen, line 660: does it work as well for interfaces, data classes, and objects?

Jolanrensen (changes requested):
good increase in robustness :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
koperagen: No for data classes - they require at least one property. But in general anything T that has no member properties / getter like functions, here we rely on Kotlin reflection

.editorconfig
koperagen: Alright, if it's somehow useful i'll keep it

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
koperagen: Let's keep for now until we figure out case with multiple properties call in this lambda. "Top level" preserve/exclude help to reuse code between them

koperagen:
> Great, this logic is now really clear! But please add a KDoc - or at least a comment for now, - to clearly define the behavior of `toDataFrame`; it's not actually trivial.

https://github.com/Kotlin/dataframe/issues/1479

Jolanrensen (approved):

PR #1483, opened 2025-10-08
Update documentation of `append`

koperagen:
This example bothered me for a long time. It's not korro, but it's better then before

### DIFF
--- a/docs/StardustDocs/topics/append.md
+++ b/docs/StardustDocs/topics/append.md
@@ -2,10 +2,27 @@
 
 Adds one or several rows to [`DataFrame`](DataFrame.md)
 ```kotlin
-df.append (
+df.append(
     "Mike", 15,
     "John", 17, 
-    "Bill", 30)
+    "Bill", 30,
+)
+```
+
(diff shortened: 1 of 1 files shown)


docs/StardustDocs/topics/append.md
```diff
@@ -2,10 +2,27 @@
...
-df.append (
+df.append(
     "Mike", 15,
     "John", 17, 
-    "Bill", 30)
+    "Bill", 30,
+)
+```
+
+If the compiler plugin is enabled, typesafe overload of `append` is available for `@DataSchema` classes.
```
Jolanrensen: *a typesafe

docs/StardustDocs/topics/append.md
```diff
@@ -2,10 +2,27 @@
...
+
+```
```
Jolanrensen: kotlin

Jolanrensen (approved):
nice addition :)

docs/StardustDocs/topics/append.md
Jolanrensen: and can you link somewhere when you say "compiler plugin"?

PR #1486, opened 2025-10-09
Improve `list.chunked()` + `List<List<T>>.toDataFrame` for parsing .srt and similar text files

koperagen:
Consider file structured as this:
```
header1
data1

header2
data2
```
I believe it's popular, one example is srt, but i personally had to deal with it a lot 

It can be parsed into dataframe:
```kt
File("data").readText().lines().chunked(3).map { it.dropLast(1) }.toDataFrame()
```

*Surprisingly* here toDataFrame is not generic Iterable<T>.toDataFrame, but completely another function. 
*Problem*: In current shape it's not helpful. 
I end up with something very close to what i want, but required change to code is somewhat non-trivial.
image

I'd either have to:
```kt
val data = File("data").readText().lines().chunked(3).map { it.dropLast(1) }
val df = buildList { 
  add(listOf("col1", "col2"))
  addAll(data)
}.toDataFrame()
```

Or switch to completely different route: 
```kt
val df = File("data").readText().lines().chunked(3).map { it.dropLast(1) }.toDataFrame("data").split("data").cast<List<*>>().into("col1", "col2")
```
with compiler plugin
```kt
val df = File("data").readText().lines().chunked(3).map { it.dropLast(1) }.toDataFrame("data").split { data }.into("col1", "col2")
```

With this API change:
```kt
val df = File("data").readText().lines().chunked(3).map { it.dropLast(1) }.toDataFrame(header = listOf("col1", "col2"))
```

Plugin will understand resulting schema too

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/common.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/common.kt
@@ -48,6 +48,7 @@ public fun catchHttpResponse(url: URL, body: (InputStream) -> AnyFrame): AnyFram
 @Deprecated(
     LISTS_TO_DATAFRAME_MIGRATION,
     ReplaceWith("this.toDataFrame(header = null, containsColumns)", "org.jetbrains.kotlinx.dataframe.api.toDataFrame"),
+    level = DeprecationLevel.WARNING,
 )
 public fun <T> List<List<T>>.toDataFrame(containsColumns: Boolean = false): AnyFrame =
     toDataFrame(header = null, containsColumns)
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
@@ -713,7 +713,7 @@ class CreateDataFrameTests {
             }
         }
 
-        val df = lines.chunked(3).toDataFrame()
+        val df = lines.chunked(3).toDataFrame(header = null)
 
         df.columnNames() shouldBe listOf("stamp", "header", "data")
         df.columnTypes() shouldBe listOf(typeOf<String>(), typeOf<String>(), typeOf<String>())


zaleslaw:
I agree that this a popular use-case, I also faced with the same and handled with File API parsing

Great, if tiny example, saying, with subtitles will be also added

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/common.kt
```diff
@@ -48,34 +50,37 @@ public fun catchHttpResponse(url: URL, body: (InputStream) -> AnyFrame): AnyFram
...
+ *                        Otherwise, each nested list is a row.
  *                        Defaults to `false`.
+ * @param header overrides extraction of column names from lists - all values are treated as data instead.
  * @return A [DataFrame] containing the data from the nested list structure.
  *         Returns an empty [DataFrame] if the input is empty or invalid.
  */
-public fun <T> List<List<T>>.toDataFrame(containsColumns: Boolean = false): AnyFrame =
+@Refine
+@Interpretable("ValuesListsToDataFrame")
+public fun <T> List<List<T>>.toDataFrame(header: List<String>? = null, containsColumns: Boolean = false): AnyFrame =
```
Jolanrensen: this is a breaking change, can we keep the old function and deprecate it?

koperagen: Deprecated and moved from io to api package

koperagen:
@zaleslaw 
> Great, if tiny example, saying, with subtitles will be also added

Updated documentation.
In the process, updated page with headers to be able to refer to individual operations
Before:
image

After:
image

docs/StardustDocs/topics/createDataFrame.md
```diff
@@ -160,7 +161,30 @@ val df = files.toDataFrame(columnName = "data")
 <inline-frame src="resources/org.jetbrains.kotlinx.dataframe.samples.api.Create.toDataFrameColumn.html" width="100%"/>
 <!---END-->
 
-Creates a [`DataFrame`](DataFrame.md) from an [`Iterable`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-iterable/) of objects:
+#### [`DataFrame`](DataFrame.md) from `List<List<T>>`:
+
+This is useful for parsing text files. For example, the `.srt` subtitle format can be parsed like this:
```
Jolanrensen, line 166: cool! :D

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/common.kt
```diff
@@ -45,48 +45,12 @@ public fun catchHttpResponse(url: URL, body: (InputStream) -> AnyFrame): AnyFram
...
- * @param containsColumns If `true`, treats each nested list as a column with its first element as the column name.
- *                        Otherwise, the first list is treated as the header.
- *                        Defaults to `false`.
- * @return A [DataFrame] containing the data from the nested list structure.
- *         Returns an empty [DataFrame] if the input is empty or invalid.
- */
+@Deprecated(
+    LISTS_TO_DATAFRAME_MIGRATION,
+    ReplaceWith("this.toDataFrame(header = null, containsColumns)", "org.jetbrains.kotlinx.dataframe.api.toDataFrame"),
+)
```
Jolanrensen, line 52: you don't need to, but I usually add an explicit DeprecationLevel.WARNING. It makes it easier to change them all to ERROR in the next cycle

Issue #1266, opened 2025-06-20
Allow `DbType` to do type conversions


Jolanrensen:
Currently, you can only specify how the JDBC types are mapped to a `KType`/`ColumnSchema`.
This must strictly follow the types of the specific implementation of the DataBase JDBC integration to prevent runtime errors, often resulting in very non-Kotlinlike types, such as `java.sql.Date`, `DuckDBBlobResult`, or `org.duckdb.JsonNode`.

Other DataFrame integrations aim to convert the types to something close to kotlin, such as KotlinX LocalDateTime.
Allowing to specify a conversion like this would also allow conversions like SomeJsonFormat -> ColumnGroup/FrameColumn on-the-fly https://github.com/Kotlin/dataframe/issues/462 and `java.sql.Array` -> `List<T>`

Issue #680, opened 2024-04-26
Change the place and resulting object for the methods of `readSql<-->Schema`


zaleslaw:
Motivation:
All these objects are placed on the DataFrame, but return not a DataFrame, but schema, also we are limited with this schema from DataFrame, but it could not keep, for example, the name of the table and other meta-parameters.

Solution:
Move these bunches of methods to the helper class `SqlSchema `and introduce static methods on it, this class could be extendable and also reimplemented for any particular Database

Issue #454, opened 2023-09-11
Improve performance for reading from database


zaleslaw:
There are a few ways to increase the speed of SQL queries and data retrieval

- [x] batches/fetching
- [x] manipulate by whole row instead of reading of each column
- [ ] databases/connection settings
- [ ] create JDBC connection pool with default as [Hikapi](https://github.com/brettwooldridge/HikariCP) for functions with connection under hood

zaleslaw:
Looks like it was closed wrongly

PR #1487, opened 2025-10-09
Refactored DataFrame JDBC API plus DataSource handling

### LINKED ISSUE #1424: Add DataSource support for JDBC integration
May we have such override?
```kt
public fun DataFrame.Companion.readSqlTable(
    dataSource: DataSource,
): AnyFrame {
    dataSource.connection.use { connection ->
        return readSqlTable(connection, ...)
    }
}
```
It looks like DataSource is a preferred API to use.
It looks like the recommended way to create SQL DB connection in production is using HikariCP, at least it's something LLM yieds for me:

```kt
fun dataSource(): HikariDataSource {
    val cfg = HikariConfig().apply {
        jdbcUrl = "jdbc:postgresql://localhost:5432/appdb?sslmode=disable"
        username = "app"
        password = "secret"
        ...
    }
    return HikariDataSource(cfg)
}
```
So, when we have the data source, we have javax.sql.DataSource , which is welcomed i.e. by exposed:
```kt
fun DataSource.listActive(): List<String> {
    Database.connect(this)
    return transaction {
        ...
    }
}
```
It would be nice to be able to use this dataSource in any possible DB wrapper, it will allow [this design](https://youtrack.jetbrains.com/issue/KTNB-1155/Database-integration) to converge.

zaleslaw:
This commit introduces new DataFrame JDBC extension functions with `DataSource` support, removing redundant duplicate utilities. It includes revisions to streamline table/schema reading and validation features, delegating reusable connection-handling logic to a dedicated utility class. Also refactor file structure for better organization of DB-related code.

Closes #1424 #680 #454 #1266

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -6808,9 +6808,13 @@ public final class org/jetbrains/kotlinx/dataframe/schema/ComparisonMode : java/
 }
 
 public abstract interface class org/jetbrains/kotlinx/dataframe/schema/DataFrameSchema {
+	public static final field Companion Lorg/jetbrains/kotlinx/dataframe/schema/DataFrameSchema$Companion;
 	public abstract fun compare (Lorg/jetbrains/kotlinx/dataframe/schema/DataFrameSchema;Lorg/jetbrains/kotlinx/dataframe/schema/ComparisonMode;)Lorg/jetbrains/kotlinx/dataframe/schema/CompareResult;
 	public static synthetic fun compare$default (Lorg/jetbrains/kotlinx/dataframe/schema/DataFrameSchema;Lorg/jetbrains/kotlinx/dataframe/schema/DataFrameSchema;Lorg/jetbrains/kotlinx/dataframe/schema/ComparisonMode;ILjava/lang/Object;)Lorg/jetbrains/kotlinx/dataframe/schema/CompareResult;
 	public abstract fun getColumns ()Ljava/util/Map;
+}
+
+public final class org/jetbrains/kotlinx/dataframe/schema/DataFrameSchema$Companion {
 }
 
 public final class org/jetbrains/kotlinx/dataframe/util/DeprecationMessagesKt {


dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -1,8 +1,8 @@
 package org.jetbrains.kotlinx.dataframe.io.db
 
 import org.jetbrains.kotlinx.dataframe.io.DbConnectionConfig
-import org.jetbrains.kotlinx.dataframe.io.TableColumnMetadata
-import org.jetbrains.kotlinx.dataframe.io.TableMetadata
+import org.jetbrains.kotlinx.dataframe.io.db.TableColumnMetadata
```
koperagen: Jupyter integration automatically imports `org.jetbrains.kotlinx.dataframe.io` package
If you want to move, add import `org.jetbrains.kotlinx.dataframe.io.db` there

zaleslaw: is it import only the direct package without subpackages? in that case, yes, it should be

koperagen: yes, only direct package

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -70,14 +73,70 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
...
+    }
+
+    /**
+     * Configures a [PreparedStatement] for optimal read performance.
+     * This method is called automatically before statement execution.
+     *
+     * @param [statement] the prepared statement to configure.
+     */
+    public open fun configureReadStatement(
+        statement: PreparedStatement
```
Jolanrensen: ktlint

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -52,153 +53,6 @@ private val logger = KotlinLogging.logger {}
...
- * Represents a column in a database table to keep all required meta-information.
- *
- * @property [name] the name of the column.
- * @property [sqlTypeName] the SQL data type of the column.
- * @property [jdbcType] the JDBC data type of the column produced from [java.sql.Types].
- * @property [size] the size of the column.
- * @property [javaClassName] the class name in Java.
- * @property [isNullable] true if column could contain nulls.
- */
-public data class TableColumnMetadata(
```
Jolanrensen, line 86: Do you want to deprecate the old one? or is it fine to move it as the JDBC module is considered "less stable" than :core?

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/TableColumnMetadata.kt
```diff
@@ -0,0 +1,20 @@
...
+ * Represents a column in a database table to keep all required meta-information.
+ *
+ * @property [name] the name of the column.
+ * @property [sqlTypeName] the SQL data type of the column.
+ * @property [jdbcType] the JDBC data type of the column produced from [java.sql.Types].
+ * @property [size] the size of the column.
+ * @property [javaClassName] the class name in Java.
+ * @property [isNullable] true if column could contain nulls.
+ */
+public data class TableColumnMetadata(
```
Jolanrensen: if we're moving this anyway. Could we try avoiding having `data class`es in the public API? They're known to cause breaking APIs when new properties are added. Just have a look at `ParserOptions` to get an example.

Jolanrensen: https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#avoid-using-data-classes-in-your-api

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/DbConnectionConfig.kt
```diff
@@ -0,0 +1,51 @@
...
+ * @property url The JDBC URL of the database, e.g., `"jdbc:postgresql://localhost:5432/mydb"`.
+ *               Must follow the standard format: `jdbc:subprotocol:subname`.
+ *
+ * @property user The username used for authentication.
+ *                Optional, default is an empty string.
+ *
+ * @property password The password used for authentication.
+ *                    Optional, default is an empty string.
+ *
+ * @property readOnly If `true` (default), the library will create the connection in read-only mode.
```
Jolanrensen: `@property` notations should be at the bottom. This messes up KDoc rendering

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readDataFrameSchema.kt
```diff
@@ -0,0 +1,437 @@
+package org.jetbrains.kotlinx.dataframe.io
```
Jolanrensen, line 1: Almost all KDocs in this file are broken(!!). Move all `@` statements to the bottom of the KDocs. If you put them half-way, the text you write below is considered part of the last `@`.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readDataFrameSchema.kt
```diff
@@ -0,0 +1,437 @@
...
+/**
+ * Retrieves the schema for an SQL table using the provided [DataSource].
+ *
+ * @param [dataSource] the [DataSource] to get a database connection from.
+ * @param [tableName] the name of the SQL table for which to retrieve the schema.
+ * @param [dbType] the type of database, could be a custom object, provided by user, optional, default is `null`,
+ * in that case the [dbType] will be recognized from the [dataSource].
+ * @return the schema of the SQL table as a [DataFrameSchema] object.
+ *
+ * ### Example with HikariCP:
```
Jolanrensen, line 47: if you write this many Kotlin samples inside KDocs, this increases the maintainability cost whenever the API changes. Consider using `@sample` instead

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/validationUtil.kt
```diff
@@ -0,0 +1,233 @@
...
+internal const val TABLE_NAME_VALID_PATTERN = "^[\\p{L}\\p{N}_]+(\\.[\\p{L}\\p{N}_]+)*$"
+
+internal fun isSqlQuery(sqlQueryOrTableName: String): Boolean {
+    val queryPattern = Regex("(?i)\\b(SELECT)\\b")
+    return queryPattern.containsMatchIn(sqlQueryOrTableName.trim())
+}
+
+internal fun isSqlTableName(sqlQueryOrTableName: String): Boolean {
+    // Match table names with optional schema and catalog (e.g., catalog.schema.table)
+    val tableNamePattern = Regex("^[a-zA-Z_][a-zA-Z0-9_]*(\\.[a-zA-Z_][a-zA-Z0-9_]*){0,2}$")
```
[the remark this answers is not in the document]
Jolanrensen: can be constant indeed

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/validationUtil.kt
```diff
@@ -0,0 +1,233 @@
...
+        }
+        return false
+    }
+
+    // Validate against forbidden patterns
+    if (hasForbiddenPatterns(normalizedSqlQuery)) {
+        return false
+    }
+
+    // Check if there are balanced quotes (single and double)
```
Jolanrensen, line 202: Can quotes be escaped in SQL? if so, we shouldn't count them when escaped

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readDataFrameSchema.kt
```diff
@@ -0,0 +1,437 @@
...
+ * ### Default Behavior:
+ * If [DbConnectionConfig.readOnly] is `true` (which is the default), the connection will be:
+ * - explicitly set as read-only via `Connection.setReadOnly(true)`
+ * - used with `autoCommit = false`
+ * - automatically rolled back after reading, ensuring no changes to the database
+ *
+ * Even if [DbConnectionConfig.readOnly] is set to `false`, the library still prevents data-modifying queries
+ * and only permits safe `SELECT` operations internally.
+ */
+public fun DataFrameSchema.Companion.getSchemaForSqlTable(
```
Jolanrensen: It's good these are now functions on `DataFrameSchema` instead of `DataFrame` :)
However, we now write "Schema" twice. 
I'd recommend renaming these to `DataFrameSchema.getForSqlTable()`

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readDataFrameSchema.kt
zaleslaw: Or, thanks

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/DbConnectionConfig.kt
zaleslaw: Ok, thanks

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readDataFrameSchema.kt
zaleslaw: Need to think, probably the naming should be changed

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readDataFrameSchema.kt
zaleslaw: I have no time and infrastructure to make it better with Database (if not using H2 for that in tests), but it valid concern - what's better to ship with example or without?

zaleslaw: Worth to create a separate ticket for that

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/TableColumnMetadata.kt
zaleslaw: Great addition, I never heard about that, thanks

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
zaleslaw: less stable, so API will significantly changed, I will reflect in a separate notebook/post

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
zaleslaw: true, will be done on the latest iteration

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readDataFrameSchema.kt
Jolanrensen: probably with example

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/validationUtil.kt
Jolanrensen: also, what about backticks?

Jolanrensen: https://learnsql.com/cookbook/how-to-escape-single-quotes-in-sql/ According to here, single quotes are usually escaped by writing them twice,
so `INSERT INTO customer (id, customer_name) VALUES (502, 'Lay''s');`. In this case your check works.
However, on MySQL and PostgreSQL, you can write `INSERT INTO customer (id, customer_name)VALUES (502, 'Lay\'s');` in which case this check will trigger a false negative. There are 3 single quotes but the statement is still valid.

Finally, this is how Oracle does it: `INSERT INTO customer (id, customer_name) VALUES (502, q'[Lay's]');` in which case your check will also trigger a false negative.

Similarly, I believe alternating is allowed, so `' hi " there '` is valid, but again a false negative is triggered because there's no even number of double quotes.

I'd remove this check altogether and find a more general solution. There's too many edge cases

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
Jolanrensen: okay, but then we won't spot failing notebook tests on TeamCity if there are any

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -267,19 +140,33 @@ public fun DataFrame.Companion.readSqlTable(
     val url = connection.metaData.url
```
Jolanrensen: unused url

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -645,6 +492,8 @@ private fun isValidTableName(tableName: String): Boolean {
  * @param [dbType] the type of database that the [ResultSet] belongs to.
  * @param [limit] the maximum number of rows to read from the [ResultSet][java.sql.ResultSet].
  * @param [inferNullability] indicates how the column nullability should be inferred.
+ * @param [configureStatement] optional lambda to configure the [PreparedStatement] before execution.
```
Jolanrensen: parameter does not exist here

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -756,6 +605,8 @@ public fun ResultSet.readDataFrame(
  * @param [inferNullability] indicates how the column nullability should be inferred.
  * @param [dbType] the type of database, could be a custom object, provided by user, optional, default is `null`,
  * in that case the [dbType] will be recognized from the [dbConfig].
+ * @param [configureStatement] optional lambda to configure the [PreparedStatement] before execution.
```
Jolanrensen, line 686: same here

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -782,296 +633,130 @@ public fun DataFrame.Companion.readAllSqlTables(
...
- * @param [connection] the database connection to read tables from.
- * @param [limit] the maximum number of rows to read from each table.
+ * @param [dataSource] the [DataSource] to get a database connection from.
  * @param [catalogue] a name of the catalog from which tables will be retrieved. A null value retrieves tables from all catalogs.
+ * @param [limit] the maximum number of rows to read from each table.
  * @param [inferNullability] indicates how the column nullability should be inferred.
  * @param [dbType] the type of database, could be a custom object, provided by user, optional, default is `null`,
- * in that case the [dbType] will be recognized from the [connection].
+ * in that case the [dbType] will be recognized from the [dataSource].
+ * @param [configureStatement] optional lambda to configure the [PreparedStatement] before execution.
```
Jolanrensen: and here, also kdocs are still ordered wrong

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -70,14 +110,72 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
...
+        statement.fetchDirection = ResultSet.FETCH_FORWARD
+    }
+
+    /**
+     * Quotes an identifier (table or column name) according to database-specific rules.
+     *
+     * Examples:
+     * - PostgreSQL: "tableName" or "schema"."table"
+     * - MySQL: `tableName` or `schema`.`table`
+     * - MS SQL: [tableName] or [schema].[table]
```
Jolanrensen: escape `[]`,

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -96,4 +194,175 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
...
+        }
+    }
+
+    /**
+     * Creates a mapping between common SQL types and their corresponding KTypes.
+     *
+     * @param tableColumnMetadata The metadata of the table column.
+     * @return The KType associated with the SQL type or a default type if no mapping is found.
+     */
+    internal fun makeCommonSqlToKTypeMapping(tableColumnMetadata: TableColumnMetadata): KType {
```
Jolanrensen: this cannot be overridden?

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -96,4 +194,175 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
...
+            } */
+            kType.classifier == Array::class -> {
+                handleArrayValues(values.toMutableList())
+            }
+
+            else -> values
+        }
+    }
+
+    internal fun handleArrayValues(values: MutableList<Any?>): List<Any> {
```
Jolanrensen: could be `private`, or `protected`

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -1229,53 +879,93 @@ private fun fetchAndConvertDataFromResultSet(
...
+            dbType = dbType,
+            inferNullability = inferNullability
         )
-    }
+    }.toDataFrame()
+
+/**
+ * Builds a single DataColumn with proper type handling.
+ */
+private fun buildDataColumn(
```
Jolanrensen: this function should likely be a part of DbType. That will make it possible to create different types of columns from the DbType, not just ValueColumns, but also FrameColumns and ColumnGroups

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -96,4 +194,175 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
...
+     * Processes the column values retrieved from the database and performs transformations based on the provided
+     * Kotlin type and column metadata. The method allows for custom post-processing logic, such as handling
+     * specific database column types, including arrays.
+     *
+     * @param values the list of raw values retrieved from the database for the column.
+     * @param kType the Kotlin type that the column values should be transformed to.
+     * @param columnMetadata the metadata of the database column, including details such as SQL type name and size.
+     * @return a list of processed column values, with transformations applied where necessary, or the original list if no transformation is needed.
+     */
+    public open fun postProcessColumnValues(
```
Jolanrensen: could probably be merged with buildDataColumn() if that function would become part of DbType

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -96,4 +194,175 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
...
+     * Extracts a value from the ResultSet for the given column.
+     * This method can be overridden by custom database types to provide specialized parsing logic.
+     *
+     * @param [rs] the ResultSet to read from
+     * @param [columnIndex] zero-based column index
+     * @param [columnMetadata] metadata for the column
+     * @param [kType] the Kotlin type for this column
+     * @return the extracted value, or null
+     */
+    public open fun extractValueFromResultSet(
```
Jolanrensen, line 208: powerful! nice :)

Jolanrensen:
I'm not a fan of the introduction of `MutableList` in `DbType`. In many cases, nothing needs to be modified, in which case returning the immutable list is just as efficient, and if values do need to be mapped, they only need to be read once after which they can be garbage collected, so again not much benefit from making it mutable.

Plus, things can easily get confusing when you mix mutable and immutable lists. Just look at:
```kt
postProcessColumnValues(values: MutableList<Any?>, ...): List<Any?>
```
How `values` will be used is unclear looking at this. Will values inside be overridden? will a new list be returned? will the types change? Very hard to tell.

You could decide to save some memory (which I think will be negligible) by making the list mutable, but then, the function should probably not return anything:
```kt
postProcessColumnValues(values: MutableList<Any?>, ...)
```

However, I'd keep it simple and stick with the immutability of DataFrame:
```kt
postProcessColumnValues(values: List<Any?>, ...): List<Any?>
```

zaleslaw:
@Jolanrensen regarding Mutablity vs Immutability for internals - the initial idea was to follow immutability style for easy-to-mantain approach, but after couple of request and communication, I've decided to optimize some obvious place to avoid additional copying before building a final DataFrame, also I'm planning to provide comment for each such situation to avoid problems in future, the same should be done for the working with metadata

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
Jolanrensen: image
still shows this in the IDE
image
but it renders correctly, so I'm happy :)

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -96,4 +194,213 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
...
+        columnIndex: Int,
+        columnMetadata: TableColumnMetadata,
+        kType: KType,
+    ): Any? =
+        try {
+            rs.getObject(columnIndex + 1)
+            // TODO: add a special handler for Blob via Streams
+        } catch (_: Throwable) {
+            // TODO: expand for all the types like in generateKType function
+            if (kType.isSupertypeOf(String::class.starProjectedType)) {
```
Jolanrensen, line 219: try to avoid `::class`, it's way heavier than `typeOf<String>()`. Also, don't you mean `subType`? instead of `superType`?

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -96,4 +194,213 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
...
+     *
+     * @param values the list of raw values retrieved from the database for the column.
+     * @param kType the Kotlin type that the column values should be transformed to.
+     * @param columnMetadata the metadata of the database column, including details such as SQL type name and size.
+     * @return a list of processed column values, with transformations applied where necessary, or the original list if no transformation is needed.
+     */
+    private fun postProcessColumnValues(
+        values: MutableList<Any?>,
+        kType: KType,
+        columnMetadata: TableColumnMetadata,
```
Jolanrensen: unused

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -96,4 +194,213 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
...
+     * Processes the column values retrieved from the database and performs transformations based on the provided
+     * Kotlin type and column metadata. The method allows for custom post-processing logic, such as handling
+     * specific database column types, including arrays.
+     *
+     * @param values the list of raw values retrieved from the database for the column.
+     * @param kType the Kotlin type that the column values should be transformed to.
+     * @param columnMetadata the metadata of the database column, including details such as SQL type name and size.
+     * @return a list of processed column values, with transformations applied where necessary, or the original list if no transformation is needed.
+     */
+    private fun postProcessColumnValues(
```
Jolanrensen: do you want DbType implementation to be able to access this function or handleArrayValues? If so, it should be `protected`, else `private` is okay.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -96,4 +194,213 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
...
+
+        return DataColumn.createValueColumn(
+            name = name,
+            values = correctedValues,
+            infer = convertNullabilityInference(inferNullability),
+            type = kType,
+        )
+    }
+
+    private fun convertNullabilityInference(inferNullability: Boolean) =
```
Jolanrensen, line 246: seems a bit too specific to be a function, don't you think? It's also just used but once

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DuckDb.kt
```diff
@@ -41,10 +41,7 @@ import org.duckdb.DuckDBResultSetMetaData
 import org.duckdb.JsonNode
 import org.jetbrains.kotlinx.dataframe.DataFrame
 import org.jetbrains.kotlinx.dataframe.io.DbConnectionConfig
```
Jolanrensen, line 43: seems the refactoring broke a lot of KDoc references in this file. But we can also fix that once we redo DuckDb integration with new better types :D

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/TableColumnMetadata.kt
```diff
@@ -0,0 +1,48 @@
...
+ * @property [isNullable] true if column could contain nulls.
+ */
+public class TableColumnMetadata(
+    public val name: String,
+    public val sqlTypeName: String,
+    public val jdbcType: Int,
+    public val size: Int,
+    public val javaClassName: String,
+    public val isNullable: Boolean = false,
+) {
```
Jolanrensen, line 20: great! But could you also add the `copy()` function from data classes? It's probably fine to go without `componentX()`, but `copy()`might still be useful if you want to reuse but only slightly change the class

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/TableMetadata.kt
```diff
@@ -0,0 +1,44 @@
...
+ * @property [name] the name of the table.
+ * @property [schemaName] the name of the schema the table belongs to (optional).
+ * @property [catalogue] the name of the catalogue the table belongs to (optional).
+ */
+public class TableMetadata {
+    public val name: String
+    public val schemaName: String?
+    public val catalogue: String?
+
+    public constructor(name: String, schemaName: String?, catalogue: String?) {
```
Jolanrensen: why isn't this the primary constructor?

Jolanrensen: same story with copy() :)

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/DbConnectionConfig.kt
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * Creates a copy of this configuration with the option to override specific properties.
+     *
+     * @param url The JDBC URL. If not specified, uses the current value.
+     * @param user The username. If not specified, uses the current value.
+     * @param password The password. If not specified, uses the current value.
+     * @param readOnly The read-only flag. If not specified, uses the current value.
+     * @return A new [DbConnectionConfig] instance with the specified changes.
+     */
+    public fun copy(
```
Jolanrensen, line 87: here you did add copy, great!

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/DbConnectionConfig.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    override fun hashCode(): Int {
+        var result = url.hashCode()
+        result = 31 * result + user.hashCode()
+        result = 31 * result + password.hashCode()
+        result = 31 * result + readOnly.hashCode()
+        return result
+    }
+
+    override fun toString(): String = "DbConnectionConfig(url='$url', user='$user', password='***', readOnly=$readOnly)"
```
Jolanrensen, line 76: appreciate the password hiding! <3

Jolanrensen (approved):

Issue #1255, opened 2025-06-16
Moving/reordering inside groups near impossible


Jolanrensen:
In addition to https://github.com/Kotlin/dataframe/issues/1021 it's very difficuly to move/reorder columns inside a column group.

Let's say we have:
![Image](https://github.com/user-attachments/assets/9120baeb-d57b-4511-870e-885ac27b171f)

`move to(Start)` moves the column outside the group. This could do with a "keepInsideGroup" argument:

![Image](https://github.com/user-attachments/assets/4f16d017-9299-4211-81ad-556a80c9a2c2)
![Image](https://github.com/user-attachments/assets/c47156f6-4998-45da-8361-c3f5c416b4ca)

We have `move after`, which works, but is a bit weird to use if you actually need a `move before` (which we don't have [yet]( https://github.com/Kotlin/dataframe/issues/1021)):

![Image](https://github.com/user-attachments/assets/82a4870d-3ea9-406c-81e1-02605db18563)

"regrouping" doesn't work in one step; the order specified is not taken into account by selecting:
![Image](https://github.com/user-attachments/assets/5544a75f-7702-4e6b-840e-527eaf776535)

you would need to ungroup first, then regroup, requiring two steps...

Reordering does seem to work, but it's weird too because you need to "reorder by" something, so like the order in a list or a number:

![Image](https://github.com/user-attachments/assets/1d992ea1-3276-4c00-9b20-47d606310933)
![Image](https://github.com/user-attachments/assets/8d73290e-9de5-4bc8-a34c-e7f4779154f5)

koperagen:
IMO this ticket describes too different things. I'd keep only `move to(Start) moves the column outside the group. This could do with a "keepInsideGroup" argument:` proposal - it will be easier to decide

koperagen:
> This could do with a "keepInsideGroup" argument

Suggest shorter version:
`df.move { group.col }.toStart(insideGroup = true)`
`df.move { group.col }.to(0, insideGroup = true)`

On a second thought,  not all selected columns can be from the same group :c

Jolanrensen:
> IMO this ticket describes too different things. I'd keep only `move to(Start) moves the column outside the group. This could do with a "keepInsideGroup" argument:` proposal - it will be easier to decide

it describes one problem, namely changing the order of columns inside a group is hard. Wether that can be solved with `reorder` or `move` is up to how we want to implement it.


> > This could do with a "keepInsideGroup" argument
> 
[...]

PR #1489, opened 2025-10-10
Move to(Start) keeping inside group

CarloMariaProietti:
About #1255
It comes with docs and a test case

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -3424,10 +3424,14 @@ public final class org/jetbrains/kotlinx/dataframe/api/MoveKt {
 	public static final fun move (Lorg/jetbrains/kotlinx/dataframe/DataFrame;[Lkotlin/reflect/KProperty;)Lorg/jetbrains/kotlinx/dataframe/api/MoveClause;
 	public static final fun move (Lorg/jetbrains/kotlinx/dataframe/DataFrame;[Lorg/jetbrains/kotlinx/dataframe/columns/ColumnReference;)Lorg/jetbrains/kotlinx/dataframe/api/MoveClause;
 	public static final fun moveTo (Lorg/jetbrains/kotlinx/dataframe/DataFrame;ILkotlin/jvm/functions/Function2;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
+	public static final fun moveTo (Lorg/jetbrains/kotlinx/dataframe/DataFrame;IZLkotlin/jvm/functions/Function2;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
+	public static final fun moveTo (Lorg/jetbrains/kotlinx/dataframe/DataFrame;IZ[Ljava/lang/String;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun moveTo (Lorg/jetbrains/kotlinx/dataframe/DataFrame;I[Ljava/lang/String;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun moveTo (Lorg/jetbrains/kotlinx/dataframe/DataFrame;I[Lkotlin/reflect/KProperty;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun moveTo (Lorg/jetbrains/kotlinx/dataframe/DataFrame;I[Lorg/jetbrains/kotlinx/dataframe/columns/ColumnReference;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun moveToEnd (Lorg/jetbrains/kotlinx/dataframe/DataFrame;Lkotlin/jvm/functions/Function2;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
+	public static final fun moveToEnd (Lorg/jetbrains/kotlinx/dataframe/DataFrame;ZLkotlin/jvm/functions/Function2;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
+	public static final fun moveToEnd (Lorg/jetbrains/kotlinx/dataframe/DataFrame;Z[Ljava/lang/String;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun moveToEnd (Lorg/jetbrains/kotlinx/dataframe/DataFrame;[Ljava/lang/String;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun moveToEnd (Lorg/jetbrains/kotlinx/dataframe/DataFrame;[Lkotlin/reflect/KProperty;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun moveToEnd (Lorg/jetbrains/kotlinx/dataframe/DataFrame;[Lorg/jetbrains/kotlinx/dataframe/columns/ColumnReference;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
@@ -3440,14 +3444,19 @@ public final class org/jetbrains/kotlinx/dataframe/api/MoveKt {
 	public static final fun moveToRight (Lorg/jetbrains/kotlinx/dataframe/DataFrame;[Lkotlin/reflect/KProperty;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun moveToRight (Lorg/jetbrains/kotlinx/dataframe/DataFrame;[Lorg/jetbrains/kotlinx/dataframe/columns/ColumnReference;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun moveToStart (Lorg/jetbrains/kotlinx/dataframe/DataFrame;Lkotlin/jvm/functions/Function2;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
+	public static final fun moveToStart (Lorg/jetbrains/kotlinx/dataframe/DataFrame;ZLkotlin/jvm/functions/Function2;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
+	public static final fun moveToStart (Lorg/jetbrains/kotlinx/dataframe/DataFrame;Z[Ljava/lang/String;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun moveToStart (Lorg/jetbrains/kotlinx/dataframe/DataFrame;[Ljava/lang/String;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun moveToStart (Lorg/jetbrains/kotlinx/dataframe/DataFrame;[Lkotlin/reflect/KProperty;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun moveToStart (Lorg/jetbrains/kotlinx/dataframe/DataFrame;[Lorg/jetbrains/kotlinx/dataframe/columns/ColumnReference;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun to (Lorg/jetbrains/kotlinx/dataframe/api/MoveClause;I)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
+	public static final fun to (Lorg/jetbrains/kotlinx/dataframe/api/MoveClause;IZ)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun toEnd (Lorg/jetbrains/kotlinx/dataframe/api/MoveClause;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
+	public static final fun toEnd (Lorg/jetbrains/kotlinx/dataframe/api/MoveClause;Z)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun toLeft (Lorg/jetbrains/kotlinx/dataframe/api/MoveClause;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun toRight (Lorg/jetbrains/kotlinx/dataframe/api/MoveClause;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun toStart (Lorg/jetbrains/kotlinx/dataframe/api/MoveClause;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
+	public static final fun toStart (Lorg/jetbrains/kotlinx/dataframe/api/MoveClause;Z)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun toTop (Lorg/jetbrains/kotlinx/dataframe/api/MoveClause;Lkotlin/jvm/functions/Function2;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static synthetic fun toTop$default (Lorg/jetbrains/kotlinx/dataframe/api/MoveClause;Lkotlin/jvm/functions/Function2;ILjava/lang/Object;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
 	public static final fun under (Lorg/jetbrains/kotlinx/dataframe/api/MoveClause;Ljava/lang/String;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
@@ -5589,6 +5598,11 @@ public final class org/jetbrains/kotlinx/dataframe/exceptions/ColumnNotFoundExce
 public final class org/jetbrains/kotlinx/dataframe/exceptions/ColumnTypeMismatchesColumnValuesException : java/lang/RuntimeException {
 	public fun <init> (Lorg/jetbrains/kotlinx/dataframe/DataColumn;Ljava/lang/Throwable;)V
 	public final fun getColumn ()Lorg/jetbrains/kotlinx/dataframe/DataColumn;
+}
+
+public final class org/jetbrains/kotlinx/dataframe/exceptions/ColumnsWithDifferentParentException : java/lang/IllegalArgumentException, org/jetbrains/kotlinx/dataframe/exceptions/DataFrameError {
+	public fun <init> (Ljava/lang/String;)V
+	public fun getMessage ()Ljava/lang/String;
 }
 
 public abstract interface class org/jetbrains/kotlinx/dataframe/exceptions/DataFrameError {


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
```diff
@@ -537,6 +537,27 @@ public fun <T, C> MoveClause<T, C>.under(
 @Interpretable("MoveTo")
 public fun <T, C> MoveClause<T, C>.to(columnIndex: Int): DataFrame<T> = moveTo(columnIndex)
 
+/**
+ * Moves columns, previously selected with [move] to a new position specified
+ * by [columnIndex] within the column group (remaining inside the group).
```
Jolanrensen: *if `insideGroup == true` that is

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
```diff
@@ -537,6 +537,27 @@ public fun <T, C> MoveClause<T, C>.under(
...
+ *
+ * Returns a new [DataFrame] with updated columns structure.
+ *
+ * For more information: {@include [DocumentationUrls.Move]}
+ *
+ * ### Examples:
+ * ```kotlin
+ * df.move { age and weight }.to(0)
+ * df.move("age", "weight").to(2)
```
Jolanrensen: how about an example with insideGroup = true?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
```diff
@@ -537,6 +537,27 @@ public fun <T, C> MoveClause<T, C>.under(
...
+ * ```
+ *
+ * @param [columnIndex] The index specifying the position in the [ColumnGroup] columns
+ *  * where the selected columns will be moved.
```
Jolanrensen: floating "*"

Jolanrensen: no information about `@param insideGroup`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
```diff
@@ -691,6 +712,24 @@ public fun <T, C> MoveClause<T, C>.toLeft(): DataFrame<T> = to(0)
 @Interpretable("MoveToStart0")
 public fun <T, C> MoveClause<T, C>.toStart(): DataFrame<T> = to(0)
 
+/**
+ * Moves columns, previously selected with [move] to the start of their [ColumnGroup] (remaining inside the group).
```
Jolanrensen: only if insideGroup = true

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
```diff
@@ -691,6 +712,24 @@ public fun <T, C> MoveClause<T, C>.toLeft(): DataFrame<T> = to(0)
...
+ *
+ * For more information: {@include [DocumentationUrls.Move]}
+ *
+ * ### Examples:
+ * ```kotlin
+ * df.move { age and weight }.toStart()
+ * df.move { colsOf<String>() }.toStart()
+ * df.move("age", "weight").toStart()
+ * ```
+ */
```
Jolanrensen, line 829: same lack of `insideGroup` in examples and `@param`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
```diff
@@ -691,6 +712,24 @@ public fun <T, C> MoveClause<T, C>.toLeft(): DataFrame<T> = to(0)
...
+@Refine
+@Interpretable("MoveToStart0")
```
Jolanrensen, line 831: We will need to add compiler plugin support for the `insideGroup` argument. Currently, we get an empty schema when this is called. We will do it after it's merged

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
```diff
@@ -124,3 +128,65 @@ internal fun <T, C> MoveClause<T, C>.moveTo(columnIndex: Int): DataFrame<T> {
             }
     return newColumnList.toDataFrame().cast()
 }
+
+internal fun <T, C> MoveClause<T, C>.moveTo(columnIndex: Int, insideGroup: Boolean): DataFrame<T> {
```
Jolanrensen: It's fine to merge `moveTo(columnIndex: Int)` and `moveTo(columnIndex: Int, insideGroup: Boolean)`. They are `internal` after all :) You could name it `moveToImpl` if you like, to adhere to our internal implementation functions convention

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
```diff
@@ -124,3 +128,65 @@ internal fun <T, C> MoveClause<T, C>.moveTo(columnIndex: Int): DataFrame<T> {
...
+    val columnsToMoveParents = columnsToMove.map { it.path.dropLast() }
+    val parentOfFirst = columnsToMoveParents.first()
+    if (columnsToMoveParents.any { it != parentOfFirst }) {
+        throw IllegalArgumentException(
+            "Cannot move columns with different parent to an index",
+        )
+    }
+
+    // if columns will be moved to top level or columns to move are at top level
+    if (!insideGroup || parentOfFirst.isEmpty()) {
```
Jolanrensen: if `insideGroup == false`, all code before does not need to run, right?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
```diff
@@ -124,3 +128,65 @@ internal fun <T, C> MoveClause<T, C>.moveTo(columnIndex: Int): DataFrame<T> {
...
+        return moveTo(columnIndex)
+    }
+
+    // columns are nested and will be eventually moved inside their own group
+    // finding reference column
+    val parentPath = df[parentOfFirst].path
+    val referenceAndSiblings = df[parentOfFirst].asColumnGroup().columns()
+    val referenceAndSiblingsPaths = referenceAndSiblings.map { parentPath + it.path }
+    val reference = if (columnIndex >= referenceAndSiblingsPaths.size){
```
Jolanrensen: linting

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
```diff
@@ -220,4 +220,87 @@ class MoveTests {
             grouped.move { "a"["b"] }.before { "a"["b"] }
         }.message shouldBe "Cannot move column 'a/b' before its own child column 'a/b'"
     }
+
```
Jolanrensen, line 223: what happens at top-level with `insideGroup = true`? I believe it should do the same regardless of `insideGroup`. Could you add a test for this?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
Jolanrensen: I'd recommend installing the KtLint IntelliJ plugin. It's auto-configured with the right settings because our project comes with an `.editorconfig` file :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
```diff
@@ -124,3 +128,65 @@ internal fun <T, C> MoveClause<T, C>.moveTo(columnIndex: Int): DataFrame<T> {
...
+        referenceAndSiblingsPaths.last()
+    } else {
+        referenceAndSiblingsPaths.get(columnIndex)
```
Jolanrensen: can be `[]`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
Jolanrensen: Also, I think it would genuinely be easier if `insideGroup = true` to `update` the group column as follows:
- convert the group to a DataFrame
- use the existing `moveTo` logic
- convert it back to a column group with the original name

I think that would be easier than reinventing the moving behavior again for moving inside groups :) Of course, some of the checks (like that they share the same parent) should be here. WDYT? Did you try something like that?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
```diff
@@ -124,3 +128,65 @@ internal fun <T, C> MoveClause<T, C>.moveTo(columnIndex: Int): DataFrame<T> {
...
+            if (columnsToMovePath.contains(it)){
+                if (referenceWasIterated)
+                    columnsAfterReferenceToMove.add(it)
+                else
+                    columnsBeforeReferenceToMove.add(it)
+            }
+        }
+    }
+
+    // move columns
```
Jolanrensen: probably easier to view the logic using a `when`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
CarloMariaProietti: Thank you for reviewing! IMO it is a good idea but i haven't tried yet a similar approach. I thought that exploiting after and before would have been a good idea because they implement the logic of a movement between nested levels, so i thought of moveTo as a particular case of moveAfter/before.

Jolanrensen: In some way that's true, however, `moveTo` is a bit simpler than move after/before and a bit different in some other cases.
For `moveTo` there are two cases now:
- either we move a subsection of columns to a certain index at the top level
    - these columns can be from any depth and of different parents
- or we move a subsection of columns to a certain index within the same group/level
    - these columns need to have the same parent

for `move after/before` there's only one case, and it's more complicated than either of those above:
- we move a subsection of columns from any depth and of potentially different parents into another place with a potentially different parent.

(btw, did we test the case `move { cols of different parents }.before { some nested col }` as well in the last PR?)

So while it would be possible to reuse move after/before for move to, it may be more efficient to avoid it.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
Jolanrensen: and lift `return` out of the when-statement

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
```diff
@@ -537,6 +538,32 @@ public fun <T, C> MoveClause<T, C>.under(
...
+ * ```kotlin
+ * df.move { age and weight }.to(0, true)
+ * df.move("age", "weight").to(2, false)
+ * ```
+ *
+ * @param [columnIndex] The index specifying the position in the [ColumnGroup] columns
+ * where the selected columns will be moved.
+ *
+ * @param [insideGroup] If true, selected columns will be moved remaining inside their group,
+ * else they will be moved on top level.
```
Jolanrensen: *to the top level

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
```diff
@@ -537,6 +538,32 @@ public fun <T, C> MoveClause<T, C>.under(
 @Interpretable("MoveTo")
 public fun <T, C> MoveClause<T, C>.to(columnIndex: Int): DataFrame<T> = moveTo(columnIndex)
 
+/**
+ * Moves columns, previously selected with [move] to a new position specified
+ * by [columnIndex]. If insideGroup is true, selected columns will be moved remaining within their [ColumnGroup],
```
Jolanrensen: "insideGroup" can have "[]"

Jolanrensen:
Oh, I see we also have these shortcut functions: 
- `public fun <T> DataFrame<T>.moveTo(newColumnIndex: Int, columns: ColumnsSelector<T, *>)` 
- `public fun <T> DataFrame<T>.moveTo(newColumnIndex: Int, vararg columns: String): DataFrame<T>`
- moveToStart(), moveToEnd()

Could you do the same there as with `.to()`? :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
CarloMariaProietti: > (btw, did we test the case `move { cols of different parents }.before { some nested col }` as well in the last PR?)
Yes, it is tested in core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt.

However, I understand that using moveTo instead of moveAfter/before is for the best. I'll try to follow your advice and implement it.

CarloMariaProietti:
Good Morning, I implemented a new logic as suggested in https://github.com/Kotlin/dataframe/pull/1489#discussion_r2426091557.
Each test works, I'll proceed to add shortcut functions as requested in
https://github.com/Kotlin/dataframe/pull/1489#issuecomment-3402042881.

CarloMariaProietti:
about https://github.com/Kotlin/dataframe/pull/1489#issuecomment-3402042881 ,

Now there are both toStart() and toEnd(), 



> Oh, I see we also have these shortcut functions:
> 
> * `public fun <T> DataFrame<T>.moveTo(newColumnIndex: Int, columns: ColumnsSelector<T, *>)`
> * `public fun <T> DataFrame<T>.moveTo(newColumnIndex: Int, vararg columns: String): DataFrame<T>`

.to() implements yet these shortcuts because it is possible to create a MoveClause with both ColumnsSelector<T, *>
 and vararg  String

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
```diff
@@ -691,6 +718,26 @@ public fun <T, C> MoveClause<T, C>.toLeft(): DataFrame<T> = to(0)
...
+ * ```kotlin
+ * df.move { age and weight }.toStart(true)
+ * df.move { colsOf<String>() }.toStart(true)
+ * df.move("age", "weight").toStart(false)
+ * ```
+ *
+ * @param [insideGroup] If true, selected columns will be moved to the start remaining inside their group,
+ * else they will be moved to the start on top level.
+ */
+public fun <T, C> MoveClause<T, C>.toStart(insideGroup: Boolean): DataFrame<T> = to(0, insideGroup)
```
Jolanrensen, line 832: oh, please keep
```kt
@Refine
@Interpretable("MoveToStart0")
```
I will add compiler plugin for this variant of the function. It can use the same interpreter as `toStart()`, but it needs to be merged first and published as dev version so we can test it works in the compiler plugin in the Kotlin repo.

Jolanrensen:
@CarloMariaProietti I'm not sure what you mean, 
We can write
```kt
public fun <T> DataFrame<T>.moveTo(newColumnIndex: Int, columns: ColumnsSelector<T, *>): DataFrame<T> =
    move(columns).to(newColumnIndex)

public fun <T> DataFrame<T>.moveTo(newColumnIndex: Int, insideGroup: Boolean, columns: ColumnsSelector<T, *>): DataFrame<T> =
    move(columns).to(newColumnIndex, insideGroup)
```
and
```kt
public fun <T> DataFrame<T>.moveTo(newColumnIndex: Int, vararg columns: String): DataFrame<T> =
    moveTo(newColumnIndex) { columns.toColumnSet() }

public fun <T> DataFrame<T>.moveTo(newColumnIndex: Int, insideGroup: Boolean, vararg columns: String): DataFrame<T> =
    moveTo(newColumnIndex, insideGroup) { columns.toColumnSet() }
```
right?

If you want, we could also add these in another PR of course.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
```diff
@@ -124,3 +130,44 @@ internal fun <T, C> MoveClause<T, C>.moveTo(columnIndex: Int): DataFrame<T> {
...
+    }
+
+    val columnsToMove = df.getColumns(columns)
+
+    // check if columns to move have the same parent
+    val columnsToMoveParents = columnsToMove.map { it.path.dropLast() }
+    val parentOfFirst = columnsToMoveParents.first()
+    if (columnsToMoveParents.any { it != parentOfFirst }) {
+        throw IllegalArgumentException(
+            "Cannot move columns with different parent to an index",
```
Jolanrensen: please specify "insideGroup", as it would be fine to move them with `insideGroup = false`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
```diff
@@ -124,3 +130,44 @@ internal fun <T, C> MoveClause<T, C>.moveTo(columnIndex: Int): DataFrame<T> {
...
+    }
+
+    // logic: remove columns to move and their siblings (from this point, sons), apply them moveTo, reinsert them
+    val parentPath = df[parentOfFirst].path
+    val sons = df[parentOfFirst].asColumnGroup()
+    // remove sons
+    val sonsWithFullPaths = sons.columns().map { parentPath + it.path }
+    val intermediateDf = df.removeImpl { sonsWithFullPaths.toColumnSet() }
+    // move sons and reinsert them
+    val columnsToMoveWithReducedPath = columnsToMove.map { it.path.last(it.path.size - parentPath.size).toPath() }
```
Jolanrensen: if all columns here have the same parent... isn't their "reduced path" simply their name?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
```diff
@@ -124,3 +130,44 @@ internal fun <T, C> MoveClause<T, C>.moveTo(columnIndex: Int): DataFrame<T> {
...
+            "Cannot move columns with different parent to an index",
+        )
+    }
+
+    // if columns will be moved to top level or columns to move are at top level
+    if (parentOfFirst.isEmpty()) {
+        return moveTo(columnIndex)
+    }
+
+    // logic: remove columns to move and their siblings (from this point, sons), apply them moveTo, reinsert them
```
Jolanrensen: It's great you made it work! But I have a feeling this is way more complicated than it needs to be.

Jolanrensen: My feeling was right, I managed to pass all your tests with this logic:
```kt
val columnsToMoveNames = columnsToMove.map { it.name() }
return df.replace { parentOfFirst.asColumnGroup() }.with {
    it.asDataFrame()
        .move { columnsToMoveNames.toColumnSet() }.to(columnIndex)
        .asColumnGroup(it.name())
}
```

Let me explain what's done here:
- we take the column names of those we want to move. This is enough, because we already checked they have the same parent.
- since we now only need to move columns around on the same level, this is just as easy as `moveTo()` on the top level. We can use `replace { parent }` to get access to just the parent group
- then, treating the column group as a DataFrame, we can simply call `move {}.to()`
- and convert it back to a column group with the right name

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
```diff
@@ -124,3 +130,44 @@ internal fun <T, C> MoveClause<T, C>.moveTo(columnIndex: Int): DataFrame<T> {
...
+        return moveTo(columnIndex)
+    }
+
+    val columnsToMove = df.getColumns(columns)
+
+    // check if columns to move have the same parent
+    val columnsToMoveParents = columnsToMove.map { it.path.dropLast() }
+    val parentOfFirst = columnsToMoveParents.first()
+    if (columnsToMoveParents.any { it != parentOfFirst }) {
+        throw IllegalArgumentException(
```
Jolanrensen: Suggestion from my colleague: Can you create your own exception that implements `DataFrameError`? that way we can catch it in the compiler plugin later :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
CarloMariaProietti: Thank you for the explanation, your solution is much more smooth. I ignored replace { } so i had to remove the level and reinsert it... If you agree, I'll paste this in the code.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/move.kt
CarloMariaProietti: Done!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
```diff
@@ -264,6 +312,17 @@ public fun <T> DataFrame<T>.moveToLeft(columns: ColumnsSelector<T, *>): DataFram
...
+/**
+ * @include [CommonMoveToStartDocs]
+ * @include [SelectingColumns.Dsl.WithExample] {@include [SetMoveToStartOperationArg]}
+ * @param [columns\] The [Columns Selector][ColumnsSelector] used to select the columns of this [DataFrame] to move.
+ * @param [insideGroup] If true, selected columns will be moved to the start remaining inside their group,
+ * else they will be moved to the start of the top level.
+ */
+@Refine
+@Interpretable("MoveToStart1")
+public fun <T> DataFrame<T>.moveToStart(columns: ColumnsSelector<T, *>, insideGroup: Boolean): DataFrame<T> = move(columns).toStart(insideGroup)
```
Jolanrensen: make sure `columns` is the last argument, else the lambda outside parentheses doesn't work. For consistency, you can do the same in the vararg String overloads

Jolanrensen: that way people can write:
```kt
df.moveToStart { cols }
// and
df.moveToStart(insideGroup = true) { cols }
```

CarloMariaProietti: Done, I fixed also moveToEnd

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
```diff
@@ -219,6 +220,53 @@ public fun <T> DataFrame<T>.moveTo(newColumnIndex: Int, vararg columns: AnyColum
 public fun <T> DataFrame<T>.moveTo(newColumnIndex: Int, vararg columns: KProperty<*>): DataFrame<T> =
     moveTo(newColumnIndex) { columns.toColumnSet() }
 
+/**
+ * Moves the specified [columns\] to a new position specified
+ * by [columnIndex]. If [insideGroup] is true selected columns
```
Jolanrensen: _Cannot resolve symbol 'columnIndex'_, if you use `[]`, you must reference the actual name of the parameter, which in this case is "newColumnIndex"

Jolanrensen: (yes, this is a mistake on our end, but unfortunately, we cannot rename it to "columnIndex" without breaking the API)

CarloMariaProietti:
> Sorry for the late response, I was on holiday :) Looks great otherwise! Thanks for handling all overloads! I'll run the CI street

It's fine, hope you had a good holiday!

Jolanrensen:
As usual, could you commit the changes after `apiDump` and `ktLintFormat`? it fails on those :)

PR #1499, opened 2025-10-14
added annotations for moveTo() to add compiler plugin support

Jolanrensen:


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/move.kt
```diff
@@ -398,6 +400,8 @@ public fun <T> DataFrame<T>.moveToEnd(vararg columns: KProperty<*>): DataFrame<T
  * @param [column] The [Column With Path Selector][ColumnsSelector] used to specify
  * a path in the [DataFrame] to move columns.
  */
+@Refine
+@Interpretable("MoveInto1")
```
koperagen: What use cases this would support?

Jolanrensen: Ah I see the issue now... I assumed:

`move { col }.into { pathOf("some", "new", "path") }`, assuming the CS DSL in the compiler plugin supports `pathOf()` and `"path"["of"]` :) 

And `move { col }.into { someColGroup["newName"] }`

but it also supports multiple cols with the new name depending on the column given...
It's like a rename.into + move.under

Jolanrensen: but that's also the problem with `rename.into {}` isn't it? we don't know how the user uses the old name or any other column data to generate the new name...

koperagen: as long as user can write arbitraty code in the lambda such powerful operations are somewhat impossible to implement in the plugin =( shortcuts like rename { }.removePrefix could cover most popular usages though. Didn't think about move into yet

So let's remove these two for now

PR #1506, opened 2025-10-16
Exclude some unused classes to reduce plugin size

koperagen:
Before
image

After
image


Jolanrensen:
I'd probably first try publishing a dev version to see if it still works with the current compiler plugin implementation. Other than that I think excluding these files is fine, as compiler plugin support (for now) only relies on closer-to-`impl` versions of these operations :)

zaleslaw:
Seems like the proposed solution makes us more fragile in terms if we want to change something, especially for statistics. Is it really worth?

Jolanrensen:
@zaleslaw I don't think fragility is impacted. The compiler plugin support for statistics functions doesn't rely on the public API, so we can safely exclude it. If we need it again at some point to support another operation, we can always include them again.

zaleslaw:
Let's see!

PR #1519, opened 2025-10-23
Update createDataFrame.md with compiler-plugin friendly examples

koperagen:
Replaced some examples that don't work with compiler plugin and use outdated API that we plan to replace, with `toDataFrame` operation. 
Now examples with random data can be copy-pasted into project with compiler plugin. Plus, generally it's much more powerful approach that can produce sophisticated dataframes:

image

image

In the process i found that our rendering doesn't work for some of the examples and fixed it.
image

image


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Create.kt
```diff
@@ -222,6 +220,70 @@ class Create : TestBase() {
...
+    @TransformDataFrameExpressions
+    fun createRandomDataFrame() {
+        // stable random + clean examples
+        @Suppress("LocalVariableName")
+        val Random = KotlinRandom(42)
+        fun <T> List<T>.random() = this.random(Random)
+        // SampleStart
+        val categories = listOf("Electronics", "Books", "Clothing")
+        // DataFrame with 4 columns and 7 rows
+        (0 until 7).toDataFrame {
```
Jolanrensen, line 234: maybe nice to show off `..<` instead of `until` :) but up to you

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Create.kt
```diff
@@ -222,6 +220,70 @@ class Create : TestBase() {
...
+            "manufacturer" {
+                "country" from { listOf("USA", "China", "Germany", "Japan").random() }
+                "yearEstablished" from { Random.nextInt(1950, 2020) }
+            }
+
+            // Frame Column
+            "reviews" from {
+                val reviewCount = Random.nextInt(0, 8)
+                (0 until reviewCount).toDataFrame {
+                    val ratings: DataColumn<Int> = expr { Random.nextInt(1, 6) }
```
Jolanrensen: could you use the Kotlin Random extension functions which take ranges instead? :) With those it's immediately clear which number is included and which is not `Random.nextInt(1..5)`, whereas with `Random.nextInt(1, 6)` I'd have to know 6 is excluded

zaleslaw:
Looks very funny and innovative!

zaleslaw (approved):

zaleslaw:
@koperagen @AndreiKingsley is it finished and could be merged?

koperagen:
@zaleslaw @AndreiKingsley 
https://github.com/Kotlin/dataframe/issues/1557

PR #1530, opened 2025-10-29
Add tests for the `count` function

### LINKED ISSUE #1496: Add unit tests for `count` function
This [function ](https://github.com/Kotlin/dataframe/blob/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt)is tested only as a part of integration tests

Create a separate file with Unit tests, achieve if possible 100% coverage

Allex-Nik:
Fixes #1496

### DIFF
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
@@ -9,25 +9,25 @@ import org.junit.BeforeClass
 import org.junit.Test
 
 /**
- * Tests the behavior of the `count` function across different DataFrame structures:
+ * Tests the behavior of the [count] function across different [DataFrame] structures:
  *
- * - **[DataColumn]**: counting all elements or elements matching a predicate,
+ * - [DataColumn]: counting all elements or elements matching a predicate,
  * including behavior on empty columns and columns with `null` values.
  *
- * - **[DataRow]**: counting elements or elements matching a predicate,
+ * - [DataRow]: counting elements or elements matching a predicate,
  * including rows containing `null` values.
  *
- * - **[DataFrame]**: counting all rows or rows matching a predicate in this [DataFrame],
- * including behavior on empty data frames and data frames with `null` values.
+ * - [DataFrame]: counting all rows or rows matching a predicate in this [DataFrame],
+ * including behavior on empty DataFrames and DataFrames with `null` values.
  *
- * - **[GroupBy]**: counting rows per group in the [GroupBy], with and without a predicate,
- * including behavior on grouped empty DataFrame and groups with `null` values.
+ * - [GroupBy]: counting rows per group in the [GroupBy], with and without a predicate,
+ * including behavior on grouped empty [DataFrame] and groups with `null` values.
  *
- * - **[Pivot]**: counting rows in each group of the [Pivot],
+ * - [Pivot]: counting rows in each group of the [Pivot],
  * including handling of `null` values and predicates.
  *
- * - **[PivotGroupBy]**: counting rows in each combined [pivot] + [groupBy] group,
- * with and without predicates, including handling of empty and `null` groups.
+ * - [PivotGroupBy]: counting rows in each combined [pivot] + [groupBy] group,
+ * with and without predicates, including handling of `null` values and predicates.
  */
 class CountTests {
 


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
```diff
@@ -0,0 +1,175 @@
...
+class CountTests {
+
+    // Test data
+
+    val df = dataFrameOf(
+        "name" to columnOf("Alice", "Bob", "Charlie"),
+        "age" to columnOf(15, 20, 25),
+        "group" to columnOf(1, 1, 2),
+    )
+    val age = df["age"].cast<Int>()
```
Allex-Nik: What would be the best way to do that without using the compiler plugin? I also tried: `val age by column<Int>()`, and then write `df[age]` but it is marked as deprecated

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
```diff
@@ -0,0 +1,175 @@
...
+        row.count { it == null } shouldBe 2
+    }
+
+    // DataFrame
+
+    @Test
+    fun `count on dataframe`() {
+        df.count() shouldBe 3
+        df.count { age > 18 } shouldBe 2
+        df.count { it["name"] == "Alice" } shouldBe 1
```
Allex-Nik, line 117: Why can I do `df.count { age > 18 }` but cannot do `df.count { name.startsWith("A") }`? `age` and `name` are obtained in the same way, and we made sure that they have the types that are appropriate for the called methods.

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
```diff
@@ -0,0 +1,175 @@
...
+        val expected = dataFrameOf(
+            "group" to columnOf(1, 2),
+            "count" to columnOf(1, 1),
+        )
+        groupedCount shouldBe expected
+    }
+
+    @Test
+    fun `count on empty grouped dataframe`() {
+        emptyDf.groupBy("group").count().count() shouldBe 0
```
Allex-Nik, line 159: If we do `emptyDf.groupBy("group").count()`, it returns a dataframe without the column `count`. Is it expected?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
Jolanrensen: I think this approach is the best for now. We deprecated the column accessor api (so `val myCol by column()`; `df[myCol]`) in favor of the compiler plugin, but we still have the String API we can use if the compiler plugin cannot be used, like here :)

You're using the String API correctly. The only alternative I can think of is `df.get { "age"<Int>() }` but it achieves exactly the same thing.

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
Jolanrensen: you're calling `CountTests.age` inside `count {}`, not actually `age` from df :)

Jolanrensen: you can use `df.count { column<Int>("age") > 18 }` to access a column from `df` by String name and type within this DSL. Or the shortcut:
`df.count { "age"<Int>() > 18 }`

Jolanrensen: you can do the same with `"name"` and then you can call startsWith :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
Jolanrensen: good catch! We should at least get an empty column named `count`. This causes runtime exceptions with the compiler plugin, but I think it's an issue deep inside aggregation itself... I'll make an issue

Jolanrensen: Thanks! https://github.com/Kotlin/dataframe/issues/1531

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
Allex-Nik: I see, thank you!

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
```diff
@@ -0,0 +1,175 @@
+package org.jetbrains.kotlinx.dataframe.api
+
+import io.kotest.matchers.shouldBe
+import org.jetbrains.kotlinx.dataframe.nrow
+import org.junit.Test
+
+class CountTests {
+
```
zaleslaw, line 33: Run ktlint for that, remove blank lines
add simple KDoc (could be AI generated with simple summary of tested scenarios)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
```diff
@@ -0,0 +1,175 @@
...
+    // Test data
+
+    val df = dataFrameOf(
```
zaleslaw: If you have some common variables reused somewhere in the class, the better pattern to follow JUnit4 hierarchy of tests

Have a look here https://github.com/Kotlin/dataframe/blob/7c189f0f5788de1dc28515c4bdd4d0c4928b2f22/dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/local/postgresTest.kt#L260

use lateinit variables

Some work done before and after all tests
https://stackoverflow.com/questions/20295578/difference-between-before-beforeclass-beforeeach-and-beforeall

zaleslaw: https://junit.org/junit4/javadoc/4.13.2/org/junit/BeforeClass.html - read more about it

This means:

1. All non-static fields (including val properties at class level in Kotlin) are reinitialized for every test method.
3. Any initialization done in the constructor or directly in property declarations runs again per test method.
5. Hence, the fact is correct: every test rebuilds df, grouped, etc. from scratch.

zaleslaw:
```
class CountTests {
    companion object {
        lateinit var df: DataFrame<*>

        @BeforeClass
        @JvmStatic
        fun setupTestData() {
            df = dataFrameOf("name" to columnOf("Alice", "Bob"))
        }
    }
}
```

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
```diff
@@ -0,0 +1,235 @@
+package org.jetbrains.kotlinx.dataframe.api
+
+import io.kotest.matchers.shouldBe
+import org.jetbrains.kotlinx.dataframe.nrow
+import org.junit.Test
+
+class CountTests {
+
+    // Test data
```
Allex-Nik: These fields with test data are reinitialized for each test method, which creates some overhead (some fields are not used by some methods). But I think it should not be significant, it is still safe, and it saves from a lot of duplication. Is it reasonable to leave it as it is?

Jolanrensen: It's just the tests, so it doesn't matter that much. In some cases we have `TestBase` so we can reuse these dataframes, but dataframes are not very heavy, so it most cases it's fine to have them duplicated. It does make it easier to see what kind of dataframe you're dealing with in the tests

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
```diff
@@ -0,0 +1,235 @@
...
+    val pivot = df.pivot("group")
+
+    val emptyDf = df.drop(df.nrow)
+
+    val dfWithNulls = df.append("Martin", null, null)
+    val ageWithNulls = dfWithNulls["age"].cast<Int?>()
+    val groupedWithNulls = dfWithNulls.groupBy("group")
+    val pivotWithNulls = dfWithNulls.pivot("group")
+
+    // DataColumn
```
Jolanrensen: oh btw, do you know region blocks?

If you write


```kt
// region DataColumn
...
// endregion
```
the regions are collapsible and they show up in the structure overview of the file

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
```diff
@@ -0,0 +1,235 @@
...
+        ageWithNulls.count { it == null } shouldBe 1
+    }
+
+    // DataRow
+
+    @Test
+    fun `count on DataRow`() {
+        val row = df[0]
+        row.count() shouldBe 3
+        (row.count { it is Number }) shouldBe 2
```
Jolanrensen: unnecessary parentheses

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
Jolanrensen: maybe explain with a comment that it's empty for now and link to the issue like "Issue #1531". Makes it easier to find our discussion later on :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
Allex-Nik: I didn't know, thank you!

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/count.kt
Allex-Nik: Thank you, I read and understood it

zaleslaw (approved):

Issue #659, opened 2024-04-15
Merge `KeyValueProperty` and `NameValuePair`


Jolanrensen:
Seems like they try to solve the same thing:

DataRowApi.kt:
```kt
@DataSchema
public data class NameValuePair<V>(val name: String, val value: V)
```
KeyValueProperty.kt:
```kt
/** A [DataSchema] interface / class can implement this if it represents a map-like data schema (so key: value). */
@DataSchema
public interface KeyValueProperty<T> {
    // needs to be explicitly overridden in @DataSchema interface, otherwise extension functions won't generate (TODO)
    public val key: String

    // needs to be explicitly overridden in @DataSchema interface, otherwise type will be read as `T` and extensions won't generate (TODO)
    @ColumnName("value")
    public val `value`: T
}
```

PR #1532, opened 2025-10-30
Merging `KeyValueProperty` and `NameValuePair`

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/659

`KeyValueProperty` and `NameValuePair` are both data schema's defining a String and a value, though they are incompatible and are used interchangeably in different parts of the library. The only difference was the name for the key (either `key` or `name`) and the fact that one is an interface and the other a data class.

This PR brings them closer together:
- making `NameValuePair` inherit `KeyValueProperty`
- Deprecating `name` in favor of `key`
- Adding deprecated overloads

At a later stage we can deprecate `NameValuePair` and introduce an instantiable `KeyValueProperty` with a better name.

@Allex-Nik `NameValuePair` is marked `@RequiredByIntellijPlugin`. Do you think these changes will cause issues?

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -1948,6 +1948,8 @@ public final class org/jetbrains/kotlinx/dataframe/api/DataFrameGetKt {
 }
 
 public final class org/jetbrains/kotlinx/dataframe/api/DataRowApiKt {
+	public static final fun NameValuePairAny_key (Lorg/jetbrains/kotlinx/dataframe/ColumnsContainer;)Lorg/jetbrains/kotlinx/dataframe/DataColumn;
+	public static final fun NameValuePairAny_key (Lorg/jetbrains/kotlinx/dataframe/DataRow;)Ljava/lang/String;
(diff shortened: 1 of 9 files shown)


Allex-Nik:
> @Allex-Nik `NameValuePair` is marked `@RequiredByIntellijPlugin`. Do you think these changes will cause issues?

Yes, it will, but it can be easily fixed. 

The type renderer for `NameValuePair` in the IDEA plugin in its current state accesses the `name` and `value` as fields, and displays them in the debugger. These changes prevent the renderer from accessing the `name` field (apparently, it cannot access the deprecated extension property `name` this way), and the fields are not displayed in the debugger for a `NameValuePair`.

However, in the renderer we can just access the `key` field in case access to `name` was unsuccessful, and it will work (I checked it with your changes), so it will be compatible with the versions of the library having new implementation of `NameValuePair` as well as with the previous versions. 

We could also try to access the deprecated `name` property via JDI in the plugin, but it seems to me that checking two fields instead of one would be a better solution :)

Jolanrensen:
@Allex-Nik Great! thanks! :) 

I tried keeping `name` as a deprecated getter to keep binary compatibility, but `@DataSchema` still detects it and generates non-deprecated accessors for the field :/ so I have to take it out of the class unfortunately

PR #1533, opened 2025-10-30
groupBy Kdocs

### LINKED ISSUE #1524: Add `groupBy` KDocs

AndreiKingsley:
close #1524

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
@@ -542,7 +542,7 @@ public interface GroupBy<out T, out G> : Grouped<G> {
      * a row from [keys] and its corresponding group of rows (as [DataFrame]).
      *
      * If [groupedColumnName] is provided, the groups will be stored
-     * in a [FrameColumn] with that name; otherwise, a default name is used.
+     * in a [FrameColumn] with that name; otherwise, a default name "group" is used.
      *
      * @param groupedColumnName The name of the column in which to store grouped data;
      * if `null`, a default name will be used.
--- a/docs/StardustDocs/topics/groupBy.md
+++ b/docs/StardustDocs/topics/groupBy.md
@@ -142,11 +142,7 @@ df.groupBy { name }.concat()
 <inline-frame src="resources/org.jetbrains.kotlinx.dataframe.samples.api.Modify.concatGroupBy.html" width="100%"/>
 <!---END-->
 
-## Reducing
-
-TODO
-
-## Aggregation
+<!---TODO ## Reducing--->
 
 To compute one or several [statistics](summaryStatistics.md) per every group of `GroupBy` use `aggregate` function. 
 Its body will be executed for every data group and has a receiver of type [`DataFrame`](DataFrame.md) that represents current data group being aggregated.


docs/StardustDocs/topics/groupBy.md
```diff
@@ -142,6 +142,10 @@ df.groupBy { name }.concat()
 <inline-frame src="resources/org.jetbrains.kotlinx.dataframe.samples.api.Modify.concatGroupBy.html" width="100%"/>
 <!---END-->
 
+## Reducing
+
+TODO
```
Jolanrensen: do you want to leave this here like this on the website?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/group.kt
```diff
@@ -24,7 +24,8 @@ import kotlin.reflect.KProperty
 // region DataFrame
 
 /**
- * Groups the specified [columns\] within the [DataFrame].
+ * Groups the specified [columns\] within the [DataFrame] into
+ * [column group][ColumnGroup].
```
Jolanrensen, line 28: *a column group

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -21,14 +36,289 @@ import kotlin.reflect.KProperty
 // region DataFrame
 
 /**
+ * Groups the rows of this [DataFrame] based on the values in one or more specified [key columns][cols].
+ * Each unique value in a key column — or a unique tuple of values for multiple columns —
+ * defines a group consisting of all rows where the column(s) contain that value combination.
```
Jolanrensen: I'd say *the group

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -21,14 +36,289 @@ import kotlin.reflect.KProperty
...
+ * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
+ *
+ * See [Selecting Columns][GroupBySelectingOptions].
+ *
+ * For more information: {@include [DocumentationUrls.GroupBy]}
+ *
+ * Don't confuse this with [group], which groups column into
+ * [column group][org.jetbrains.kotlinx.dataframe.columns.ColumnGroup].
+ */
+internal interface GroupByDocs {
```
Jolanrensen, line 70: These docs are amazingly useful!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -21,14 +36,289 @@ import kotlin.reflect.KProperty
...
+     * {@include [LineBreak]}
+     * {@include [DslGrammarLink]}
+     * {@include [LineBreak]}
+     *
+     * ### Create and transform [GroupBy]
+     *
+     * [**`groupBy`**][groupBy]**`(`**`moveToTop: `[`Boolean`][Boolean]**` = true)  {  `**`columns: `[`ColumnsSelector`][ColumnsSelector]**` }`**
+     *
+     * {@include [Indent]}
+     * `\[ `__`.`__[**`sortByGroup`**][GroupBy.sortByGroup]**`() `**
```
Jolanrensen: there are no closing "]"s on any of the lines

Jolanrensen: oh it's an or-or thing, uhhm

Jolanrensen: Actually, this is wrong :) you can sort multiple times, which can make sense in case of ties. So I'd just put `[ .sortByX() ]` all on separate lines

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -21,14 +36,289 @@ import kotlin.reflect.KProperty
...
+internal interface GroupByDocs {
+    /**
+     * ## [groupBy][groupBy] Operation Grammar
+     * {@include [LineBreak]}
+     * {@include [DslGrammarLink]}
+     * {@include [LineBreak]}
+     *
+     * ### Create and transform [GroupBy]
+     *
+     * [**`groupBy`**][groupBy]**`(`**`moveToTop: `[`Boolean`][Boolean]**` = true)  {  `**`columns: `[`ColumnsSelector`][ColumnsSelector]**` }`**
```
Jolanrensen: `` ` = true)  {  ` `` requires an additional space in front of `=`. The https://github.github.com/gfm/#code-spans rules of MD state that code spans starting and ending with a space character will have one surrounding space trimmed, making the specific example look like `"moveToTop: Boolean= true) { columns:..."`. This is currently implemented differently in the KDoc markdown renderer of IntelliJ, so for now it renders correctly, but I suspect it's only a matter of time since they claim to follow the GFM standard.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
Jolanrensen: Also, all `()` have an unneeded space after them

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -21,14 +36,289 @@ import kotlin.reflect.KProperty
...
+     * `| `__`.`__[**`sortByCountAsc`**][GroupBy.sortByCountAsc]**`() `**
+     *
+     * {@include [Indent]}
+     * `| `__`.`__[**`sortByKey`**][GroupBy.sortByKey]**`() `**
+     *
+     * {@include [Indent]}
+     * `| `__`.`__[**`sortByKeyDesc`**][GroupBy.sortByKeyDesc]**`() `**`]`
+     *
+     * {@include [Indent]}
+     * `\[ `__`.`__[**`sortBy`**][GroupBy.sortBy]**`  {  `**`columns: `[`ColumnsSelector`][ColumnsSelector]**` } `**
```
Jolanrensen: unneeded space after last `}`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -21,14 +36,289 @@ import kotlin.reflect.KProperty
...
+     *
+     * {@include [Indent]}
+     * `| `__`.`__[**`sortByDesc`**][GroupBy.sortByDesc]**`  {  `**`columns: `[`ColumnsSelector`][ColumnsSelector]**` } `**`]`
```
Jolanrensen: same story with the 'or'. I'd just write `[]`, it's hard to tell the difference between `[` and `|`, plus `[]` is more accurate

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -21,14 +36,289 @@ import kotlin.reflect.KProperty
...
+     *
+     * {@include [Indent]}
+     * `\[ `__`.`__[**`updateGroups`**][GroupBy.updateGroups]**`  {  `**`frameExpression`**` } `**`]`
```
Jolanrensen: closing `}` requires extra surrounding spaces due to spaces trimming GFM rule

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -21,14 +36,289 @@ import kotlin.reflect.KProperty
...
+     *
+     * {@include [Indent]}
+     * `\[ `__`.`__[**`filter`**][GroupBy.filter]**` {  `**`predicate: `[`GroupedRowFilter`][GroupedRowFilter]**` } `**`]`
```
Jolanrensen: one extra space needed in front of `{`, and `}` requires one extra before and after it

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -21,14 +36,289 @@ import kotlin.reflect.KProperty
...
+     *
+     * {@include [Indent]}
+     * `\[ `__`.`__[**`add`**][GroupBy.add]**`(`**`column: `[`DataColumn`][DataColumn]**`)  {  `**`rowExpression: `[`RowExpression`][RowExpression]**` } `**`]`
```
Jolanrensen: `{` has one space too many in front and after it. They won't be trimmed since `` `) `` doesn't start with a space. `}` needs extra surrounding spaces

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -21,14 +36,289 @@ import kotlin.reflect.KProperty
...
+     * ### Reduce [GroupBy] into [DataFrame]
+     *
+     * {@include [Indent]}
+     * [GroupBy][GroupBy]`.`[**`minBy`**][GroupBy.minBy]**`  {  `**`column: `[`ColumnSelector`][ColumnSelector]**` }`**
+     *
+     * {@include [Indent]}
+     * `| `__`.`__[**`maxBy`**][GroupBy.maxBy]**`  {  `**`column: `[`ColumnSelector`][ColumnSelector]**` }`**
+     *
+     * {@include [Indent]}
+     * `| `__`.`__[**`first`**][GroupBy.first]`  \[ `**` {  `**`rowCondition: `[`RowFilter`][RowFilter]**` } `**`]`
```
Jolanrensen: space usage again:
- `` `  \[ `**` {  `**`rowCondition: ` ``

will render like:
- ` [{ rowCondition: `

but we want it to render like:
- ` [ { rowCondition: `

similar with `}`

Jolanrensen: I won't check the space usage below, you got it :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -21,14 +36,289 @@ import kotlin.reflect.KProperty
...
+     *
+     * {@include [Indent]}
+     * `| `__`.`__[**`last`**][GroupBy.last]`  \[ `**`{  `**`rowCondition: `[`RowFilter`][RowFilter]**` } `**`]`
+     *
+     * {@include [Indent]}
+     * __`.`__[**`concat`**][ReducedGroupBy.concat]**`() `**
```
Jolanrensen, line 129: does this need a `|`?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -21,14 +36,289 @@ import kotlin.reflect.KProperty
...
+     * `| `__`.`__[**`values`**][Grouped.values]**`  {  `**`valueColumns: `[`ColumnsSelector`][ColumnsSelector]**` }`**
+     *
+     * {@include [Indent]}
+     * `| `__`.`__[**`count`**][Grouped.count]**`() `**
+     *
+     * {@include [Indent]}
+     * `| `__`.`__[**`aggregate`**][Grouped.aggregate]**`  {  `**`aggregations: `[`AggregateDsl`][AggregateDsl]**` }`**
+     *
+     * {@include [Indent]}
+     * `| `__`.`__[<aggregation_statistic>][AggregationStatistics]
```
Jolanrensen, line 158: interesting new DSL Grammar concept :) but clear and clickable! Me like

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/pivot.kt
```diff
@@ -275,6 +276,34 @@ internal fun <T> Pivot<T>.reduce(reducer: Selector<DataFrame<T>, DataRow<T>?>):
...
+     * [PivotGroupBy] can be [reduced][PivotGroupByDocs.Reducing]
+     * or [aggregated][PivotGroupByDocs.Aggregation].
+     *
+     * Check out [PivotGroupBy Grammar][PivotGroupByDocs.Grammar].
+     *
+     * For more information: {@include [DocumentationUrls.PivotGroupBy]}
+     */
+    interface CommonDescription
+
+    interface Grammar
```
Jolanrensen, line 303: maybe put something like `TODO Issue #xxx` so we won't forget?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -21,14 +36,289 @@ import kotlin.reflect.KProperty
...
+     * * [median][Grouped.median]/[medianOf][Grouped.medianOf]/[medianFor][Grouped.medianFor]
+     * * [percentile][Grouped.percentile]/[percentileOf][Grouped.percentileOf]/[percentileFor][Grouped.percentileFor]
+     */
+    interface AggregationStatistics
+
+    /**
+     * ### [GroupBy] transformations
+     *
+     * A [GroupBy] can be transformed into a new [GroupBy] using one of the following methods:
+     * * [sortByGroup][GroupBy.sortByGroup] / [sortByGroupDesc][GroupBy.sortByGroupDesc] — sorts the **order
```
Jolanrensen, line 201: since these first 4 have a different default for asc/desc, could you mention something about which order is used by default? Or do you think this is clear enough when shown like this?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -21,14 +36,289 @@ import kotlin.reflect.KProperty
...
+     *   of groups** (and their corresponding keys) by values computed with a [DataFrameExpression] applied to each group;
+     * * [sortByCount][GroupBy.sortByCount] / [sortByCountAsc][GroupBy.sortByCountAsc] — sorts the **order
+     *   of groups** (and their corresponding keys) by the number of rows they contain;
+     * * [sortByKey][GroupBy.sortByKey] / [sortByKeyDesc][GroupBy.sortByKeyDesc] — sorts the **order
+     *   of groups** (and their corresponding keys) by the grouping key values;
+     * * [sortBy][GroupBy.sortBy] / [sortByDesc][GroupBy.sortByDesc] — sorts the **order of rows within each group**
+     *   by one or more column values;
+     * * [updateGroups][GroupBy.updateGroups] — transforms each group into a new one;
```
Jolanrensen, line 209: what's up with all the semicolons at the end of the summations here and in the kdocs below? I don't think they add much, as it's already clear you're summing up options with the dots at the beginning of each line :) I'd either write a period or just nothing

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -48,21 +347,91 @@ public fun <T> DataFrame<T>.groupBy(vararg cols: AnyColumnReference, moveToTop:
...
+ * ```kotlin
+ * // Create a new column "newName" based on existing "oldName" values
+ * // and use it as a grouping key:
+ * pivot.groupBy { expr("newName") { oldName.drop(5) } }
+ * ```
+ *
+ * @include [PivotGroupByDocs.CommonDescription]
+ */
+@ExcludeFromSources
+internal interface GroupByForPivotDocs
```
Jolanrensen: can be private too

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -48,21 +347,91 @@ public fun <T> DataFrame<T>.groupBy(vararg cols: AnyColumnReference, moveToTop:
...
+
 // region Pivot
 
+/**
+ * {@include [CommonGroupByForPivotDocs]}
+ * @include [SelectingColumns.Dsl]
+ *
+ * #### For example:
+ *
+ * `pivot.`{@get [OPERATION]}` { length `[and][ColumnsSelectionDsl.and]` age }`
```
Jolanrensen, line 383: why are you using `@get` here? They only serve purpose if these docs are included somewhere else and you need a different operation here. `[groupBy][groupBy]` is shorter to write even ;P

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -48,21 +347,91 @@ public fun <T> DataFrame<T>.groupBy(vararg cols: AnyColumnReference, moveToTop:
...
 @AccessApiOverload
 public fun <T> Pivot<T>.groupBy(vararg columns: AnyColumnReference): PivotGroupBy<T> = groupBy { columns.toColumnSet() }
 
+/**
+ * {@include [CommonGroupByForPivotDocs]}
+ * @include [SelectingColumns.ColumnNames]
+ *
+ * #### For example:
+ *
+ * `df.`{@get [OPERATION]}`("length", "age")`
```
Jolanrensen, line 411: same here

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -71,42 +440,160 @@ public fun <T> Pivot<T>.groupByOther(): PivotGroupBy<T> {
...
+/**
+ * An alternative representation of a [GroupBy.Entry], holding a key–group pair.
+ *
+ * @property key The key represented as a [DataRow].
+ * @property group The [DataFrame] containing the rows belonging to this group.
+ */
 public data class GroupWithKey<T, G>(val key: DataRow<T>, val group: DataFrame<G>)
 
+/**
+ * A dataframe-like structure that contains all unique combinations of key values
```
Jolanrensen: *key-values?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -71,42 +440,160 @@ public fun <T> Pivot<T>.groupByOther(): PivotGroupBy<T> {
...
     public fun filter(predicate: GroupedRowFilter<T, G>): GroupBy<T, G>
 
+    /**
+     * Converts this [GroupBy] into a [DataFrame].
+     *
+     * Each row of the resulting [DataFrame] represents a unique key–group pair:
+     * a row from [keys] and its corresponding group of rows (as [DataFrame]).
+     *
+     * If [groupedColumnName] is provided, the groups will be stored
+     * in a [FrameColumn] with that name; otherwise, a default name is used.
```
Jolanrensen: what's that name?

Jolanrensen (changes requested):

docs/StardustDocs/topics/groupBy.md
AndreiKingsley: Ok, will comment it, should be added later (#1535)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
AndreiKingsley: `[groupedColumnName]` argument

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
AndreiKingsley: Ok, now I added `]` to the end of each line, so we need a space in the end.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
AndreiKingsley: I added ] to the end of each line, so we need a space in the end.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
AndreiKingsley: same

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
AndreiKingsley: yep, it's selection of one of the aggregation methods: concat or into or values

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
AndreiKingsley: Should we add it to Grammar spec? :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
AndreiKingsley: I believe it's ok when we write it like this, but we indeed should mentions this in KDocs of these functions

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
AndreiKingsley: There is a rule that when writing such a list in English, short phrases should end with a semicolon (the last one with a period). Here should be ';', fixed

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
AndreiKingsley: Fair enough, but I'd like to fix it further (not in ths PR) to use Dsl.WithExample with a custom variable name (`df` by default but with ability to customize it)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
Jolanrensen: sure :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
Jolanrensen: ah, interesting, didn't know that

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
Jolanrensen: no I mean "a default name", what's the default?

AndreiKingsley: added

Issue #1535, opened 2025-10-31
Update `groupBy` and `pivot` website documentation


AndreiKingsley:
* update example with a new iframes
* add info about reducing
* add missing methods (`filter`, `meanBy` and others, see all operations in Grammar added in #1533)

PR #1537, opened 2025-10-31
Aggregation of empty grouped df should still generate columns

### LINKED ISSUE #1531: Aggregating an empty DF does not make aggregation columns appear
Let's say we have

```kt
val df = dataFrameOf("a")("a")
val counted = df.drop(1).groupBy { a }.count()
```
or
```kt
val counted = df.drop(1).groupBy { a }.aggregate { count() into "count" }
```

In this case `counted` does not obtain the column `count`, it just has the empty column `a`.

This is an issue, because the compiler plugin cannot tell a dataframe is empty, so it will happily suggest `.count` and cause runtime exceptions:

Image

We should at least get an empty `Int` column `count` in this case.

Thanks for the find @Allex-Nik!

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1531

The reason
```kt
emptyDf.groupBy { name }.aggregate {
    count() into "count"
}
```
or
```kt
emptyDf.groupBy { name }.count()
```

would not generate a `count` column, was because the aggregation was run once for each group and then concatenated. This means it was never called if the dataframe was empty and there were no groups.

This is unexpected behavior and causes runtime exceptions when accessing the result of the aggregation with the compiler plugin (it cannot tell whether a DF is empty or not).

I fixed it by calling the `aggregate {}` body once with an empty group if there are no groups. This generates all columns as expected even though there's no data inside.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/aggregate.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/aggregate.kt
@@ -21,4 +21,8 @@ public fun <T, R> Pivot<T>.aggregate(separate: Boolean = false, body: Selector<A
 @Refine
 @Interpretable("Aggregate")
 public fun <T, R> Grouped<T>.aggregate(body: AggregateGroupedBody<T, R>): DataFrame<T> =
-    aggregateGroupBy((this as GroupBy<*, *>).toDataFrame(), { groups.cast() }, removeColumns = true, body).cast<T>()
+    aggregateGroupBy(
+        df = (this as GroupBy<*, *>).toDataFrame(),
+        selector = { groups.cast() },
+        body = body,
+    ).cast()
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/GroupByImpl.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/GroupByImpl.kt
@@ -75,7 +75,6 @@ internal class GroupByImpl<T, G>(
 internal fun <T, G, R> aggregateGroupBy(
     df: DataFrame<T>,
     selector: ColumnSelector<T, DataFrame<G>?>,
-    removeColumns: Boolean,
     body: AggregateGroupedBody<G, R>,
 ): DataFrame<T> {
     val defaultAggregateName = "aggregated"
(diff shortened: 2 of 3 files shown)


AndreiKingsley (approved):
I believe we need to add some info about `groupBy` / aggregation behavior on empty dataframe in our documentation.

Jolanrensen:
> I believe we need to add some info about `groupBy` / aggregation behavior on empty dataframe in our documentation.

hmm maybe yes, but on the other hand, it's not very common and I believe this new behavior is what people will expect; if you aggregate an empty dataframe, your aggregators will have to operate on empty columns. Makes sense, right?

But we could mention it as part of https://github.com/Kotlin/dataframe/issues/1535

koperagen:
What if aggregate would throw exception if df is empty? "Fail fast"

Jolanrensen:
@koperagen We could, but I think it's best to not throw an exception if we don't have to. If users create a pipeline for `DataFrame<TheirType>`, it should behave the same in terms of types regardless of the df being empty or not.
In this case, all statistical aggregation functions have `orNull {}` versions specifically for empty columns (or they work fine on empty ones, like `sum()` or `count()`), that way the user can handle empty dataframes gracefully if they expect them.

Plus, most functions, like `max {}`, already throw an exception if they get an empty column as input, suggesting that users use `maxOrNull {}` instead.

...come to think of it. Actually, using `-OrNull` overloads makes zero sense here
image

After running this, `max` doesn't have to be of type `DataColumn<String?>`, it can remain `DataColumn<String>`, it's just empty. The problem is that `max { col }` throws an exception when `col` is empty, so we can't execute it...

I need to think about this for a bit... Maybe throwing and exception earlier is better after all...

We could really use Rich Errors here! That way `max {}` could return `ColumnIsEmptyError<T> | T` and we could make an empty column of type `T`... When an exception is thrown, the other lines in the aggregate body are not executed so we don't see them at all :/

... actually, another way to look at it: we kinda already use "rich errors" for these statistics functions. We filter out all nulls from the input of columns before performing the operation, so if `maxOrNull` returns `null`, this is comparable to returning a `ColumnIsEmptyError`. This holds for all statistical operations. The compiler plugin would just need to be adjusted to know that "-OrNull overloads of statistics will not introduce nullability to the columns" :)

... I have no idea how, though, because users can also introduce `nulls` in aggregations whenever they please:
`emptyGroupBy.aggregate { null into "nulls" }.nulls` and then `nulls` should definitely be of type `Nothing?`.

koperagen:
> The compiler plugin would just need to be adjusted to know that "-OrNull overloads of statistics will not introduce nullability to the columns" :)

interesting!

zaleslaw (approved):

koperagen:
Faced this issue :( 
```kt
val interestingResources = dataFrameOf("inventoryId", "id", "count")(1,"2", 3).drop(1)
val resourceCounts = interestingResources.groupBy { id }.aggregate {
    sumOf { count }
} // id, aggregated
val totalValue = resourceCounts.sum() // aggregated
totalValue.aggregated // fails here
```
So yes, i'd say i want an empty "aggregated" column. It would fix my pipeline

Jolanrensen:
@koperagen One problem is still our implementation itself:
- Aggregation is run for each row, and the results are combined into columns.
- We'd need to run `.aggregate {}` at least once to get any result
- If the DataFrame is empty, though, we have nothing to run it on.
- For some statistics, like `sum` we could get away with returning `0` for an empty `Int` column, but what is there to return for `max { "myEmptyCol"<String>() }`?

Issue #1423, opened 2025-09-01
Replace documentation interfaces with (nested) typealiases to shrink jar size


koperagen:
There are unused symbols in the jar for example in the api package:
<img width="1956" height="1392" alt="Image" src="https://github.com/user-attachments/assets/7f8b4f56-15eb-4081-a012-ff1422d848e1" />

I believe most of them are utility interaces for Kodex not used for navigation. Let's check it and try to remove unused ones, and maybe convert other to typealiases? https://kotlinlang.org/docs/type-aliases.html#nested-type-aliases

Jolanrensen:
Good idea to run an [`@ExcludeFromSources`](https://github.com/Kotlin/dataframe/blob/master/KDOC_PREPROCESSING.md#excludefromsources-annotation-excluding-code-content-from-sources) check :)

We will need to be careful about which documentation interfaces we can exclude though. Interfaces which users must be able to access via clicking (such as DSL Grammars) should not be excluded.

Examples of interfaces that are safe for exclusion are [Link Interfaces](https://github.com/Kotlin/dataframe/blob/master/KDOC_PREPROCESSING.md#link-interfaces), [Arg Interfaces](https://github.com/Kotlin/dataframe/blob/master/KDOC_PREPROCESSING.md#arg-interfaces), [URLs](https://github.com/Kotlin/dataframe/blob/master/KDOC_PREPROCESSING.md#urls) (I think they're already excluded), and `XxxDocs` interfaces (only if they are solely `@include`d, never just linked to, and their potentially nested interfaces are also all excludable).

I like the idea of type aliases :D they are now supported in KoDEx too! (Though this would only correspond to the "leaf" interfaces, still that's quite many!)

(Need to test whether Dokka/KoDEx supports nested ones though, one moment...)

Jolanrensen:
<img width="960" height="540" alt="Image" src="https://github.com/user-attachments/assets/092ba6ef-3e34-45d2-b7ac-6f33ae91662f" />

:)

<img width="1264" height="1101" alt="Image" src="https://github.com/user-attachments/assets/066bd4cf-c948-452d-8142-012b57cd4bb3" />

:(

Created an issue for it at KoDEx: https://github.com/Jolanrensen/KoDEx/issues/91

Jolanrensen:
Long story short: We may need to bump to Gradle 9.x to be able to use nested type aliases from KoDEx

Jolanrensen:
[...]

PR #1543, opened 2025-11-03
(Nested) typealias KDocs

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1423 by enabling nested type aliases and replacing all bodiless documentation interfaces with type aliases of `Nothing`. Type aliases, compared to interfaces, are "transparent" in the JVM, as they don't exist in Java. This saves quite some bytes in the final jar.

With this refactor, we save 400kB :)


koperagen:
Very cool!

Jolanrensen:
just need to check notebooks etc don't have issues with nested typealiases

Jolanrensen:
image
:smiling_face_with_tear:

Looks like we have to wait a bit

Jolanrensen:
Now that https://github.com/Kotlin/dataframe/issues/1650 is merged, I redid the conversion and overwrote this branch :)

PR #1545, opened 2025-11-04
`NameValueProperty`

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/659 2nd attempt. (see first attempt [here](https://github.com/Kotlin/dataframe/pull/1532))

After discussing with @koperagen, this time I did it the other way around; keeping the names 'name' and 'value' instead of 'key' and 'value'.

Deprecated `KeyValueProperty` in favor of `NameValueProperty`. Made `NameValuePair` inherit `NameValueProperty`, bringing them closer together.

@Allex-Nik same question as for the first attempt :)

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -13,7 +13,7 @@ libsPublisher = "2.2.0-dev-61"
 
 # "Bootstrap" version of the dataframe, used in the build itself to generate @DataSchema APIs,
 # dogfood Gradle / KSP plugins in tests and idea-examples modules
-dataframe = "1.0.0-dev-8314"
+dataframe = "1.0.0-dev-8808"
 korro = "0.1.6"
 
 binaryCompatibilityValidator = "0.17.0"


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/NameValueProperty.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    // needs to be explicitly overridden in @DataSchema interface, otherwise type will be read as `T` and extensions won't generate (TODO)
+    @ColumnName("value")
+    public val `value`: T
+}
+
+// region Deprecated
+
+/** A [DataSchema] interface / class can implement this if it represents a map-like data schema (so key: value). */
+@Deprecated(KEY_VALUE_PROPERTY, ReplaceWith("NameValueProperty<T>"))
```
koperagen, line 27: Just curious, is it working as expected? Never tried replace with + generics

Jolanrensen: image
image
yep :)

Allex-Nik:
> @Allex-Nik same question as for the first attempt :)

These changes should not cause any issues with the IDEA plugin. `NameValuePair` with the same names of fields is still there, and `AnyRow.namedValues()` still returns `List<NameValuePair<Any?>>`, so the plugin has access to the fields it refers to.  

I checked locally how rendering of `NameValuePair` in the plugin works with these changes - it works fine :)

koperagen (approved):

Issue #1546, opened 2025-11-06
DataFrame with `Nothing?` column is detected as `FormattedFrame<*>?` in notebooks


Jolanrensen:
As reported here: https://kotlinlang.slack.com/archives/C05333T208Y/p1762352545398199 

This is the direct result of [`hasFormattedColumns()`](https://github.com/Kotlin/dataframe/blob/a2c7f5dc9a82fbd4cacc65d09d40773b0954a8e8/dataframe-jupyter/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/JupyterHtmlRenderer.kt#L103):
```kt
internal fun AnyFrame.hasFormattedColumns() =
    this.getColumns { colsAtAnyDepth().colsOf<FormattedFrame<*>?>() }.isNotEmpty()
```

As `colsOf<T>()` includes all subtypes of `T`, `colsOf<FormattedFrame<*>>()?`  will include `null` filled columns, of type `Nothing?`. This behavior is correct, though in `hasFormattedColumns()` this causes a problem, as it will return `true` if the dataframe has a `Nothing?` column.

Let's check the library for other `colsOf<.*?>()` usages to prevent similar issues.

ileasile:
Please check such DF in the notebook after fix - it's important that it's rendered fine when it's not considered formatted

Jolanrensen:
While in most cases, the potential `Nothing?` column is handled well, it can still show up unexpected in some cases, like:

<img width="1003" height="1229" alt="Image" src="https://github.com/user-attachments/assets/295c8be5-35cb-49ac-9871-20872765a134" />

or

<img width="2476" height="1272" alt="Image" src="https://github.com/user-attachments/assets/87b4741f-c6bf-4de5-960d-0aaba7b2ba70" />

btw, this is a weird bug too actually, you should be able to put any instance in a `Nothing` column

Edit: no, it actually isn't. `fillNulls {}` uses `update {}` which doesn't allow changing types of columns. `fillNulls { nothing? }` simply doesn't make sense as there's nothing you can fill it up with without changing the type.

Issue #686, opened 2024-05-02
`GroupBy.take()` and other missing functions


Jolanrensen:
Currently, we can't do:
```kt
groupedDf
    .take(10)
    .concat()
```
to only concatenate the values of the first 10 groups. Instead, we'll have to convert to a normal DF first and convert back:
```kt
groupedDf
    .toDataFrame().take(10).asGroupBy()
    .concat()
```
The only row-based function that's available is `filter(GroupedRowFilter)` which can allow you to write `.filter { it.index() <= 10 }` but seems a bit odd.

Jolanrensen:
Other missing functions include `size()`, `drop()`, `first()` etc.
Maybe we could make it an `AnyFrame` or a `DataColumn<GroupedDataRow>`

Jolanrensen:
interestingly `.filter {}` runs on a `GroupedRowFilter<T, G>`, where `T` is the original DF type. This allows type-safe access to all key columns, but also to all non-key columns which don't exist in the GroupBy object, causing Exceptions... This might need a slight redesign.

koperagen:
> interestingly `.filter {}` runs on a `GroupedRowFilter<T, G>`, where `T` is the original DF type. This allows type-safe access to all key columns, but also to all non-key columns which don't exist in the GroupBy object, causing Exceptions... This might need a slight redesign.

https://github.com/Kotlin/dataframe/pull/663

Jolanrensen:
`GroupBy` has some other curiosities I found while digging into it; It has a bunch of (extension) functions for it which can be divided into three categories:

1. Functions that run on the individual groups, for example
  - Those that transform the groups, like `add {}`, `cumSum {}`, etc.
  - Reducing functions like `minBy {}`, `first()`, `last {}`
  - `aggregate {}` belongs a bit in the first two categories, but the functions inside its DSL run on individual groups
2. Functions that run on the entire `GroupBy` by columns or do something special
  - Like `aggregate {}`, `pivot {}`, `concat()`/`concatWithKeys()`, `into()`/`toDataFrame()`
  - `xs()`, filters columns both from the keys and groups
  - `sortBy {}`
 3. Functions that run on the entire `GroupBy`, treating it per row of keys+group
  - `forEach {}`, `map {}`/`mapToFrames {}`/`mapToRows {}`, `filter {}`
  - They are not mentioned on https://kotlin.github.io/dataframe/groupby.html
  - They have 3 different ways of representing a "row" of keys + group

[...]

PR #1547, opened 2025-11-06
Documentation and tests for the `first` and `firstOrNull` functions

### LINKED ISSUE #1279: Add KDocs and missed tests for `first` and `firstOrNull` functions
https://github.com/Kotlin/dataframe/blob/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt

Also check the coverage and add appropriate unit test (it's mentioned only in integration tests)

Allex-Nik:
Fixes #1279

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
@@ -139,7 +139,7 @@ public fun <T> DataFrame<T>.firstOrNull(): DataRow<T>? = if (nrow > 0) first() e
 /**
  * Returns the first [row][DataRow] in this [DataFrame] that satisfies the given [predicate].
  *
- * {@include [RowFilterDescription]}
+ * @include [RowFilterDescription]
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
  *
@@ -173,7 +173,7 @@ public inline fun <T> DataFrame<T>.first(predicate: RowFilter<T>): DataRow<T> =
  * Returns `null` if the [DataFrame] contains no rows matching the [predicate]
  * (including the case when the [DataFrame] is empty).
  *
- * {@include [RowFilterDescription]}
+ * @include [RowFilterDescription]
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
  *
@@ -241,7 +241,7 @@ public fun <T, G> GroupBy<T, G>.first(): ReducedGroupBy<T, G> = reduce { firstOr
  * the corresponding row in [ReducedGroupBy] will contain `null` values for all columns in the group,
  * except the grouping key.
  *
- * {@include [RowFilterDescription]}
+ * @include [RowFilterDescription]
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
  *
@@ -293,7 +293,7 @@ public fun <T> Pivot<T>.first(): ReducedPivot<T> = reduce { firstOrNull() }
  *
  * For more information about [Pivot] with examples: {@include [DocumentationUrls.Pivot]}
  *
- * {@include [RowFilterDescription]}
+ * @include [RowFilterDescription]
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
  *
@@ -355,13 +355,13 @@ public fun <T> PivotGroupBy<T>.first(): ReducedPivotGroupBy<T> = reduce { firstO
  * in the resulting [ReducedPivotGroupBy] it will be represented by a [row][DataRow] with `null` values
  * (except the grouping key).
  *
- * {@include [DocumentationUrls.PivotGroupBy]}
+ * @include [DocumentationUrls.PivotGroupBy]
  *
- * {@include [DocumentationUrls.Pivot]}
+ * @include [DocumentationUrls.Pivot]
  *
- * {@include [DocumentationUrls.GroupBy]}
+ * @include [DocumentationUrls.GroupBy]
  *
- * {@include [RowFilterDescription]}
+ * @include [RowFilterDescription]
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
  *


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -36,8 +42,227 @@ class FirstTests : ColumnsSelectionDslTests() {
             df.select { "name".firstCol { col -> col.any { it == "Alice" } } },
             df.select { Person::name.firstCol { col -> col.any { it == "Alice" } } },
             df.select { NonDataSchemaPerson::name.firstCol { col -> col.any { it == "Alice" } } },
+            df.select { pathOf("name").firstCol() },
             df.select { pathOf("name").firstCol { col -> col.any { it == "Alice" } } },
             df.select { it["name"].asColumnGroup().firstCol { col -> col.any { it == "Alice" } } },
         ).shouldAllBeEqual()
     }
+
+    @Test
```
Allex-Nik: Is it fine to put these tests in the same class with `FirstColumnsSelectionDsl`? Or is it better to put them in a different class?
If it is better to put them in a different class, is it worth inheriting from `TestBase` to reuse `df`?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -36,8 +42,227 @@ class FirstTests : ColumnsSelectionDslTests() {
...
+    fun `firstOrNull on DataFrame`() {
+        df.firstOrNull()?.name?.lastName shouldBe "Cooper"
+        df.drop(df.nrow).firstOrNull() shouldBe null
+
+        df.firstOrNull { !isHappy }?.name?.lastName shouldBe "Daniels"
+        df.firstOrNull { age > 50 } shouldBe null
+        df.drop(df.nrow).firstOrNull { isHappy } shouldBe null
+    }
+
+    @Test
```
Allex-Nik, line 137: For now I haven't added a test on an empty dataframe since if we do `emptyDf.groupBy { age }.first()`, we do not get the `group` column. This is a known issue, Jolan fixed it, but it is not merged yet.

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -36,8 +42,227 @@ class FirstTests : ColumnsSelectionDslTests() {
...
+    fun `first on GroupBy`() {
+        val grouped = df.groupBy { isHappy }
+        val reducedGrouped = grouped.first()
+        val firstHappy = reducedGrouped.values()[0]
+        val firstUnhappy = reducedGrouped.values()[1]
+
+        firstHappy shouldBe dataFrameOf(
```
Allex-Nik, line 143: `grouped.first().values()[0].name.lastName shouldBe "Cooper"` also worked (and similar approach in some further tests), but I decided to refrain from using generated properties, as they were only generated for `df` without transformations, and attempts to access these properties in some further transformations cause issues. I think not using the compiler plugin further in the file is more correct.

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -36,8 +42,227 @@ class FirstTests : ColumnsSelectionDslTests() {
...
+                "lastName" to columnOf("Daniels")
+            ),
+            "age" to columnOf(20),
+            "city" to columnOf("Moscow"),
+            "weight" to columnOf(null),
+        )[0]
+    }
+
+    @Test
+    fun `first on GroupBy with predicate`() {
```
Allex-Nik, line 166: Might be an issue here: 

`df.groupBy { isHappy }.first{ age > 10 }` 

works fine: ReducedGroupBy contains columns `isHappy`, `name` (`firstName`, `lastName`), `age`, `city`, `weight`, but

`df.groupBy { isHappy }.first{ age > 100 }`

 returns ReducedGroupBy without the `name` column.

For example, this test passes, but it seems to me that it should not:

`grouped.first { age > 100 }.values()[0] shouldBe dataFrameOf("isHappy", "age", "city", "weight")(true, null, null, null)[0]`.

That's why for now I haven't added a test for a predicate that doesn't match any row.

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -36,8 +42,227 @@ class FirstTests : ColumnsSelectionDslTests() {
...
+                "firstName" to columnOf("Alice"),
+                "lastName" to columnOf("Wolf")
+            ),
+            "age" to columnOf(20),
+            "city" to columnOf(null),
+            "weight" to columnOf(55),
+        )[0]
+    }
+
+    @Test
```
Allex-Nik, line 225: I faced an issue here:

`ReducedPivot` (i.e. `pivot.first()`) and `ReducedPivotGroupBy` are not rendered correctly in notebooks when there is `null` in any row resulting from `first()`.

If I replace `null` in such a row with some value, it is displayed correctly.

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -36,8 +42,227 @@ class FirstTests : ColumnsSelectionDslTests() {
...
+                "firstName" to columnOf("Charlie"),
+                "lastName" to columnOf("Daniels")
+            ),
+            "age" to columnOf(20),
+            "city" to columnOf("Moscow"),
+            "weight" to columnOf(null),
+        )[0]
+    }
+
+    @Test
```
Allex-Nik, line 304: I took a simpler dataframe here for readability. Otherwise it was more laborious for a reader to validate the result.

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
Allex-Nik, line 83: Is it fine to put these tests in the same class with FirstColumnsSelectionDsl? Or is it better to put them in a different class?
If it is better to put them in a different class, is it worth inheriting from TestBase to reuse df?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -27,32 +29,141 @@ import kotlin.reflect.KProperty
 
 // region DataColumn
 
+/**
+ * Returns the first value in this [DataColumn].
+ *
+ * @param T The type of the values in the [DataColumn].
```
koperagen: Line seems redundant to me because param T is always inferred from DataColumn<T>

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -62,26 +173,209 @@ public inline fun <T> DataFrame<T>.firstOrNull(predicate: RowFilter<T>): DataRow
 
 // region GroupBy
 
+/**
+ * Selects the first row from each group of the given [GroupBy]
+ * and returns a [ReducedGroupBy] containing these rows
+ * (one row per group, each row is the first row in its group).
```
koperagen: or null if group is empty

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -62,26 +173,209 @@ public inline fun <T> DataFrame<T>.firstOrNull(predicate: RowFilter<T>): DataRow
...
+ *
+ * ### Example
+ * ```kotlin
+ * // Select the first employee from each group formed by the job title
+ * employees.groupBy { jobTitle }.first()
+ * ```
+ *
+ * @param T The type of the values in the [GroupBy].
```
koperagen: I think these lines are redundant because they are always infered

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -62,26 +173,209 @@ public inline fun <T> DataFrame<T>.firstOrNull(predicate: RowFilter<T>): DataRow
...
+ *
+ * See also:
+ * - [pivot];
+ * - common [reduce][Pivot.reduce].
+ *
+ * ### Example
+ * ```kotlin
+ * // Select the first row for each city.
+ * // Returns a ReducedPivot with one column per city and the first row from the group in each column.
+ * df.pivot { city }.first()
```
koperagen: Please see if you can come up with representative example. Like, in what situation you'd use this function? What df typically it will be and what ideas one can draw from the result? Will be good if example can convey this

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -62,26 +173,209 @@ public inline fun <T> DataFrame<T>.firstOrNull(predicate: RowFilter<T>): DataRow
...
+ *   corresponding to the values of that key;
+ * - each value contains the first row from that group.
+ *
+ * The original [Pivot] column structure is preserved.
+ * If the [Pivot] was created using multiple or nested keys
+ * (e.g., via [and][PivotDsl.and] or [then][PivotDsl.then]),
+ * the structure remains unchanged — only the contents of each group
+ * are replaced with the first row from that group.
+ *
+ * Equivalent to `reduce { firstOrNull() }`.
```
koperagen: reduce is internal function, people won't be able to use it like this

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -62,26 +173,209 @@ public inline fun <T> DataFrame<T>.firstOrNull(predicate: RowFilter<T>): DataRow
...
 
 // endregion
 
 // region Pivot
 
+/**
+ * Reduces this [Pivot] by selecting the first row from each group.
+ *
+ * Returns a [ReducedPivot] where:
+ * - each column corresponds to a [pivot] group — if multiple pivot keys were used,
```
koperagen: I think text explanations of pivot make it more scary than it is. For `first` i suggest to not include common pivot logic and refer to pivot kdoc instead
Reference to website with HTML tables or ascii tables might do a better job conveying what's going on

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
koperagen: It's ok to put it here. It's common to have functions with same name in the same file, even if it's CS DSL and DF API

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
AndreiKingsley: Yes, I'd omit (everywhere) a `@param` with type parameter, and add `@return`!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -27,32 +29,141 @@ import kotlin.reflect.KProperty
 
 // region DataColumn
 
+/**
+ * Returns the first value in this [DataColumn].
+ *
+ * @param T The type of the values in the [DataColumn].
+ *
```
AndreiKingsley, line 39: Please, add here and in all other places "See also" section with related operations. For example
```
See also [firstOrNull], [last], [take].
```

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
Allex-Nik: I have faced an issue in this case that might be unexpected behavior.

I started with `df.take(0).groupBy { age }.first()` - it doesn't return `null` (returns ReducedGroupBy), which is probably fine because there are no groups at all.

But now I have tried
```kt
val grouped = df.groupBy { age }
grouped.updateGroups {
    if (it == grouped.groups[0]) {
        it.take(0)
    } else it
}.first()
```
to make the first group empty. And applying `first` in this case causes an exception:
``The problem is found in one of the loaded libraries: check library renderers
java.lang.IllegalStateException: Can not insert column `age` because column with this path already exists in DataFrame``.

This problem does not occur if every group has at least one row, or if I remove the column `age` from every group. For example, this works:
```kt
grouped.updateGroups {
    val new = it.remove { age }
    if (it == grouped.groups[0]) {
        new.take(0)
    } else new
}.first()
```
We get `null` for an empty group and the first row for others.

But is it expected behavior that we get such an error about conflicting columns? Or maybe I am just obtaining an empty group incorrectly.

The `df` in this case is:
```kt
val df = dataFrameOf(
    "name" to columnOf("Alice", "Bob", "Charlie"),
    "age" to columnOf(15, 20, 25),
)
```

Or, to use a bit more natural example, we can make a `fullJoin` of `df` with `val ages = dataFrameOf("age" to columnOf(30))`, then group by `age`, and filter out the row with `null` value in the group for the key 30. And applying `first` in this case causes the same exception.

I am reporting this just in case it is a not known issue :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
Jolanrensen: related to https://github.com/Kotlin/dataframe/issues/1531

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -27,32 +29,141 @@ import kotlin.reflect.KProperty
...
+ * @param T The type of the values in the [DataColumn].
+ */
 public fun <T> DataColumn<T>.firstOrNull(): T? = if (size > 0) first() else null
 
+/**
+ * Returns the first value in this [DataColumn] that matches the given [predicate].
+ *
+ * ### Example
+ * ```kotlin
+ * // Select from the column "age" the first value where the age is greater than 17
```
Jolanrensen: "select" is confusing, as we also have the `select` operation. I'd say "get"

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -27,32 +29,141 @@ import kotlin.reflect.KProperty
...
+ * df.age.first { it > 17 }
+ * ```
+ *
+ * @param T The type of the values in the [DataColumn].
+ * @param predicate A lambda expression used to select a value
```
Jolanrensen: *the first value

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -27,32 +29,141 @@ import kotlin.reflect.KProperty
...
+ * that satisfies a condition specified in this expression.
+ * This predicate takes a value from the [DataColumn] as an input
+ * and returns `true` if the value satisfies the condition or `false` otherwise.
+ *
+ * @throws [NoSuchElementException] if the [DataColumn] contains no element matching the [predicate]
```
Jolanrensen: Helpful! We don't add these enough. Though I would add "@see firstOrNull" somewhere around here so people will know how to avoid this exception

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -27,32 +29,141 @@ import kotlin.reflect.KProperty
...
+ * and returns `true` if the value satisfies the condition or `false` otherwise.
+ */
 public fun <T> DataColumn<T>.firstOrNull(predicate: (T) -> Boolean): T? = values.firstOrNull(predicate)
 
 // endregion
 
 // region DataFrame
 
+/**
+ * Returns the first row in this [DataFrame].
```
Jolanrensen: you could link to `DataRow` from "row" :) may be helpful

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
Jolanrensen: holds for the other functions as well

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
Jolanrensen: same applies below

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -62,26 +173,209 @@ public inline fun <T> DataFrame<T>.firstOrNull(predicate: RowFilter<T>): DataRow
...
+ * - common [reduce][Pivot.reduce].
+ *
+ * {@include [RowFilterDescription]}
+ *
+ * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
+ *
+ * ### Example
+ * ```kotlin
+ * // Select the first row for each city where the population is greater than 100 000.
+ * df.pivot { city }.first { population > 100000 }
```
Jolanrensen: 100_000 is better readable ;P (and compiles!)

Jolanrensen: or `10e5` or something ;P

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -62,26 +173,209 @@ public inline fun <T> DataFrame<T>.firstOrNull(predicate: RowFilter<T>): DataRow
...
+ *
+ * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
+ *
+ * ### Example
+ * ```kotlin
+ * // From each combination of faculty and enrollment year select the first student older than 21.
+ * students.pivot { faculty }.groupBy { enrollmentYear }.first { age > 21 }
+ * ```
+ *
+ * @param predicate A lambda expression used to select a value
```
Jolanrensen: oh you can actually also link to `RowFilter` :) it has some useful kdoc too

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
Jolanrensen: *gradle plugin.

The compiler plugin is not enabled here. If it were, you would indeed be able to use extension properties with transformations :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
Jolanrensen: Same reason as https://github.com/Kotlin/dataframe/issues/1531 I assume?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -36,8 +42,227 @@ class FirstTests : ColumnsSelectionDslTests() {
...
+    fun `first on GroupBy with predicate`() {
+        val grouped = df.groupBy { isHappy }
+        val reducedGrouped = grouped.first{ it["age"] as Int > 17 && it["city"] != "Moscow" }
```
Jolanrensen: Don't forget to lint :) I recommend using the KtLint plugin. You can also call `ktLintFormat` from gradle

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
Jolanrensen: how do they look when they are rendered incorrectly?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
Jolanrensen: Your first issue can also be reproduced in notebook by 
```kt
df.groupBy { age }.updateGroups { it }.first()
```
I'm not entirely sure why..

Jolanrensen: Ah! `df.groupBy { age }.updateGroups { it }.first().concat()` works fine again. So the issue is in the renderer for `ReducedGroupBy` in notebooks

Jolanrensen: Can be reproduced outside notebooks with `df.groupBy { age }.updateGroups { it }.first().values()`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
Allex-Nik: I see that it might look a bit ambiguous (especially because of "used"), but when I wrote this, my idea was that the logic of determining the first value is outside the scope of the predicate, and the predicate as a function just returns true if the input satisfies the condition. Do you think we still need to change it to "the first value"?

Maybe it's not that important, but I am clarifying because this part occurs in several places in this file and in `last.kt` as well :)

Jolanrensen: It may be important information to know that it will stop calling the predicate after the first `true` is is found. 
That's not conveyed when you say it's "used to select a value" :)
So yes, you explained in the broad sense what a "predicate" is, but I think it will be more valuable if you describe more specifically what it's used for in this specific case.

(similarly when you have a function that takes `age: Int` and you start explaining what an `Int` is ;P)

Allex-Nik: I see, will fix it, thank you!

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
Allex-Nik: Maybe, but it is not completely clear to me why it is `name` that disappears, and, for example, for
```kt
val students = dataFrameOf(
    "name" to columnOf("Alice", "Bob"),
    "age" to columnOf(15, 20),
    "group" to columnOf(1, 2)
)

students.groupBy { name }.count { name == "Charlie" }
```
we get the column `count` with 0s.
Or when we do `students.groupBy { name }.first { age == 30 }`, no column disappears.

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
Allex-Nik: image

I meant this format.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -27,32 +29,141 @@ import kotlin.reflect.KProperty
...
+ *
+ * ### Example
+ * ```kotlin
+ * // Select the first row where the value in the "age" column is greater than 17
+ * // and the "name/firstName" column starts with 'A'
+ * df.first { age > 17 && name.firstName.startsWith("A") }
+ * ```
+ *
+ * @param T The type of the [DataFrame].
+ * @param predicate A lambda expression used to select a value
```
zaleslaw: I could only suggest to make it more readable yet with adding [] brackets for the parameter names `@param [predicate] A lambda....`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -27,32 +29,141 @@ import kotlin.reflect.KProperty
...
+ * @param T The type of the values in the [DataColumn].
+ * @param predicate A lambda expression used to select a value
+ * that satisfies a condition specified in this expression.
+ * This predicate takes a value from the [DataColumn] as an input
+ * and returns `true` if the value satisfies the condition or `false` otherwise.
+ *
+ * @throws [NoSuchElementException] if the [DataColumn] contains no element matching the [predicate]
+ * (including the case when the [DataColumn] is empty).
+ */
 public fun <T> DataColumn<T>.first(predicate: (T) -> Boolean): T = values.first(predicate)
```
Allex-Nik, line 75: Is there any reason we do not make this function `inline`? `last` in the same case is `inline` (the same goes for `DataColumn<T>.firstOrNull(predicate: (T) -> Boolean)`)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
Jolanrensen:
```kt
students.groupBy { name }.count { name == "Charlie" }
```
is a shortcut for
```kt
students.groupBy { name }.aggregate { count { name == "Charlie" } into "count" }
```
`count()` is an aggregator; it operates on each individual group and aggregates all their values per column into single values, usually based on statistics; in this case a count. The result is a normal `DataFrame` again.


whereas `first()`/`last()`/`minBy`/`maxBy` are reducers, they work per row, producing a `ReducedGroupBy`, offering a couple different ways of turning their result back into a dataframe:

You can `concat()` the reduced rows, turn them `into("") {}` a columnGroup, or take the `values {}` of specific columns (which, as opposed to concat, produce nulls if no value is present) and then concatenate them (Notebooks show the `.values()` version of `ReducedGroupBy` if you display it)

I agree this distinction is quite arbitrary and unclear looking just at the function names. There's yet another set of functions that operate on the entire `GroupBy`, like `sortBy {}` and `forEach {}`, complicating things even further... I would vote for a GroupBy overhaul but we can only do that in 1.1 most likely: https://github.com/Kotlin/dataframe/issues/686#issuecomment-3196206489

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
Jolanrensen: Ah yes, another one of https://github.com/Kotlin/dataframe/issues/1546 in the wild :) It's been fixed

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -62,26 +173,209 @@ public inline fun <T> DataFrame<T>.firstOrNull(predicate: RowFilter<T>): DataRow
 
 // region GroupBy
 
+/**
+ * Selects the first row from each group of the given [GroupBy]
```
Jolanrensen: (Now that I understand a bit more about GroupBy) maybe it makes sense to phrase it like:

"[Reduces][GroupByDocs.Reducing] the groups by taking the first from each..."

This points people a bit more in the right direction.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -62,26 +173,209 @@ public inline fun <T> DataFrame<T>.firstOrNull(predicate: RowFilter<T>): DataRow
...
+ * and returns a [ReducedGroupBy] containing these rows
+ * (one row per group, each row is the first row in its group).
+ *
+ * ### Example
+ * ```kotlin
+ * // Select the first employee from each group formed by the job title
+ * employees.groupBy { jobTitle }.first()
```
Jolanrensen: A `ReducedGroupBy` is rarely the end. It may make sense to add `concat()`, `into()`, or `values()` after it

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -62,26 +173,209 @@ public inline fun <T> DataFrame<T>.firstOrNull(predicate: RowFilter<T>): DataRow
...
+ * @param T The type of the values in the [GroupBy].
+ * @param G The type of the groups in the [GroupBy].
+ *
+ * @return A [ReducedGroupBy] containing the first row from each group.
+ */
 @Interpretable("GroupByReducePredicate")
 public fun <T, G> GroupBy<T, G>.first(): ReducedGroupBy<T, G> = reduce { firstOrNull() }
 
+/**
+ * Selects from each group of the given [GroupBy] the first row satisfying the given [predicate],
```
Jolanrensen: I would also use "Reduces each group..." here

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
Jolanrensen: you may also look at #1554; Andrei and I have been discussing a lot about how to explain Pivot from text. It may have some good example phrases you can reuse :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -62,26 +218,192 @@ public inline fun <T> DataFrame<T>.firstOrNull(predicate: RowFilter<T>): DataRow
...
+ *
+ * If a group in this [GroupBy] is empty,
+ * the corresponding [row][DataRow] in the resulting [ReducedGroupBy] will contain `null` values
+ * for all columns in the group, except the grouping key.
+ *
+ * ### Example
+ * ```kotlin
+ * // In a DataFrame of orders sorted by date and time,
+ * // find the first order placed by each customer
+ * df.groupBy { customerId }.first().concat()
```
Allex-Nik, line 223: Does the usage of `concat` here and `concat` or `values` in the following examples make sense?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -27,32 +30,185 @@ import kotlin.reflect.KProperty
...
 public fun <T> DataColumn<T>.firstOrNull(predicate: (T) -> Boolean): T? = values.firstOrNull(predicate)
 
 // endregion
 
 // region DataFrame
 
+/**
+ * Returns the first [row][DataRow] in this [DataFrame].
+ *
+ * @see [firstOrNull]
```
Allex-Nik: I was thinking of adding a receiver for all cases other than DataColumn, for example  `@see [DataFrame.lastOrNull]` in region DataFrame. Asked about this here: https://github.com/Kotlin/dataframe/pull/1561#discussion_r2569670648
If the receiver is not specified, the link always leads to the version of the function that is defined on DataColumn.

Jolanrensen: that's a good idea. However, inside `@see` they you cannot hide the `DataFrame` part. It will be shown in any case. However, this is probably not an issue. It's more important it links to the right function

Allex-Nik: Yes, I tried it, and the `DataFrame` part is visible in the documentation. I tried something like `[firstOrNull][DataFrame.firstOrNull]` but it didn't work. However, it provides the correct link.
documentation

If visibility of a receiver is not an issue, I will update these `@see` here then :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
```diff
@@ -27,32 +30,185 @@ import kotlin.reflect.KProperty
...
 public fun <T> DataColumn<T>.firstOrNull(predicate: (T) -> Boolean): T? = values.firstOrNull(predicate)
 
 // endregion
 
 // region DataFrame
 
+/**
+ * Returns the first [row][DataRow] in this [DataFrame].
+ *
+ * @see [DataFrame.firstOrNull]
```
AndreiKingsley: Do you actually need to provide receiver here?
Anyway, please, omit it. Just `[firstOrNull]` should work, but if not, please use 
`[firstOrNull][DataFrame.firstOrNull]`

AndreiKingsley: Do it in all other place, please

AndreiKingsley (changes requested):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
Allex-Nik: We discussed it here: https://github.com/Kotlin/dataframe/pull/1547#discussion_r2569994610

`@see [firstOrNull][DataFrame.firstOrNull]` still leads to showing the whole thing: `DataFrame.firstOrNull`.

But I think we can also do it in this style:
```
 See [lastOrNull][DataFrame.lastOrNull],
 [first][DataFrame.first],
 [take][DataFrame.take],
 [takeLast][DataFrame.takeLast],
 [takeWhile][DataFrame.takeWhile]
```

It would look like here:
picture

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/first.kt
Jolanrensen: nope, please inline it too :) that will allow people to call suspend functions inside the body etc

Jolanrensen (approved):

PR #1551, opened 2025-11-06
Nothing cols fixes

Jolanrensen:
This PR contains a handful of fixes and optimizations regarding `Nothing(?)` in column types:

- Fixes https://github.com/Kotlin/dataframe/issues/1546 and adds a helpful note to `colsOf` to prevent future confusion
- `DataColumn.empty()` created a `Unit`-typed column. I have no idea why, I made it `Nothing`, which better represents that `emptyCol.get(0): Nothing` will always throw an exception. Added `DataColumn.emptyOf<T>()` if you need a specifically typed empty column.
- [#1546](https://github.com/Kotlin/dataframe/issues/1546) also discovered that exceptions in `update` provided little info, so I added `UpdateException` for this.
- While inspecting Column DataCollector for null-filled or empty columns, I rewrote `::class`-logic to `typeOf`'s (for efficiency), and added redundancy checks:
  - (If the data collector should produce a `Nothing` column, it can only do that if it has no values; so the resulting column must be empty)
  - (if the data collector should produce a `Nothing?` column, it can only do that if all its values are `null`)

I also checked usages of `colsOf()` querying for nullable types to see if similar issues like [#1546](https://github.com/Kotlin/dataframe/issues/1546) existed. Fortunately, in most cases, including null-filled columns are not an issue! All statistics functions simply skip nulls, for instance.

There are some unfortunate cases, like:
```kt
dataFrameOf(
    "a" to columnOf(1, 2, 3, null),
    "b" to columnOf(null, null, null, null),
    "c" to columnOf(7, 3, 2, 65),
).fillNulls { colsOf<Int?>() }.with { 0 }
```
which throw `java.lang.IllegalArgumentException: Can not add value of class kotlin.Int to column of type kotlin.Nothing?. Value = 0` because column `b: Nothing?` is included by `colsOf<Int?>()`. However, hopefully, the note to add `.filter { !it.allNulls() }` is found :). Alternatively, we could decide to add an argument `colsOf<Int?>(exactType = true)` in the future.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataColumn.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataColumn.kt
@@ -218,7 +218,9 @@ public interface DataColumn<out T> : BaseColumn<T> {
         ): DataColumn<T> = createByType(name, values, typeOf<T>(), infer)
 
         /**
-         * Creates an empty [DataColumn] with given [name].
+         * Creates an empty [DataColumn] with given [name] of type [Nothing].
+         * If you want to specify another type, use [`emptyOf<T>()`][emptyOf].
+         *
          * @see emptyOf
          */
         public fun empty(name: String = ""): DataColumn<Nothing> =


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
```diff
@@ -94,28 +95,36 @@ private fun <C, R> ColumnGroup<C>.replaceRowsIf(
         .asColumnGroup()
         .cast()
 
+public class UpdateException(override val message: String, cause: Throwable? = null) :
+    IllegalStateException(message, cause),
+    DataFrameError
```
koperagen: Better to add this supertype only if you have example of code that triggers this in compiler plugin

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataColumn.kt
```diff
@@ -216,8 +217,16 @@ public interface DataColumn<out T> : BaseColumn<T> {
             infer: Infer = Infer.None,
         ): DataColumn<T> = createByType(name, values, typeOf<T>(), infer)
 
-        /** Creates an empty [DataColumn] with given [name]. */
-        public fun empty(name: String = ""): AnyCol = createValueColumn(name, emptyList<Unit>(), typeOf<Unit>())
+        /**
+         * Creates an empty [DataColumn] with given [name].
+         * @see emptyOf
+         */
+        public fun empty(name: String = ""): DataColumn<Nothing> =
```
koperagen, line 226: We have two operations that use empty: concatImpl and getColumn. Please see if they can be affected by this change. Otherwise it's quite neat

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
Jolanrensen: Ah, that might be difficult, as it most often relies on runtime information. (besides maybe `update {}.with { error() }`). I thought we now preferred to use the `DataFrameError` type for any new exception we create

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataColumn.kt
Jolanrensen: seems `concat` fails on dataframes without rows, but not because of `Nothing` columns, but because it uses `::class` reflection in `createColumnGuessingType()`, interesting

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
koperagen: I think about it as a way to opt-in reporting warnings for exceptions, that we know for sure make sense for user. In theory we could report any exception. DataFrameError will help us to validate them first

Jolanrensen: soo... DataFrameWarning? ;P

koperagen: Could be

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataColumn.kt
Jolanrensen: fixed it in the next commit :)

koperagen (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/DataColumn.kt
```diff
@@ -216,8 +217,16 @@ public interface DataColumn<out T> : BaseColumn<T> {
             infer: Infer = Infer.None,
         ): DataColumn<T> = createByType(name, values, typeOf<T>(), infer)
 
-        /** Creates an empty [DataColumn] with given [name]. */
-        public fun empty(name: String = ""): AnyCol = createValueColumn(name, emptyList<Unit>(), typeOf<Unit>())
+        /**
+         * Creates an empty [DataColumn] with given [name].
```
AndreiKingsley: I believe it's important to adjust kdoc here, that 
1) "it creates an empty column of type `Nothing`"
2) "to create an empty column with the specified type use `emptyOf`"

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/colsOf.kt
```diff
@@ -96,6 +96,11 @@ public interface ColsOfColumnsSelectionDsl {
      *
      * This function operates solely on columns at the top-level.
      *
+     * __NOTE:__ Null-filled columns of type [Nothing?][Nothing] will be included when selecting [`colsOf`][colsOf]`<T?>()`.
+     *   This is because [Nothing][Nothing] is considered a subtype of all other types in Kotlin.
+     *   To exclude these columns, call `.`[filter][ColumnsSelectionDsl.filter]` { !it.`[allNulls][DataColumn.allNulls]`() }`
```
AndreiKingsley, line 101: Why `colsOf<T?> except colsOf<Nothing>` is not an option here?

AndreiKingsley (changes requested):
Great! I've missed that behavior a few times.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/colsOf.kt
Jolanrensen: `Nothing` cannot be used as reified argument

Jolanrensen: just have a look what we have to do to achieve `typeOf<Nothing>()` in dataframe XD https://github.com/Kotlin/dataframe/blob/fc5033c75615766d06fdda9ed141d390539e0175/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/TypeUtils.kt#L563

Jolanrensen: we would need to have a shortcut for that `nothingCols()` for instance, but I'm not sure it's common enough

Jolanrensen: or, you know, they type
```kt
colsOf<T?>() except colsOf(typeOf<List<Nothing>>().arguments.first().type!!)
```
you know, elegant

PR #1553, opened 2025-11-07
Remove redundant type parameter from transpose

koperagen:
It's not used at all and all usages still compile
Makes it clearer:
image

vs

image


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/transpose.kt
```diff
@@ -13,7 +12,7 @@ import kotlin.reflect.typeOf
 
 // region DataRow
 
-public fun <T> DataRow<T>.transpose(): DataFrame<NameValuePair<*>> {
+public fun AnyRow.transpose(): DataFrame<NameValuePair<*>> {
```
Jolanrensen, line 15: I wonder if this is a binary change or not, otherwise please deprecate the old one as ERROR :) (or HIDDEN even)

koperagen: No changes in api dump. Generic are erased in bytecode, so they should not affect binary compatibility

Jolanrensen: all good then :)

PR #1556, opened 2025-11-10
Compare DataFrames

### LINKED ISSUE #658: Comparing two data frame
I have two data frames having same same schema, Is there way to compare the two data frames ? so that it provide the added , deleted and modified rows. It may take some single/group of Key columns and Ignore columns.

CarloMariaProietti:
FIXES #658 
It makes possible to compare DataFrame by exploiting Myers difference algotithm whose cost is O((M+N)*D) .
M is length of dfA, N is length of dfB, D is length of shortest edit script to get B from A.

Returns a DataFrame< ComparisonDescription >, 
ComparisonDescription is a schema created specifically for this use case.

It comes with a proper test case.

About Myers difference algotithm:
https://neil.fraser.name/writing/diff/myers.pdf

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
@@ -1,6 +1,7 @@
 package org.jetbrains.kotlinx.dataframe.impl.api
 
 import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.DataRow
 import org.jetbrains.kotlinx.dataframe.annotations.DataSchema
 import org.jetbrains.kotlinx.dataframe.api.DataRowSchema
 import org.jetbrains.kotlinx.dataframe.api.concat
@@ -9,11 +10,12 @@ import org.jetbrains.kotlinx.dataframe.api.emptyDataFrame
 import org.jetbrains.kotlinx.dataframe.nrow
 
 @DataSchema
-internal class ComparisonDescription(
+internal class ComparisonDescription<T>(
     val rowAtIndex: Int,
     val of: DataFrameOfComparison,
     val modification: RowOfComparison,
     val insertedAfterRow: Int?,
+    val modifiedRowContent: DataRow<T>,
 ) : DataRowSchema
 
 internal enum class DataFrameOfComparison {
@@ -30,8 +32,8 @@ internal enum class RowOfComparison {
  * Returns a DataFrame whose rows explain the differences between dfA and dfB.
  * One must think of the set of commands in a script as being executed simultaneously
  */
-internal fun <T> compareDataFramesImpl(dfA: DataFrame<T>, dfB: DataFrame<T>): DataFrame<ComparisonDescription> {
-    var comparisonDf = emptyDataFrame<ComparisonDescription>()
+internal fun <T> compareDataFramesImpl(dfA: DataFrame<T>, dfB: DataFrame<T>): DataFrame<ComparisonDescription<T>> {
+    var comparisonDf = emptyDataFrame<ComparisonDescription<T>>()
     // compare by exploiting Myers difference algorithm
     val shortestEditScript = myersDifferenceAlgorithmImpl(dfA, dfB)
     for (i in 1 until shortestEditScript.size) {
@@ -44,6 +46,7 @@ internal fun <T> compareDataFramesImpl(dfA: DataFrame<T>, dfB: DataFrame<T>): Da
             xPrev + 1 == x && yPrev + 1 != y -> {
                 val indexOfRemovedRow = x - 1
                 val sourceDfOfRemovedRow = DataFrameOfComparison.DFA
+                val rowContent = dfA[indexOfRemovedRow]
                 comparisonDf = comparisonDf.concat(
                     dataFrameOf(
                         ComparisonDescription(
@@ -51,6 +54,7 @@ internal fun <T> compareDataFramesImpl(dfA: DataFrame<T>, dfB: DataFrame<T>): Da
                             sourceDfOfRemovedRow,
                             RowOfComparison.REMOVED,
                             null,
+                            rowContent,
                         ),
                     ),
                 )
@@ -61,6 +65,7 @@ internal fun <T> compareDataFramesImpl(dfA: DataFrame<T>, dfB: DataFrame<T>): Da
                 val indexOfInsertedRow = y - 1
                 val sourceDfOfInsertedRow = DataFrameOfComparison.DFB
                 val indexOfReferenceRow = x - 1
+                val rowContent = dfB[indexOfInsertedRow]
                 comparisonDf = comparisonDf.concat(
                     dataFrameOf(
                         ComparisonDescription(
@@ -68,6 +73,7 @@ internal fun <T> compareDataFramesImpl(dfA: DataFrame<T>, dfB: DataFrame<T>): Da
                             sourceDfOfInsertedRow,
                             RowOfComparison.INSERTED,
                             indexOfReferenceRow,
+                            rowContent,
                         ),
                     ),
                 )
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/compareDataFrames.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/compareDataFrames.kt
@@ -38,11 +38,11 @@ class CompareDataFramesTest {
         )
         val comparison = compareDataFramesImpl(dfA, dfB)
         comparison shouldBe dataFrameOf(
-            ComparisonDescription(0, DataFrameOfComparison.DFA, RowOfComparison.REMOVED, null),
-            ComparisonDescription(1, DataFrameOfComparison.DFA, RowOfComparison.REMOVED, null),
-            ComparisonDescription(1, DataFrameOfComparison.DFB, RowOfComparison.INSERTED, 2),
-            ComparisonDescription(5, DataFrameOfComparison.DFA, RowOfComparison.REMOVED, null),
-            ComparisonDescription(5, DataFrameOfComparison.DFB, RowOfComparison.INSERTED, 6),
+            ComparisonDescription(0, DataFrameOfComparison.DFA, RowOfComparison.REMOVED, null, dfA[0]),
+            ComparisonDescription(1, DataFrameOfComparison.DFA, RowOfComparison.REMOVED, null, dfA[1]),
+            ComparisonDescription(1, DataFrameOfComparison.DFB, RowOfComparison.INSERTED, 2, dfB[1]),
+            ComparisonDescription(5, DataFrameOfComparison.DFA, RowOfComparison.REMOVED, null, dfA[5]),
+            ComparisonDescription(5, DataFrameOfComparison.DFB, RowOfComparison.INSERTED, 6, dfB[5]),
         )
     }
 
@@ -76,12 +76,12 @@ class CompareDataFramesTest {
         )
         val comparison = compareDataFramesImpl(dfA, dfB)
         comparison shouldBe dataFrameOf(
-            ComparisonDescription(0, DataFrameOfComparison.DFA, RowOfComparison.REMOVED, null),
-            ComparisonDescription(1, DataFrameOfComparison.DFA, RowOfComparison.REMOVED, null),
-            ComparisonDescription(2, DataFrameOfComparison.DFA, RowOfComparison.REMOVED, null),
-            ComparisonDescription(0, DataFrameOfComparison.DFB, RowOfComparison.INSERTED, 2),
-            ComparisonDescription(1, DataFrameOfComparison.DFB, RowOfComparison.INSERTED, 2),
-            ComparisonDescription(2, DataFrameOfComparison.DFB, RowOfComparison.INSERTED, 2),
+            ComparisonDescription(0, DataFrameOfComparison.DFA, RowOfComparison.REMOVED, null, dfA[0]),
+            ComparisonDescription(1, DataFrameOfComparison.DFA, RowOfComparison.REMOVED, null, dfA[1]),
+            ComparisonDescription(2, DataFrameOfComparison.DFA, RowOfComparison.REMOVED, null, dfA[2]),
+            ComparisonDescription(0, DataFrameOfComparison.DFB, RowOfComparison.INSERTED, 2, dfB[0]),
+            ComparisonDescription(1, DataFrameOfComparison.DFB, RowOfComparison.INSERTED, 2, dfB[1]),
+            ComparisonDescription(2, DataFrameOfComparison.DFB, RowOfComparison.INSERTED, 2, dfB[2]),
         )
     }
 


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
```diff
@@ -0,0 +1,196 @@
+package org.jetbrains.kotlinx.dataframe.impl.api
+
+import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.api.DataRowSchema
+import org.jetbrains.kotlinx.dataframe.api.concat
+import org.jetbrains.kotlinx.dataframe.api.dataFrameOf
+import org.jetbrains.kotlinx.dataframe.api.emptyDataFrame
+import org.jetbrains.kotlinx.dataframe.nrow
+
+internal class ComparisonDescription(
```
Jolanrensen: data schemas are created with `@DataSchema`, not with `: DataRowSchema`

CarloMariaProietti: Thank you for reviewing! I added `@DataSchema` , however with the current implementation `: DataRowSchema`
is still necessary because of lines 41-42..

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
```diff
@@ -0,0 +1,196 @@
...
+            }
+        }
+        // try with a longer edit script
+        d++
+    }
+    val immutablePath = path.toList()
+    return immutablePath
+}
+
+internal fun <T> recoursivePathFill(
```
Jolanrensen: *recursive?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
```diff
@@ -0,0 +1,196 @@
...
+    // Return value
+    val path = mutableListOf<Pair<Int, Int>>()
+    // 'ses' stands for shortest edit script, next var is never returned, it is in the code
+    // to show the capabilities of the algorithm
+    var sesLength: Int?
+    val sumOfLength = dfA.nrow + dfB.nrow
+    // matrix containing the endpoint of the furthest reaching D-path ending in diagonal k
+    // for each d-k couple of interest
+    val v = arrayListOf<IntArray>()
+    for (d in 0..sumOfLength) {
```
Jolanrensen: d is not used

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/compareDataFrames.kt
```diff
@@ -0,0 +1,183 @@
+package org.jetbrains.kotlinx.dataframe.api
+
+import io.kotest.matchers.shouldBe
+import org.jetbrains.kotlinx.dataframe.impl.api.ComparisonDescription
+import org.jetbrains.kotlinx.dataframe.impl.api.compareDataFramesImpl
+import org.jetbrains.kotlinx.dataframe.impl.api.myersDifferenceAlgorithmImpl
+import org.junit.Test
+import kotlin.Pair
+
+private class SchemaForThisTest(val integer: Int, val string: String) : DataRowSchema
```
Jolanrensen: Also needs to be `@DataSchema`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
```diff
@@ -0,0 +1,196 @@
...
+    for (i in 1 until shortestEditScript.size) {
+        x = shortestEditScript[i].first
+        y = shortestEditScript[i].second
+        xPrev = shortestEditScript[i - 1].first
+        yPrev = shortestEditScript[i - 1].second
+        when {
+            // row at index 'x-1' of dfA was removed
+            xPrev + 1 == x && yPrev + 1 != y -> {
+                comparisonDf = comparisonDf.concat(
+                    dataFrameOf
```
Jolanrensen: strange formatting, can you lint this file?

Jolanrensen: also probably best to add argument names to ComparisonDescription, "magic" argumentless constants are often a source of bugs :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
```diff
@@ -0,0 +1,196 @@
...
+
+/**
+ * Returns a DataFrame whose rows explain the differences between dfA and dfB.
+ * One must think of the set of commands in a script as being executed simultaneously
+ */
+internal fun <T> compareDataFramesImpl(dfA: DataFrame<T>, dfB: DataFrame<T>): DataFrame<ComparisonDescription> {
+    var comparisonDf = emptyDataFrame<ComparisonDescription>()
+    // compare by exploiting Myers difference algorithm
+    val shortestEditScript = myersDifferenceAlgorithmImpl(dfA, dfB)
+    var x: Int?
```
Jolanrensen: these don't need to be nullable

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
```diff
@@ -0,0 +1,196 @@
...
+ *
+ * snake: non-diagonal edge and then a possibly empty sequence of diagonal edges
+ * D-path: a path starting at (0,0) that has exactly D non-diagonal edges
+ */
+internal fun <T> myersDifferenceAlgorithmImpl(dfA: DataFrame<T>, dfB: DataFrame<T>): List<Pair<Int, Int>> {
+    // Return value
+    val path = mutableListOf<Pair<Int, Int>>()
+    // 'ses' stands for shortest edit script, next var is never returned, it is in the code
+    // to show the capabilities of the algorithm
+    var sesLength: Int?
```
Jolanrensen: not null

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
```diff
@@ -0,0 +1,196 @@
...
+    var isOver = false
+    // starting the algorithm
+    // 0 position is -(M+N) position in the alg's paper -> need to normalize each access to v
+    val normalizer = sumOfLength
+    v[0][1 + normalizer] = 0 // fitticious
+    // d is the number of non-diagonal edges
+    var d = 0
+    while (d <= sumOfLength && !isOver) {
+        for (k in -d..d step 2) {
+            var x: Int?
```
Jolanrensen: not null

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
```diff
@@ -0,0 +1,196 @@
...
+) {
+    // Enlist my self
+    val xCurrent = v[d][k + normalizer]
+    val yCurrent = xCurrent - k
+    path.add(Pair(xCurrent, yCurrent))
+    // I look for endpoint I was built from, it is represented by kPrev.
+    // It will be an argument of the next recoursive step.
+    // Moreover, I need to enlist the points composing the snake that precedes me (it may be empty).
+    if (d > 0) {
+        var kPrev: Int? = null
```
Jolanrensen: not null, nor do they need to be initialized

Jolanrensen: can be `val kPrev: Int`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
Jolanrensen: you can also put them inside the for-loop and make them `val`. This looks very java-like :P

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
```diff
@@ -0,0 +1,196 @@
...
+internal fun <T> myersDifferenceAlgorithmImpl(dfA: DataFrame<T>, dfB: DataFrame<T>): List<Pair<Int, Int>> {
+    // Return value
+    val path = mutableListOf<Pair<Int, Int>>()
+    // 'ses' stands for shortest edit script, next var is never returned, it is in the code
+    // to show the capabilities of the algorithm
+    var sesLength: Int?
+    val sumOfLength = dfA.nrow + dfB.nrow
+    // matrix containing the endpoint of the furthest reaching D-path ending in diagonal k
+    // for each d-k couple of interest
+    val v = arrayListOf<IntArray>()
```
Jolanrensen: `mutableListOf` is the kotlin way

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
Jolanrensen: can be `tailrec`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
```diff
@@ -0,0 +1,196 @@
...
+    for (d in 0..sumOfLength) {
+        v.add(IntArray(sumOfLength * 2 + 1))
+    }
+    var isOver = false
+    // starting the algorithm
+    // 0 position is -(M+N) position in the alg's paper -> need to normalize each access to v
+    val normalizer = sumOfLength
+    v[0][1 + normalizer] = 0 // fitticious
```
Jolanrensen: "fictitious"?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
```diff
@@ -0,0 +1,196 @@
...
+import org.jetbrains.kotlinx.dataframe.api.DataRowSchema
+import org.jetbrains.kotlinx.dataframe.api.concat
+import org.jetbrains.kotlinx.dataframe.api.dataFrameOf
+import org.jetbrains.kotlinx.dataframe.api.emptyDataFrame
+import org.jetbrains.kotlinx.dataframe.nrow
+
+internal class ComparisonDescription(
+    val rowAtIndex: Int,
+    val of: String,
+    val wasRemoved: Boolean?,
```
Jolanrensen: this results in `true` or `null`... what's wrong with `false`? Same below

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
```diff
@@ -0,0 +1,196 @@
...
+import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.api.DataRowSchema
+import org.jetbrains.kotlinx.dataframe.api.concat
+import org.jetbrains.kotlinx.dataframe.api.dataFrameOf
+import org.jetbrains.kotlinx.dataframe.api.emptyDataFrame
+import org.jetbrains.kotlinx.dataframe.nrow
+
+internal class ComparisonDescription(
+    val rowAtIndex: Int,
+    val of: String,
```
Jolanrensen: if this only contains "dfA" or "dfB" this could better be an enum. Unfortunately dataframes cannot have a title :/ that would have been even better

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
```diff
@@ -0,0 +1,196 @@
...
+    val wasRemoved: Boolean?,
+    val wasInserted: Boolean?,
+    val afterRow: Int?,
```
Jolanrensen: maybe call this `insertedAfterRow` or something more expressive. That explains why it's `null` if `wasRemoved == true`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
Jolanrensen: so... rows can either be removed or inserted, right? nothing else. Let's turn these two booleans in an enum as well then

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
Jolanrensen: I see :) but then please still add @DataSchema as well

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
```diff
@@ -0,0 +1,196 @@
...
+) : DataRowSchema
```
Jolanrensen, line 19: I wonder if we could include the modified `DataRow<*>` in the resulting DataFrame as well. That could make it a bit easier

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
CarloMariaProietti: Yes, sorry for the grammar error.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
CarloMariaProietti: I choose for null instead of false to remark that if wasInserted is true, wasRemoved column does not even make sense (and vive-versa).

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/compareDataFrames.kt
CarloMariaProietti: Adding `@DataSchema` forced the schema to be internal/public. I did it the same.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
CarloMariaProietti: Myers difference algorithm exploits the idea of comparison in a boolean sense,
a member (row) of the data structure (df) is equal or not-equal to another member.
In this logic a modified row is represented by two `DataRow<ComparisonDescription>` ,
One row is like: _**row n of dfA was removed**_  and the other: _**row m of dfB was inserted after row n-1 (of dfA)**_.

CarloMariaProietti: Imo representing explicitly modified rows means customing Myers Alg logic by introducing a 3-possible-output non boolean logic of comparison: Equal, Non Equal, Similar. Similar means that compared row differ for a limited number of elements (that may be proportional to row's length). 
In a lower abstraction level,  'similar' looks like a 'flagged' diagonal move in the edit script graph
-> list of `Pair` would be no more enough to represent the result, we would need a list of 3-ple.
Anyway, imo, this slution has the following weakness: the result is less neutral because the previous  definition of 'Similarity'  can't determine with certainty wheter a row was actually modified.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
```diff
@@ -124,9 +149,9 @@ internal fun <T> myersDifferenceAlgorithmImpl(dfA: DataFrame<T>, dfB: DataFrame<
     return immutablePath
 }
 
-internal fun <T> recoursivePathFill(
+internal fun <T> tailrec(
```
Jolanrensen: I meant `internal tailrec fun <T> recursivePathFill()`: https://kotlinlang.org/docs/functions.html#tail-recursive-functions

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
Jolanrensen: I see; well, in that case an enum makes more sense I think

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
Jolanrensen: Correct me if I'm wrong, but honestly, this seems a thing AI would do wrong. If you're using AI, that's okay; however, you remain the one responsible for the code.

At first you had:
```kt
val wasRemoved: Boolean?
val wasInserted: Boolean?
val afterRow: Int?
```

I remarked here: https://github.com/Kotlin/dataframe/pull/1556#discussion_r2571643349 both values had two options: either `null` or `true`. This is an odd thing to do with booleans which, as you know, have 2 values already: `true` or `false`, however, you explained why, so that makes slightly more sense now :)

What I don't understand is why you now have two nullable `RowOfComparison`s:
```kt
val wasRemoved: RowOfComparison?
val wasInserted: RowOfComparison?
```
What would these even contain? A row is either 'removed' (`WAS_REMOVED`) or 'inserted' (`WAS_INSERTED_AFTER_ROW`), right? So why not just make one column called `modification: RowOfComparison`?

Next, I meant you to rename `afterRow` to `insertedAfterRow`, not `wasInserted`.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
Jolanrensen: I'm not (yet) asking for in-row differences. Simply for adding the original row to a new column when that row was "Not Equal", so either "removed" or "inserted". A `ComparisonDescription` like **row n of dfA was removed** doesn't really help me if I can't see what "row n of dfA" actually contains. Similar to **row m of dfB**.

Does that make sense?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
CarloMariaProietti: Honestly, no AI at all was used in this code. I misunderstood https://github.com/Kotlin/dataframe/pull/1556#discussion_r2571651459,
I thought that I had to create an enum whose constants had an exact correspondance with boolean values true and false (so that the schema could remain the same) .
However, now I understood what You meant :), I proceed to correct.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
CarloMariaProietti: Yes, it would make the comparison output more independent. I can try to implement it.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
Jolanrensen: Then I take back what I said :) Thanks for the explanation!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/compareDataFrames.kt
CarloMariaProietti: > Yes, it would make the comparison output more independent. I can try to implement it.

Done, i added a  `DataRow<T>` column to  `ComparisonDescription<T> `  representing the content of the row.

CarloMariaProietti:
If there is anything I can do, I am at full disposal :)

Jolanrensen:
> If there is anything I can do, I am at full disposal :)

Sure! However, in the team we currently give more priority to fixing bugs for the 1.0 release rather than adding new features like this. This will likely be a 1.1 thing, so it has lower priority for us; we'll likely revisit it later.

However, if you want, you could explore how DataFrame users would actually use this new functionality. Your PR only provides internal implementation code, after all. Similar to `describe()`, how would users actually compare two dataframes? Both in "normal" Kotlin code and in notebooks. What would be a good API for this, and does your implementation provide all information the users would expect to be there? Would you use it yourself?

CarloMariaProietti:
> > If there is anything I can do, I am at full disposal :)
> 
> Sure! However, in the team we currently give more priority to fixing bugs for the 1.0 release rather than adding new features like this. This will likely be a 1.1 thing, so it has lower priority for us; we'll likely revisit it later.
> 
> However, if you want, you could explore how DataFrame users would actually use this new functionality. Your PR only provides internal implementation code, after all. Similar to `describe()`, how would users actually compare two dataframes? Both in "normal" Kotlin code and in notebooks. What would be a good API for this, and does your implementation provide all information the users would expect to be there? Would you use it yourself?

I think that comparing dataframes could be very usefull when the goal is to monitor something in the time.
For example: I have 2 DFs, representing my customer base, one refers to today  and the other refers to last year. In this context I would like to know the characteristics of the the clients I gained/lost and I would like the comparison to return a DF whose rows represent these clients. 
Current implementation of `compare` allows to do this by properly quering returned DF; however the only usefull columns would be  `modifiedRowContent` and `modification`.
Regarding the API, if I were an user I would like to do something like `dfRepresentingToday.compareTo(dfRepresentingLastYear)`.

Jolanrensen:
> I think that comparing dataframes could be very usefull when the goal is to monitor something in the time. For example: I have 2 DFs, representing my customer base, one refers to today and the other refers to last year. In this context I would like to know the characteristics of the the clients I gained/lost and I would like the comparison to return a DF whose rows represent these clients. Current implementation of `compare` allows to do this by properly quering returned DF; however the only usefull columns would be `modifiedRowContent` and `modification`. Regarding the API, if I were an user I would like to do something like `dfRepresentingToday.compareTo(dfRepresentingLastYear)`.

Yeah I think that makes sense :) maybe give it a try with a draft API and a notebook using it (you can publish DF to maven Local and run it in a notebook using:
```kts
USE { repositories(mavenLocal()) }
%use dataframe(v=1.0.0-dev)
```
)

PR #1558, opened 2025-11-11
 Replace `File` with `Path`

### LINKED ISSUE #527: Replace io.File on nio.Path in DataFrame.read methods
Now we are using File class from java.io in our API for reading DataFrames instead of Path from java.nio

For example

```kotlin
public interface SupportedDataFrameFormat : SupportedFormat {
    public fun readDataFrame(stream: InputStream, header: List<String> = emptyList()): DataFrame<*>

    public fun readDataFrame(file: File, header: List<String> = emptyList()): DataFrame<*>
}
```

Both of the classes are adopted in Kotlin with some of the extension functions, but usage of Path is more modern and highly-recommended for library creators.

I suggest replacing Path to File or temporarily add new method with Path for 1-2 releases to be aligned with modern recommendation 

I want to add two additional links
https://discuss.kotlinlang.org/t/file-vs-path-in-kotlin/26349
https://www.baeldung.com/java-path-vs-file

AndreiKingsley:
closes #527.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/guess.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/guess.kt
@@ -16,6 +16,7 @@ import java.io.File
 import java.io.FileNotFoundException
 import java.io.InputStream
 import java.net.URL
+import java.nio.file.Path
 import java.util.ServiceLoader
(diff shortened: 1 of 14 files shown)


Jolanrensen:
best to have both File and Path, like in `readCsv()`

PR #1554, opened 2025-11-07
Pivot kdocs

AndreiKingsley:
Closes #1525 
Closes #1552

PR #1561, opened 2025-11-12
Add documentation and tests for the `last` and `lastOrNull` functions

### LINKED ISSUE #1497: Add unit tests and documentation for `last.kt` function
This file contains a functions which are not covered by unit tests and highly likely has no documentation

Allex-Nik:
Fixes #1497

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/last.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/last.kt
@@ -108,7 +108,7 @@ public inline fun <T> DataColumn<T>.lastOrNull(predicate: (T) -> Boolean): T? = 
  * Returns `null` if the [DataFrame] contains no rows matching the [predicate]
  * (including the case when the [DataFrame] is empty).
  *
- * {@include [RowFilterDescription]}
+ * @include [RowFilterDescription]
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
  *
@@ -138,7 +138,7 @@ public inline fun <T> DataFrame<T>.lastOrNull(predicate: RowFilter<T>): DataRow<
 /**
  * Returns the last [row][DataRow] in this [DataFrame] that satisfies the given [predicate].
  *
- * {@include [RowFilterDescription]}
+ * @include [RowFilterDescription]
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
  *
@@ -237,7 +237,7 @@ public fun <T, G> GroupBy<T, G>.last(): ReducedGroupBy<T, G> = reduce { lastOrNu
  * the corresponding row in [ReducedGroupBy] will contain `null` values for all columns in the group,
  * except the grouping key.
  *
- * {@include [RowFilterDescription]}
+ * @include [RowFilterDescription]
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
  *
@@ -289,7 +289,7 @@ public fun <T> Pivot<T>.last(): ReducedPivot<T> = reduce { lastOrNull() }
  *
  * For more information about [Pivot] with examples: {@include [DocumentationUrls.Pivot]}
  *
- * {@include [RowFilterDescription]}
+ * @include [RowFilterDescription]
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
  *
@@ -352,13 +352,13 @@ public fun <T> PivotGroupBy<T>.last(): ReducedPivotGroupBy<T> = reduce { lastOrN
  * in the resulting [ReducedPivotGroupBy] it will be represented by a [row][DataRow] with `null` values
  * (except the grouping key).
  *
- * {@include [DocumentationUrls.PivotGroupBy]}
+ * @include [DocumentationUrls.PivotGroupBy]
  *
- * {@include [DocumentationUrls.Pivot]}
+ * @include [DocumentationUrls.Pivot]
  *
- * {@include [DocumentationUrls.GroupBy]}
+ * @include [DocumentationUrls.GroupBy]
  *
- * {@include [RowFilterDescription]}
+ * @include [RowFilterDescription]
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
  *


Allex-Nik:
The checks fail for now because in the documentation I refer to `[DocumentationUrls.Pivot]` which is added by @AndreiKingsley in the PR #1554  but not merged yet. Is it fine to have it in this state until #1554 is merged, or should I delete `[DocumentationUrls.Pivot]` for now and add it later?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/last.kt
```diff
@@ -1,12 +1,37 @@
...
+ * - **GroupBy**: reducing each group to its last row or the last row matching a predicate,
+ * with handling groups that contain no matching rows.
+ * - **Pivot**: reducing each group in the pivot to its last row or the last row matching a predicate,
+ * with handling groups that contain no matching rows.
+ * - **PivotGroupBy**: reducing each combined [pivot] + [groupBy] group to its last row
+ * or the last row matching a predicate, with handling [pivot] + [groupBy] combinations that contain no matching rows.
+ */
 class LastTests : ColumnsSelectionDslTests() {
+
+    // region ColumnsSelectionDsl
```
Allex-Nik, line 43: I was thinking that a test file could benefit from these regions as well. Is it a good idea to add it here?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/last.kt
```diff
@@ -60,26 +188,186 @@ public fun <T> DataFrame<T>.last(): DataRow<T> {
 
 // region GroupBy
 
+/**
+ * Gets the last [row][DataRow] from each group of the given [GroupBy]
+ * and returns a [ReducedGroupBy] containing these rows
+ * (one row per group, each row is the last row in its group).
+ *
+ * If the group in [GroupBy] is empty,
```
Allex-Nik: I mentioned that it will return null for empty groups, but there is still the issue I mentioned for this case in #1547:
https://github.com/Kotlin/dataframe/pull/1547/files#r2505104735

Jolanrensen:
> The checks fail for now because in the documentation I refer to `[DocumentationUrls.Pivot]` which is added by @AndreiKingsley in the PR #1554 but not merged yet. Is it fine to have it in this state until #1554 is merged, or should I delete `[DocumentationUrls.Pivot]` for now and add it later?

I expect it to be merged soon. If it takes too long, you can also copy the code from #1554 and let git handle the merge. Whatever you like best :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/last.kt
Jolanrensen: could you create an issue for that? I probably should have done that myself when finding the reproducer... But it makes it easier to track and refer to :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/last.kt
```diff
@@ -27,28 +30,153 @@ import kotlin.reflect.KProperty
 
 // region DataColumn
 
+/**
+ * Returns the last value in this [DataColumn].
+ *
+ * See also [lastOrNull], [first], [take], [takeLast].
```
Jolanrensen, line 36: There's also the `@see` tag if you just want to refer to other functions.

The only downside of `@see` is that you cannot add "extra" text to it. That's why we often write things like "See [something] for some reason."

However, if you don't need a description, you can simply write
```
@see [lastOrNull]
@see [first]
```
etc.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/last.kt
Allex-Nik: Yes, sure! I probably should have done it right away, I just wasn't 100% sure :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/last.kt
Allex-Nik: Does it make sense to add a receiver (for all cases other than DataColumn)? For example, `@see [DataFrame.lastOrNull]` in region DataFrame. Otherwise the link always leads to the version on DataColumn

AndreiKingsley (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/last.kt
```diff
@@ -60,26 +202,187 @@ public fun <T> DataFrame<T>.last(): DataRow<T> {
...
+/**
+ * [Reduces][PivotGroupByDocs.Reducing] this [PivotGroupBy]
+ * by taking from each combined [pivot] + [groupBy] group the last [row][DataRow] satisfying the given [predicate].
+ * Returns a [ReducedPivotGroupBy] that contains the last [row][DataRow], matching the [predicate],
+ * from each corresponding group.
+ * If any combined [pivot] + [groupBy] group in [PivotGroupBy] does not contain any rows matching the [predicate],
+ * in the resulting [ReducedPivotGroupBy] it will be represented by a [row][DataRow] with `null` values
+ * (except the grouping key).
+ *
+ * {@include [DocumentationUrls.PivotGroupBy]}
```
Jolanrensen: It doesn't matter much, but you can put all these `include`s directly below each other without brackets. KoDEx should handle it well, it's just as clear, and it's a bit more compact :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/last.kt
```diff
@@ -60,26 +202,187 @@ public fun <T> DataFrame<T>.last(): DataRow<T> {
...
 public fun <T> Pivot<T>.last(): ReducedPivot<T> = reduce { lastOrNull() }
 
+/**
+ * [Reduces][PivotDocs.Reducing] this [Pivot] by taking from each group the last [row][DataRow]
+ * satisfying the given [predicate], and returns a [ReducedPivot] that contains the last [row][DataRow],
+ * matching the [predicate], from the corresponding group in each column.
+ *
+ * For more information about [Pivot] with examples: {@include [DocumentationUrls.Pivot]}
+ *
+ * {@include [RowFilterDescription]}
```
Jolanrensen: no need to add brackets. Those are only needed if you want to write an @ statement after some other text.

Jolanrensen: same in other places.

Jolanrensen: It doesn't hurt, but it's best to be consistent

PR #1572, opened 2025-11-13
Fixes pivot shortcuts to use `PivotDsl`

### LINKED ISSUE #1548: Fix `pivot` functions API
Now most of `pviot`, `pivotMatches`, `pivotCounts` overloads use common `ColumnsSelector` instead of `PivotDsl` columns selector.

Jolanrensen:
Fixes #1548 by making all pivot shortcuts use `PivotDsl`. I believe this will not cause issues as `PivotDsl : ColumnsSelectionDsl`, and all shortcuts simply call `pivot {}.operation()`.


AndreiKingsley (changes requested):
Please, add 
```
Select or express pivot columns using the [PivotDsl].

@include [PivotDslDocs]
```
 in these overloads.
 
 Also there's a mistake in `columns` argument description in 939th line.

PR #1573, opened 2025-11-14
Fixing bug with extracting metadata for many columns

### LINKED ISSUE #1351: Teradata JDBC Performance Problem during reading data from the DatabaseMetaData
Read metadata from JDBC ResultSet.metaData instead of DatabaseMetaData (or allow both) in readSqlQuery()

- to solve Teradata-JDBC performance problem
- to collect metaData for simple sql joins for Teradata-JDBC (and other JDBC drivers?)

(I am new here and this is my first issue. 
I hope it helps and you can use my solution.)

My problem:

`fun getTableColumnsMetadata()` in `readJdbc.kt` reads the metadata from DatabaseMetaData for each column.
This causes Teradata-JDBC to execute an extremely slow query for each column.
Ultimately, this leads to a timeout when there are too many columns.
Teradata's DatabaseMetaData.getColumns() is useful only 
for collecting metadata for a set of columns in one call.

Here are the details of my setup:

I tested DataFrame-Beta2 with Teradata JDBC: 20.00.00.46.
As usual, I created a custom DbType object.

```kotlin
    public object MyTeradb : DbType("teradata") {
            get() = "com.teradata.jdbc.TeraDriver"
    
    // ...  dummies from the example for HSQLDB
    
    }
```

I tried reading data from an SQL table with about 90 columns.

```kotlin
    val teraSql = "select top 100 * from my_tab_with90cols"
    val df = DataFrame.readSqlQuery( conn, teraSql, MyTeradb, strictValidation = false )
```

My solution:

I created a modified version of `fun getTableColumnsMetadata()`,
so that it reads metadata from the ResultSet object,
This works perfectly for Teradata JDBC 20.00.00.46.

I commented out the old code so you can compare the two versions.

As you can see, only the `isNullable` flag is collected
from DatabaseMetaData in the old version.

The isNullable information is available in `ResultSetMetaData`

```java
//java.sql.ResultSetMetaData
public abstract int isNullable(int column);
```

I used that function to retrieve the isNullable information:

```kotlin
/**
 * Retrieves the metadata of the columns in the result set.
 * 
 * metadata are collected from resultSet.metaData 
 * to avoid DataBaseMetaData-queries per column
 * (queries per column are a problem for teradata-jdbc e.g.)
 *
 * @param rs the result set
 * @return a mutable list of [TableColumnMetadata] objects,
 *         where each TableColumnMetadata object contains information such as the column type,
 *         JDBC type, size, and name.
 */
private fun getTableColumnsMetadata(rs: ResultSet): MutableList<TableColumnMetadata> {
    val metaData: ResultSetMetaData = rs.metaData
    val numberOfColumns: Int = metaData.columnCount
    val tableColumns = mutableListOf<TableColumnMetadata>()
    val columnNameCounter = mutableMapOf<String, Int>()
    //val databaseMetaData: DatabaseMetaData = rs.statement.connection.metaData
    //val catalog: String? = rs.statement.connection.catalog.takeUnless { it.isNullOrBlank() }
    //val schema: String? = rs.statement.connection.schema.takeUnless { it.isNullOrBlank() }

    for (i in 1 until numberOfColumns + 1) {
        //val tableName = metaData.getTableName(i)
        val columnName = metaData.getColumnName(i)

        // orig versions queries databaseMetaData to get isNullable.
        // this does not work on teradata for tables with many columns.

        //  // this algorithm works correctly only for SQL Table and ResultSet opened on one SQL table
        //val columnResultSet: ResultSet =
        //    databaseMetaData.getColumns(catalog, schema, tableName, columnName)
        //val isNullable = if (columnResultSet.next()) {
        //    columnResultSet.getString("IS_NULLABLE") == "YES"
        //} else {
        //    true // we assume that it's nullable by default
        //}
        
        // for teradata-jdbc: use isNullable() from metaData
        val isNullable = when (metaData.isNullable(i)) {
            ResultSetMetaData.columnNullable -> true
            ResultSetMetaData.columnNullableUnknown -> true
            ResultSetMetaData.columnNoNulls -> false
            else -> true // default to nullable
        }


        val name = manageColumnNameDuplication(columnNameCounter, columnName)
        val size = metaData.getColumnDisplaySize(i)
        val type = metaData.getColumnTypeName(i)
        val jdbcType = metaData.getColumnType(i)
        val javaClassName = metaData.getColumnClassName(i)

        tableColumns += TableColumnMetadata(name, type, jdbcType, size, javaClassName, isNullable)
    }
    return tableColumns
}
```

This modification works perfectly with my Teradata JDBC setup.
Now, I can execute simple queries or joins 
and collect the data in a data frame.

(Teradata-JDBC reports the nullable information correctly in joins if there are no subqueries.)

However, there may be side effects with other databases
if they don't support ResultSetMetaData.isNullable(int column)
as needed.

One option is to introduce a boolean flag like
`useMetaDataFromResultSet = false` in `DataFrame.readSqlQuery(...)`
to maintain the previous version.

zaleslaw:
Fixes #1351 

In this PR - I have just removed and cleaned function from one class to another, giving user ability to override it, because it's a vendor - specific behaviour

### DIFF
--- a/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
+++ b/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
@@ -412,9 +412,9 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
      * where `DatabaseMetaData.getColumns` is known to be slow).
      *
      * @param resultSet The [ResultSet] containing query results.
-     * @return A mutable list of [TableColumnMetadata] objects.
+     * @return A list of [TableColumnMetadata] objects.
      */
-    public open fun getTableColumnsMetadata(resultSet: ResultSet): MutableList<TableColumnMetadata> {
+    public open fun getTableColumnsMetadata(resultSet: ResultSet): List<TableColumnMetadata> {
         val rsMetaData = resultSet.metaData
         val connection = resultSet.statement.connection
         val dbMetaData = connection.metaData
@@ -446,8 +446,9 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
             } catch (_: Exception) {
                 // Some drivers may throw for unsupported features
                 // In that case, fallback to DatabaseMetaData
-                val cols = dbMetaData.getColumns(catalog, schema, tableName, columnName)
-                if (cols.next()) cols.getString("IS_NULLABLE") == "YES" else true
+                dbMetaData.getColumns(catalog, schema, tableName, columnName).use { cols ->
+                    if (cols.next()) !cols.getString("IS_NULLABLE").equals("NO", ignoreCase = true) else true
+                }
             }
 
             val columnType = rsMetaData.getColumnTypeName(index)
--- a/dataframe-jdbc/api/dataframe-jdbc.api
+++ b/dataframe-jdbc/api/dataframe-jdbc.api
@@ -109,6 +109,7 @@ public abstract class org/jetbrains/kotlinx/dataframe/io/db/DbType {
 	public fun getDefaultFetchSize ()I
 	public fun getDefaultQueryTimeout ()Ljava/lang/Integer;
 	public abstract fun getDriverClassName ()Ljava/lang/String;
+	public fun getTableColumnsMetadata (Ljava/sql/ResultSet;)Ljava/util/List;
 	public fun getTableTypes ()Ljava/util/List;
 	public abstract fun isSystemTable (Lorg/jetbrains/kotlinx/dataframe/io/db/TableMetadata;)Z
 	public fun makeCommonSqlToKTypeMapping (Lorg/jetbrains/kotlinx/dataframe/io/db/TableColumnMetadata;)Lkotlin/reflect/KType;
--- a/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
+++ b/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
@@ -852,7 +852,7 @@ private fun readTableAsDataFrame(
 }
 
 internal fun getTableColumnsMetadata(resultSet: ResultSet, dbType: DbType): MutableList<TableColumnMetadata> =
-    dbType.getTableColumnsMetadata(resultSet)
+    dbType.getTableColumnsMetadata(resultSet).toMutableList()
 
 /**
  * Fetches and converts data from a ResultSet into a mutable map.


dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -397,4 +398,100 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
...
+     * it falls back to using [DatabaseMetaData.getColumns] for affected columns.
+     *
+     * Override this method in subclasses to provide database-specific behavior
+     * (for example, to disable fallback for databases like Teradata or Oracle
+     * where `DatabaseMetaData.getColumns` is known to be slow).
+     *
+     * @param resultSet The [ResultSet] containing query results.
+     * @return A mutable list of [TableColumnMetadata] objects.
+     */
+    public open fun getTableColumnsMetadata(resultSet: ResultSet): MutableList<TableColumnMetadata> {
```
Jolanrensen: Still unsure about the mutable list, but we'll see

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -397,4 +398,100 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
...
+        val connection = resultSet.statement.connection
+        val dbMetaData = connection.metaData
+        val catalog = connection.catalog.takeUnless { it.isNullOrBlank() }
+        val schema = connection.schema.takeUnless { it.isNullOrBlank() }
+
+        val columnCount = rsMetaData.columnCount
+        val columns = mutableListOf<TableColumnMetadata>()
+        val nameCounter = mutableMapOf<String, Int>()
+
+        for (index in 1..columnCount) {
```
Jolanrensen, line 428: It's 1-indexed? Interesting

Jolanrensen: But same logic as the old one so should be fine

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -397,4 +398,100 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
...
+                        true
+                    }
+
+                    else -> true
+                }
+            } catch (_: Exception) {
+                // Some drivers may throw for unsupported features
+                // In that case, fallback to DatabaseMetaData
+                val cols = dbMetaData.getColumns(catalog, schema, tableName, columnName)
+                if (cols.next()) cols.getString("IS_NULLABLE") == "YES" else true
```
Jolanrensen: I had to read this line a couple times. 
Since we default to "true", maybe it's clearer to invert the statement to 

```kt
if (cols.next()) cols.getString("IS_NULLABLE") != "NO" // I assume getString returns null when there is no such column? In that case we should also return true 
else true
```

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
zaleslaw: I suppose you are right here

PR #1574, opened 2025-11-18
Join kdocs

### LINKED ISSUE #1387: Add KDocs for `Join` and `JoinWith`
https://github.com/Kotlin/dataframe/blob/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
https://github.com/Kotlin/dataframe/blob/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt

- [ ] Also, please check coverage for Joins and Joins DSL in Unit-test, test missed cases

AndreiKingsley:
Closes #1387.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
@@ -66,9 +66,9 @@ private interface JoinDocs
 private interface JOIN_METHOD
 
 /**
- * [JoinDsl] defines the columns used for joining [DataFrame]s
+ * [JoinDsl] allows you to define the columns used for joining [DataFrame]s
  * and provides methods to match columns with different names
- * between the left and right sides.
+ * between the left and right side.
  *
  * Provides the left [DataFrame] both as the receiver (`this`) and as the argument (`it`),
  * allowing you to reference its columns directly.
@@ -85,6 +85,9 @@ private interface JOIN_METHOD
  * // Join by one column with different names —
  * // "firstName" in the left dataframe and "name" in the right one
  * dfLeft.{@get [JoinMethod] join}(dfRight) { left -> left.firstName match right.name }
+ *
+ * // Match columns using String API
+ * dfLeft.{@get [JoinMethod] join}(dfRight) { "symbol" match right.getValue<Char>("char") }
  * ```
  */
 @ExcludeFromSources
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
@@ -88,7 +88,7 @@ public typealias JoinExpression<A, B> = Selector<JoinedDataRow<A, B>, Boolean>
  * Each join type has a corresponding shortcut function:
  * [innerJoinWith], [leftJoinWith], [rightJoinWith], [fullJoinWith], [filterJoinWith], and [excludeJoinWith].
  *
- * See also [join], which performs a join by exact values equality in the selected columns.
+ * See also [join], which performs a join by exact value equality in the selected columns.
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
  *
@@ -119,7 +119,7 @@ public fun <A, B> DataFrame<A>.joinWith(
  * See also general [joinWith] as well as other shortcuts with each of join types:
  * [leftJoinWith], [rightJoinWith], [fullJoinWith], [filterJoinWith], [excludeJoinWith].
  *
- * See also [join], which performs a join by simply matching columns.
+ * See also [join], which performs a join by exact value equality in the selected columns.
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
  *
@@ -146,7 +146,7 @@ public fun <A, B> DataFrame<A>.innerJoinWith(right: DataFrame<B>, joinExpression
  * See also general [joinWith] as well as other shortcuts with each of join types:
  * [innerJoinWith], [rightJoinWith], [fullJoinWith], [filterJoinWith], [excludeJoinWith].
  *
- * See also [join], which performs a join by simply matching columns.
+ * See also [join], which performs a join by exact value equality in the selected columns.
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
  *
@@ -173,7 +173,7 @@ public fun <A, B> DataFrame<A>.leftJoinWith(right: DataFrame<B>, joinExpression:
  * See also general [joinWith] as well as other shortcuts with each of join types:
  * [innerJoinWith], [leftJoinWith], [fullJoinWith], [filterJoinWith], [excludeJoinWith].
  *
- * See also [join], which performs a join by simply matching columns.
+ * See also [join], which performs a join by exact value equality in the selected columns.
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
  *
@@ -200,7 +200,7 @@ public fun <A, B> DataFrame<A>.rightJoinWith(right: DataFrame<B>, joinExpression
  * See also general [joinWith] as well as other shortcuts with each of join types:
  * [leftJoinWith], [rightJoinWith], [innerJoinWith], [filterJoinWith], [excludeJoinWith].
  *
- * See also [join], which performs a join by simply matching columns.
+ * See also [join], which performs a join by exact value equality in the selected columns.
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
  *
@@ -227,7 +227,7 @@ public fun <A, B> DataFrame<A>.fullJoinWith(right: DataFrame<B>, joinExpression:
  * See also general [joinWith] as well as other shortcuts with each of join types:
  * [leftJoinWith], [rightJoinWith], [fullJoinWith], [innerJoinWith], [excludeJoinWith].
  *
- * See also [join], which performs a join by simply matching columns.
+ * See also [join], which performs a join by exact value equality in the selected columns.
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
  *
@@ -254,7 +254,7 @@ public fun <A, B> DataFrame<A>.filterJoinWith(right: DataFrame<B>, joinExpressio
  * See also general [joinWith] as well as other shortcuts with each of join types:
  * [leftJoinWith], [rightJoinWith], [fullJoinWith], [filterJoinWith], [innerJoinWith].
  *
- * See also [join], which performs a join by simply matching columns.
+ * See also [join], which performs a join by exact value equality in the selected columns.
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
  *


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -12,13 +13,95 @@ import org.jetbrains.kotlinx.dataframe.columns.ColumnWithPath
...
 import kotlin.reflect.KProperty
 
+/**
+ * If no join columns are specified, all columns with matching names in both [DataFrame]s are used.
+ *
+ * If both [DataFrame]s contain columns with the same name that are *not* part of the join keys,
+ * such columns are treated as distinct. Such a column from the right [DataFrame] will be automatically
+ * renamed in the resulting [DataFrame].
+ */
+private interface JoinBehavior
```
Jolanrensen, line 35: `@ExcludeFromSources`, check other places too :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -12,13 +13,95 @@ import org.jetbrains.kotlinx.dataframe.columns.ColumnWithPath
...
 import org.jetbrains.kotlinx.dataframe.impl.columns.ColumnListImpl
 import org.jetbrains.kotlinx.dataframe.util.DEPRECATED_ACCESS_API
 import kotlin.reflect.KProperty
 
+/**
+ * If no join columns are specified, all columns with matching names in both [DataFrame]s are used.
+ *
+ * If both [DataFrame]s contain columns with the same name that are *not* part of the join keys,
+ * such columns are treated as distinct. Such a column from the right [DataFrame] will be automatically
+ * renamed in the resulting [DataFrame].
```
Jolanrensen: renamed how?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -12,13 +13,95 @@ import org.jetbrains.kotlinx.dataframe.columns.ColumnWithPath
...
+ */
+private interface JoinBehavior
+
+/**
+ * Joins this [DataFrame] with [other][\other] [DataFrame] using selected key columns.
```
Jolanrensen: *the other, the selected key columns

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -162,30 +557,99 @@ internal class ColumnMatchImpl<C>(override val left: ColumnReference<C>, overrid
...
+ * * [JoinType.Inner] (default) — {@include [JoinType.Inner]}
+ * * [JoinType.Left] — {@include [JoinType.Left]}
+ * * [JoinType.Right] — {@include [JoinType.Right]}
+ * * [JoinType.Full] — {@include [JoinType.Full]}
+ *
+ * **Non-merging joins:**
+ * * [JoinType.Filter] — {@include [JoinType.Filter]}
+ * * [JoinType.Exclude] — {@include [JoinType.Exclude]}
+ */
+internal interface JoinTypeDescription
```
Jolanrensen, line 686: helpful!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -12,13 +13,95 @@ import org.jetbrains.kotlinx.dataframe.columns.ColumnWithPath
...
+ * See [Selecting Columns][SelectingColumns].
+ *
+ * For more information, {@include [DocumentationUrls.Join]}.
+ *
+ * ### This `join` overload
+ */
+private interface JoinDocs
+
+// `join` method used in the example
+private interface JoinMethod
```
Jolanrensen: *JOIN_METHOD

Jolanrensen: and can be excluded from sources as well

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -12,13 +13,95 @@ import org.jetbrains.kotlinx.dataframe.columns.ColumnWithPath
...
+
+/**
+ * [JoinDsl] defines the columns used for joining [DataFrame]s
```
Jolanrensen: *allows you to define the columns...

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -12,13 +13,95 @@ import org.jetbrains.kotlinx.dataframe.columns.ColumnWithPath
...
+ * and provides methods to match columns with different names
+ * between the left and right sides.
```
Jolanrensen: *side

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -12,13 +13,95 @@ import org.jetbrains.kotlinx.dataframe.columns.ColumnWithPath
...
+ * // Join by one column with different names —
+ * // "firstName" in the left dataframe and "name" in the right one
+ * dfLeft.{@get [JoinMethod] join}(dfRight) { left -> left.firstName match right.name }
+ * ```
+ */
+@ExcludeFromSources
+internal interface JoinDslDescription
+
+/**
+ * Select join columns (incl. with different names in this and [other][other] [DataFrame]s)
```
Jolanrensen: incl. means "included"? But then it says "included with". Shouldn't it then be just "including different names in this..." or something like that?

Jolanrensen: "this" can get []

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -27,89 +110,386 @@ public fun <A, B> DataFrame<A>.join(
...
+/**
+ * Performs [inner join][JoinType.Inner] of this [DataFrame] with [other][\other] [DataFrame]
+ * using selected key columns.
+ * @include [JoinType.Inner]
+ *
+ * This is a shortcut for [join] with [JoinType.Inner].
+ *
+ * @include [JoinBehavior]
+ *
+ * See also general [join] as well as other shortcuts with each of join types:
```
Jolanrensen: general join, as well

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -12,13 +13,95 @@ import org.jetbrains.kotlinx.dataframe.columns.ColumnWithPath
...
+ * See also [Columns selection via DSL][SelectingColumns.Dsl].
+ *
+ * ### Examples
+ * ```kotlin
+ * // Join by two columns with the same names in both dataframes
+ * dfLeft.{@get [JoinMethod] join}(dfRight) { name and city }
+ *
+ * // Join by one column with different names —
+ * // "firstName" in the left dataframe and "name" in the right one
+ * dfLeft.{@get [JoinMethod] join}(dfRight) { left -> left.firstName match right.name }
```
Jolanrensen, line 87: could you add an example of using Strings to refer to columns inside this DSL? It would be useful for people to know they can use `match` using the String API as well

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -27,89 +110,386 @@ public fun <A, B> DataFrame<A>.join(
...
+ * Joins all [DataFrame]s in this iterable into a single [DataFrame].
+ *
+ * Sequentially applies the [join] operation to each [DataFrame] in order.
+ * Returns `null` if the iterable is empty.
+ *
+ * @param [joinType] [JoinType] defining how rows are matched and combined.
+ * @param [selector] optional [JoinColumnsSelector] specifying key columns.
+ * @return resulting [DataFrame], or `null` if the iterable is empty.
+ */
 public fun <T> Iterable<DataFrame<T>>.joinOrNull(
```
Jolanrensen, line 451: I wonder why we only have `.joinOrNull()` and not just `.join()` that throws an exception when the iterable is empty...

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -27,89 +110,386 @@ public fun <A, B> DataFrame<A>.join(
...
+     * @receiver column from the left [DataFrame].
+     * @param [other] column from the right [DataFrame].
+     * @return [ColumnMatch] representing the column pair used for joining.
+     */
+    @ExcludeFromSources
+    private interface MatchDocs
+
+    /**
+     * @include [MatchDocs]
+     */
```
Jolanrensen: You can inline small KDocs if you like:
`/** @include [MatchDocs] */`

It works exactly the same, (KoDEx formats resulting KDoc automatically) it's just a bit more compact in the source

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -150,6 +542,9 @@ public interface JoinDsl<out A, out B> : ColumnsSelectionDsl<A> {
     }
 }
 
+/**
+ * A special [ColumnSet] that specifies [column match][JoinDsl.match] for [join] operation.
```
Jolanrensen: *the join operation

Jolanrensen: *a column match

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -162,30 +557,99 @@ internal class ColumnMatchImpl<C>(override val left: ColumnReference<C>, overrid
         throw UnsupportedOperationException()
 }
 
+/**
+ * Creates a [ColumnMatch].
+ *
+ * Not intended for public API consumption. Please, use [match][JoinDsl.match] instead.
```
Jolanrensen: Image lol

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -162,30 +557,99 @@ internal class ColumnMatchImpl<C>(override val left: ColumnReference<C>, overrid
...
+ */
 public fun <C> ColumnMatch(left: ColumnReference<C>, right: ColumnReference<C>): ColumnMatch<C> =
```
Jolanrensen, line 567: Hmm I wonder if we could hide this somehow. This does pollute the public scope

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
```diff
@@ -5,14 +5,96 @@ import org.jetbrains.kotlinx.dataframe.DataRow
 import org.jetbrains.kotlinx.dataframe.Selector
 import org.jetbrains.kotlinx.dataframe.annotations.Interpretable
 import org.jetbrains.kotlinx.dataframe.annotations.Refine
+import org.jetbrains.kotlinx.dataframe.documentation.DocumentationUrls
+import org.jetbrains.kotlinx.dataframe.documentation.DocumentationUrls.Join
```
Jolanrensen: unused imports

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
```diff
@@ -5,14 +5,96 @@ import org.jetbrains.kotlinx.dataframe.DataRow
...
+ * rows should be matched based on custom logic rather than simple equality.
+ *
+ * Creates a new [DataFrame] by combining [rows][DataRow]
+ * from both inputs according to the [\joinExpression] matching rule.
+ */
+@ExcludeFromSources
+private interface JoinWithCommonDescription
+
+// `joinWith` method used in the example
+private interface JoinWithMethod
```
Jolanrensen: *JOIN_WITH_METHOD and exclude from sources in multiple places

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
```diff
@@ -5,14 +5,96 @@ import org.jetbrains.kotlinx.dataframe.DataRow
...
+import org.jetbrains.kotlinx.dataframe.documentation.ExcludeFromSources
+import org.jetbrains.kotlinx.dataframe.documentation.SelectingColumns
+import org.jetbrains.kotlinx.dataframe.documentation.SelectingColumns.ColumnGroupsAndNestedColumnsMention
 import org.jetbrains.kotlinx.dataframe.impl.api.joinWithImpl
 
+/**
+ * A [JoinExpression] defines the matching condition between [rows][DataRow] of the two [DataFrame]s.
+ * It provides access to row values from both the left and right [DataFrame]s
+ * and expects a [Boolean] result indicating whether the rows match.
+ * All combinations of rows from the left and right [DataFrame]s that satisfies
```
Jolanrensen: No "s" needed. It's an ellipse of "the left DataFrame and the right DataFrame" -> "the left- and the right DataFrame", so singular

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
```diff
@@ -5,14 +5,96 @@ import org.jetbrains.kotlinx.dataframe.DataRow
...
+ * this condition are matched.
+ *
+ * This method is useful when the data was not originally designed relationally, or when
```
Jolanrensen: "not originally designed relationally" the what? could you simplify this sentence? haha

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
```diff
@@ -5,14 +5,96 @@ import org.jetbrains.kotlinx.dataframe.DataRow
...
+ * ### Examples
+ * ```kotlin
+ * // Join rows where the `fullName` value in the left DataFrame
+ * // contains the `firstName` value in the right DataFrame.
+ * dfLeft.{@get [JoinWithMethod] joinWith}(dfRight) { left -> left.fullName.contains(right.firstName) }
+ *
+ * // Join rows where the `date` value in the right DataFrame
+ * // falls within the interval defined by the `startDate` and `endDate`
+ * // values in the left DataFrame.
+ * dfLeft.{@get [JoinWithMethod] joinWith}(dfRight) { right.date in startDate..endDate }
```
Jolanrensen, line 44: same as the other example, can you add another with Strings?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
Jolanrensen: come to think of it, `pivot` could probably also use this well

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
```diff
@@ -5,14 +5,96 @@ import org.jetbrains.kotlinx.dataframe.DataRow
...
+/**
+ * A special [row][DataRow] expression used to define
+ * the row-matching condition in a [joinWith] operation.
+ *
+ * Provides the [row][DataRow] of the left [DataFrame] both
+ * as the receiver (`this`) and as the argument (`it`),
+ * allowing you to reference its values directly.
+ *
+ * The [row][DataRow] of the right [DataFrame] is available
+ * via [right][JoinedDataRow.right].
```
Jolanrensen: *as

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
```diff
@@ -5,14 +5,96 @@ import org.jetbrains.kotlinx.dataframe.DataRow
...
+ * using the provided [\joinExpression].
+ *
+ * @include [JoinWithCommonDescription]
+ *
+ * {@include [JoinTypeDescription]}
+ *
+ * Each join type has a corresponding shortcut function:
+ * [innerJoinWith], [leftJoinWith], [rightJoinWith], [fullJoinWith], [filterJoinWith], and [excludeJoinWith].
+ *
+ * See also [join], which performs a join by simply matching columns.
```
Jolanrensen: "by exact values" or something like that. Emphasizes the difference better

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
```diff
@@ -5,14 +5,96 @@ import org.jetbrains.kotlinx.dataframe.DataRow
...
+ *
+ * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
+ *
+ * For more information, {@include [DocumentationUrls.JoinWith]}.
+ *
+ * @include [JoinWithExample]
+ * @param [right] [DataFrame] to join with.
+ * @param [type] [JoinType] defining how rows are matched and combined.
+ * @param [joinExpression] [JoinExpression] specifying rows join condition.
```
Jolanrensen: *the

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
```diff
@@ -21,31 +103,163 @@ public fun <A, B> DataFrame<A>.joinWith(
...
+ * @return joined [DataFrame].
+ */
 @Refine
 @Interpretable("FullJoinWith")
 public fun <A, B> DataFrame<A>.fullJoinWith(right: DataFrame<B>, joinExpression: JoinExpression<A, B>): DataFrame<A> =
     joinWith(right, JoinType.Full, joinExpression)
 
+/**
+ * Performs [filter join][JoinType.Filter] of this [DataFrame] with the [right][\right] [DataFrame]
```
Jolanrensen: I'm wondering... Is "Performs **a** filter join of this DataFrame with the right DataFrame" better? Depends whether we think of "join" as a special concept or as a generic action. You can say "I performed a cleanup", but at the same time "I performed `gradlew clean`".

AndreiKingsley: Actually, a good question!
I believe you’re right — “performs a join” is better since join is used here as a technical operation rather than just an action.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
AndreiKingsley: Oh, we need an explanation of auto renaming principles. Add it.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
AndreiKingsley: [this] doesn't refer to anything!

AndreiKingsley: it means "columns, including those that have different names in different dfs", there's a difference with a common `ColumnsSelectionDsl`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
AndreiKingsley: We could add it!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
AndreiKingsley: Does compiler plugin use it?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
AndreiKingsley: Oh, it looks really awful in this DSL

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
Jolanrensen: `[this\]` does refer to the receiver of the `join` function when this KDoc is included there

Jolanrensen: ah okay, please rewrite it like that :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
Jolanrensen: @koperagen ?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
Jolanrensen: still, it's important for people to see it's possible

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
AndreiKingsley: image

No, it doesn't!

AndreiKingsley: image

`[this]` doesn't refer to the reciever

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -162,30 +555,136 @@ internal class ColumnMatchImpl<C>(override val left: ColumnReference<C>, overrid
...
 public enum class JoinType {
-    Left, // all data from left dataframe, nulls for mismatches in right dataframe
-    Right, // all data from right dataframe, nulls for mismatches in left dataframe
-    Inner, // only matched data from right and left dataframe
-    Filter, // only matched data from left dataframe
-    Full, // all data from left and from right dataframe, nulls for any mismatches
-    Exclude, // mismatched rows from left dataframe
+
+    /**
+     * @include [LeftJoinTypeDocs]
```
Jolanrensen: KoDEx cannot touch KDocs on enum entries https://github.com/Jolanrensen/KoDEx/issues/93

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -27,89 +116,378 @@ public fun <A, B> DataFrame<A>.join(
...
+ * @include [JoinStringApiExample] {@set [JOIN_METHOD] filterJoin}
+ * @param other [DataFrame] to join with.
+ * @param columns [Column Names][String] specifying join columns.
+ * @return joined [DataFrame].
+ */
 public fun <A, B> DataFrame<A>.filterJoin(other: DataFrame<B>, vararg columns: String): DataFrame<A> =
     filterJoin(other) { columns.toColumnSet() }
 
+/**
+ * Performs a [exclude join][JoinType.Exclude] of this [DataFrame] with [other][\other] [DataFrame]
```
Jolanrensen: *an

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -27,89 +116,378 @@ public fun <A, B> DataFrame<A>.join(
...
+ * @return joined [DataFrame].
+ */
 public fun <A, B> DataFrame<A>.join(
     other: DataFrame<B>,
     vararg columns: String,
     type: JoinType = JoinType.Inner,
 ): DataFrame<A> = join(other, type) { columns.toColumnSet() }
 
+/**
+ * Performs a [inner join][JoinType.Inner] of this [DataFrame] with the [other][\other] [DataFrame]
```
Jolanrensen: *an

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -12,13 +13,101 @@ import org.jetbrains.kotlinx.dataframe.columns.ColumnWithPath
...
+ *
+ * For more information, {@include [DocumentationUrls.Join]}.
+ *
+ * ### This `join` overload
+ */
+@ExcludeFromSources
+private interface JoinDocs
+
+// `join` method used in the example
+@Suppress("ktlint:standard:class-naming")
```
Jolanrensen: If you `@Suppress("ClassName")` you suppress both intelliJ and KtLint at the same time :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/AutoRenaming.kt
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * ## Auto-renaming in [DataFrame]
+ *
+ * In some operations, multiple columns with the same name may appear
+ * in the resulting [DataFrame].
+ *
+ * In such cases, columns with duplicate names are automatically renamed
+ * using the pattern `"\$name\$n"`, where `name` is the original column name
+ * and `n` is a unique index (1, 2, 3, and so on); the first name column goes without a number.
```
Jolanrensen: the first time the name of the column is encountered, no number is appended.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -27,89 +116,378 @@ public fun <A, B> DataFrame<A>.join(
...
 ): DataFrame<A> = joinImpl(other, type, addNewColumns = type.addNewColumns, selector)
 
+/**
+ * ### Example
+ * ```kotlin
+ * // Join by two columns with the same names in both dataframes
+ * dfLeft.{@get [JoinMethod] join}(dfRight, "name", "city")
+ * ```
+ */
+private interface JoinStringApiExample
```
Jolanrensen, line 130: exclude from sources

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/AutoRenaming.kt
AndreiKingsley: you mean "not added"?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -27,89 +116,378 @@ public fun <A, B> DataFrame<A>.join(
...
+ * @include [JoinStringApiExample] {@set [JOIN_METHOD] leftJoin}
+ * @param other [DataFrame] to join with.
+ * @param columns [Column Names][String] specifying join columns.
+ * @return joined [DataFrame].
+ */
 public fun <A, B> DataFrame<A>.leftJoin(other: DataFrame<B>, vararg columns: String): DataFrame<A> =
     leftJoin(other) { columns.toColumnSet() }
 
+/**
+ * Performs a [right join][JoinType.Right] of this [DataFrame] with [other][\other] [DataFrame]
```
Jolanrensen: *the other

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
```diff
@@ -27,89 +116,378 @@ public fun <A, B> DataFrame<A>.join(
...
+ * using selected key columns.
```
Jolanrensen: the selected key columns

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
```diff
@@ -5,14 +5,101 @@ import org.jetbrains.kotlinx.dataframe.DataRow
...
+ * using the provided [\joinExpression].
+ *
+ * @include [JoinWithCommonDescription]
+ *
+ * {@include [JoinTypeDescription]}
+ *
+ * Each join type has a corresponding shortcut function:
+ * [innerJoinWith], [leftJoinWith], [rightJoinWith], [fullJoinWith], [filterJoinWith], and [excludeJoinWith].
+ *
+ * See also [join], which performs a join by exact values equality in the selected columns.
```
Jolanrensen: *exact value equality?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
```diff
@@ -21,31 +108,163 @@ public fun <A, B> DataFrame<A>.joinWith(
     joinExpression: JoinExpression<A, B>,
 ): DataFrame<A> = joinWithImpl(right, type, addNewColumns = type.addNewColumns, joinExpression)
 
+/**
+ * Performs a [inner join][JoinType.Inner] of this [DataFrame] with the [right][\right] [DataFrame]
```
Jolanrensen: *an

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
AndreiKingsley: why?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
Jolanrensen: check correct usage of "a" vs "an" :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
Jolanrensen: check other places too

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
Jolanrensen: !

Jolanrensen: the dsl doesn't define anything itself. It's a tool, using which users can define columns

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
Jolanrensen: !

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/join.kt
Jolanrensen: turn off KoDEx ;) it doesn't understand `[this]` it tries to look for `org.jetbrains.kotlinx.dataframe.api.this` which doesn't exist. But on the user's side it's resolved correctly

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
```diff
@@ -5,14 +5,101 @@ import org.jetbrains.kotlinx.dataframe.DataRow
...
+ * dfLeft.{@get [JoinWithMethod] joinWith}(dfRight) { left -> left.fullName.contains(right.firstName) }
+ *
+ * // Join rows where the `date` value in the right `DataFrame`
+ * // falls within the interval defined by the `startDate` and `endDate`
+ * // values in the left `DataFrame`.
+ * dfLeft.{@get [JoinWithMethod] joinWith}(dfRight) { right.date in startDate..endDate }
+ *
+ * // String API; join rows where `score` value in the left `DataFrame` is higher than 3.4
+ * // and the `passed` value in the right `DataFrame` is `true`.
+ * dfLeft.{@get [JoinWithMethod] joinWith}(dfRight) { "score"<Int>() > 3.4 && right["passed"] as Boolean }
```
Jolanrensen, line 48: Nice!

You can also avoid the cast like this by writing:
`right[{ "passed"<Boolean>() }]`
or
`right.getValue<Boolean>("passed")`
whatever you like best :)

please also add this example to the "normal" join

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/AutoRenaming.kt
Jolanrensen: yes

Jolanrensen: "appending" is adding something at the end

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/joinWith.kt
Jolanrensen: Like "I sort my directory by _file size_" not "files size". It's an attribute of a single file, so it should be in singular form. (In Dutch/German it would be 1 word)

It's a bit strange with equality, I agree, as equality can only be expressed between two values. So maybe it could be better written like "which performs a join based on the exact equality between values in the selected columns"

PR #1580, opened 2025-11-20
Update Guide-for-backend-SQL-developers.md

koperagen:
Syntax for pivot is incorrect (don't compile). From SQL i think this is what it should look like
image

image

There's an option to make resulting dataframe flat, but probably it's ok with ColumnGroup
image

image


Jolanrensen:
Another reason why we should always compile our examples before writing docs :)

PR #1584, opened 2025-11-25
1.0.0 beta4n

AndreiKingsley:
do not merge!

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 projectName=dataframe
-version=1.0.0
+version=1.0.0-Beta4n
 jupyterApiTCRepo=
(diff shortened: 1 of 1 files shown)


dataframe-jupyter/build.gradle.kts
```diff
@@ -65,7 +65,7 @@ tasks.processJupyterApiResources {
 }
 
 tasks.test {
-    maxHeapSize = "2048m"
+    maxHeapSize = "8192m"
```
Jolanrensen, line 68: don't forget to make all these changes in master as well!

Usually these release branches just contain the version change commit (and maybe the -n patch) but nothing else

PR #1588, opened 2025-11-26
Added a fallback mechanism for the reading metadata

### LINKED ISSUE #1220: table name could not be available for Kotlin dataframe JDBC
From   https://github.com/Kotlin/dataframe/blob/5f5f6a1c253e609666bcb1f0665342d7fa132f08/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt#L923

```
private fun getTableColumnsMetadata(rs: ResultSet): MutableList<TableColumnMetadata> {
    val metaData: ResultSetMetaData = rs.metaData
    val numberOfColumns: Int = metaData.columnCount
    val tableColumns = mutableListOf<TableColumnMetadata>()
    val columnNameCounter = mutableMapOf<String, Int>()
    val databaseMetaData: DatabaseMetaData = rs.statement.connection.metaData
    val catalog: String? = rs.statement.connection.catalog.takeUnless { it.isNullOrBlank() }
    val schema: String? = rs.statement.connection.schema.takeUnless { it.isNullOrBlank() }

    for (i in 1 until numberOfColumns + 1) {
        val tableName = metaData.getTableName(i)
```

` val tableName = metaData.getTableName(i)`  could be not available for some DB vendors, such as Hive JDBC with `throw new SQLFeatureNotSupportedException("Method not supported");` 

It is possible to make table name empty if exception thrown. 

```
val tableName = try {
            metaData.getTableName(i)
        }   catch (e: Exception) {
            ""
        }
```

zaleslaw:
Fixes #1220 + some discovered bugs regarding type mappings

### DIFF
--- a/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/MariaDb.kt
+++ b/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/MariaDb.kt
@@ -21,10 +21,11 @@ public object MariaDb : DbType("mariadb") {
         // Force BIGINT to always be Long, regardless of javaClassName
         // MariaDB JDBC driver may report Integer for small BIGINT values
         // TODO: investigate the corner case
-        /*if (tableColumnMetadata.jdbcType == java.sql.Types.BIGINT) {
-            val kType = Long::class.createType(nullable = tableColumnMetadata.isNullable)
-            return ColumnSchema.Value(kType)
-        }*/
+
+        // if (tableColumnMetadata.jdbcType == java.sql.Types.BIGINT) {
+        //    val kType = Long::class.createType(nullable = tableColumnMetadata.isNullable)
+        //    return ColumnSchema.Value(kType)
+        // }
 
         if (tableColumnMetadata.sqlTypeName == "INTEGER UNSIGNED" ||
             tableColumnMetadata.sqlTypeName == "INT UNSIGNED"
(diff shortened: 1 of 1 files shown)


zaleslaw:
@copilot make a review of fallback mechanism from the point of supporting some JDBC sources which are not supported yet

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -418,16 +430,44 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
...
-            val columnName = rsMetaData.getColumnName(index)
-            val tableName = rsMetaData.getTableName(index)
+            // Try to getColumnName, fallback to getColumnLabel, then generate name
+            val columnName = try {
+                rsMetaData.getColumnName(index)
+            } catch (_: Exception) {
+                try {
+                    rsMetaData.getColumnLabel(index)
+                } catch (_: Exception) {
+                    "column_$index"
```
koperagen: column$index, let's avoid python style identifier

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/MariaDb.kt
```diff
@@ -18,6 +18,20 @@ public object MariaDb : DbType("mariadb") {
         get() = "org.mariadb.jdbc.Driver"
 
     override fun convertSqlTypeToColumnSchemaValue(tableColumnMetadata: TableColumnMetadata): ColumnSchema? {
+        // Force BIGINT to always be Long, regardless of javaClassName
+        // MariaDB JDBC driver may report Integer for small BIGINT values
+        if (tableColumnMetadata.jdbcType == java.sql.Types.BIGINT) {
+            val kType = Long::class.createType(nullable = tableColumnMetadata.isNullable)
```
koperagen: will this small BIGINT, runtime value actually be Long or Integer?
please add a test for it if possible

PR #1593, opened 2025-11-28
Suppress readExcel StatusLogger from log4j in 2025.2.4 notebooks

koperagen:
This message doesn't appear in 2025.3 anymore, but for 2025.2.+ fix is needed. Let's agree if we want it

### DIFF
--- a/dataframe-jupyter/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
+++ b/dataframe-jupyter/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
@@ -46,10 +46,14 @@ import org.jetbrains.kotlinx.jupyter.api.KotlinKernelHost
 import org.jetbrains.kotlinx.jupyter.api.Notebook
 import org.jetbrains.kotlinx.jupyter.api.VariableName
 import org.jetbrains.kotlinx.jupyter.api.declare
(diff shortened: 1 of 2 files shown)


dataframe-jupyter/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
```diff
@@ -200,6 +204,9 @@ internal class Integration(private val notebook: Notebook, private val options:
             declare("dataFrameConfig" to config)
         }
 
+        // Needed to suppress ERROR Log4j API could not find a logging provider in 2025.2.4
+        loadLog4jIntoNotebookClasspath(notebook)
```
Jolanrensen, line 208: Does calling this function in 2025.3+ cause any issues? Or could it cause issues in the future?

As it's only meant for 2025.2-. Can we somehow check which version the notebook is running and only call it if it's necessary?

koperagen: > Does calling this function in 2025.3+ cause any issues? Or could it cause issues in the future?

No, StatusLogger is suppressed by notebooks there. I don't think this issue will reappear

> Can we somehow check which version the notebook is running and only call it if it's necessary?

We have KotlinNotebookPluginUtils.IdeBuildNumber. I'll try

koperagen:
Update: after testing again fix doesn't work

PR #1597, opened 2025-11-28
update kandy dependencies

AndreiKingsley:
Kandy dependcies updates from #1584

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -62,8 +62,7 @@ geotools = "33.2"
 jai-core = "1.1.3"
 jts = "1.20.0"
 
-kandy = "0.8.3-dev-92"
-kandy-stats= "0.5.1-dev-2"
+kandy = "0.8.3"
 
 exposed = "1.0.0-beta-2"
 hibernate = "6.5.2.Final"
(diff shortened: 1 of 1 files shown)


gradle/libs.versions.toml
```diff
@@ -62,12 +62,8 @@ geotools = "33.2"
 jai-core = "1.1.3"
 jts = "1.20.0"
 
-# Normal examples Kandy versions
-kandy = "0.8.2-dev-90"
-
-# Example notebooks Kandy versions
-kandy-notebook = "0.8.1n"
-kandy-stats-notebook = "0.5.0n"
+kandy = "0.8.3-dev-92"
```
Jolanrensen: so, we only use kandy in notebooks? so only the -n version?

AndreiKingsley: No, kandy doesn't have a "-n" version now.

Jolanrensen: And that works fine? Does it include DataFrame as `compileOnly()` now?

AndreiKingsley: It works fine both in gradle and notebooks, but you need to add `dataframe` dependency first.
And no, DataFrame is included as `implementation`

AndreiKingsley: So we need to ask user to `%use dataframe` before `kandy`!

Jolanrensen: What about adding something like this to our jupyter integration?
```kt
onLoaded {
    val alreadyLoadedKandy = try {
        execute("kandyConfig").value
        true
    } catch (_: Exception) {
        false
    }

    if (alreadyLoadedKandy) {
        display(
            "Kandy was found on the notebook classpath. When combined with DataFrame, make sure DataFrame is loaded first; then Kandy.",
            null,
        )
    }
}
```

I think something like that could prevent some confusion :)

gradle/libs.versions.toml
```diff
@@ -62,12 +62,8 @@ geotools = "33.2"
...
+kandy-stats= "0.5.1-dev-2"
```
koperagen: Can we update to 0.8.3 too?

PR #1601, opened 2025-12-02
Optin experimental Instant in codegen if this type is used

koperagen:


dataframe-jupyter/src/test/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/JupyterCodegenTests.kt
```diff
@@ -33,6 +33,24 @@ class JupyterCodegenTests : JupyterReplTestCase() {
         res2["value"].type shouldBe typeOf<List<Any?>>()
     }
 
+    @Test
+    fun `opt in experimental Instant`() {
```
Jolanrensen, line 37: This test seems to pass, even without the fix. The errors are only logged to console

koperagen: Yes, no idea how to make it fail though. I made sure to check that with my fix logger output has no errors

koperagen: Ideally we just need to revert this fix in the next release when it's no longer experimental

Jolanrensen: I have an idea :)

Jolanrensen: https://github.com/Kotlin/dataframe/pull/1602

PR #1603, opened 2025-12-03
Add internal DataFrame.toCode utility to help convert examples

koperagen:
dataFrameOf is compiler plugin-friendly way to serialize dataframes without FrameColumns, but converting existing row-based dataFrameOf(names)(values), etc, is bothersome. This API could help to quickly go from notebook to gradle project or from tests to project
I put it in tests module with intention to maybe later expose it as API

### DIFF
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/Utils.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/Utils.kt
@@ -1,7 +1,9 @@
 package org.jetbrains.kotlinx.dataframe
 
+import org.jetbrains.kotlinx.dataframe.api.getColumns
 import org.jetbrains.kotlinx.dataframe.api.print
 import org.jetbrains.kotlinx.dataframe.api.schema
+import org.jetbrains.kotlinx.dataframe.columns.ColumnGroup
 import org.jetbrains.kotlinx.dataframe.io.renderToString
 import org.jetbrains.kotlinx.dataframe.types.UtilTests
 import java.net.URL
@@ -28,3 +30,45 @@ fun <T : DataFrame<*>> T.alsoDebug(println: String? = null, rowsLimit: Int = 20)
         print(borders = true, title = true, columnTypes = true, valueLimit = -1, rowsLimit = rowsLimit)
         schema().print()
     }
+
+fun DataFrame<*>.toCode(variableName: String = "df"): String =
+    buildString {
+        append("val $variableName = dataFrameOf(\n")
+        appendColumns(this@toCode, indent = 1)
+        append(")")
+    }
+
+private fun StringBuilder.appendColumns(df: DataFrame<*>, indent: Int) {
+    val pad = "    ".repeat(indent)
+    df.getColumns { colsAtAnyDepth().simplify() }.forEach { col ->
+        if (col is ColumnGroup<*>) {
+            append("$pad\"${col.name()}\" to columnOf(\n")
+            appendColumns(col, indent + 1)
+            append("$pad),\n")
+        } else {
+            appendColumn(col, pad)
+        }
+    }
+}
+
+private fun StringBuilder.appendColumn(column: DataColumn<Any?>, pad: String) {
+    append("$pad\"${column.name()}\" to columnOf(")
+    append(column.values().joinToString(", ") { it.toLiteral() })
+    append("),\n")
+}
+
+private fun Any?.toLiteral(): String =
+    when (this) {
+        null -> "null"
+        is String -> "\"${escape()}\""
+        is Char -> "'$this'"
+        is Long -> "${this}L"
+        is Float -> "${this}f"
+        else -> toString()
+    }
+
+private fun String.escape() =
+    replace("\\", "\\\\")
+        .replace("\"", "\\\"")
+        .replace("\n", "\\n")
+        .replace("\t", "\\t")
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/types/UtilTests.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/types/UtilTests.kt
@@ -4,6 +4,9 @@ import io.kotest.matchers.shouldBe
 import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.DataRow
 import org.jetbrains.kotlinx.dataframe.api.columnOf
+import org.jetbrains.kotlinx.dataframe.api.dataFrameOf
+import org.jetbrains.kotlinx.dataframe.api.group
+import org.jetbrains.kotlinx.dataframe.api.into
 import org.jetbrains.kotlinx.dataframe.documentation.UnifyingNumbers
 import org.jetbrains.kotlinx.dataframe.impl.asArrayAsListOrNull
 import org.jetbrains.kotlinx.dataframe.impl.commonParent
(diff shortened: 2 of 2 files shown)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/Utils.kt
```diff
@@ -28,3 +30,45 @@ fun <T : DataFrame<*>> T.alsoDebug(println: String? = null, rowsLimit: Int = 20)
         print(borders = true, title = true, columnTypes = true, valueLimit = -1, rowsLimit = rowsLimit)
         schema().print()
     }
+
+fun DataFrame<*>.toCode(variableName: String = "df"): String =
```
Jolanrensen, line 34: - Could this return a `CodeString`? similar to `generateInterfaces()` etc.
- I was a bit worried about the name, but now I look in the autocomplete; among `toHtml()`, `toList()`, `toJson()`, "`toCode()`" is not that cluttering or weird :) (actually, it's the `generateX()` functions that stand out...)
- I do think that the `val %variableName =` in the beginning should be omitted. We probably can't assume that people who want to use it will use it solely as an assignment. Maybe they want to use the generated dataframe-creating code to be an expression, like:
```kt
exec("val list = listOf(${df.toCode()})")
```

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/Utils.kt
```diff
@@ -28,3 +30,45 @@ fun <T : DataFrame<*>> T.alsoDebug(println: String? = null, rowsLimit: Int = 20)
...
+            appendColumns(col, indent + 1)
+            append("$pad),\n")
+        } else {
+            appendColumn(col, pad)
+        }
+    }
+}
+
+private fun StringBuilder.appendColumn(column: DataColumn<Any?>, pad: String) {
+    append("$pad\"${column.name()}\" to columnOf(")
```
Jolanrensen, line 55: maybe it's safer to include the type in `columnOf` explicitly :) It solves a lot of the `.toLiteral()` cases already, like `columnOf<Long>(1, 2, 3)`, but even ones you don't yet catch, like `columnOf<Short>(1, 2, 3)`

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/Utils.kt
```diff
@@ -28,3 +30,45 @@ fun <T : DataFrame<*>> T.alsoDebug(println: String? = null, rowsLimit: Int = 20)
...
+    append(column.values().joinToString(", ") { it.toLiteral() })
+    append("),\n")
+}
+
+private fun Any?.toLiteral(): String =
```
Jolanrensen, line 60: You can supply a KType to the function, which saves instance checks :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/Utils.kt
```diff
@@ -28,3 +30,45 @@ fun <T : DataFrame<*>> T.alsoDebug(println: String? = null, rowsLimit: Int = 20)
...
+    when (this) {
+        null -> "null"
+        is String -> "\"${escape()}\""
+        is Char -> "'$this'"
+        is Long -> "${this}L"
+        is Float -> "${this}f"
+        else -> toString()
```
Jolanrensen, line 67: Lists might be another nice (and commonly used) addition. Maybe maps?

Actually, that does point out a limitation of the current implementation; it's not extensible and easy to break with unsupported types.

Maybe we could add a lambda to this function so people can supply their own 'toCode renderer'. Something like

```kt
typealias ValueToCodeRenderer = (value: Any?, type: KType) -> CodeString
val toStringRenderer: ValueToCodeRenderer = { value, _ -> value.toString() }

fun DataFrame<*>.toCode(customRenderer: ValueToCodeRenderer = toStringRenderer): CodeString
```

wdyt?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/Utils.kt
koperagen: > Could this return a CodeString? similar to generateInterfaces() etc. 

I think yes. I'd also consider CodeString.copyToClipboard()

> I do think that the val %variableName = in the beginning should be omitted.

What if nullable parameter: variableName: String?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/Utils.kt
koperagen: I'll add Short and other literals, would prefer to avoid redundant type parameter so it's easy to copy-paste as is

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/Utils.kt
koperagen: List and Map seem reasonable by default. Code renderer too

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/Utils.kt
koperagen: i think instance check is very cheap on JVM and we have a smartcast as a result

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/Utils.kt
Jolanrensen: Yes, good idea for both :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/Utils.kt
Jolanrensen: As long as we can somehow guarantee that the provided literals/class initializations are always the expected type, it will be fine. Maybe we could also parameterize this option if we actually want to promote this method to the public API?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/Utils.kt
Jolanrensen: Would you prefer cheap or free? ;P but yeah, the smart casts are nice

koperagen: How free? :o Maybe i don't get the idea

koperagen: let's keep instance checks, i added a lot of branches where smart casts are useful

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/Utils.kt
koperagen: I'll setup extensive testing with Kotest https://kotest.io/docs/proptest/property-based-testing.html and Jupyter Kernel to execute generated code and compare resulting instance to original + compilation

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/Utils.kt
Jolanrensen: It's free because you get the values from a `DataColumn` where the `KType` is already known. So if we would supply this `KType` to `toLiteral()` it's simply an equality check instead of an instance check.

PR #1606, opened 2025-12-04
update readme

AndreiKingsley:


README.md
```diff
@@ -200,6 +214,8 @@ other recommended versions.
 | 1.0.0-Beta2              | 8 / 11               | 2.0.20         | 0.12.0-383             | 18.1.0               | 2.2.20-dev-3524         | 0.8.1-dev-66             |
 | 1.0.0-Beta3n (notebooks) | 8 / 11               | 2.2.20         | 0.15.0-587 (K1 only)   | 18.3.0               | -                       | 0.8.1n                   |
 | 1.0.0-Beta3              | 8 / 11               | 2.2.20         | 0.15.0-587             | 18.3.0               | 2.2.20 / IDEA 2025.2+   | 0.8.1                    |
+| 1.0.0-Beta4n (notebooks) | 8 / 11               | 2.2.21         | 0.16.0-736 (K1 only)   | 18.3.0               | -                       | 0.8.3                    |
```
Jolanrensen: I think we can drop the "K1 only" note :) DataFrame now seems to work in k2 mode as well for the latest 0.17.x notebook version. Though, users will probably not even know how to switch modes yet, as it's still hidden under a registry flag

Issue #1258, opened 2025-06-18
Add pdf docs (for AI models learning).


AndreiKingsley:

AndreiKingsley:
https://github.com/Kotlin/kandy/issues/394

PR #1615, opened 2025-12-05
Enable llms.txt generation in docs

koperagen:
Related to https://github.com/Kotlin/dataframe/issues/1258
More info about this file: https://www.jetbrains.com/help/writerside/generate-llms-txt.html
I couldn't test it locally, but in GH Actions use writerside builder 2025.04.8412﻿
 that includes this feature: https://www.jetbrains.com/help/writerside/8312.html
 
 Let's see the content, decide if we need it. Although single artifact for LLM is a thing, LLM providers seem to use our *.md topics from GitHub instead of documentation website.

### DIFF
--- a/docs/StardustDocs/cfg/buildprofiles.xml
+++ b/docs/StardustDocs/cfg/buildprofiles.xml
@@ -18,6 +18,7 @@
         <browser-edits-url>https://github.com/Kotlin/dataframe/edit/master/docs/StardustDocs/</browser-edits-url>
         <include-in-head>include-head.html</include-in-head>
     </variables>
+    <llms-txt single-file="true"/>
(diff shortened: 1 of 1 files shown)


Jolanrensen:
[llms.txt](https://github.com/user-attachments/files/24036279/llms.txt)

this is what it spits out :) 

While it looks rather complete, I am not sure if it may be too large for users to include in their context... Maybe we could try `<llms-text/>` which also produces individual text files:

[_llms.zip](https://github.com/user-attachments/files/24036344/_llms.zip)

koperagen:
Hmm.. i don't know. I'm yet to fully grasp the idea, need to read more https://llmstxt.org/

Issue #1614, opened 2025-12-05
Make `:examples` be separate projects and use the compiler plugin


Jolanrensen:
Because `:examples` are `include()`'d in from the main `settings.gradle.kts`, they are built using the "raw" dataframe sources. This is great for making sure changing our API does not break our tests, but it comes with some limits:
- we need to inherit versions of libraries, like the Kotlin version
- Enabling the compiler plugin in modules inside `:dataframe` is [not that easy](https://github.com/Kotlin/dataframe/pull/1456/files#diff-c51a3140f40e36ed849bdc6bad83ce621f1af525c6b6b48f8ca14ab45d37c42d) and requires a lot of boilerplate build-code that I don't think our users would appreciate to find in our examples.

Some other examples, like `:examples/android-example` and `:examples/kotlin-dataframe-plugin-example` use `includeBuild()` to avoid these issues. However, they come with the downside of using their own versions (which are easily forgotten to be bumped after releases), and as they use their own version of dataframe, they will not fail when we break our API until it's too late.

Maybe we can find a middle ground in Gradle where we dó use `includeBuild` but can still substitute or override dependencies, or we make a generic "enable the compiler plugin for local modules" function we can call from all local modules in such a way that users can identify and omit them for themselves.

Jolanrensen:
We could also reuse the idea we have with our notebook examples:
- We build examples in a `/dev` folder with our actual modules.
- Upon a new version published, we move them out of the /dev folder and give them a fixed DF version that users can easily copy. Possibly combined with https://github.com/Kotlin/dataframe/pull/1616

koperagen:
Maintaining two versions seems to be the only option. If we publish zip, it should depend on stable release. At the same time another copy will depend on master (using substitute) and help track changes

koperagen:
https://github.com/Kotlin/dataframe/pull/1627/commits/61c06c4be3b0d3e691116b15915b1de76931ba9a
Please take this into account

Jolanrensen:
I've been experimenting with gradle convention plugins and composite builds: https://github.com/Jolanrensen/GradleConventionsExperiment

I believe this sample project can be used as a good base for the DataFrame project itself:
- (type safe) convention plugins allow us to remove duplicate build code and avoid using `allProjects {}`
[...]

PR #1616, opened 2025-12-05
Add GH Action to produce direct links to download example subprojects

koperagen:
There's a problem that even include build subprojects are a part of dataframe repo, so one need to download and build it entirely to run them.
New GH Action will build zip archives from subproject directories and push them to separate orphan branch. We will be able to share links directly to download. 
It'll make these subprojects more easily available and will enable sharing of more complex projects while keeping everything under one project for nice development experience

### DIFF
--- a/.github/workflows/archive-examples.yml
+++ b/.github/workflows/archive-examples.yml
@@ -1,0 +1,81 @@
+  name: Archive Example Projects
+
+  on:
+    push:
+      branches:
+        - master
+      paths:
+        - 'examples/android-example/**'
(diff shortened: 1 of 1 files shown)


Jolanrensen:
I think this should probably be combined with https://github.com/Kotlin/dataframe/issues/1614. I like the idea, but maybe make it an issue first, so we can consider them together :)

The other examples would also benefit from this!

koperagen:
We can extend this approach to other examples after migrating them to subprojects

Jolanrensen:
That's true. Okay, let's enable it for these two for now, so we can already put it on the website :)

Issue #265, opened 2023-02-15
Fill the information about the Column Arithmetic


zaleslaw:
The [page ](https://kotlin.github.io/dataframe/columnarithmetics.html) is empty

Jolanrensen:
depends whether we want to remove or keep column arithmetics

koperagen:
Please refer to this discussion when implementing https://github.com/Kotlin/dataframe/pull/1617

PR #1617, opened 2025-12-05
Update column operations documentation

koperagen:
column arithmetics are essentially obsolete IMO, doesn't make sense to keep this page around

### DIFF
--- a/docs/StardustDocs/d.tree
+++ b/docs/StardustDocs/d.tree
@@ -177,7 +177,6 @@
             <toc-element topic="joinWith.md"/>
         </toc-element>
         <toc-element topic="columnOperations.md">
-            <toc-element topic="columnArithmetics.md" toc-title="Arithmetics"/>
             <toc-element topic="columnStatistics.md" toc-title="Statistics"/>
             <toc-element topic="asIterable.md"/>
             <toc-element topic="asSequenceColumn.md" toc-title="asSequence"/>
--- a/docs/StardustDocs/topics/columnArithmetics.md
+++ b/docs/StardustDocs/topics/columnArithmetics.md
@@ -1,3 +1,0 @@
-[//]: # (title: Column arithmetics)
-
-// TODO
--- a/docs/StardustDocs/topics/columnOperations.md
+++ b/docs/StardustDocs/topics/columnOperations.md
@@ -1,3 +1,7 @@
-[//]: # (title: Column operations)
+[//]: # (title: Column Operations)
+<show-structure depth="3"/>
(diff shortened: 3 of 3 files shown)


Jolanrensen:
related issues: https://github.com/Kotlin/dataframe/issues/530, https://github.com/Kotlin/dataframe/issues/265

I wouldn't say they're obsolete actually. I've seen AI recommend `colA + colB` on multiple occasions. If we can, why shouldn't we support such a thing?

koperagen:
I think they are obsolete in a sense that we don't have cases where we'd recommend using it. At least i don't. dataframe is not pandas, and to be useful in Kotlin most of the time you need to operate on individual objects / values. Which it the default for majority of our operations, except for rare `groupBy { s.length() }`. But even there i'd recommend `groupBy { expr { s.length } }` because we do have `DataColumn<String>.length()`, but not `DataColumn<List>.size()`. 

We can keep these functions around (i don't suggest to deprecate them), but i prefer to migrate examples from using it whenever i see. If you have specific examples when you see it to be useful, please share

Jolanrensen:
@koperagen I think most useful examples are from Kandy:

Let's say you have some data with values and corrections of those values:
```kt
val df = dataFrameOf(
    "ID" to columnOf(1L, 2L, 3L),
    "score" to columnOf(10, 20, 30),
    "correction" to columnOf(-0.5, 0.0, 0.5),
)
```

It would be nice to plot the immediate results of the scores with correction without having to add a new "temporary" column, just for plotting:

```kt
df.plot { 
    bars { 
        x(ID)
        y(score + correction)
    }
}
```
(this is currently not possible between 2 columns, and there is no easy way to zip two columns... There's `add {}` but this requires an extra cell in notebooks (for now))

"`column` **operator** `value`" or "`value` **operator** `value`" can usually be replaced with something like `expr { column + value }` or a `column.map { it + value }`. But it may be nice to be able to do a `x(columnA / 100.0)` at once in Kandy.

koperagen:
> @koperagen I think most useful examples are from Kandy:

True! I can easily imagine that, i remember writing similar code myself. 
I suggest to remove the page, and when we decide to introduce it back - let's provide "goal" of that API
1. That most of the time column operations are not needed in dataframe unlike pandas, because we provide row-based APIs
2. Recommended "adapter" functions like map or expr
3. Example where it might still be useful, and only then what APIs we provide

Jolanrensen:
@koperagen good plan :) Certainly better to remove it than have a `// TODO` on a page

Issue #1619, opened 2025-12-05
Improve reading of Parquet files with nested schema


Allex-Nik:
blocked by #536 

There is a schema of a Parquet file:
```
{
  "name": "book",
  "type": "record",
  "fields": [
    {"name": "id", "type": "int"},
    {"name": "title", "type": "string"},
    {
      "name": "author",
      "type": {
        "type": "record",
        "name": "author",
        "fields": [
          {"name": "id", "type": "int"},
          {"name": "firstName", "type": "string"},
          {"name": "lastName", "type": "string"}
        ]
      }
    },
    {"name": "genre", "type": "string"},
    {"name": "publisher", "type": "string"}
  ]
}
```

The field `author` is nested. 
The schema is parsed as `org.apache.avro.Schema` and `AvroParquetWriter` is used to write the Parquet file.
When this file is read with `DataFrame.readParquet()`, the nested field `author` is represented in a DataFrame as a ValueColumn containing a map in each cell:

![dataframe](https://github.com/user-attachments/assets/c11dfdbd-2bec-44f3-b760-377f9b18f9d1)

This kind of fields, however, could be represented as a ColumnGroup.

The Parquet file mentioned above and the Kotlin Notebook in the screenshot are attached below (as a zip archive, as GitHub does not accept .parquet files).

[parquet_file_and_notebook.zip](https://github.com/user-attachments/files/23967545/parquet_file_and_notebook.zip)

Issue #536, opened 2023-12-12
Add inner / Struct type support in Arrow


fb64:
Arrow Struct type is read as a `Map<String, Any?>` object :
https://github.com/Kotlin/dataframe/blob/86b80e0c9cd372334e8eff05115a7c50b6ea61bc/dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt#L171-L173

But write does not support Map Object and by defaut value is serialized as a `String` :
https://github.com/Kotlin/dataframe/blob/86b80e0c9cd372334e8eff05115a7c50b6ea61bc/dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowTypesMatching.kt#L93-L95

The following test fail  because `c` column is a LinkedHashMap in a SingletonList on excepted Dataframe and a single String in an Arraylist on readIpc object

```kotlin
    @Test
    fun testReadIPC(){
        val a by columnOf("one")
        val b by columnOf(2.0)
        val c by listOf(
            mapOf(
                "c1" to Text("inner"),
                "c2" to 4.0,
                "c3" to 50.0,
            ) as Map<String, Any?>
        ).toColumn()
        val d by columnOf("four")
        val expected =  dataFrameOf(a, b, c, d)
        val readIpc = DataFrame.readArrowIPC(expected.saveArrowIPCToByteArray())
        readIpc shouldBe expected
    }
```

![image](https://github.com/Kotlin/dataframe/assets/3306862/1da1602e-bf54-4fc8-a860-b513ef62d2de)

![image](https://github.com/Kotlin/dataframe/assets/3306862/38b6986b-2344-46ee-b08f-2fc43732531b)

It could be relevant to add the support of inner type by Writing `Map<String,Any?>` in a [Struct field](https://arrow.apache.org/docs/cpp/api/datatype.html#_CPPv4N5arrow4Type4type6STRUCTE). 
Some points need to be addressed before implementation : 
- Should it support inner of inner... recursively ?
- Should it support only `Map<String, Any?>` ? 
- Struct field is it the better choice as Arrow support also [Map type](https://arrow.apache.org/docs/cpp/api/datatype.html#_CPPv4N5arrow4Type4type3MAPE)
- Maybe inner type should be specified for the dataframe core and be implemented in a consistent way for all supported types

[...]

Issue #271, opened 2023-02-16
writeArrowFeather not working with nested type ?


phodal:
Hi, in my case, I want to create a arrow file in client side, then pass to server side. But when I just try run `writeArrowFeather`,  will show the `IndexOutOfBoundsException` issues.

```java
Exception in thread "main" java.lang.IndexOutOfBoundsException: index: 31393, length: 2320 (expected: range(0, 32768))
	at org.apache.arrow.memory.ArrowBuf.checkIndex(ArrowBuf.java:701)
	at org.apache.arrow.memory.ArrowBuf.setBytes(ArrowBuf.java:765)
	at org.apache.arrow.vector.BaseVariableWidthVector.setBytes(BaseVariableWidthVector.java:1244)
	at org.apache.arrow.vector.BaseVariableWidthVector.set(BaseVariableWidthVector.java:1059)
	at org.apache.arrow.vector.VarCharVector.set(VarCharVector.java:255)
	at org.jetbrains.kotlinx.dataframe.io.ArrowWriterImpl$infillVector$1.invoke(ArrowWriterImpl.kt:111)
	at org.jetbrains.kotlinx.dataframe.io.ArrowWriterImpl$infillVector$1.invoke(ArrowWriterImpl.kt:111)
	at org.jetbrains.kotlinx.dataframe.api.ForEachKt.forEachIndexed(forEach.kt:34)
	at org.jetbrains.kotlinx.dataframe.io.ArrowWriterImpl.infillVector(ArrowWriterImpl.kt:111)
	at org.jetbrains.kotlinx.dataframe.io.ArrowWriterImpl.allocateVectorAndInfill(ArrowWriterImpl.kt:197)
	at org.jetbrains.kotlinx.dataframe.io.ArrowWriterImpl.allocateVectorSchemaRoot(ArrowWriterImpl.kt:223)
	at org.jetbrains.kotlinx.dataframe.io.ArrowWriter$DefaultImpls.writeArrowFeather(ArrowWriter.kt:114)
	at org.jetbrains.kotlinx.dataframe.io.ArrowWriterImpl.writeArrowFeather(ArrowWriterImpl.kt:61)
	at org.jetbrains.kotlinx.dataframe.io.ArrowWriter$DefaultImpls.writeArrowFeather(ArrowWriter.kt:125)
	at org.jetbrains.kotlinx.dataframe.io.ArrowWriterImpl.writeArrowFeather(ArrowWriterImpl.kt:61)
[...]

PR #1621, opened 2025-12-08
Apache Arrow/Parquet: ColumnGroup <-> StructVector reading and writing

koperagen:
fixes https://github.com/Kotlin/dataframe/issues/536
fixes https://github.com/Kotlin/dataframe/issues/1619
related to https://github.com/Kotlin/dataframe/issues/271, but it also requires List reading/writing


koperagen:
Yes, let's do ListVector too. It's mentioned in #271 that won't be closed by this PR

PR #1628, opened 2025-12-08
KDocs fixes for `distinct` and `distinctBy`

### LINKED ISSUE #1434: Distinct KDocs are wrong
The KDocs of `distinct` actually show the docs of `distinctBy {}`:

Image

This is not how `distinct { cols }` functions. Distinct with columns is a shortcut for `select { cols }.distinctBy { all() }`.
So, 
`distinctBy`: Removes duplicated rows based on the specified columns.
`distinct`: Selects the specified columns keeping only distinct rows

Allex-Nik:
Fixes #1434

The documentation for the `distinct` function with parameters described the functionality of `distinctBy`: it suggested that `distinct` removes duplicated rows omitting the fact that it also selects the specified columns and the result contains only these columns.

In this PR I:
* Fixed this issue by changing descriptions of `distinct`
* Added `distinctBy` to DocumentationUrls
* Made some other minor fixes

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
@@ -10,8 +10,8 @@ import org.jetbrains.kotlinx.dataframe.annotations.Refine
 import org.jetbrains.kotlinx.dataframe.api.DistinctDocs.DESCRIPTION
 import org.jetbrains.kotlinx.dataframe.api.DistinctDocs.DISTINCT_PARAM
 import org.jetbrains.kotlinx.dataframe.api.DistinctDocs.DISTINCT_RETURN
-import org.jetbrains.kotlinx.dataframe.api.DistinctDocs.FUNCTION
 import org.jetbrains.kotlinx.dataframe.api.DistinctDocs.PHRASE_ENDING
+import org.jetbrains.kotlinx.dataframe.api.DistinctDocs.SEE_ALSO
 import org.jetbrains.kotlinx.dataframe.api.Select.SelectSelectingOptions
 import org.jetbrains.kotlinx.dataframe.columns.ColumnSet
 import org.jetbrains.kotlinx.dataframe.columns.toColumnSet
@@ -29,12 +29,12 @@ import kotlin.reflect.KProperty
 // region DataFrame
 
 /**
- * {@get [DESCRIPTION] Removes duplicated rows based on {@get [PHRASE_ENDING]}}.
+ * {@get [DESCRIPTION] Removes duplicated rows based on $[PHRASE_ENDING]=all columns.}
  *
  * The [rows][DataRow] in the resulting [DataFrame] are in the same order
  * as they were in the original [DataFrame].
  *
- * See also {@get [FUNCTION] [distinctBy] that removes duplicated rows based on the specified columns
+ * See also {@get [SEE_ALSO] [distinctBy] that removes duplicated rows based on the specified columns
  * and keeps all the columns in the resulting [DataFrame].}
  *
  * @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
@@ -43,41 +43,45 @@ import kotlin.reflect.KProperty
  *
  * For more information:
  *
- * {@include [DocumentationUrls.Distinct]}
+ * @include [DocumentationUrls.Distinct]
  *
- * {@include [DocumentationUrls.DistinctBy]}
+ * @include [DocumentationUrls.DistinctBy]
  *
- * {@get [DISTINCT_PARAM]}
+ * @get [DISTINCT_PARAM]
  *
  * @return {@get [DISTINCT_RETURN] A new [DataFrame] containing only distinct rows.}
  */
 @ExcludeFromSources
 @Suppress("ClassName")
 private interface DistinctDocs {
+    // Parameter of the function (the `@param` part of the KDoc)
     interface DISTINCT_PARAM
 
+    // Value returned by the function (the `@return` part of the KDoc)
     interface DISTINCT_RETURN
 
+    // Description of what the function does
     interface DESCRIPTION
 
+    // Part of the description that can be customized for a specific function
     interface PHRASE_ENDING
 
-    interface FUNCTION
+    // Reference to a related function (see also)
+    interface SEE_ALSO
 }
 
 /**
- * {@include [DistinctDocs]}
- * {@set [PHRASE_ENDING] all columns}
- * {@set [DISTINCT_PARAM]}
+ * @include [DistinctDocs]
+ * @set [DISTINCT_PARAM]
  */
 public fun <T> DataFrame<T>.distinct(): DataFrame<T> = distinctBy { all() }
 
 /**
- * {@include [DistinctDocs]}
- * {@set [DESCRIPTION] Selects the specified columns and keeps only distinct rows based on these selected columns}
- * {@set [DISTINCT_PARAM] @param [columns\] The [ColumnsSelector] used to select columns
- * that will be included in the resulting [DataFrame] and considered for evaluating distinct rows.}
- * {@set [DISTINCT_RETURN] A new [DataFrame] containing only selected columns and distinct rows.}
+ * @include [DistinctDocs]
+ * @set [DESCRIPTION] Selects the specified columns and keeps only distinct rows based on these selected columns.
+ * @set [DISTINCT_PARAM] @param [columns\] The [ColumnsSelector] used to select columns
+ * that will be included in the resulting [DataFrame] and considered for evaluating distinct rows.
+ * @set [DISTINCT_RETURN] A new [DataFrame] containing only selected columns and distinct rows.
  */
 @Refine
 @Interpretable("Distinct0")
@@ -92,11 +96,11 @@ public fun <T> DataFrame<T>.distinct(vararg columns: KProperty<*>): DataFrame<T>
     }
 
 /**
- * {@include [DistinctDocs]}
- * {@set [DESCRIPTION] Selects the specified columns and keeps only distinct rows based on these selected columns}
- * {@set [DISTINCT_PARAM] @param [columns\] The names of the columns to select
- * and to consider for evaluating distinct rows.}
- * {@set [DISTINCT_RETURN] A new [DataFrame] containing only selected columns and distinct rows.}
+ * @include [DistinctDocs]
+ * @set [DESCRIPTION] Selects the specified columns and keeps only distinct rows based on these selected columns.
+ * @set [DISTINCT_PARAM] @param [columns\] The names of the columns to select
+ * and to consider for evaluating distinct rows.
+ * @set [DISTINCT_RETURN] A new [DataFrame] containing only selected columns and distinct rows.
  */
 public fun <T> DataFrame<T>.distinct(vararg columns: String): DataFrame<T> = distinct { columns.toColumnSet() }
 
@@ -111,13 +115,13 @@ public fun <T> DataFrame<T>.distinctBy(vararg columns: KProperty<*>): DataFrame<
     distinctBy { columns.toColumnSet() }
 
 /**
- * {@include [DistinctDocs]}
- * {@set [PHRASE_ENDING] the specified columns}
- * {@set [FUNCTION] [distinct] that selects the specified columns
+ * @include [DistinctDocs]
+ * {@set [PHRASE_ENDING] the specified}
+ * @set [SEE_ALSO] [distinct] that selects the specified columns
  * (if the columns are not specified, selects all columns)
- * and keeps only distinct rows based on these selected columns.}
- * {@set [DISTINCT_PARAM] @param [columns\]
- * The names of the columns to consider for evaluating distinct rows.}
+ * and keeps only distinct rows based on these selected columns.
+ * @set [DISTINCT_PARAM] @param [columns\]
+ * The names of the columns to consider for evaluating distinct rows.
  */
 public fun <T> DataFrame<T>.distinctBy(vararg columns: String): DataFrame<T> = distinctBy { columns.toColumnSet() }
 
@@ -127,13 +131,13 @@ public fun <T> DataFrame<T>.distinctBy(vararg columns: AnyColumnReference): Data
     distinctBy { columns.toColumnSet() }
 
 /**
- * {@include [DistinctDocs]}
- * {@set [PHRASE_ENDING] the specified columns}
- * {@set [FUNCTION] [distinct] that selects the specified columns
+ * @include [DistinctDocs]
+ * {@set [PHRASE_ENDING] the specified}
+ * @set [SEE_ALSO] [distinct] that selects the specified columns
  * (if the columns are not specified, selects all columns)
- * and keeps only distinct rows based on these selected columns.}
- * {@set [DISTINCT_PARAM] @param [columns\] The [ColumnsSelector] used to select columns
- * that will be considered for evaluating distinct rows.}
+ * and keeps only distinct rows based on these selected columns.
+ * @set [DISTINCT_PARAM] @param [columns\] The [ColumnsSelector] used to select columns
+ * that will be considered for evaluating distinct rows.
  */
 public fun <T, C> DataFrame<T>.distinctBy(columns: ColumnsSelector<T, C>): DataFrame<T> {
     val cols = get(columns)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
```diff
@@ -23,24 +25,26 @@ import kotlin.reflect.KProperty
...
  *
  * @see [Selecting Columns][SelectSelectingOptions].
  * @see {@include [DocumentationUrls.Distinct]}
+ * @see {@include [DocumentationUrls.DistinctBy]}
  */
 @ExcludeFromSources
 @Suppress("ClassName")
 private interface DistinctDocs {
     interface DISTINCT_PARAM
+    interface DISTINCT_RETURN
```
Allex-Nik, line 61: Does it make sense to add it here? I introduced this tag to separate cases when we return a DataFrame with distinct rows and when we return a DataFrame which contains only selected columns and distinct rows.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
```diff
@@ -52,15 +56,21 @@ public fun <T> DataFrame<T>.distinct(): DataFrame<T> = distinctBy { all() }
...
  */
 @Refine
 @Interpretable("Distinct0")
 public fun <T, C> DataFrame<T>.distinct(columns: ColumnsSelector<T, C>): DataFrame<T> = select(columns).distinct()
 
 /**
  * {@include [DistinctDocs]}
- * {@set PHRASE_ENDING the specified columns}.
+ * {@set DESCRIPTION It selects the specified columns and keeps only distinct rows based on these selected columns}
+ * {@set [DistinctDocs.DISTINCT_PARAM] @param [columns] The names of the columns to select
```
Allex-Nik: I changed the description of the parameter for some functions to mention that it is also used to select columns.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
```diff
@@ -87,7 +103,10 @@ public fun <T> DataFrame<T>.distinct(vararg columns: AnyColumnReference): DataFr
 
 /**
  * {@include [DistinctDocs]}
+ * {@set NAME DistinctBy}
  * {@set PHRASE_ENDING the specified columns}.
+ * {@set [DistinctDocs.DISTINCT_PARAM] @param [columns]
```
Allex-Nik: I specified this explicitly for this and the following functions (instead of keeping the default option in the common section in `DistinctDocs`) to avoid the issue with `[columns]` I mentioned in the description of the PR.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
```diff
@@ -23,24 +25,27 @@ import kotlin.reflect.KProperty
 // region DataFrame
 
 /**
- * ## The Distinct Operation
+ * ## The {@get NAME Distinct} Operation
```
AndreiKingsley: Remove header, we decided not to add headers at all

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
AndreiKingsley: Yes, it's ok

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
```diff
@@ -23,24 +25,27 @@ import kotlin.reflect.KProperty
...
  *
- * It removes duplicated rows based on {@get PHRASE_ENDING}.
+ * {@get DESCRIPTION It removes duplicated rows based on {@get PHRASE_ENDING}}.
```
AndreiKingsley: Please, change in all kdocs "It removes .." -> "Removes .."

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
```diff
@@ -23,24 +25,27 @@ import kotlin.reflect.KProperty
...
  *
- * __NOTE:__ The rows in the resulting [DataFrame] are in the same order as they were in the original [DataFrame].
+ * __NOTE:__ The [rows][DataRow] in the resulting [DataFrame] are in the same order
```
AndreiKingsley: Nice, but let's remove "__NOTE:__"

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
```diff
@@ -23,24 +25,27 @@ import kotlin.reflect.KProperty
...
+ * as they were in the original [DataFrame].
  *
- * {@get [DISTINCT_PARAM] @param [columns]
- * The names of the columns to consider for evaluating distinct rows.}
+ * {@get [DISTINCT_PARAM]}
  *
- * @return A new DataFrame containing only distinct rows.
+ * {@get [DISTINCT_RETURN] @return A new [DataFrame] containing only distinct rows.}
  *
  * @see [Selecting Columns][SelectSelectingOptions].
```
AndreiKingsley: Let's rewrite this section with our template: https://github.com/Kotlin/dataframe/blob/5d6b7af77de4d92de9b1b3adb686e8af23eacae2/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt#L109

Something like that

```
See also [distinctBy], that (write difference here)

 @include [SelectingColumns.ColumnGroupsAndNestedColumnsMention]
 
 See [Selecting Columns][ConvertSelectingOptions].
 
 For more information: {@include [DocumentationUrls.Distinct]}
```

AndreiKingsley:
@Allex-Nik To avoid this problem with `[columns]` you can write `[\columns]` instead. We use this trick in many places.

I'd even say it's better to write `[\columns]` in such places always.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
```diff
@@ -87,7 +104,10 @@ public fun <T> DataFrame<T>.distinct(vararg columns: AnyColumnReference): DataFr
 
 /**
  * {@include [DistinctDocs]}
+ * {@set NAME DistinctBy}
  * {@set PHRASE_ENDING the specified columns}.
```
AndreiKingsley: change to` [\columns]` here and in other places

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
```diff
@@ -23,24 +25,27 @@ import kotlin.reflect.KProperty
...
- * __NOTE:__ The rows in the resulting [DataFrame] are in the same order as they were in the original [DataFrame].
+ * __NOTE:__ The [rows][DataRow] in the resulting [DataFrame] are in the same order
+ * as they were in the original [DataFrame].
  *
- * {@get [DISTINCT_PARAM] @param [columns]
- * The names of the columns to consider for evaluating distinct rows.}
+ * {@get [DISTINCT_PARAM]}
  *
- * @return A new DataFrame containing only distinct rows.
+ * {@get [DISTINCT_RETURN] @return A new [DataFrame] containing only distinct rows.}
```
Jolanrensen: Seems you can "lift out" the `@return`, as you include it everywhere you set `[DISTINCT_RETURN]`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
```diff
@@ -23,24 +25,27 @@ import kotlin.reflect.KProperty
...
  *
  * @see [Selecting Columns][SelectSelectingOptions].
  * @see {@include [DocumentationUrls.Distinct]}
```
Jolanrensen: never used `@see` like that :o seems to work well!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
Jolanrensen: (and you never set `[DISTINCT_RETURN]` to empty)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
```diff
@@ -52,15 +57,21 @@ public fun <T> DataFrame<T>.distinct(): DataFrame<T> = distinctBy { all() }
 
 /**
  * {@include [DistinctDocs]}
- * {@set PHRASE_ENDING the specified columns}.
+ * {@set DESCRIPTION It selects the specified columns and keeps only distinct rows based on these selected columns}
+ * {@set [DistinctDocs.DISTINCT_PARAM] @param [columns] The names of the columns to select
```
Jolanrensen: it's not "The names of the columns" though, it's a columns selector. Check how we write the `@param` of columnsSelectors in other operations :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
Jolanrensen: it's also not "the names" here

Jolanrensen: but as this function is deprecated, you can probably remove the kdocs altogether

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
Jolanrensen: Please make sure all `@get/@set` keys are `[references]`. This will work too, but references are refactor- and typo-safe :)

Jolanrensen: See https://github.com/Kotlin/dataframe/blob/master/KDOC_PREPROCESSING.md#arg-interfaces

Jolanrensen:
> Specified the parameter explicitly for every function to avoid incorrect resolution of the columns parameter mentioned in DistinctDocs

TL;DR: write it like `[columns\]`

Long explanation:

Whenever some KDocs is `@include`d, all references mentioned in that doc are expanded to their fully qualified path, if possible. This solves the issue that a reference to `[a]` in one doc is not necessarily resolvable as `[a]` in another doc. But it may be resolvable as `[a][path.to.a]`.

If a reference cannot be found, it's left unchanged as `[a]`.

Unfortunately, this system is not perfect (resolving symbols by path in kotlin myself is hard XD), so when you write
```kt
/** @param [columns] The names of the columns to consider for evaluating distinct rows. */
interface DistinctDocs

/** @include [DistinctDocs] */
public fun <T> DataFrame<T>.distinctBy...
```

KoDEx tries to find `[columns]` in the scope of `DistinctDocs`, finds it, and expands it to `[columns][org.jetbrains.kotlinx.dataframe.columns]` before including it at `distinctBy()`.

Luckily, we have the [`\` escape character](https://github.com/Kotlin/dataframe/blob/67a1ef868ed3b0c2e4e90e6c7ca3729df39952db/KDOC_PREPROCESSING.md#-escape-character) :) which allows us to stop KoDEx from doing "clever" things. These are removed from the KDocs in the last phase. These allow you to 'break' tags like `\@inlude X`, or `\$something`, or references like `\[columns]` so they aren't processed anymore :). (You can put the `\` anywhere in the reference I believe, actually)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
Jolanrensen: It's probably best to just write the `@param` line at the bottom of this KDoc, as each overload needs a different line of text for it. Yes, it will be below the `@return` then, but the IDE still renders it well.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
Allex-Nik: Then I need to include `DESCRIPTION` and `PHRASE_ENDING` into `DistinctDocs` like this, right?

```kotlin
private interface DistinctDocs {
    interface DISTINCT_PARAM

    interface DISTINCT_RETURN

    interface DESCRIPTION

    interface PHRASE_ENDING

    interface FUNCTION
}
```

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
Allex-Nik: After the solution with `[columns\]` was pointed out, the KDocs for deprecated functions were deleted, and the parameter descriptions were individualized for different parameters, this is no longer relevant.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
Allex-Nik: I fixed the descriptions of the parameters, thank you :)
Why is it best to write the `@param` line at the bottom, below the `@return`? If I set the `@return` after the `@param`, everything seems to work for all cases in the file.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
Allex-Nik: Is it fine to do `[columns\]`? I saw `[columns\]` in other functions I checked, and the effect seems to be the same.

Allex-Nik:
> TL;DR: write it like `[columns\]`
> 
> Long explanation:

@Jolanrensen, thank you! I understand :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
Jolanrensen: yes, doesn't matter where you put the `\` :) as long as you break the reference

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
```diff
@@ -23,45 +29,60 @@ import kotlin.reflect.KProperty
 // region DataFrame
 
 /**
- * ## The Distinct Operation
+ * {@get [DESCRIPTION] Removes duplicated rows based on {@get [PHRASE_ENDING]}}.
```
Jolanrensen: interesting, I've never used a `{@get}` inside the "default" section of another `{@get}` XD but it works :). Did you know you can also write `$[PRASE_ENDING]` btw? just an alternative notation, does the same.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
```diff
@@ -23,45 +29,60 @@ import kotlin.reflect.KProperty
...
 private interface DistinctDocs {
     interface DISTINCT_PARAM
+
+    interface DISTINCT_RETURN
+
+    interface DESCRIPTION
+
+    interface PHRASE_ENDING
+
+    interface FUNCTION
```
Jolanrensen: I usually write a tiny comment (no KDoc) next to each parameter, to clarify what they mean

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
```diff
@@ -23,45 +29,60 @@ import kotlin.reflect.KProperty
...
  * {@include [DistinctDocs]}
- * {@set PHRASE_ENDING all columns}.
- * {@set [DistinctDocs.DISTINCT_PARAM]}
+ * {@set [PHRASE_ENDING] all columns}
+ * {@set [DISTINCT_PARAM]}
  */
 public fun <T> DataFrame<T>.distinct(): DataFrame<T> = distinctBy { all() }
 
 /**
  * {@include [DistinctDocs]}
```
Jolanrensen: None of the tags in this KDoc need `{}` around them, as they are not inside another line. It doesn't hurt though, so feel free to write it how you think is best

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
Allex-Nik: I didn't know about `$[PRASE_ENDING]`, it looks better, I used it :)

Jolanrensen: you can also provide a "default" with that notation, but it needs to be small, as spaces split it up already, like `$key=default` :)

Allex-Nik: Nice! I implemented it as well, thank you!
Now `$[PHRASE_ENDING]=all`, and in some places I just replace "all" with "the specified", so "columns" is no longer a part of the `[PHRASE_ENDING]` and is only a part of `[DESCRIPTION]` :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
Allex-Nik: Done

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/distinct.kt
Allex-Nik: Done

AndreiKingsley (approved):
Nice!

Issue #1587, opened 2025-11-26
Refactor type inference in JDBC module


zaleslaw:
For example
1) convertSqlTypeToColumnSchemaValue and convertSqlTypeToKType should be defined once

Issue #1273, opened 2025-06-24
Impossible to provide nested List/Array type in `DbType`


Jolanrensen:
Only today I found out DataFrame JDBC does some limited auto-conversion of `java.sql.Array` to `kotlin.Array<T>`, however, trying to leverage that for my own `DbType` proved to be impossible.

In `convertSqlTypeToKType`:
- If I give the type `java.sql.Array`, it gets read `like java.sql.Array`, as expected
- If I give it the type `kotlin.Array<*>`, it gets read like `kotlin.Array<*>`
- If I give it the type `kotlin.Array<Any?>`, it gets read like `kotlin.Array<Any?>`, auto-conversion, interesting
- If I give it the type `kotlin.Array<String?>` (or some other `T`), it gets read as `java.sql.Array` again?

This means I have no way to define the type of data in the array, even though my database provides this information.

The last case is caused by [this check here](https://github.com/Kotlin/dataframe/blob/62b48942b1aef35f939f2d0aff407872028fd177/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt#L1017). In Kotlin the classifier of an array is not equal to itself. It can be solved by writing it like `if ((kotlinTypesForSqlColumns[index]!!.classifier as KClass<*>).qualifiedName == Array::class.qualifiedName)`.

Still, doing that fix does not solve the problem, because [`handleArrayValues()`](https://github.com/Kotlin/dataframe/blob/62b48942b1aef35f939f2d0aff407872028fd177/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt#L1038) tries to infer the nested type using reflection, by getting the `componentType` of the Java class. This only works for primitive arrays, not when the array is an `Object[]`, which is often the case.

Probably solved best in combination with https://github.com/Kotlin/dataframe/issues/1266, because all conversions could better be done in a single place.

Jolanrensen:
Additionally, in notebooks, creating table in DuckDB without type-map like this:
```sql
CREATE TABLE IF NOT EXISTS test_table (
    id INTEGER PRIMARY KEY,
    name VARCHAR,
    age INTEGER,
    salary DOUBLE,
    hire_date DATE,
    pets VARCHAR[],
    map MAP(INT, VARCHAR),
)
```

causes:
```
Error compiling code: Unresolved reference: T
@DataSchema
interface _DataFrameType {
[...]

Issue #461, opened 2023-10-04
Add support for date/time types conversion in JDBC integration


zaleslaw:
The proposed idea adds data/time auto-discovering in the spirit of the CSV integration
[the link](https://github.com/Kotlin/dataframe/blob/913d6a0fe0fe2f30573b1f0c7696146457392f39/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/parse.kt#L218)

Jolanrensen:
I think it's different from parsing from strings, like in CSV; JDBC already has types for dates, so converting should be done via `convert`.

Should probably be handled by https://github.com/Kotlin/dataframe/issues/1266.

It might be a good idea to add `java.sql.time*` types in our `convert` implementation so it's easier to handle type conversions by `DbType`s

PR #1632, opened 2025-12-11
DbType2

### LINKED ISSUE #537: Add support for PostgreSQL geometry types
- [ ] PGbox
- [ ] PGcircle
- [ ] PGline
- [ ] PGlseg
- [ ] PGpath
- [ ] PGpoint
- [ ] PGpolygon
- [ ] PGmoney
- [ ] PGinterval

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1273
Fixes https://github.com/Kotlin/dataframe/issues/1587
Fixes https://github.com/Kotlin/dataframe/issues/461
Fixes #537

Helps https://github.com/Kotlin/dataframe/issues/387

Takes care of https://github.com/Kotlin/dataframe/issues/462, but only for DuckDB.

Follows up on https://github.com/Kotlin/dataframe/issues/1266 and https://github.com/Kotlin/dataframe/issues/462

### `DbType` was overhauled greatly:
- Removed `convertSqlTypeToColumnSchemaValue` and `convertSqlTypeToKType` in favor of `getExpectedJdbcType`
- removed  `extractValueFromResultSet` in favor of `getValueFromResultSet`
- introduced a new streamlined structure for reading from a `ResultSet` in `DbType`:
    - `getValueFromResultSet()` retrieves a value from a `ResultSet` given a column and row with the type given by `getExpectedJdbcType()`
    - Next, the value is optionally preprocessed by `preprocessValue()` with the new type given by `getPreprocessedValueType()`
    - Finally, the values are turned into a DataFrame column by `buildDataColumn()` with the new schema given by `getTargetColumnSchema()`
- In debug mode, the schema is checked at runtime.
- Column names might change, according to #387
- By default, we now preprocess Java's `LocalDateTime` to Kotlin's `LocalDateTime`, `java.sql.TimeStamp` to Kotlin `Instant` and UUIDs to Kotlin `Uuid`. More preprocessing might follow later.

### AdvancedDbType
- Introduced `AdvancedDbType` + `JdbcToDataFrameConverter` for easier Type-conversion logic

### DuckDB
- Completely rewritten to `AdvancedDbType`
- Supports all DuckDB types including Structs, Maps, Lists/Arrays (with recursive conversion!) and JSON #462
- This now makes it the most advanced SQL database we support completely!

### Other:
- `BuildConfig` now includes the module name to not clash with other modules.
- Added support for all PostgreSQL extension types https://github.com/Kotlin/dataframe/issues/537
- Ran all local DB tests

### DIFF
--- a/dataframe-jdbc/api/dataframe-jdbc.api
+++ b/dataframe-jdbc/api/dataframe-jdbc.api
@@ -255,6 +255,7 @@ public final class org/jetbrains/kotlinx/dataframe/io/db/PostgreSql : org/jetbra
 	public fun buildTableMetadata (Ljava/sql/ResultSet;)Lorg/jetbrains/kotlinx/dataframe/io/db/TableMetadata;
 	public fun getDriverClassName ()Ljava/lang/String;
 	public fun getExpectedJdbcType (Lorg/jetbrains/kotlinx/dataframe/io/db/TableColumnMetadata;)Lkotlin/reflect/KType;
+	public fun getValueFromResultSet (Ljava/sql/ResultSet;ILorg/jetbrains/kotlinx/dataframe/io/db/TableColumnMetadata;Lkotlin/reflect/KType;)Ljava/lang/Object;
 	public fun isSystemTable (Lorg/jetbrains/kotlinx/dataframe/io/db/TableMetadata;)Z
 	public fun quoteIdentifier (Ljava/lang/String;)Ljava/lang/String;
 }
--- a/dataframe-jdbc/build.gradle.kts
+++ b/dataframe-jdbc/build.gradle.kts
@@ -15,6 +15,7 @@ dependencies {
     api(projects.core)
     compileOnly(libs.duckdb.jdbc)
     compileOnly(libs.sqlite)
+    compileOnly(libs.postgresql)
     implementation(libs.kotlinLogging)
     testImplementation(libs.mariadb)
     testImplementation(libs.sqlite)
--- a/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/PostgreSql.kt
+++ b/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/PostgreSql.kt
@@ -1,10 +1,20 @@
 package org.jetbrains.kotlinx.dataframe.io.db
 
+import org.postgresql.geometric.PGbox
+import org.postgresql.geometric.PGcircle
+import org.postgresql.geometric.PGline
+import org.postgresql.geometric.PGlseg
+import org.postgresql.geometric.PGpath
+import org.postgresql.geometric.PGpoint
+import org.postgresql.geometric.PGpolygon
+import org.postgresql.util.PGInterval
+import org.postgresql.util.PGmoney
+import org.postgresql.util.PGobject
 import java.sql.ResultSet
 import java.util.Locale
 import kotlin.reflect.KType
+import kotlin.reflect.full.starProjectedType
 import kotlin.reflect.full.withNullability
-import kotlin.reflect.typeOf
 
 /**
  * Represents the PostgreSql database type.
@@ -16,14 +26,53 @@ public object PostgreSql : DbType("postgresql") {
     override val driverClassName: String
         get() = "org.postgresql.Driver"
 
-    override fun getExpectedJdbcType(tableColumnMetadata: TableColumnMetadata): KType {
-        // because of https://github.com/pgjdbc/pgjdbc/issues/425
-        if (tableColumnMetadata.sqlTypeName == "money") {
-            return typeOf<String>().withNullability(tableColumnMetadata.isNullable)
-        }
-        // TODO: Composite types like tableColumnMetadata.sqlTypeName = ROW("a" INTEGER, "b" CHARACTER VARYING(10))
+    /**
+     * Map of [PostgreSQL object][PGobject] types by name
+     * containing both their [Java class][Class] and [Kotlin type][KType].
+     *
+     * These types need to be retrieved explicitly with [Java class][Class] in [ResultSet.getObject], else
+     * their return type is unpredictable.
+     */
+    private val pgObjectTypes =
+        listOf(
+            PGbox(),
+            PGcircle(),
+            PGline(),
+            PGlseg(),
+            PGpath(),
+            PGpoint(),
+            PGpolygon(),
+            PGmoney(),
+            PGInterval(),
+        ).map(::PgObjectType)
+            .associateBy { it.typeName }
 
-        return super.getExpectedJdbcType(tableColumnMetadata)
+    // TODO: Composite types like tableColumnMetadata.sqlTypeName = ROW("a" INTEGER, "b" CHARACTER VARYING(10))
+    override fun getExpectedJdbcType(tableColumnMetadata: TableColumnMetadata): KType {
+        val typeName = tableColumnMetadata.sqlTypeName.lowercase()
+        return if (typeName in pgObjectTypes) {
+            pgObjectTypes[typeName]!!.kType.withNullability(tableColumnMetadata.isNullable)
+        } else {
+            super.getExpectedJdbcType(tableColumnMetadata)
+        }
+    }
+
+    /**
+     * Overridden so [PGobject] types are retrieved explicitly with [Java class][Class],
+     * else their return type is unpredictable.
+     */
+    override fun <J> getValueFromResultSet(
+        rs: ResultSet,
+        columnIndex: Int,
+        tableColumnMetadata: TableColumnMetadata,
+        expectedJdbcType: KType,
+    ): J {
+        val typeName = tableColumnMetadata.sqlTypeName.lowercase()
+        return if (typeName in pgObjectTypes) {
+            rs.getObject(columnIndex + 1, pgObjectTypes[typeName]!!.javaClass) as J
+        } else {
+            super.getValueFromResultSet(rs, columnIndex, tableColumnMetadata, expectedJdbcType)
+        }
     }
 
     override fun isSystemTable(tableMetadata: TableMetadata): Boolean =
@@ -40,5 +89,13 @@ public object PostgreSql : DbType("postgresql") {
     override fun quoteIdentifier(name: String): String {
         // schema.table -> "schema"."table"
         return name.split(".").joinToString(".") { "\"$it\"" }
+    }
+
+    private data class PgObjectType(val typeName: String, val kType: KType, val javaClass: Class<*>) {
+        constructor(pgObject: PGobject) : this(
+            typeName = pgObject.type,
+            kType = pgObject::class.starProjectedType,
+            javaClass = pgObject::class.java,
+        )
     }
 }


dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DuckDb.kt
```diff
@@ -79,83 +84,151 @@ public object DuckDb : DbType("duckdb") {
     /** the name of the class of the DuckDB JDBC driver */
     override val driverClassName: String = "org.duckdb.DuckDBDriver"
 
+    override fun generateTypeInformation(tableColumnMetadata: TableColumnMetadata): AnyTypeInformation =
```
koperagen: Please consider: ColumnValuesConverter, JdbcColumnConverter, ...
I think it's important to convey that we, most importantly, want to convert values from JDBC classes to something Kotlin-friendly - providing KType / ColumnSchema is just a side effect, right? ValuePreprocessor and ColumnPostprocessor are good in this regard

Jolanrensen: Actually, no, the main use case of this function is building the `ColumnSchema`, aka the type information for DataFrame, given some JDBC `TableColumnMetadata`, so we know what kind of DataFrame will pop out when some SQL is read.

Actually, most values can directly go from JDBC into a ValueColumn, no conversion needed. However, in the cases where conversion needs to be done (like Date -> Kotlin Date, Struct -> DataRow), this conversion can be provided in the `TypeInformation` class that's returned, either in the form of a value-preprocessor, or, if you need all values to be gathered first, in the form of a column-postprocessor. Still, they are strictly tied to the `TableColumnMetadata`.

Does that make sense? :) I've tried several names already for this `TypeInformation` class, (DbTypeInformation, DbColumnType...) but none really fit or become too large. But I'd like to convey that it does not háve to convert. If you do want to convert, you can create a `TypeInformation` instance with `typeInformationWithPreprocessingFor(...)`.

We could also split up providing the type information and actually converting them. However, I liked the idea of providing the type information and converting in one place, because it forces you to write a JDBC type, the conversion, and the schema that's created at once, keeping the logic together.

If you'd ignore what I made thus far, how would you make an API like this where you can define how JDBC types are mapped (and potentially converted) to column types?

koperagen: Ok, if i'm thinking from scratch: looks like you're referring to ResultSet (kind of Row) -> value extraction, that also can be database specific apparently? 
So people might need to specify what class their Jdbc driver will return to our "generic value extractor" of some sort. We're not converting anything at this step, just calling rs.getObject, and we need to know what to expect. 
If my description is accurate, can we call it JdbcDriverSpec? Can it be a separate entity from "converters"?
ExpectedTypes?

Jolanrensen: Yes, it can definitely be a separate step. 
However, the converters will still need access to the original TableColumnMetadata to function properly, and they might do some duplicate logic:

So, we have a couple cases:
- The first case is the easiest: We just want to read a column from JDBC, no conversion necessary. So the Db implementation could simply give a KType representing the type of the values coming from that column, and a `ValueColumn<ThatType>` will be created.
- The second case is a bit more complicated: We want to read a column of values from JDBC, but they need to be converted, like Dates. In this case, the implementation could give a KType again representing the type of the values coming from that column. A preprocessor could take those values and KType and convert them (and return a new KType) before a `DataColumn<NewType>` is created. The preprocessor might need the original `TableColumnMetadata` though, as it might have information that cannot be represented by just the KType.
- The next case is comparable to the previous one, but now we want to post-process the column, like in a column of arrays where we want to infer their common type. So the implementation will give a KType of `java.sql.Array`, it may be preprocessed, a `ValueColumn` is created, and then the postprocessor can do its magic, converting the `ValueColumn` with values to any sort of column it likes. It might need the original `TableColumnMetadata`, and `KType` and return the new column and ColumnSchema.
- The final case is where a structured column needs to be created: We read a `STRUCT` column from JDBC; the first step returns the KType `java.sql.Struct`. ~The preprocessor can convert each value to a `DataRow<*>` based on the KType and `TableColumnMetadata`, so a `DataColumn<DataRow<*>`, aka a `ColumnGroup` can be created. Though, we still need to report the new `ColumnSchema.Group` somewhere so we can do `TableColumnMetadata -> ColumnSchema` without reading actual data. Maybe in the postprocessor?~ The preprocessor should do nothing, making a `ValueColumn<Struct>`, then the post-processor should turn it in a `ColumnGroup`, returning the resulting `ColumnSchema.Group` as well.

In the final case, we can see the `TableColumnMetadata` needing being parsed up to 3 separate times: In the first step to generate a `KType` we're not even using, as it's just `typeOf<AnyRow>()`, in the second step, to convert each value to a correct `DataRow`, and in the final step to produce the right `ColumnSchema.Group`. This can be quite tedious, as `TableColumnMetadata` can contain recursive types as well... It's also hard to track the logic of a single type across multiple separate functions...
But maybe we can find a hybrid between these two approaches? Allowing logic to be grouped, but functions to be separate?

koperagen: > The final case is where a structured column needs to be created: We read a STRUCT column from JDBC; the first step returns the KType java.sql.Struct. The preprocessor can convert each value to a DataRow<*> based on the KType and TableColumnMetadata, so a DataColumn<DataRow<*>, aka a ColumnGroup can be created. Though, we still need to report the new ColumnSchema.Group somewhere so we can do TableColumnMetadata -> ColumnSchema without reading actual data. Maybe in the postprocessor

As a side note, in recent Map.toDataFrame PR i noticed that map.toDataRow that uses type inference hits a very obvious bottleneck in reflective type inference that is called for each row, each value individually. In that case column-based creation of ColumnGroup reduces time from 17s to 1.5s! 
I think efficient transformation of ResultSet with Struct value should be done in ColumnPostprocessor maybe? Like DataColumn<Struct> -> ColumnGroup.

koperagen: > However, the converters will still need access to the original TableColumnMetadata to function properly, and they might do some duplicate logic:

Makes sense

Jolanrensen: :o 

Yes, it makes sense to postpone it to the post-processing step indeed! A `DataRow<*>` is a `DataFrame` with one row after all, so forming a DF with 1000 rows will create 1000 intermediate DFs with type inference. We'd need https://github.com/Kotlin/dataframe/issues/1541 to be able to do this efficiently.

But this just shows it's good to have both pre- and post-processing :) we need both.

When a `FrameColumn` is created, it does make sense to create `DataFrame`s in the preprocessing step

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -883,38 +919,69 @@ internal fun fetchAndConvertDataFromResultSet(
...
+                expectedJdbcType = expectedJdbcType,
+            )
+            val preprocessedValue = dbType.preprocessValue<Any, Any>(
+                value = value,
+                tableColumnMetadata = tableColumnMetadata,
+                expectedJdbcType = expectedJdbcType,
+                expectedPreprocessedValueType = preprocessedValueType,
             )
-            columnData[columnIndex].add(value)
+            columnData[name]!!.add(preprocessedValue)
```
zaleslaw: Is the name unique here? i prefer to be built on indices

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -866,15 +874,43 @@ internal fun getTableColumnsMetadata(resultSet: ResultSet, dbType: DbType): Muta
...
     dbType: DbType,
+    tableColumns: List<TableColumnMetadata>,
+    rs: ResultSet,
     limit: Int?,
     inferNullability: Boolean,
 ): AnyFrame {
-    val columnKTypes = buildColumnKTypes(tableColumns, dbType)
-    val columnData = readAllRowsFromResultSet(rs, tableColumns, columnKTypes, dbType, limit)
-    val dataFrame = buildDataFrameFromColumnData(columnData, tableColumns, columnKTypes, dbType, inferNullability)
+    val expectedJdbcTypes = getExpectedJdbcTypes(
```
zaleslaw, line 898: ExpectedJdbcTypes could be more complex structure, but better to keep order of column according indicies for debugging, not name-based, it could be edge object with fields: index, name, KType for example

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
```diff
@@ -866,15 +874,43 @@ internal fun getTableColumnsMetadata(resultSet: ResultSet, dbType: DbType): Muta
...
+        dbType = dbType,
+        tableColumns = tableColumns,
+    )
+    val preprocessedValueTypes = getPreprocessedValueTypes(
```
zaleslaw, line 902: Describe somewhere the processes

Expected->Preprocessed (what's the difference and why we need this step, what does it give)

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readDataFrameSchema.kt
```diff
@@ -98,6 +95,8 @@ public fun DataFrameSchema.Companion.readSqlTable(
 ): DataFrameSchema {
     val determinedDbType = dbType ?: extractDBTypeFromConnection(connection)
 
+    // TODO don't need to read 1 row, take it just from TableColumnMetadatas
```
zaleslaw, line 104: It's very safe and cheap way for any database, I moved from taking info from TableColumnMetadata, also lead to less error - prone in our codebase and make it flexible

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/MySql.kt
```diff
@@ -18,12 +16,11 @@ public object MySql : DbType("mysql") {
     override val driverClassName: String
         get() = "com.mysql.jdbc.Driver"
 
-    override fun convertSqlTypeToColumnSchemaValue(tableColumnMetadata: TableColumnMetadata): ColumnSchema? {
+    override fun getExpectedJdbcType(tableColumnMetadata: TableColumnMetadata): KType {
```
zaleslaw, line 19: I want to change keep naming a la convertJdbcTypeToKType

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/MariaDb.kt
```diff
@@ -50,26 +46,6 @@ public object MariaDb : DbType("mariadb") {
             tables.getString("table_cat"),
         )
 
-    override fun convertSqlTypeToKType(tableColumnMetadata: TableColumnMetadata): KType? {
```
zaleslaw, line 53: Great refactoring

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -82,10 +95,210 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
...
+     * @param [columnIndex] zero-based column index
+     * @return the extracted value, or null
      */
-    public abstract fun convertSqlTypeToColumnSchemaValue(tableColumnMetadata: TableColumnMetadata): ColumnSchema?
+    public open fun <J : Any> getValueFromResultSet(
+        rs: ResultSet,
+        columnIndex: Int,
+        tableColumnMetadata: TableColumnMetadata,
+        expectedJdbcType: KType,
+    ): J? =
```
zaleslaw: J?

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -82,10 +95,210 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
...
+
+    // TODO add preprocessors for common types, like sql Arrays, Java datetimes, etc.
+    public open fun <J : Any, D : Any> preprocessValue(
+        value: J?,
+        tableColumnMetadata: TableColumnMetadata,
+        expectedJdbcType: KType,
+        expectedPreprocessedValueType: KType,
+    ): D? =
+        when (tableColumnMetadata.jdbcType) {
+            Types.TIMESTAMP if tableColumnMetadata.javaClassName == "java.time.LocalDateTime" ->
```
zaleslaw, line 415: Why are Time types processed on that stage?

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
```diff
@@ -196,213 +401,6 @@ public abstract class DbType(public val dbTypeInJdbcUrl: String) {
...
-            Types.VARBINARY to ByteArray::class,
-            Types.LONGVARBINARY to ByteArray::class,
-            Types.NULL to String::class,
-            Types.JAVA_OBJECT to Any::class,
-            Types.DISTINCT to Any::class,
-            Types.STRUCT to Any::class,
-            Types.ARRAY to Array::class,
-            Types.BLOB to ByteArray::class,
-            Types.CLOB to Clob::class,
-            Types.REF to Ref::class,
```
zaleslaw, line 343: Is it kept as is or changed somewhere?

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
Jolanrensen: of course it's unique, columns are always uniquely accessible by name, no? In DataFrame this is very obvious as each DataFrame is a Map of String to Column. I'd say names are even more stable than indices. Is this any different for JDBC tables?

Jolanrensen: From the JDBC sources: `Column names used as input to getter methods are case insensitive. When a getter method is called with a column name and several columns have the same name, the value of the first matching column will be returned. The column name option is designed to be used when column names are used in the SQL query that generated the result set. For columns that are NOT explicitly named in the query, it is best to use column numbers. If column names are used, the programmer should take care to guarantee that they uniquely refer to the intended columns, which can be assured with the SQL AS clause`

So you're right :) we should use indices indeed! Plus, we should have a system in place that handles columns with the same or no name. I however do think a `Map<Int, KType>` we had before is a waste of resources. A simple `List<KType>` achieves the same thing without having to store integers explicitly as keys

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readJdbc.kt
Jolanrensen: That's certainly possible! Though that would look a bit more like `AdvancedDbType` which, for each column, requires you to provide an `AnyJdbcToDataFrameConverter` containing all information needed to read and convert that column. Maybe the concept could be merged

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
Jolanrensen: J-DBC. It's clearer in the context of `AdvancedDbType`:
- `J`-DBC type
- `D`-ataFrame type
- `P`-ost processed type

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/MySql.kt
Jolanrensen: There's no converting happening at this stage yet. As a user you're supposed to provide the KTypes of the values coming directly from JDBC

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt
Jolanrensen: As is. It's the expected types coming from JDBC but I use KTypes now instead of `::class` because it's more efficient and flexible.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readDataFrameSchema.kt
Jolanrensen: Except for empty databases that just have a schema and no data. When building just the schema, we shouldn't have to look at the actual database contents. It's up to the `DbType` implementor to provide a watertight `getExpectedJdbcType()`, `getPreprocessedValueType()`, and `getTargetColumnSchema()` from just `TableColumnMetadata`. So if we manage to construct `TableColumnMetadata`s from `connection`, `tableName` and `dbType`, we can simply call those functions to get a schema without accessing any data.

Jolanrensen: ...and for databases you may have no query access to

Jolanrensen:
@zaleslaw I'm not sure what to do about `java.time.OffsetDateTime` and `java.time.OffsetTime`. Do you know if there's a kotlin equivalent?

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readDataFrameSchema.kt
Jolanrensen: https://github.com/Kotlin/dataframe/issues/1681

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/AdvancedDbType.kt
```diff
@@ -0,0 +1,84 @@
...
+ * This API is experimental and subject to change.
+ */
+public abstract class AdvancedDbType(dbTypeInJdbcUrl: String) : DbType(dbTypeInJdbcUrl) {
+
+    protected abstract fun generateConverter(tableColumnMetadata: TableColumnMetadata): AnyJdbcToDataFrameConverter
+
+    private val converterCache = mutableMapOf<TableColumnMetadata, AnyJdbcToDataFrameConverter>()
+
+    protected fun getConverter(tableColumnMetadata: TableColumnMetadata): AnyJdbcToDataFrameConverter =
+        converterCache.getOrPut(tableColumnMetadata) {
```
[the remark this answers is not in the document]
Jolanrensen: wow, good point!

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DuckDb.kt
```diff
@@ -196,6 +338,17 @@ public object DuckDb : DbType("duckdb") {
...
 
+    /** Parses "STRUCT(v VARCHAR, i INTEGER)" into [("v", "VARCHAR"), ("i", "INTEGER")] */
+    internal fun parseStructType(typeString: String): Map<String, String> {
+        if (!typeString.startsWith("STRUCT(")) {
+            error("invalid STRUCT type: $typeString")
+        }
+        return typeString.removeSurrounding("STRUCT(", ")")
+            .split(",")
+            .map { it.trim().split(" ") }
+            .associate { (name, type) -> name to type }
```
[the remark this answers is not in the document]
Jolanrensen: again good point

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/local/duckDbTest.kt
```diff
@@ -310,7 +328,19 @@ class DuckDbTest {
             subset = DataFrame.readSqlQuery(connection, """SELECT test_table.name, test_table.age FROM test_table""")
         }
 
-        schema.compare(Person.expected.schema()).isSuperOrMatches() shouldBe true
+        withClue({
```
zaleslaw, line 331: It looks like we have  code in commonTestScenarios    with the same intention, could we move this schema validation there unify it together somehow?

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/h2/postgresH2Test.kt
```diff
@@ -123,6 +123,18 @@ class PostgresH2Test {
 
             connection.createStatement().execute(createTableQuery.trimIndent())
 
+            // Table with STRUCT/ROW column to verify JDBC Types.STRUCT handling
+            @Language("SQL")
```
zaleslaw, line 127: Please port these PostgreSQL test scenarios to the `PostgresTest` and test them locally.

Basically the PostgreSQL test scenarios is green for me

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/h2/mysqlH2Test.kt
```diff
@@ -228,8 +230,8 @@ class MySqlH2Test {
                     st.setDouble(11, i * 10.0)
                     st.setBigDecimal(12, BigDecimal(i * 10))
                     st.setDate(13, java.sql.Date(System.currentTimeMillis()))
-                    st.setTimestamp(14, java.sql.Timestamp(System.currentTimeMillis()))
-                    st.setTimestamp(15, java.sql.Timestamp(System.currentTimeMillis()))
+                    st.setTimestamp(14, Timestamp(System.currentTimeMillis()))
```
zaleslaw, line 233: Install locally MySQL and port these test changes, the same for MariaDB, for the proper tests `mariadbTest`, `mysqlTest`

Now it fails 

```
        schema.columns["datetimeCol"]!!.type shouldBe typeOf<java.sql.Timestamp>()
        schema.columns["timestampCol"]!!.type shouldBe typeOf<java.sql.Timestamp>()
```

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/h2/mssqlH2Test.kt
```diff
@@ -200,11 +201,11 @@ class MSSQLH2Test {
         schema.columns["nvarcharColumn"]!!.type shouldBe typeOf<String?>()
         schema.columns["nvarcharMaxColumn"]!!.type shouldBe typeOf<String?>()
         schema.columns["realColumn"]!!.type shouldBe typeOf<Float?>()
-        schema.columns["smalldatetimeColumn"]!!.type shouldBe typeOf<java.sql.Timestamp?>()
+        schema.columns["smalldatetimeColumn"]!!.type shouldBe typeOf<Instant?>()
```
zaleslaw, line 204: The same thing with fixing tyme types

Jolanrensen:
@zaleslaw I reverted `DbType.getValueFromResultset()` back to its original implementation. It turns out `rs.getObject(i, javaClass)` fails when `javaClass` is a supertype the actual type of the column in question. That delivers a whole bunch of unnecessary exceptions.

Instead, I fixed #537 and its "money" type by only calling `rs.getObject(i, javaClass)` inside `PostgreSql : DbType` for the `PGobject` extension types. This satisfies all tests consistently without causing potential issues with other SQL types.

All local tests pass now

PR #1634, opened 2025-12-11
Remove unused testing utility ColumnAccessTracker

koperagen:
Not much to add. Used once in a tiny test, but wastes CPU cycles on every column (row) access. Time to go


Jolanrensen:
I mean, it's a nice idea for the debug mode... if we would actually use it

koperagen:
Before:
25.525 ±(99.9%) 2.955 ms/op [Average]
  (min, avg, max) = (25.093, 25.525, 26.890), stdev = 0.767

After:
20.654 ±(99.9%) 2.057 ms/op [Average]
  (min, avg, max) = (19.959, 20.654, 21.360), stdev = 0.534

```kt
package org.jetbrains.kotlinx.dataframe

import kotlinx.benchmark.Benchmark
import kotlinx.benchmark.BenchmarkMode
import kotlinx.benchmark.BenchmarkTimeUnit
import kotlinx.benchmark.Blackhole
import kotlinx.benchmark.Measurement
import kotlinx.benchmark.Mode
import kotlinx.benchmark.OutputTimeUnit
import kotlinx.benchmark.Scope
import kotlinx.benchmark.Setup
import kotlinx.benchmark.State
import kotlinx.benchmark.Warmup
import org.jetbrains.kotlinx.dataframe.api.dataFrameOf
import org.jetbrains.kotlinx.dataframe.api.forEach

@State(Scope.Benchmark)
@Warmup(iterations = 5, time = 1)
@Measurement(iterations = 5, time = 1)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(BenchmarkTimeUnit.MILLISECONDS)
open class RowValueAccessBenchmark {

    private lateinit var df: AnyFrame

    @Setup
    fun setup() {
        val i = 1_000_000
        df = dataFrameOf(
            "name" to (1..i).map { "Person$it" },
            "age" to (1..i).map { 20 + it % 50 },
            "city" to (1..i).map { "City${it % 100}" },
            "salary" to (1..i).map { 50000.0 + it * 10 },
            "isActive" to (1..i).map { it % 2 == 0 }
        )
    }

    @Benchmark
    fun stringLookup(blackhole: Blackhole) {
        df.forEach { row ->
            val name = row["name"] as String
            val age = row["age"] as Int
            val city = row["city"] as String
            blackhole.consume(name)
            blackhole.consume(age)
            blackhole.consume(city)
        }
    }
}
```

Issue #90, opened 2022-02-17
Create DataFrame from list of rows where each row is Map


PoslavskySV:
Hi guys, 

it would be nice to add a method for creating a DataFrame from a list of rows represented as general `Map`s. Right now when I do:
```kotlin
val rows : List<Map<String, Any?>>

val df = rows.toDataFrame()
```
I get a wired result - DataFrame with columns obtained from the properties of `Map` class. But it is more intuitive to get a `DataFrame` with columns obtained from the keys of `Map`s. Does it make sense for you?

koperagen:
Does this map only have primitive values? I'm asking because in that case we can simply create an overload for List<Map<String, Any?>>. But if you need to convert objects inside this Map into dataframe structures, then Map class should be supported as a special case in existing Iterable<*>.toDataFrame()

PoslavskySV:
I am not sure that I got your question. Map values can have arbitrary type (same for each Map of iterable). For example, suppose I have a list (or any other iterable) of maps with the following structure:
```
{
    "col1" -> String,
    "col2" -> Double,
    "col3" -> MyEnum
}
```
then from the `rows.toDataFrame()` I would expect to obtain `AnyFrame` with the columns `col1` of type `String`, `col2` of type Double etc. Does it sound meaningful?

koperagen:
Yes. Let me clarify the question. If you have col4 in your map with the type `class Name(val firstName: String, val lastName: String)`, we can convert it in two ways:
1. To `DataColumn<Name>` 
2. To `ColumnGroup`  with 2 columns, firstName and lastName (it's like `Iterable<*>.toDataFrame(depth = 2)` would work for classes)
Do you need 1 or 2?

PoslavskySV:
Thank you for clarification, now I see! I need option 1, and generally I feel that it is more natural here.

ian-k:
@PoslavskySV 
Till the team implements the proper solution you can use this extension function:
fun List<Map<String, Any>>.toDataFrame(): AnyFrame {
    val columns: MutableMap<String, MutableList<Any?>> = mutableMapOf()
    val notNullCols: MutableSet<String> = mutableSetOf()
    val columnSize=this.size

    this.forEachIndexed{ ri,row ->
        run {

            row.keys.forEach { col ->
                run {
                    if (columns[col] == null) {
                        columns[col] = mutableListOf()
                    }

                    val value=if (row[col].let { it is String && it.isEmpty() }) null else row[col]
[...]

PR #1635, opened 2025-12-11
Optimize Iterable<Map>.toDataFrame conversion

koperagen:
Now it hits a lot of heavy reflection calls like `type.isSubtypeOf<AnyRow?>()`, `AnyFrame` on each `Map -> DataRow`. 
Collecting values first, inferring type for all gives a visible improvement

17 s -> 1.5 s

Code i used for testing:

```kt
val rowCount: Int = 1_000_000
val columnCount: Int = 10
val columns = (0 until columnCount).map { "col$it" }
val maps: List<Map<String, Any?>> = (0 until rowCount).map { rowIdx ->
    columns.associate { col ->
        col to when (columns.indexOf(col) % 4) {
            0 -> rowIdx
            1 -> "value_$rowIdx"
            2 -> rowIdx * 1.5
            3 -> rowIdx % 2 == 0
            else -> null
        }
    }
}

maps.toDataFrame()
```

fixes https://github.com/Kotlin/dataframe/issues/90

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
@@ -65,7 +65,25 @@ public fun <T> Iterable<DataRow<T>>.toDataFrame(): DataFrame<T> {
 }
 
 @JvmName("toDataFrameMapStringAnyNullable")
-public fun Iterable<Map<String, *>>.toDataFrame(): DataFrame<*> = map { it.toDataRow() }.toDataFrame()
+public fun Iterable<Map<String, Any?>>.toDataFrame(): AnyFrame {
+    val list = asList()
+    if (list.isEmpty()) return DataFrame.empty()
+
+    val allKeys = linkedSetOf<String>()
+    for (row in this) {
+        allKeys.addAll(row.keys)
+    }
+
+    val columns = allKeys.map { key ->
+        val values = ArrayList<Any?>(list.size)
+        for (row in this) {
+            values.add(row[key])
+        }
+        DataColumn.createByInference(key, values)
+    }
+
+    return columns.toDataFrame()
+}
 
 @JvmName("toDataFrameAnyColumn")
 public fun Iterable<AnyBaseCol>.toDataFrame(): AnyFrame = dataFrameOf(this)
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
```diff
@@ -65,7 +65,25 @@ public fun <T> Iterable<DataRow<T>>.toDataFrame(): DataFrame<T> {
 }
 
 @JvmName("toDataFrameMapStringAnyNullable")
-public fun Iterable<Map<String, *>>.toDataFrame(): DataFrame<*> = map { it.toDataRow() }.toDataFrame()
+public fun Iterable<Map<String, Any?>>.toDataFrame(): AnyFrame {
+    val list = asList()
+    if (list.isEmpty()) return DataFrame.empty()
+
+    val allKeys = linkedSetOf<String>()
```
Jolanrensen, line 72: :o TIL what a linked HashSet is

Jolanrensen: though, it seems `mutableSetOf` produces the same thing in Kotlin :) so I thought all sets behaved this way. Still it's nice to be expressive :D

Jolanrensen (approved):

Jolanrensen:
I wonder... could we do the same trick with `Iterable<DataRow<*>>.toDataFrame()`?

Currently, it's only optimized when all DataRows originate from the same DF, but else it just calls `iterable.map { it.toDataFrame() }.concat()` :(

Actually, we can just call `map { it.toMap() }.toDataFrame().cast()` now XD

koperagen:
> Actually, we can just call map { it.toMap() }.toDataFrame().cast() now XD

Lol. How tables have turned! Yes, i'm positively surprised by how relatively straightforward this is, at least without recursive value conversion. With DataRow we probably need to create nested column groups, so it will be a bit trickier

Jolanrensen:
> > Actually, we can just call map { it.toMap() }.toDataFrame().cast() now XD
> 
> Lol. How tables have turned! Yes, i'm positively surprised by how relatively straightforward this is, at least without recursive value conversion. With DataRow we probably need to create nested column groups, so it will be a bit trickier

ah yes, of course! let's do that separately :) I like this recent performance-improvement-trend ;P

PR #1636, opened 2025-12-11
Lazy statistics for ValueColumn

### LINKED ISSUE #1492: Lazy statistics for columns
Let's say you write

```kt
df.filter { someValue > df().myColumn.max() }
```

This is way faster:
```kt
val myColumnMax = df.myColumn.max()
df.filter { someValue > myColumnMax }
```

Maybe we could solve this by having lazily calculated stats stored inside `ValueColumn`s. Columns are immutable after all, so it would be safe to do so and the performance gain should be significant!

Of course, this wouldn't work when you write:
```kt
df.filter { someValue > (myColumn + 1).max() }
// or
df.filter { someValue > myColumn.maxOf { it + 1 } }
```

but that's okay I think. We can't have it all :)

CarloMariaProietti:
Fix #1492 
The idea is the following:
ValueColumnInternal is an interface for statistic values, which in this way are not exposed as public.
Implementations of ValueColumnInternal contain the actual cache.

It was necessary to have two caches for each stat (for the moment only max) because computing the stat may give different outputs basing on skipNaN boolean parameter.

I implemented the solution by overloading `aggregateSingleColumn`, this overload exploits the original `aggregateSingleColumn` by wrapping it so that it is possible to exploit caches.

For the moment there is only `max`, however it would be easy to do the same with `min`, `sum`,  `mean `and `median`.
For `percentile `and `std `it could be done something similar.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnImpl.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnImpl.kt
@@ -17,10 +17,8 @@ public class ParameterValue(public val parameter: Any?) {
         val otherAsParameterValue = other as ParameterValue?
         val that = otherAsParameterValue?.parameter
         if (parameter is Boolean && that is Boolean) {
-            println("my_equals")
             return this.parameter == that
         }
-        println("default_equals")
         return super.equals(other)
     }
 


git
Jolanrensen, line 1: please remove this from the commit

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnImpl.kt
```diff
@@ -8,14 +8,26 @@ import org.jetbrains.kotlinx.dataframe.columns.ValueColumn
 import kotlin.reflect.KType
 import kotlin.reflect.full.withNullability
 
+public class WrappedStatistic(
```
Jolanrensen: this class should not be public, should it?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
```diff
@@ -33,7 +34,11 @@ public fun <T : Comparable<T>> DataColumn<T?>.max(skipNaN: Boolean = skipNaNDefa
     maxOrNull(skipNaN).suggestIfNull("max")
 
 public fun <T : Comparable<T>> DataColumn<T?>.maxOrNull(skipNaN: Boolean = skipNaNDefault): T? =
-    Aggregators.max<T>(skipNaN).aggregateSingleColumn(this)
+    if (this is ValueColumnInternal<*>) {
```
Jolanrensen: I'm not so fond of this solution, as it requires a lot of refactoring in other functions, plus it does not work when you write `df.max { myCol }`, as I mentioned in https://github.com/Kotlin/dataframe/issues/1492#issuecomment-3562802137

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnImpl.kt
```diff
@@ -8,14 +8,26 @@ import org.jetbrains.kotlinx.dataframe.columns.ValueColumn
...
+    public var wasComputedSkippingNaN: Boolean = false,
+    public var wasComputedNotSkippingNaN: Boolean = false,
+    public var statisticComputedSkippingNaN: Any? = null,
+    public var statisticComputedNotSkippingNaN: Any? = null,
+)
+
+internal interface ValueColumnInternal<T> : ValueColumn<T> {
+    val max: WrappedStatistic
```
Jolanrensen: I would make this a `var` and nullable, so we can initialize it to `null` and don't need to instantiate a class for each statistic when a column is created.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnImpl.kt
Jolanrensen: this class should not be public, should it?

Also, I think, if you make the other a `var`, this can be a `data class` with `var`'s. It's a bit more kotlin-like :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnImpl.kt
```diff
@@ -8,14 +8,26 @@ import org.jetbrains.kotlinx.dataframe.columns.ValueColumn
 import kotlin.reflect.KType
 import kotlin.reflect.full.withNullability
 
+public class WrappedStatistic(
+    public var wasComputedSkippingNaN: Boolean = false,
+    public var wasComputedNotSkippingNaN: Boolean = false,
+    public var statisticComputedSkippingNaN: Any? = null,
+    public var statisticComputedNotSkippingNaN: Any? = null,
+)
```
Jolanrensen: what about `std` and `percentile` that take extra arguments?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
Jolanrensen: Instead. I'd do this check inside the original `aggregateSingleColumn()`. Each `Aggregator` has a `name` which you could use to query the `ValueColumnInternal` for the right `WrappedStatistic` if they are stored in a `Map<String, WrappedStatistic>` in `ValueColumnImpl`.
Though I suppose each `Aggregator` will also need to store any other provided arguments like `skipNaN: Boolean` and `percentile: Double` when needed... In a `Map<String, Any?>` maybe?

That way we could store our "Statistics Cache" in `ValueColumnImpl` as a 
```kt
Map<String, Map<Map<String, Any?>, Any?>>
```
so the result cache could look like:
```
{
   "max" : {
        { "skipNaN": true } : 312.4
    },
   "min" : {},
    "std" : {
        { "std": 0.9, "skipNaN": false } : Double.NaN,
        { "std": 0.9, "skipNaN": true } : 12.3
    }
}
```

The challenge may lie in doing this neatly ;P

Jolanrensen (changes requested):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
CarloMariaProietti: Thank you for reviewing!

Imo using `Map<String, Map<Map<String, Any?>, Any?>>` introduces a problem,
making a query to this structure does not allow to know if the statistic was computed 'in the past'.
Computing the stat using `aggregateSequence` implies that the stat can be null, so making a query and getting null does not tell me whether the stat was computed yet.

Maybe it could be `Map<String, Map<Map<String, Any?>, WrappedStatistic>>`
where WrappedStatistic has two fileds : `wasComputed: Boolean `and `actualStatistic: Any?` ?

Jolanrensen: If you're gonna put the statistic result in a wrapped class anyway, you can also make the value type of the map nullable :) Like

`value class StatisticResult(val value: Any?)` and  `Map<..., StatisticResult?>`. Then if `map[key] == null`, we know it has not been calculated, else we can take `map[key].value`. Saves you a lot of boolean-juggling :)

CarloMariaProietti: Using` value class StatisticResult(val value: Any?)`,
at the moment this is how I implemented aggregateSingleColumn 
```
public fun aggregateSingleColumn(column: DataColumn<Value?>): Return {
        if (column is ValueColumnInternal<*>) {
            //println("ValueColumnInternal")
            // cache check, cache is dynamically created
            val aggregator = this.aggregator ?: throw IllegalStateException("Aggregator is required")
            val desiredStatisticNotConsideringParameters = column.statistics.getOrPut(aggregator.name) {
                mutableMapOf<Map<String, ParameterValue?>, StatisticResult>()
            }
            // can't compare maps whose Values are Any? -> ParameterValue which overrides equals instead
            val desiredStatistic = desiredStatisticNotConsideringParameters[aggregator.statisticsParameters]
            // if desiredStatistic is null, statistic was never calculated
            if (desiredStatistic != null) {
                //println("cache hit")
                return desiredStatistic.value as Return
            }
            //println("cache miss")
            val statistic = aggregateSequence(
                values = column.asSequence(),
                valueType = column.type().toValueType(),
            )
            desiredStatisticNotConsideringParameters.put(aggregator.statisticsParameters, StatisticResult(statistic))
            return aggregateSingleColumn(column)
        }
        return aggregateSequence(
            values = column.asSequence(),
            valueType = column.type().toValueType(),
        )
    }
```

CarloMariaProietti:
I have done some refactoring at aggregators level, now it should work according to https://github.com/Kotlin/dataframe/pull/1636#discussion_r2614613670, now also `min` exploits cache.
Current implementation should also accomplish https://github.com/Kotlin/dataframe/pull/1636#discussion_r2614643350 and https://github.com/Kotlin/dataframe/pull/1636#discussion_r262758260.
By exploiting this implementation, it should be easy to exploit cache also for other statistics.

`ParameterValue `is a class created so that it can override equals and hashCode -> I can correctly query the statistic cache.

CarloMariaProietti:
I have closed and reopened this PR beacause I thought 'Files changed' section was 'bugged', I was wrong :)

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/AggregatorAggregationHandler.kt
```diff
@@ -26,13 +29,37 @@ public interface AggregatorAggregationHandler<in Value : Any, out Return : Any?>
...
      * Aggregates the data in the given column and computes a single resulting value.
-     * Calls [aggregateSequence].
+     * Calls [aggregateSequence]. It tries to exploit a cache for statistics which is proper of
+     * [ValueColumnInternal]
      */
-    public fun aggregateSingleColumn(column: DataColumn<Value?>): Return =
-        aggregateSequence(
+    public fun aggregateSingleColumn(column: DataColumn<Value?>): Return {
+        if (column is ValueColumnInternal<*>) {
+            println("ValueColumnInternal")
```
Jolanrensen, line 37: please remove these printlns

CarloMariaProietti: This PR is bugged, the prints (but non only) are not in the actuale code (what I see in my IDE). I'm sorry for this inconvenience. I should create a new PR I think, it also appears that 11 files was modified but this is not true.

CarloMariaProietti: #1656 is the fixed PR which contains the actual implementation.

Jolanrensen: you can also try force pushing to your branch to make sure the version in this PR is the same as your local version. It's cleaner than creating a new PR

CarloMariaProietti: It appears that my local branch is up to date with remote one, so pushing makes no effect.

Jolanrensen: force push

Jolanrensen: that overwrites the remote branch entirely, making it equal

CarloMariaProietti: I tried `git push --force my_fork lazy_stats` but still got `Everything up-to-date`.
Yesterday I had to delete and re-clone the repository beacause of building issues. Maybe I made some mistakes in this process. However the new PR  is 'clean'. I created the new branch by simply starting from master and copy-pasting the code.

Jolanrensen:
Alright, continued by https://github.com/Kotlin/dataframe/pull/1656

PR #1638, opened 2025-12-12
Avoid sending unused HTML tables in internal Kotlin Notebook <-> Kernel table widget JSON data exchange

koperagen:
I noticed that on every new page we send all our HTML resources+rendered table to Kotlin Notebook plugin.
Thing is, requests from Kotlin Notebook do not update "static" output of the cell, so it's unused, won't be seen by anyone

Let's update mime result of getRowsSubsetForRendering to only return JSON. Change is made to not affect anything else by default


Jolanrensen:
@Allex-Nik should probably check whether this breaks the intellij plugin. Adding a field to a data class is a breaking change.

Jolanrensen (approved):

Allex-Nik:
It doesn't break the plugin, the plugin works with these changes :)

Issue #1436, opened 2025-09-09
Effecient KotlinNotebokPluginUtils.sortByColumns


koperagen:
Take 10_000_000 rows, open it in a table widget in notebooks and sort by a column
Sorting and loading take a lot of time. I don't have a profile so i can't say for sure where's the actual bottleneck, it needs to be investigated
However this method performs sorting of the entire dataframe, all 10 million rows even when only 20 or 100 are going to be displayed. There's more efficient algorithms for such situations, for example least or greatest from Guava: https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/Comparators.html#least(int,java.util.Comparator)

Jolanrensen:
Good idea!

We use a similar algorithm for quick select in our percentile/median/quantile implementation: https://github.com/Kotlin/dataframe/blob/b46524691922c1c49c5258b2f74d7ac8aa817c85/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/math/quantile.kt#L281 though this only returns a single element.

According to the [TopKSelect source](https://fossies.org/dox/guava-33.4.8/android_2guava_2src_2com_2google_2common_2collect_2TopKSelector_8java_source.html) their solution uses less memory than quickselect :)

koperagen:
Worth to look at serialization/deserialization too, just sorting 10 million rows shouldn't really take long, so i suspect something else affecting the performance

PR #1639, opened 2025-12-12
Fast sorting of columns/table in Kotlin Notebook

koperagen:
I suggest to look at this PR by individual commits to see how code evolved
First iteration is simple and managed to achieve sub 1s speed for 1 m rows, which is acceptable
Later i introduced wrapper that exploits the fact that Kotlin Notebook takes small part of sorted dataframe, so after sorting we can avoid materializing the entire thing
And, since we can only sort 1 column at a time now, added special case of sorting just 1 column. Comparator can be a bit faster. + It was easy to apply Arrays.parallelSort to that, which all combined gave us another magnitude of speedup
So now sorting of 1 m rows is within ~30-50 ms on my PC, depends on how much core you have.

Bonus: it's now possible to sort columns with DataFrame and List values by their size

Intermediate results after removing reflection from sorting loop:
| Type | Size | eager |
| :--- | :--- | :--- |
| int | 1000000 | 66.746 |
| category | 1000000 | 130.657 |
| double | 1000000 | 519.548 |
| string | 1000000 | 366.065 |
| list | 1000000 | 252.999 |
| frame | 1000000 | 246.892 |

Final results with lazy materialization + improved comparators for optimal bytecode where possible + Arrays.parallelSort:

| Type | Size | optimized |
| :--- | :--- | :--- |
| int | 1000000 | 3.562 |
| category | 1000000 | 7.268 |
| double | 1000000 | 30.330 |
| string | 1000000 | 42.588 |
| list | 1000000 | 32.648 |
| frame | 1000000 | 42.252 |

fixes https://github.com/Kotlin/dataframe/issues/1436

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
@@ -73,6 +73,10 @@ public object KotlinNotebookPluginUtils {
     /**
      * Sorts a dataframe-like object by multiple columns.
      * If a column type is not comparable, sorting by string representation is applied instead.
+     * Sorts DataFrames by their size because looking at the smallest / biggest groups after groupBy is very popular.
+     *
+     * Returns "lazily materialized" dataframe, which means get, getRows, take operation must be applied to turn it to a valid sorted dataframe.
+     * "lazily materialized" - after sorting 1 million of rows and given the page size = 100, a dataframe with only 100 rows is created.
      *
      * @param dataFrameLike The dataframe-like object to sort.
      * @param columnPaths The list of columns to sort by. Each element in the list represents a column path


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
```diff
@@ -236,6 +283,7 @@ public object KotlinNotebookPluginUtils {
 
             is FormattedFrame<*> -> dataframeLike.df
 
+            // мб повыше перенести
```
Jolanrensen: left some russian in ;P

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
```diff
@@ -103,59 +108,101 @@ public object KotlinNotebookPluginUtils {
...
-        val firstValue = row1.getValueOrNull(key) as Comparable<Any?>?
-        val secondValue = row2.getValueOrNull(key) as Comparable<Any?>?
-
-        return when {
-            firstValue == null && secondValue == null -> 0
-            firstValue == null -> if (desc) 1 else -1
-            secondValue == null -> if (desc) -1 else 1
-            desc -> secondValue.compareTo(firstValue)
-            else -> firstValue.compareTo(secondValue)
+    private fun createColumnComparator(column: AnyCol, desc: Boolean): Comparator<DataRow<*>> {
```
Jolanrensen, line 170: Can you write the logic for how this comparison is done in the KDoc of `sortByColumns`? It's a good default, but there are multiple ways we could have handled sorting these types of columns, so it's good to explain what we picked

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
```diff
@@ -103,59 +108,101 @@ public object KotlinNotebookPluginUtils {
...
+            column.valuesAreComparable() -> compareBy(nullsLast()) {
+                column[it] as Comparable<Any?>?
+            }
+
+            // Comparator shows a slight improvement in performance for this case
+            column.isFrameColumn() -> Comparator { r1, r2 ->
+                column[r1].rowsCount().compareTo(column[r2].rowsCount())
+            }
+
+            column.isList() -> compareBy { (column[it] as? List<*>)?.size ?: 0 }
```
Jolanrensen, line 181: what about arrays, maps etc? do we care about sorting those too?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
Jolanrensen: Actually, it looks like this function does some things similar to our sort implementation https://github.com/Kotlin/dataframe/blob/36ea0c85bc5dd788e2616cff4d2e9282f07a60db/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/sort.kt#L57 I have a feeling these can and should be merged (maybe in a follow-up pr). I think it would also be greatly beneficial to be able to sort by DF size when people call `df.sortBy { frameCol }`, and it would be odd if you could sort a certain column in the table component, but not using the dataframe api itself, don't you think?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
```diff
@@ -103,59 +108,101 @@ public object KotlinNotebookPluginUtils {
             ColumnPath(path)
         }
 
-        val comparator = createComparator(sortKeys, isDesc)
+        if (sortKeys.size == 1) {
+            val column = df.getColumn(sortKeys[0])
 
-        return df.sortWith(comparator)
+            // Not sure how to have generic logic that would produce Comparator<Int> and Comparator<DataRow> without overhead
+            // For now Comparator<DataRow> is needed for fallback case of sorting multiple columns. Although it's now impossible in UI
```
Jolanrensen, line 119: not impossible if we'd move this to the sortBy {} operation :)

koperagen: By "not possible" i mean table widget doesn't allow picking multiple columns to sort at once

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
koperagen: lol. it says "maybe move it a bit higher" 
i thought, isn't AnyFrame the most likely dataframeLike? :) But it'd save maybe 0,000001%

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
koperagen: Interesting. Yes, why not

koperagen: > Can you write the logic for how this comparison is done in the KDoc of sortByColumns? It's a good default, but there are multiple ways we could have handled sorting these types of columns, so it's good to explain what we picked

You mean Frame and List columns, for instance?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
koperagen: I guess we do, it will be very easy to expand for Collections and Arrays. I decided to cut the branch on List+Frame because at this point PR became mostly about performance. But i don't see why not compare everything that makes sense

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
koperagen: updated

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
Jolanrensen: I know :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
Jolanrensen: That will probably be negligible haha

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
Jolanrensen: > You mean Frame and List columns, for instance?

Yes :) It's a win-win; we won't have to maintain two different sets of code regarding to sorting datarows, and users outside notebooks will gain more sorting capabilities!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
```diff
@@ -68,6 +73,10 @@ public object KotlinNotebookPluginUtils {
     /**
      * Sorts a dataframe-like object by multiple columns.
      * If a column type is not comparable, sorting by string representation is applied instead.
+     * Sorts DataFrames by their size because looking at the smallest / biggest groups after groupBy is very popular.
```
Jolanrensen, line 76: and lists

Jolanrensen (approved):

core/build.gradle.kts
```diff
@@ -89,6 +91,17 @@ dependencies {
     testImplementation(projects.dataframeCsv)
 }
 
+benchmark {
```
zaleslaw, line 94: Happy to see first benchmark in our project!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/KotlinNotebookPluginUtils.kt
```diff
@@ -103,60 +112,100 @@ public object KotlinNotebookPluginUtils {
...
+
+                column.isList() -> compareBy { (column[it] as? List<*>)?.size ?: 0 }
+
+                else -> compareBy { column[it]?.toString() ?: "" }
+            }
+
+            val finalComparator = if (isDesc[0]) comparator.reversed() else comparator
+
+            val permutation = Array(column.size()) { it }
+            Arrays.parallelSort(permutation, finalComparator)
```
zaleslaw, line 136: @copilot is it thread-safe to use parallesort here?

zaleslaw:
https://stackoverflow.com/questions/36128533/why-is-the-minimum-granularity-defined-as-8192-in-java8-in-order-to-switch-from

```
val threshold = 8192 // Arrays.parallelSort uses this internally
if (permutation.size >= threshold) {
    Arrays.parallelSort(permutation, finalComparator)
} else {
    Arrays.sort(permutation, finalComparator)
}
```

ForkJoinPool under the hood uses the common thread pool and also, thinking am I - should it live inside Kotlin kernel thread (I have no idea about how it's organized in Kotlin Notebook), where exactly the child threads will be created

koperagen:
> ForkJoinPool under the hood uses the common thread pool and also, thinking am I - should it live inside Kotlin kernel thread (I have no idea about how it's organized in Kotlin Notebook), where exactly the child threads will be created

Two options. First, default - Kotlin Kernel lives in its own JVM process. Second is attached mode where Kernel lives inside intellij process and uses common pool from there. But IMO in both cases it's no different from using Stream.parallelStream which is fairly widespread and, to my knowledge, recommended

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataFrameGet.kt
```diff
@@ -45,11 +45,9 @@ public fun <T> DataFrame<T>.getColumns(vararg columns: String): List<AnyCol> = g
 
 public fun <T> DataFrame<T>.getColumnIndex(col: AnyCol): Int = getColumnIndex(col.name())
 
-public fun <T> DataFrame<T>.getRows(range: IntRange): DataFrame<T> =
-    if (range == indices()) this else columns().map { col -> col[range] }.toDataFrame().cast()
+public fun <T> DataFrame<T>.getRows(range: IntRange): DataFrame<T> = get(range)
 
-public fun <T> DataFrame<T>.getRows(indices: Iterable<Int>): DataFrame<T> =
-    columns().map { col -> col[indices] }.toDataFrame().cast()
+public fun <T> DataFrame<T>.getRows(indices: Iterable<Int>): DataFrame<T> = get(indices)
```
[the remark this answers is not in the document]
koperagen: getRows is not a member function

zaleslaw:
@koperagen, after analysis, I didn't see a reason to reject for now. I understand that we could easily fix any issues on our side if something goes wrong due to a user request, and some other threads could create a race for obtaining resources from a common pool.

But great to keep in mind

Issue #1565, opened 2025-11-13
`writeGeoJson` breaks ordering of rows:


AndreiKingsley:
<img width="915" height="727" alt="Image" src="https://github.com/user-attachments/assets/a242fd85-a170-449c-8d50-a81ed1f15292" />

Issue #1187, opened 2025-05-12
`crs` assignment  after `read..` or creating methods for GeoDataFrame


AndreiKingsley:
1. Add optional crs parameter to `read..` methods for GeoDataFrame
Need some research here, is it valid or not in general.
The Shapefile always contains explicit CRS field, so no additional parameter is needed in `readShapefile()`
The GeoJSON Specification [claims](https://datatracker.ietf.org/doc/html/rfc7946#section-4) that CRS always should be WGS84.
2. Add `crs` field in `GeoDataFrame` not nullable.

AndreiKingsley:
So, some important updates after discussion of https://github.com/Kotlin/dataframe/pull/1640/files/8302a66491a8a9069d997975b1cc9f44c0fce1e4#r2622618050

And the problem is  
_Current GeoJSON support in GeoTools is **not compliant with the GeoJSON specification (RFC 7946)** and produces ambiguous and incorrect results._

Namely:
1) ["crs" field is deprecated](https://datatracker.ietf.org/doc/html/rfc7946#appendix-B.1). As mentioned before,
> the GeoJSON Specification [claims](https://datatracker.ietf.org/doc/html/rfc7946#section-4) that CRS always should be WGS84.

  However, it allows to store geometries in any CRS (at your own risk / by some agreements), but AFAIU  "crs" field should not be used nevertheless.

GeoTools writes and reads "crs" field. Moreover, **it does it wrong!** -  it parses WGS84 CRS from geojson with the latitude first, however GeoJson points **must** contain longitude first. So, now `write` then `read` works wrong for GeoJson. 

_(also I don't note it anywhere but GeoDataFrame is designed to be geojson like, including longitude-first approach)_

2) It doesn't work for 3D geometries (with Z axis), however it's allowed in GeoJSON spec.


In future we should add our own GeoJSON IO implementation - in fact it should be really easy, we just need to `readJson()` into DF and slightly modify it!

But now we can do more light temporary fixes:

1) make `crs` field in `GeoDataFrame` not null.
   * Shapefile, as well as other popular formats always contain it.
   * For `GeoJson` reading it should be WGS84 by default;   
2) Add `readGeoJson()` overload with next arguments:
   * `readOutdatedCrs: Boolean` (with Warning deprecation) - if user still wants to get crs from geojson
3) Add possibility  to override crs of the GeoDf (now you can only do it with `geodf.df.toGeo(yourCrs)`). Not sure how it shoud be done
[...]

PR #1640, opened 2025-12-15
Generate lexicographically ordered IDs for writeGeojson features

koperagen:
fixes https://github.com/Kotlin/dataframe/issues/1565

### DIFF
--- a/dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/geotools/toSimpleFeatureCollection.kt
+++ b/dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/geotools/toSimpleFeatureCollection.kt
@@ -40,14 +40,16 @@ fun GeoDataFrame<*>.toSimpleFeatureCollection(
     val featureCollection = ListFeatureCollection(featureType)
 
     val featureBuilder = SimpleFeatureBuilder(featureType)
-
+    // if ID is present, SortedMap in DefaultFeatureCollection sorts rows by ID lexicographically
+    // I couldn't disable writing it, so let's generate lexicographically sorted IDs
+    val format = "f%0${df.rowsCount().toString().length}d"
     df.forEach { row ->
         val geometry = row["geometry"]
         featureBuilder.add(geometry)
         df.columnNames().filter { it != "geometry" }.forEach { colName ->
             featureBuilder.add(row[colName])
         }
-        val feature: SimpleFeature = featureBuilder.buildFeature(null)
+        val feature: SimpleFeature = featureBuilder.buildFeature(String.format(format, index()))
         featureCollection.add(feature)
     }
 
--- a/dataframe-geo/src/test/kotlin/org/jetbrains/kotlinx/dataframe/geo/io/IOTest.kt
+++ b/dataframe-geo/src/test/kotlin/org/jetbrains/kotlinx/dataframe/geo/io/IOTest.kt
@@ -1,5 +1,6 @@
 package org.jetbrains.kotlinx.dataframe.geo.io
 
+import org.geotools.referencing.CRS
 import org.jetbrains.kotlinx.dataframe.api.dataFrameOf
 import org.jetbrains.kotlinx.dataframe.geo.GeoDataFrame
 import org.jetbrains.kotlinx.dataframe.geo.toGeo
@@ -35,17 +36,18 @@ class IOTest {
         assert(geodf.crs == null)
     }
 
-    /*
-    TODO: doesn't work for now - writers adds ids that breaks order when read features
     @Test
     fun writeGeoJson() {
         val tempFile = Files.createTempFile("simple_points", ".json").toFile()
         simplePointsGeoDf.writeGeoJson(tempFile)
 
-        assertEquals(simplePointsGeoDf, GeoDataFrame.readGeoJson(tempFile.toURI().toURL()))
+        val loadedGeoDataFrame = GeoDataFrame.readGeoJson(tempFile.toURI().toURL())
+        assertEquals(simplePointsGeoDf.df, loadedGeoDataFrame.df)
+        // TODO: Doesn't work because of how equality between CRS is checked by geotools
+        // assertEquals(simplePointsGeoDf, loadedGeoDataFrame)
 
         tempFile.deleteOnExit()
-    }*/
+    }
 
     @Test
     fun readShapefile() {


dataframe-geo/src/test/kotlin/org/jetbrains/kotlinx/dataframe/geo/io/IOTest.kt
```diff
@@ -35,17 +36,18 @@ class IOTest {
...
 
-    /*
-    TODO: doesn't work for now - writers adds ids that breaks order when read features
     @Test
     fun writeGeoJson() {
         val tempFile = Files.createTempFile("simple_points", ".json").toFile()
         simplePointsGeoDf.writeGeoJson(tempFile)
 
-        assertEquals(simplePointsGeoDf, GeoDataFrame.readGeoJson(tempFile.toURI().toURL()))
+        val loadedGeoDataFrame = GeoDataFrame.readGeoJson(tempFile.toURI().toURL())
```
AndreiKingsley, line 44: We need to add more IO overloads ;)

dataframe-geo/src/test/kotlin/org/jetbrains/kotlinx/dataframe/geo/io/IOTest.kt
```diff
@@ -35,17 +36,18 @@ class IOTest {
...
+        assertEquals(simplePointsGeoDf.df, loadedGeoDataFrame.df)
+        // TODO: Doesn't work because of how equality between CRS is checked by geotools
```
AndreiKingsley, line 46: The proper way to check CRS equality in Geotools is 
`CRS.equals`, or `CRS.equalsIgnoreMetadata` if you don't want to compare metedata.

But there, shouldn't they be equal?

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/geotools/toSimpleFeatureCollection.kt
```diff
@@ -40,14 +40,16 @@ fun GeoDataFrame<*>.toSimpleFeatureCollection(
     val featureCollection = ListFeatureCollection(featureType)
 
     val featureBuilder = SimpleFeatureBuilder(featureType)
-
+    // if ID is present, SortedMap in DefaultFeatureCollection sorts rows by ID lexicographically
```
AndreiKingsley, line 43: Oh, ok. Is that a known issue in Geotools? Is there an official workaround?

AndreiKingsley (changes requested):

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/geotools/toSimpleFeatureCollection.kt
koperagen: I think they did it intentionally, doesn't look like a bug
Got pure geotools repro here. I see that just reading and writing back the same collection is idempotent, i.e. rows go the same order. But problem arises when we extract data from collection (create dataframe) and then re-create feature collection from it. Please have a look. 

Output of the program:
```
Original order:
  feature-0 -> Point 1
  feature-1 -> Point 2

Built feature id: fid-1a9d9e5e_19b2837675d_-8000
Built feature id: fid-1a9d9e5e_19b2837675d_-7fff

Written JSON:
{"type":"FeatureCollection","features":[{"type":"Feature","geometry":{"type":"Point","coordinates":[31.5,51.5]},"properties":{"name":"Point 2"},"id":"fid-1a9d9e5e_19b2837675d_-7fff"},{"type":"Feature","geometry":{"type":"Point","coordinates":[30.5,50.5]},"properties":{"name":"Point 1"},"id":"fid-1a9d9e5e_19b2837675d_-8000"}]}

After write+read:
  fid-1a9d9e5e_19b2837675d_-7fff -> Point 2
  fid-1a9d9e5e_19b2837675d_-8000 -> Point 1
```

Somewhat interesting to see that in original feature collections feature-0 and feature-1 are IDs - but no ID in the original file

```
package org.jetbrains.kotlinx.dataframe.geo

import org.geotools.data.simple.SimpleFeatureCollection
import org.geotools.feature.DefaultFeatureCollection
import org.geotools.feature.simple.SimpleFeatureBuilder
import org.geotools.feature.simple.SimpleFeatureTypeBuilder
import org.geotools.geojson.feature.FeatureJSON
import org.locationtech.jts.geom.Geometry
import java.io.File

fun main() {
    val geojson = """
     {
      "type": "FeatureCollection",
      "features": [
        {
          "type": "Feature",
          "geometry": {
            "type": "Point",
            "coordinates": [30.5, 50.5]
          },
          "properties": {
            "name": "Point 1"
          }
        },
        {
          "type": "Feature",
          "geometry": {
            "type": "Point",
            "coordinates": [31.5, 51.5]
          },
          "properties": {
            "name": "Point 2"
          }
        }
      ]
    }
    """.trimIndent()

    val featureJSON = FeatureJSON()

    // Read original
    val original = featureJSON.readFeatureCollection(geojson.reader()) as SimpleFeatureCollection
    println("Original order:")
    printNames(original)

    // Extract to intermediate storage (simulating DataFrame)
    val data = mutableListOf<Pair<String, Geometry>>()
    original.features().use { iter ->
        while (iter.hasNext()) {
            val f = iter.next()
            data.add(f.getAttribute("name") as String to f.defaultGeometry as Geometry)
        }
    }

    // Rebuild FeatureCollection from scratch (like toSimpleFeatureCollection does)
    val typeBuilder = SimpleFeatureTypeBuilder()
    typeBuilder.name = "geodata"
    typeBuilder.add("the_geom", Geometry::class.java)
    typeBuilder.add("name", String::class.java)
    val featureType = typeBuilder.buildFeatureType()

    val rebuilt = DefaultFeatureCollection()
    val featureBuilder = SimpleFeatureBuilder(featureType)
    for ((name, geometry) in data) {
        featureBuilder.add(geometry)
        featureBuilder.add(name)
        val feature = featureBuilder.buildFeature(null) // ID is auto-generated
        println("Built feature id: ${feature.id}")
        rebuilt.add(feature)
    }

    // Write to file
    val tempFile = File.createTempFile("test", ".geojson")
    tempFile.outputStream().use { featureJSON.writeFeatureCollection(rebuilt, it) }

    println("\nWritten JSON:")
    println(tempFile.readText())

    // Read back
    val reloaded = featureJSON.readFeatureCollection(tempFile) as SimpleFeatureCollection
    println("\nAfter write+read:")
    printNames(reloaded)

    tempFile.delete()
}

fun printNames(fc: SimpleFeatureCollection) {
    fc.features().use { iter ->
        while (iter.hasNext()) {
            val f = iter.next()
            println("  ${f.id} -> ${f.getAttribute("name")}")
        }
    }
}
```

dataframe-geo/src/test/kotlin/org/jetbrains/kotlinx/dataframe/geo/io/IOTest.kt
koperagen: It doesn't help :(
`println(CRS.equalsIgnoreMetadata(simplePointsGeoDf.crs, loadedGeoDataFrame.crs))`

false

They are in fact identical, two lines just go in different order
image

AndreiKingsley: Wait, then there's problem here! In GeoDataFrame (as in GeoJson specification) longitude should always go first! That's important!

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/geotools/toSimpleFeatureCollection.kt
AndreiKingsley: Ok, then this solution looks good. Thank you!

AndreiKingsley (changes requested):

dataframe-geo/src/test/kotlin/org/jetbrains/kotlinx/dataframe/geo/io/IOTest.kt
koperagen: Interesting. This is how CRS looks right after reading of features: 
image

Order of longitude is broken? 
image

Content of the file we're reading from:
"crs":{"type":"name","properties":{"name":"EPSG:4326"}}

I see that our default CRS is created with longitudeFirst = true: CRS.decode("EPSG:4326", true)

image

AndreiKingsley: Related to
https://github.com/Kotlin/dataframe/issues/1187#issuecomment-3669893557
Now we can leave it as it is.

AndreiKingsley (approved):

Issue #722, opened 2024-06-04
Add Maven and Amper configurations in the documentation


koperagen:
Coroutines have configuration for Maven
https://github.com/Kotlin/kotlinx.coroutines

As i said here, for Maven we don't have a plugin now and we can mention it
https://github.com/Kotlin/dataframe/discussions/717#discussioncomment-9665763

But for Amper it should be possible to use Gradle plugin too
[Gradle interop](https://github.com/JetBrains/amper/blob/release/0.3/docs/Documentation.md#gradle-interop), including combining Amper and Gradle modules in one project*.

Jolanrensen:
I'd like to bump up this issue. Especially since the compiler plugin now works regardless of the build system used!

Showing off Amper would be good for dogfooding and showing Maven would be good for JVM ecosystem integration, so both have benefits! Not everyone uses Gradle, after all ;P

zaleslaw:
Should be added:

- [ ] Test projects for the KDA-Test-Projects
- [ ] Example projects for idea-projects chapter

AndreiKingsley:
Close this issue with creating two new separate issues #1686 and #1687

PR #1643, opened 2025-12-16
Added Maven project example + Maven setup instructions

AndreiKingsley:
Helps #722

### DIFF
--- a/docs/StardustDocs/topics/setup/SetupMaven.md
+++ b/docs/StardustDocs/topics/setup/SetupMaven.md
@@ -80,7 +80,7 @@ fun main() {
 on-the-fly in Maven projects, making development with Kotlin DataFrame faster,
 more convenient, and fully type- and name-safe.
 
-> Requires Kotlin 2.2.20-Beta1 or higher.  
+> Requires Kotlin 2.2.20-Beta1 or higher and IntelliJ IDEA 2025.3 or higher.  
 > { style = "note" }
 
 To enable the plugin in your Maven project, add it to the `plugins` section:
@@ -110,6 +110,9 @@ To enable the plugin in your Maven project, add it to the `plugins` section:
 ## Project Example
 
 See [the Maven example project with the Kotlin DataFrame Compiler Plugin enabled on GitHub](https://github.com/Kotlin/dataframe/tree/master/examples/kotlin-dataframe-plugin-maven-example).
+
+You can also 
+[download this project](https://github.com/Kotlin/dataframe/raw/example-projects-archives/kotlin-dataframe-plugin-maven-example.zip).
 
 
 ## Next Steps
--- a/README.md
+++ b/README.md
@@ -114,12 +114,12 @@ Or manually specify the version:
 > ```
 
 Refer to the
-[Get started with Kotlin DataFrame in Kotlin Notebook](https://kotlin.github.io/dataframe/gettingstartedkotlinnotebook.html)
+[Setup Kotlin DataFrame in Kotlin Notebook](https://kotlin.github.io/dataframe/setupkotlinnotebook.html)
 for details.
 
 ### Gradle
 
-Add dependencies in the build.gradle.kts script:
+Add dependencies in the `build.gradle.kts` script:
 
 ```kotlin
 dependencies {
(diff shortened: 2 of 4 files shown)


docs/StardustDocs/topics/setup/SetupMaven.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+## Kotlin DataFrame Compiler Plugin
+
+[Kotlin DataFrame Compiler Plugin](Compiler-Plugin.md) enables automatic generation of
+[extension properties](extensionPropertiesApi.md) and updates [data schemas](schemas.md)
+on-the-fly in Maven projects, making development with Kotlin DataFrame faster,
+more convenient, and fully type- and name-safe.
+
+> Requires Kotlin 2.2.20-Beta1 or higher.
```
koperagen: And IntelliJ 2025.3

koperagen (approved):

koperagen:
Thanks!

koperagen:
Please see if there's a good place to mention link to Maven setup in our main README.md

koperagen (approved):

examples/kotlin-dataframe-plugin-maven-example/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+    </repositories>
+
+    <build>
+        <sourceDirectory>src/main/kotlin</sourceDirectory>
+        <testSourceDirectory>src/test/kotlin</testSourceDirectory>
+        <plugins>
+            <plugin>
+                <groupId>org.jetbrains.kotlin</groupId>
+                <artifactId>kotlin-maven-plugin</artifactId>
+                <version>2.3.0-RC3</version>
```
Jolanrensen, line 31: we might actually need an "example generator" at this point. This will be yet another version to manually update each time

examples/kotlin-dataframe-plugin-maven-example/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+            <artifactId>junit-jupiter</artifactId>
+            <version>5.10.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-stdlib</artifactId>
+            <version>2.3.0-RC3</version>
+        </dependency>
+        <!-- DataFrame and Kandy dependencies -->
```
Jolanrensen, line 99: and some more...

examples/kotlin-dataframe-plugin-maven-example/pom.xml
```diff
@@ -0,0 +1,112 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Jolanrensen, line 1: Also, since we don't build this example, we cannot track when it breaks. Maybe something like https://stackoverflow.com/questions/35795250/maven-project-as-dependency-in-gradle-project could work?

examples/kotlin-dataframe-plugin-maven-example/pom.xml
koperagen: I suggest a Gradle task that would read Kotlin version from our toml and update required files. We can run it on assemble, for instance, so it will be hard to miss. It also very fast and no additional infra required

docs/StardustDocs/topics/setup/SetupMaven.md
```diff
@@ -0,0 +1,126 @@
...
+3. Name the new project and change its location, if necessary.
+
+   > Select the **Create Git repository** checkbox to place the new project under version control. 
+   > You can enable this later at any time.
+   > {type="tip"}
+
+4. From the **Language** list, select **Kotlin**.
+5. Select the **Maven** build system.
+6. From the **JDK list**, select the [JDK](https://www.oracle.com/java/technologies/downloads/) 
+that you want to use in your project. The minimum supported version is JDK 8.
```
Jolanrensen, line 30: maybe we should recommend newer versions as well, as some parts of DF require 11

docs/StardustDocs/topics/Compiler-Plugin.md
```diff
@@ -55,16 +55,15 @@ Setup library dependency:
 implementation("org.jetbrains.kotlinx:dataframe:%dataFrameVersion%")
 ```
 
-Add this line to `gradle.properties`:
+Due to the [known issue](https://youtrack.jetbrains.com/issue/KT-66735), incremental compilation must be disabled for now.
+Add the following line to your `gradle.properties` file:
+
 ```properties
 kotlin.incremental=false
```
Jolanrensen, line 62: @koperagen this isn't a problem in Maven?

docs/StardustDocs/v.list
```diff
@@ -2,5 +2,5 @@
 <!DOCTYPE vars SYSTEM "https://resources.jetbrains.com/writerside/1.0/vars.dtd">
 <vars>
     <var name="dataFrameVersion" value="1.0.0-Beta4" type="string"/>
-    <var name="compilerPluginKotlinVersion" value="2.3.0-RC" type="string"/>
```
Jolanrensen, line 5: 2.3.0 is released

examples/kotlin-dataframe-plugin-maven-example/pom.xml
Jolanrensen: how would we 'mark' where the versions need to be overwritten though? A Korro-like notation?

docs/StardustDocs/topics/Compiler-Plugin.md
koperagen: Not a problem there

examples/kotlin-dataframe-plugin-maven-example/pom.xml
Jolanrensen: Something like `%compilerPluginKotlinVersion%` in writerside would also be nice, but that would require two versions of all files. A "template" version and a "compiled" version with the actual values

Jolanrensen: actually, that's pretty easily done!

For each file we want to process, like `pom.xml` we create `pom.xml.template` with `%compilerPluginKotlinVersion%` notations, and then we have a top-level gradle task that replaces all files `a.b` that have an `a.b.template` counterpart with the right versions!

examples/kotlin-dataframe-plugin-maven-example/pom.xml
AndreiKingsley: I left comments only for things not generated by project wizard!

examples/kotlin-dataframe-plugin-maven-example/pom.xml
AndreiKingsley: 🤔

PR #1647, opened 2025-12-18
Squash annotations

koperagen:
I thought, let's employ a bit shorted syntax to reduce noise
Maybe there are other annotations that can be squashed, i squashed only the most popular

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/Nulls.kt
@@ -366,8 +366,7 @@ private interface CommonDropNullsFunctionDoc
  * @include [DropNulls.WhereAllNullParam]
  * @include [DropDslParam]
  */
-@Refine
-@Interpretable("DropNulls0")
+@[Refine Interpretable("DropNulls0")]
 public fun <T> DataFrame<T>.dropNulls(whereAllNull: Boolean = false, columns: ColumnsSelector<T, *>): DataFrame<T> {
     val cols = this[columns]
     return if (whereAllNull) {
@@ -377,8 +376,7 @@ public fun <T> DataFrame<T>.dropNulls(whereAllNull: Boolean = false, columns: Co
     }
 }
 
-@Refine
-@Interpretable("DropNulls0")
+@[Refine Interpretable("DropNulls0")]
 @Deprecated(
     "DataFrame conventional name for filterNot* functions is drop*",
     ReplaceWith("dropNulls(columns = columns)"),
@@ -391,8 +389,7 @@ public fun <T> DataFrame<T>.filterNotNull(columns: ColumnsSelector<T, *>): DataF
  * This overload operates on all columns in the [DataFrame].
  * @include [DropNulls.WhereAllNullParam]
  */
-@Refine
-@Interpretable("DropNulls1")
+@[Refine Interpretable("DropNulls1")]
 public fun <T> DataFrame<T>.dropNulls(whereAllNull: Boolean = false): DataFrame<T> = dropNulls(whereAllNull) { all() }
 
 /**
@@ -485,8 +482,7 @@ private interface CommonDropNAFunctionDoc
  * @include [DropNA.WhereAllNAParam]
  * @include [DropDslParam]
  */
-@Refine
-@Interpretable("DropNa0")
+@[Refine Interpretable("DropNa0")]
 public fun <T> DataFrame<T>.dropNA(whereAllNA: Boolean = false, columns: ColumnsSelector<T, *>): DataFrame<T> {
     val cols = this[columns]
     return if (whereAllNA) {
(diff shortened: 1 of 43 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/filter.kt
```diff
@@ -84,13 +84,11 @@ public fun <T> DataFrame<T>.filterBy(column: ColumnSelector<T, Boolean>): DataFr
 public fun <T> DataFrame<T>.filterBy(column: String): DataFrame<T> = filterBy { column.toColumnOf() }
 
 @Suppress("DEPRECATION_ERROR")
```
Jolanrensen, line 86: there's another ;P

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/frameCol.kt
```diff
@@ -452,24 +444,21 @@ public interface FrameColColumnsSelectionDsl<out _UNUSED> {
...
 
     /**
      * @include [FrameColKPropertyDocs] {@set [CommonFrameColDocs.RECEIVER] myColumnGroup.}
      */
-    @Deprecated(DEPRECATED_ACCESS_API)
-    @AccessApiOverload
+    @[Deprecated(DEPRECATED_ACCESS_API) AccessApiOverload]
     @Suppress("INAPPLICABLE_JVM_NAME")
     @JvmName("frameColDataFrameKProperty")
     public fun <C> AnyColumnGroupAccessor.frameCol(property: KProperty<DataFrame<C>>): ColumnAccessor<DataFrame<C>> =
```
Jolanrensen, line 464: two more

Jolanrensen:
I would try to follow/create a little style guide for all our annotations. I just submitted an issue for ktlint https://github.com/pinterest/ktlint/issues/3212 for this, as it doesn't seem to know `@[]` notations.

You now create:
```kt
@[Deprecated(DEPRECATED_ACCESS_API) AccessApiOverload]
@Suppress("INAPPLICABLE_JVM_NAME")
@JvmName("frameColDataFrameKProperty")
public fun <C> AnyColumnGroupAccessor.frameCol...
```
I would recommend putting the `@[]` at the bottom. It looks a bit cleaner, in my opinion:
You now create:
```kt
@Suppress("INAPPLICABLE_JVM_NAME")
@JvmName("frameColDataFrameKProperty")
@[Deprecated(DEPRECATED_ACCESS_API) AccessApiOverload]
public fun <C> AnyColumnGroupAccessor.frameCol...
```
Also, we don't really honor an order. What about just lexicographically?
```kt
@JvmName("frameColDataFrameKProperty")
@Suppress("INAPPLICABLE_JVM_NAME")
@[AccessApiOverload Deprecated(DEPRECATED_ACCESS_API)]
public fun <C> AnyColumnGroupAccessor.frameCol...
```
Next, the plan is to "compact" multiple annotations that are often used together?

We could probably include the JvmName one then too:
```kt
@[AccessApiOverload Deprecated(DEPRECATED_ACCESS_API)]
@[JvmName("frameColDataFrameKProperty") Suppress("INAPPLICABLE_JVM_NAME")]
public fun <C> AnyColumnGroupAccessor.frameCol...
```

wdyt?

AndreiKingsley:
@Jolanrensen to be honest, it doesn't look very pretty to me, however the idea of grouping annotations seems great!

zaleslaw:
Honestly, I'd prefer the next order:

```
@Suppress("INAPPLICABLE_JVM_NAME")
@JvmName("frameColDataFrameKProperty") 
```

- at the bottom, because they are services and do not give so much to developers, the most significant things, like deprecation, should be at the top, and this is how it's usually added by IDE, isn't it?

zaleslaw (commented):
Sorry, I want to argue with this approach

Although @[A B] is valid Kotlin syntax, it is not a conventional style. It is rarely used outside of use-site targets, is not shown in official Kotlin documentation examples, and is not commonly found in real Kotlin projects or codebases. It is also not adopted by common tools or JetBrains/Android codebases.

In this case, it replaces the familiar form

```kotlin
@Deprecated(...)
@AccessApiOverload
```
with
```kotlin
@[Deprecated(...) AccessApiOverload]
```

which mixes a standard Kotlin annotation with a project-specific one, reduces readability, makes diffs harder to read, and introduces an uncommon idiom without any clear benefit.

Moreover, it's absolutely JVM, Kotlin and other approaches have the list of these `aspects` of the function as separate annotations

koperagen:
We agreed to not follow up on this idea

Jolanrensen:
Let's return to it once the community reaches a consensus on what is recommended annotation notation, like with KtLint: https://github.com/pinterest/ktlint/issues/3212 

Personally, it makes sense for me to group annotations that are related in some way, but it should be done consistently everywhere as to keep the codebase clean and predictable :)

PR #1648, opened 2025-12-18
Migrate most of the examples in 40 puzzles from delegate style dataframeOf(Columns) to dataFrameOf(Pair<String, Column>)

koperagen:
Didn't do it for random-related examples because I'm still not sure about cleanest replacement for these specific cases


AndreiKingsley (approved):
Is this now most recommended  way to create a `DataFrame` (from raw values) ?

koperagen:
> Is this now most recommended way to create a `DataFrame` (from raw values) ?

Yes. It's easier for creating nested column groups and we avoid showing ColumnAccessors too much - most APIs that use it are deprecated, would be confusing.

PR #1649, opened 2025-12-18
Remove deprecated and unused access api examples and migrate multi-column add/map

koperagen:
They're already not used on the website for a while.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/AccessApi.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/AccessApi.kt
@@ -9,8 +9,7 @@ import org.jetbrains.kotlinx.dataframe.documentation.AccessApi.AnyApiLinks
  * or deleted while wrangling. Kotlin, in contrast, is a statically typed language and all types are defined and verified
  * ahead of execution. That's why creating a flexible, handy, and, at the same time, safe API to a dataframe is tricky.
  *
- * In `Kotlin DataFrame` we provide four different ways to access columns, and, while they're essentially different, they
- * look pretty similar in the data wrangling DSL. These include:
+ * In `Kotlin DataFrame` we provide two different ways to access columns:
  * @include [AnyApiLinks]
  *
  * For more information: {@include [DocumentationUrls.AccessApis]}
@@ -21,8 +20,6 @@ internal interface AccessApi {
 
     /**
      * - {@include [ExtensionPropertiesApiLink]}
-     * - {@include [KPropertiesApiLink]}
-     * - {@include [ColumnAccessorsApiLink]}
      * - {@include [StringApiLink]}
      */
     interface AnyApiLinks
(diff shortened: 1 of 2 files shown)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/JoinWith.kt
```diff
@@ -144,7 +144,7 @@ class JoinWith : TestBase() {
             cellRenderer = renderer,
             configuration = SamplesDisplayConfiguration.copy(
                 cellFormatter = { row, col ->
-                    val value = row[col]
+                    val value = col[row]
```
AndreiKingsley, line 147: xD

AndreiKingsley (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/AccessApi.kt
```diff
@@ -81,8 +72,9 @@ internal interface AccessApi {
      *
      * For more information: {@include [DocumentationUrls.AccessApis.ExtensionPropertiesApi]}
      *
-     * For example: {@comment This works if you include the test module when running KoDEx}
+     * For example, in notebooks extension properties are generated from runtime data after the cell is executed: {@comment This works if you include the test module when running KoDEx}
```
Jolanrensen, line 75: You can put the `@comment` on the next line, if you write it without `{}`, there should be no newline in between `executed:` and `@sample`

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Modify.kt
```diff
@@ -933,8 +934,8 @@ class Modify : TestBase() {
     fun addMany_strings() {
         // SampleStart
         df.add {
-            "year of birth" from 2021 - "age"<Int>()
-            "age"<Int>() gt 18 into "is adult"
+            "year of birth" from { 2021 - "age"<Int>() }
+            expr { "age"<Int>() > 18 } into "is adult"
```
Jolanrensen, line 938: maybe we could alternate with other, more kotlinlike notations, such as `"age<Int>().map { it > 18 } into "is adult"`

Jolanrensen: I do like we have more `expr` examples :) Feels like the concept is not well known

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Schemas.kt
```diff
@@ -88,7 +88,7 @@ class Schemas {
         // SampleEnd
     }
 
-    fun DataFrame<Person>.countAdults() = count { it[Person::age] > 18 }
+    fun DataFrame<Person>.countAdults() = count { it.age > 18 }
```
Jolanrensen, line 91: doesn't need `it` right?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Modify.kt
koperagen: "age<Int>().map { it > 18 } into "is adult"
let's keep this syntax in mind, it's not bad

PR #1656, opened 2025-12-19
Lazy Statistics

CarloMariaProietti:
Fixed version of #1636

> Fixes https://github.com/Kotlin/dataframe/issues/1492
> The idea is the following:
> ValueColumnInternal is an interface for statistic values, which in this way are not exposed as public.
> Implementations of ValueColumnInternal contain the actual cache.
> 
> It was necessary to have two caches for each stat (for the moment only max) because computing the stat may give different outputs basing on skipNaN boolean parameter.
> 
> I implemented the solution by overloading aggregateSingleColumn, this overload exploits the original aggregateSingleColumn by wrapping it so that it is possible to exploit caches.
> 
> For the moment there is only max, however it would be easy to do the same with min, sum, mean and median.
> For percentile and std it could be done something similar.

### DIFF
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/statistics.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/statistics.kt
@@ -1041,7 +1041,7 @@ class StatisticsTests {
         // derived columns
         val renamedColumn = valueColumn.rename("newName").asValueColumn().internalValueColumn()
         val colWithDifferentType = ((valueColumn as ResolvingValueColumn).source as ValueColumnImpl)
-            .changeType(typeOf<Double>())
+            .changeType(typeOf<Number>())
         // tests
         valueColumn.asValueColumn().internalValueColumn()
             .getStatisticCacheOrNull("min", mapOf("skipNaN" to false))?.value shouldBe 1


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/Aggregator.kt
```diff
@@ -75,13 +77,30 @@ public class Aggregator<in Value : Any, out Return : Any?>(
...
             AggregatorProvider { name ->
                 Aggregator(
                     aggregationHandler = aggregationHandler,
                     inputHandler = inputHandler,
                     multipleColumnsHandler = multipleColumnsHandler,
                     name = name,
+                    statisticsParameters = statisticsParameters,
+                )
+            }
+
```
CarloMariaProietti: The following version of `invoke` is used by stats that do not exploit cache yet, It shall be removed when each stat  will support cache.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnImpl.kt
```diff
@@ -8,14 +8,41 @@ import org.jetbrains.kotlinx.dataframe.columns.ValueColumn
 import kotlin.reflect.KType
 import kotlin.reflect.full.withNullability
 
+@JvmInline
+internal value class StatisticResult(val value: Any?)
+
```
CarloMariaProietti, line 13: `ParameterValue `class was created so that it could override `equals `and `hashCode`. In this way it is possible to correctly compare two `Map<String, ParameterValue>`.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnImpl.kt
```diff
@@ -8,14 +8,53 @@ import org.jetbrains.kotlinx.dataframe.columns.ValueColumn
...
+public class ParameterValue(public val parameter: Any?) {
```
Jolanrensen: Why is this necessary? booleans, doubles and integers are primitives. They already have a consistent hashcode and correctly working `equals`

Jolanrensen: It's also perfectly fine to use a map of primitives as key of another map, as the equality and hashcode of a Map depends solely on its contents: 
image

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnImpl.kt
Jolanrensen: if you comment, please leave a comment at exactly the relevant line :) Looking at the comment, I thought it belonged to `StatisticResult`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnImpl.kt
CarloMariaProietti: I understand, that class was actually useless. Fixed with the new commit. I choose for a non-nullable `Any` beacuse each time an entry is created the actual parameter is non-nullable.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/Aggregator.kt
```diff
@@ -75,13 +76,30 @@ public class Aggregator<in Value : Any, out Return : Any?>(
...
+        internal operator fun <Value : Any, Return : Any?> invoke(
```
Jolanrensen: this overload is unnecessary, simply add `= emptyMap()` to the statisticsParameters parameter above

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/Aggregators.kt
```diff
@@ -117,8 +125,9 @@ public object Aggregators {
         by withOneOption { skipNaN: Boolean ->
             twoStepSelectingForAny<Comparable<Any>, Comparable<Any>?>(
                 getReturnType = minTypeConversion,
-                stepOneSelector = { type -> minOrNull(type, skipNaN) },
                 indexOfResult = { type -> indexOfMin(type, skipNaN) },
+                stepOneSelector = { type -> minOrNull(type, skipNaN) },
+                statisticsParameters = mapOf<String, Any>(Pair("skipNaN", skipNaN)),
```
Jolanrensen: I've never seen anyone create a map like this before, haha. What about `mapOf("skipNaN" to skipNaN)`?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnImpl.kt
```diff
@@ -48,10 +56,13 @@ internal open class ValueColumnImpl<T>(
     override fun defaultValue() = defaultValue
 
     override fun forceResolve() = ResolvingValueColumn(this)
+
+    override val statistics = mutableMapOf<String, MutableMap<Map<String, Any>, StatisticResult>>()
```
Jolanrensen: hmm maybe a name like `statisticsCache` is a bit more descriptive

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnImpl.kt
```diff
@@ -8,14 +8,22 @@ import org.jetbrains.kotlinx.dataframe.columns.ValueColumn
 import kotlin.reflect.KType
 import kotlin.reflect.full.withNullability
 
+@JvmInline
+internal value class StatisticResult(val value: Any?)
+
+internal interface ValueColumnInternal<T> : ValueColumn<T> {
+    val statistics: MutableMap<String, MutableMap<Map<String, Any>, StatisticResult>>
```
Jolanrensen: It's a bit dangerous to expose a mutable map, especially one as complicated as this one, for other parts of the library to modify.

I would move the logic of getting/storing statistics here and only call those functions in Aggregators.kt

Jolanrensen: this way `ValueColumnInternal` could only expose functions like `putStatisticCache(name, arguments, value)`, and `getStatisticCacheOrNull(name, arguments)` and make the MutableMap private inside `ValueColumnImpl`. It's less bug prone that way :)

Jolanrensen: and it allows to you avoid names like `desiredStatisticNotConsideringParameters` which I have difficulties with comprehending ;P

CarloMariaProietti: > It's a bit dangerous to expose a mutable map, especially one as complicated as this one, for other parts of the library to modify.
> 
> I would move the logic of getting/storing statistics here and only call those functions in Aggregators.kt

I'm not sure I understood. At the moment getting/storing statistics is done in AggregatorAggregationHandler.kt
(`aggregateSingleColumn`) and in that same function there is also the logic of getting/storing.
The logic should be moved in `ValueColumnInternal `but why should the functions be called in Aggregators.kt?

CarloMariaProietti: > this way `ValueColumnInternal` could only expose functions like `putStatisticCache(name, arguments, value)`, and `getStatisticCacheOrNull(name, arguments)` and make the MutableMap private inside `ValueColumnImpl`. It's less bug prone that way :)

If the mutable map is private inside `ValueColumnImpl`, how could I access that field inside `ValueColumnInternal`?

CarloMariaProietti: Maybe `statisticsCache `could be a `Map` instead of `MutableMap` and getting/storing functions in `ValueColumnInternal `could eventually make a cast to `MutableMap`?

Jolanrensen: > The logic should be moved in ValueColumnInternal

No, into `ValueColumnImpl`. Interfaces describe the structure of a class, an API without implementation if you will. They should not implement something or dictate how it should be implemented.

So just something like:
```kt
internal interface ValueColumnInternal<T> : ValueColumn<T> {
    fun putStatisticCache(statName: String, arguments: Map<String, Any>, value: StatisticResult)
    fun getStatisticCacheOrNull(statName: String, arguments: Map<String, Any>): StatisticResult?
}
```

Then if `ValueColumnImpl` wants to implement `ValueColumnInternal` it will need to override those functions and write the logic for it. `ValueColumnImpl` can then hold the actual `Map` with the statistic cache but keep it private.

We also have a convention in DataFrame if you add `internal` functions to a `public` concept, like `ValueColumn`, to add an extension function like: `internal fun <T> ValueColumn<T>.internal(): ValueColumnInternal<T> = this as ValueColumn<T>`. That way you can simply call `someValueCol.internal().putStatisticCache(...)` from elsewhere in the internal parts of the library, like in the aggregators :)

Is that a bit clearer?

CarloMariaProietti: Thank you for the explenation, now this is very clear.
I'm still a bit confused  about the following statement :
> I would move the logic of getting/storing statistics here and only call those functions in Aggregators.kt

Why should these function be called in Aggregators.kt? Should not they be called in `aggregateSingleColumn` which is the only function that need that logic?

Jolanrensen: ah, I'm sorry for the confusion. I indeed meant `aggregateSingleColumn` :) so AggregatorAggregationHandler.kt

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnImpl.kt
```diff
@@ -17,8 +17,6 @@ internal interface ValueColumnInternal<T> : ValueColumn<T> {
     fun getStatisticCacheOrNull(statName: String, arguments: Map<String, Any>): StatisticResult?
 }
 
-internal fun <T> ValueColumn<T>.internal() = this as ValueColumnInternal<T>
```
Jolanrensen: What do you mean, "it breaks for `ValueColumnWithParent`"? Ah, probably because `ValueColumnWithParent` only implements `ValueColumn`... Instead of removing this overload (which doesn't fix the issue), you should better make sure all implementations of `ValueColumn` also implement `ValueColumnInternal` :)

Jolanrensen: otherwise, only `ValueColumn`s which specifically use the `ValueColumnImpl` implementation will work with this statistics cache.

CarloMariaProietti: > What do you mean, "it breaks for `ValueColumnWithParent`"? 

Compilation problem in ValueColumnWithParent.kt due to the following lines;

```
override fun changeType(type: KType) =
        ValueColumnWithParent(parent, source.internal().changeType(type).asValueColumn())
```

Jolanrensen: ah, it clashes with `DataColumn.internal()`... Well, one solution for that would be to make ValueColumnInternal inherit DataColumnInternal as well.

CarloMariaProietti: this solution introduces a new problem. `ValueColumnInternal `would inherit `rename` from both `ValueColumn `and `DataColumnInternal` but these functions have different return Type -> compilation error.

Jolanrensen: Okay, then just name it `internalValueColumn()` or something :) We'll look at it later. It's just a convenience function after all :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/sort.kt
```diff
@@ -114,7 +116,20 @@ internal class SortColumnDescriptor<C>(
     val column: ValueColumn<C>,
     val direction: SortDirection = SortDirection.Asc,
     val nullsLast: Boolean = false,
-) : ValueColumn<C> by column
+) : ValueColumn<C> by column,
+    ValueColumnInternal<C> {
```
Jolanrensen: *`ValueColumnInternal<C> by column.internalValueColumn()`

If you checked each implementation of `ValueColumn` also implements `ValueColumnInternal`, it's safe to delegate this `by column`, allowing you to skip writing duplicate logic in all `ValueColumnInternal` classes except `ValueColumnImpl`.

You can do the same trick in all other places

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnImpl.kt
```diff
@@ -48,10 +60,22 @@ internal open class ValueColumnImpl<T>(
     override fun defaultValue() = defaultValue
 
     override fun forceResolve() = ResolvingValueColumn(this)
+
+    private val statisticsCache = mutableMapOf<String, MutableMap<Map<String, Any>, StatisticResult>>()
+
+    override fun putStatisticCache(statName: String, arguments: Map<String, Any>, value: StatisticResult) {
+        statisticsCache.getOrPut(statName) {
+            mutableMapOf<Map<String, Any>, StatisticResult>()
```
Jolanrensen: you can remove the explicit type arguments

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/AggregatorAggregationHandler.kt
```diff
@@ -26,13 +28,32 @@ public interface AggregatorAggregationHandler<in Value : Any, out Return : Any?>
 
     /**
      * Aggregates the data in the given column and computes a single resulting value.
-     * Calls [aggregateSequence].
+     * Calls [aggregateSequence]. It tries to exploit a cache for statistics which is proper of
```
Jolanrensen: *a property?

Jolanrensen:
Alright, it seems the implementation is in place. However, we never test the cache is actually being used. For all we know, it's still being recalculated every time. Please add some tests in the statistics tests to verify that the cache is actually being set when a new statistic is calculated.

You could also test the retrieving part by saving something wrong into the cache manually and then check if it reappears when the statistic is requested again.

(now you can see why it makes sense to have the logic of interacting with the cache inside the ValueColumn instead of on the aggregator side :) Makes it easier to test)

CarloMariaProietti:
> (now you can see why it makes sense to have the logic of interacting with the cache inside the ValueColumn instead of on the aggregator side :) Makes it easier to test)

Yes, much better. Until now I tested the behaviour of cache only by using prints and it seems to work correctly.
However, I wrote a message on the slack datascience channel where I tell about Google Summer of Code, I would be very gratefull if you could read that proposal and give me your feedback :)

Jolanrensen:
> However, I wrote a message on the slack datascience channel where I tell about Google Summer of Code, I would be very gratefull if you could read that proposal and give me your feedback :)

We'll discuss this in the team first and get back to you later, if that's okay

CarloMariaProietti:
> > However, I wrote a message on the slack datascience channel where I tell about Google Summer of Code, I would be very gratefull if you could read that proposal and give me your feedback :)
> 
> We'll discuss this in the team first and get back to you later, if that's okay

Yes, of course.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/AggregatorAggregationHandler.kt
CarloMariaProietti: I meant that the cache is accessed by ValueColumnInternal

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/statistics.kt
```diff
@@ -955,4 +957,34 @@ class StatisticsTests {
         val percentile35 = res3["name"] as String
         percentile35 shouldBe "Charlie"
     }
+
+    @Test
+    fun `statistics cache for ValueColumn`() {
```
Jolanrensen: Yes, testing whether statistics functions actually read the cache is complete :)

But now also test that they actually write to it

Jolanrensen:
To be able to close https://github.com/Kotlin/dataframe/issues/1492 we also need some test that verify the original issue now works. So please try the example with `df.filter { x < df["valueCol"].cast<Int>().max() }` or something like that, where `.max()` would be called for each row in the DF and now should benefit from the cache being there.

Also some dataframe-wide statistic function like `df.max { "valueCol"<Int>() }` (and then checking that `df["valueCol"].asValueColumn().internalValueColumn().getStatisticsCache...` actually works).

This actually made me think of something: Try renaming a column that has some statistics cache :) Even though the contents of the column are exactly the same as before, the cache will be gone as a new `ValueColumnImpl` is instantiated. This also applies to `changeType()`. I believe you could supply your statisticsCache to the new one :) (I also think you don't even need to 'copy' the cache, just pass it on. As `values` is immutable, anything calculated from the same values will hold, even if the column is renamed and inside another dataframe)

CarloMariaProietti:
Hey, I'm sorry for my absence, during the last week I have been very busy with university exams. I made a commit trying to accomplish https://github.com/Kotlin/dataframe/pull/1656#issuecomment-3785201929.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/Aggregators.kt
```diff
@@ -132,6 +144,7 @@ public object Aggregators {
                 getReturnType = maxTypeConversion,
                 stepOneSelector = { type -> maxOrNull(type, skipNaN) },
                 indexOfResult = { type -> indexOfMax(type, skipNaN) },
+                statisticsParameters = mapOf<String, Any>("skipNaN" to skipNaN),
```
Jolanrensen: You can drop all types for `mapOf`, the types are inferred. Also check other places

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/aggregation/aggregators/Aggregators.kt
```diff
@@ -140,17 +153,30 @@ public object Aggregators {
         skipNaN: Boolean,
         ddof: Int,
         ->
-        flattenReducingForNumbers(stdTypeConversion) { type ->
-            std(type, skipNaN, ddof)
-        }
+        flattenReducingForNumbers(
+            getReturnType = stdTypeConversion,
+            statisticsParameters = mapOf<String, Any>(
+                ("skipNaN" to skipNaN),
```
Jolanrensen: unnecessary `()` parentheses. Also check other places.

In Kotlin, we create maps like:
```kt
mapOf(
    "something" to value1,
    "somethingElse" to value2,
)
```

Jolanrensen (approved):
Besides my latest comment, I think it's good to go :) I'll ask another colleague for a review as well.
Please rebase this PR on the master so the /generated-sources won't cause any conflicts. (usually we don't commit this folder, as it's updated automatically when an update is pushed to the master branch, but it's hard to remove it now)

Jolanrensen:
@koperagen raised a very good point: `ValueColumn` is a public interface which can be implemented by anyone. While we don't encourage this, it does appear in the wild: https://github.com/SciProgCentre/dataforge-core/blob/main/tables-kt/tables-kt-dataframe/src/main/kotlin/space/kscience/dataforge/dataframe/TableAsDataFrame.kt
This means that `fun <T> ValueColumn<T>.internalValueColumn() = this as ValueColumnInternal<T>` will fail if someone implements their own version of `ValueColumn`.

Might I suggest a small workaround for this case?
```kt
internal interface ValueColumnInternal<T> : ValueColumn<T> {
    fun putStatisticCache(statName: String, arguments: Map<String, Any>, value: StatisticResult)

    fun getStatisticCacheOrNull(statName: String, arguments: Map<String, Any>): StatisticResult?

    /**
     * Tiny wrapper for [ValueColumn]s that are not [ValueColumnImpl],
     * which can occur in case a user implements the [ValueColumn] interface.
     */
    class Wrapper<T>(val source: ValueColumn<T>) :
        ValueColumn<T> by source,
        ValueColumnInternal<T> {
        val statisticsCache: MutableMap<String, MutableMap<Map<String, Any>, StatisticResult>> = mutableMapOf()

        override fun putStatisticCache(statName: String, arguments: Map<String, Any>, value: StatisticResult) {
            statisticsCache.getOrPut(statName) { mutableMapOf() }[arguments] = value
        }

        override fun getStatisticCacheOrNull(statName: String, arguments: Map<String, Any>): StatisticResult? =
            statisticsCache[statName]?.get(arguments)
    }
}

internal fun <T> ValueColumn<T>.internalValueColumn() =
    this as? ValueColumnInternal<T> ?: ValueColumnInternal.Wrapper(this)
```

CarloMariaProietti:
> @koperagen raised a very good point: `ValueColumn` is a public interface which can be implemented by anyone. While we don't encourage this, it does appear in the wild: https://github.com/SciProgCentre/dataforge-core/blob/main/tables-kt/tables-kt-dataframe/src/main/kotlin/space/kscience/dataforge/dataframe/TableAsDataFrame.kt This means that `fun <T> ValueColumn<T>.internalValueColumn() = this as ValueColumnInternal<T>` will fail if someone implements their own version of `ValueColumn`.
> 
> Might I suggest a small workaround for this case? 

Thanks to @koperagen for his review. New commit integrates this code.

Jolanrensen:
@CarloMariaProietti Thanks for your contribution! I'll run our CI and merge the PR for you

Jolanrensen:
@CarloMariaProietti Seems the CI fails the bottom-most statistics cache test with [debug mode enabled](https://github.com/Kotlin/dataframe/blob/master/CONTRIBUTING.md#pr-workflow):

```
Values of ValueColumn '' have types '[Int]' which are not compatible given with column type 'kotlin.Double'
```
The debug mode is enabled for our CI tests and does some extra checks, such as failing when the values in a column mismatch their declared type (which is the case when you change the type of the Int column to Double).
I'll change it to `.changeType(typeOf<Number>())` so the resulting column is valid :)

In the future, you can add `kotlin.dataframe.debug=true` in gradle.properties to run these checks locally.

CarloMariaProietti:
> @CarloMariaProietti Thanks for your contribution! I'll run our CI and merge the PR for you

Thank you for your precious help in this feature's implementation :)

Issue #1650, opened 2025-12-18
Update to Kotlin 2.3.0


Jolanrensen:
https://kotlinlang.org/docs/whatsnew23.html

- [ ] Update .toml and subprojects
- [ ] Also, update all examples
- [ ] Update docs
- [ ] Finish https://github.com/Kotlin/dataframe/pull/1543 now nested type aliases are stable
- [ ] `Instant` should now be stable too!

Jolanrensen:
One small thing; we cannot bump KSP because of https://github.com/Kotlin/dataframe/issues/1445 and we potentially plan on removing KSP anyway (well, the gradle plugin)

Jolanrensen:
Blocked by https://youtrack.jetbrains.com/issue/KT-83292/Nothing-KType-not-equal-to-other-Nothing-KType-in-2.3.0-anymore

Jolanrensen:
Or more specifically https://youtrack.jetbrains.com/issue/KT-81024

PR #1657, opened 2025-12-19
Kotlin 2.3, kotlin.time.Instant

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1650

- Bumps to Kotlin 2.3.10
- Enables kotlin Instant parsing by default; updates docs accordingly
- ~~Excludes KSP-, gradle-, and compiler plugins from the build as KSP1 is incompatible with Kotlin 2.3, and/or they have become obsolete to build~~
- ~~Removed reliance on ksp/gradle plugins for `:core`, and `geo` modules.~~
    - ~~I moved all generated code into a separate folder under `/src` called `generated-dataschema-accessors`. Please have a look if you like this or if we should put it somewhere else :)~~
- This change was cherry-picked into https://github.com/Kotlin/dataframe/pull/1670 already
- I did not yet touch the `:examples`. They still use KSP (with a lot of warnings) until we do https://github.com/Kotlin/dataframe/issues/1614. (I tried using the same thing as `:core`, but I think it's better to update them to show users how to generate their own data schemas and use the compiler plugin than to invest time now making a non-ksp non-compiler-plugin hybrid version we don't plan to keep anyways)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,7 +2,7 @@ projectName=dataframe
 version=1.0.0
 jupyterApiTCRepo=
 kotlin.jupyter.add.scanner=false
-org.gradle.jvmargs=-Xmx16G -Duser.language=en -Duser.country=US -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx8G -Duser.language=en -Duser.country=US -Dfile.encoding=UTF-8
 
 # build.number.detection=false
 # build.number=0.8.0


Jolanrensen:
Blocked by https://youtrack.jetbrains.com/issue/KT-83292/Nothing-KType-not-equal-to-other-Nothing-KType-in-2.3.0-anymore. We have 18 new failing tests because of this bug.

Putting it on hold for now. Maybe 2.3.10 will be the one ;P

Jolanrensen:
Currently, we're blocked by KSP1; our Gradle plugin and KSP plugin. We will need to follow up on our [deprecation](https://kotlin.github.io/dataframe/gradle-plugin.html) of the gradle/ksp plugin to move forward.

I'll see if I can disable usage of the ksp plugin in our modules as well, though, disabling it for our examples requires https://github.com/Kotlin/dataframe/issues/1614

AndreiKingsley:
@Jolanrensen 
I believe we need to add info about `parse` here
https://kotlin.github.io/dataframe/migrationto-1-0.html#migration-to-standard-library-instant

Jolanrensen:
Now just waiting for 2.3.10 to drop :) then we can merge

Issue #1661, opened 2026-01-20
Streamline our infrastructure with Gradle Convention plugins


Jolanrensen:
Related to https://github.com/Kotlin/dataframe/issues/1614

The DataFrame project infrastructure has become large, complex, and fractured, making it difficult to maintain, and we often forget to bump versions in places because they are not synced.

I've been experimenting with gradle convention plugins and composite builds: https://github.com/Jolanrensen/GradleConventionsExperiment

I believe this sample project can be used as a good base for the DataFrame project itself:

- (type safe) convention plugins allow us to remove duplicate build code and avoid using allProjects {}
(It's also the [recommended way](https://docs.gradle.org/current/userguide/best_practices_structuring_builds.html#favor_composite_builds) to structure builds according to Gradle and [Kotlin](https://kotlinlang.org/docs/gradle-best-practices.html#use-convention-plugins))
- It fixes the issue where we have to manually sync versions to subprojects like `/plugins/keywords-generator`. We can simply use a build-settings convention plugin that finds and sets up the correct version catalog :)
- See https://github.com/Jolanrensen/GradleConventionsExperiment for more information and links

We could introduce it gradually:

- [x] Enable build-settings-logic and build-logic with just version catalogs, start using it for nested projects (fixed by: https://github.com/Kotlin/dataframe/pull/1662)
- [x] Move `allProjects {}` logic into dedicated conventions (like `dfbuild.kotlinJava8` and `dfbuild.kotlinJava11`, for instance) https://github.com/Kotlin/dataframe/pull/1670
- [x] Fix our examples situation: https://github.com/Kotlin/dataframe/issues/1614 and #1610, WIP: https://github.com/Kotlin/dataframe/pull/1677
- [ ] Create "common dataframe IO plugin" convention which provides and sets up plugins we use everywhere, like KtLint, binary compatibility validator, publisher, build config
- [ ] Move Korro logic into dedicated convention (we now have duplicate logic in :core and :samples)
- [x] Move KoDEx logic into dedicated convention (duplicated in :core and :dataframe-csv) #1765
- [ ] Make the setup for running the compiler plugin inside local modules a convention plugin (currently worked out in :samples, but we might need it in other places later as well)
- [ ] Move `:plugins:expressions-converter`, `:plugins:keywords-generator`, and `:plugins:public-api-modifier` (if we actually use that one) into `build-logic`. They are only used from local build files, so essentially, they're already convention plugins.
[...]

PR #1662, opened 2026-01-20
Gradle conventions init and shared version catalogs

Jolanrensen:
Finishes first step of https://github.com/Kotlin/dataframe/issues/1661

Initializing build-logic and build-settings-logic with dfbuild/dfsettings prefix. Following https://github.com/Jolanrensen/GradleConventionsExperiment and inspired by Dokka. 

Sharing same version catalog and repositories between all subprojects.

### DIFF
--- a/build-logic/README.md
+++ b/build-logic/README.md
@@ -1,0 +1,12 @@
+## :build-logic
+
+This project contains all shared logic for build logic (`build.gradle.kts` files).
+
+The entire DataFrame project is built
+using [Composite Builds](https://docs.gradle.org/current/userguide/composite_builds.html)
+and [Pre-compiled Script Plugins](https://docs.gradle.org/current/userguide/implementing_gradle_plugins_precompiled.html)
+acting as [Convention Plugins](https://docs.gradle.org/current/userguide/implementing_gradle_plugins_convention.html).
+
+### Plugins:
+- `dfbuild.base`: common build logic for all projects.
+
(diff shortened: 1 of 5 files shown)


build-settings-logic/src/main/kotlin/dfsettings.base.settings.gradle.kts
```diff
@@ -0,0 +1,32 @@
+val jupyterApiTCRepo: String? by settings
+
+dependencyResolutionManagement {
```
Jolanrensen, line 3: Using `dependencyResolutionManagement.repositories`  (https://docs.gradle.org/current/userguide/centralizing_repositories.html), while incubating, is the [recommended way to go](https://docs.gradle.org/current/userguide/best_practices_dependencies.html#set_up_repositories_in_settings). 

I like the idea of defining repositories in a single settings file, but what I don't like is that there is no 'merge' option (like mentioned [here](https://github.com/gradle/gradle/issues/32443#issuecomment-2665808893)). So if you want to add another repository for a submodule, you can only overwrite áll repositories defined here in the settings file. I hope merging is added soon :)

Alternatively, we could put repositories in `dfbuild.base.gradle.kts` and apply the `base` convention plugin in all submodules. Also fine by me, but it's not following the best practices.

Jolanrensen:
Thanks for reviewing @zaleslaw @koperagen @AndreiKingsley ! I'll add two small readme's in the new folders and then merge :)

Issue #1555, opened 2025-11-07
readCSV deprecation, how do we specify charset now?


corlaez:
I couldn't find a way to specify charset at all, my file is utf-8. The deprecated method has a charset parameter but the new method doesn't offer any way to change the charset, not even through parser options.

am i missing something? why not keep the charset param?

Jolanrensen:
Hi! I've moved this issue to DataFrame, as that holds the `readCsv()` function.

The new `readCsv()` implementation is based on the blazing fast [Deephaven CSV library](https://github.com/deephaven/deephaven-csv/tree/main). This library reads files at the byte level, so it handles charsets itself. As far as I know, Deephaven and its [Double parser](https://github.com/wrandelshofer/FastDoubleParser) are based on UTF-8, so you should be good to go :)

If you need some other charset, I would recommend manually reading the file to String first and then calling `DataFrame.readCsvStr()`.

corlaez:
From what I see the new method is not using UTF-8

I have a UTF encoded csv file and I see the following:
deprecated:
<img width="905" height="399" alt="Image" src="https://github.com/user-attachments/assets/cd1803c1-2fa1-4064-8917-0a5830c41a09" />
new:
<img width="788" height="284" alt="Image" src="https://github.com/user-attachments/assets/07e9f6fa-042d-43eb-9357-200376b0e95d" />

(Sure, I shouldn't be using the weird " for a json, but it serves to illustrate the wrong encoding)

Jolanrensen:
@corlaez could you send a sample of your data? Since it's the `"` character breaking, it might be because of the `quote =` parameter, which is `"` by default :)

corlaez:
```
Id,InsertedAt,RefTimestamp,mutableData
1,2000-01-01 01:01:01,2000-01-01 01:01:01,"{“mg”: 15, “socialOnly”: false}"
```
`\u201c`
and 
`\u201d`

corlaez:
`"ȭ"` in mutable data displays as 

old:

<img width="765" height="194" alt="Image" src="https://github.com/user-attachments/assets/acc44f1f-8f9b-4e6c-97c1-9c7faf368e76" />

new:
<img width="784" height="191" alt="Image" src="https://github.com/user-attachments/assets/3f2827e0-1470-446e-8401-4bdaab24a497" />

Jolanrensen:
[...]

PR #1665, opened 2026-01-22
Custom Csv charsets support

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1555#issuecomment-3780867178

- Adds `charset` argument to all readCsv-like functions
- Deprecated old functions as HIDDEN for binary compatibility (breaks unnamed function call compatibility, but with a function as large as this, I hope people already came to the conclusion that named arguments are superior for it ;P)
- Detects charsets automatically from the BOM of UTF-* files if they exist!


dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readDelim.kt
```diff
@@ -103,6 +108,7 @@ public fun DataFrame.Companion.readDelim(
  */
 public fun DataFrame.Companion.readDelim(
     file: File,
+    charset: Charset? = CHARSET,
```
koperagen: Do you have a preference on the order? At first i thought it's sorted, but looks like order is arbitrary? IMO maybe charset should go after delimiter and header based on how frequently it could be used. Personally i don't remember having to change charset from default to read a CSV file.

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentationCsv/DelimParams.kt
```diff
@@ -42,7 +43,7 @@ internal object DelimParams {
      */
     interface FILE_OR_URL_READ
 
-    /** @param inputStream Represents the file to read. */
+    /** @param inputStream Represents the file to read. Can be encoded in UTF-8 or ASCII. */
```
koperagen: Is it accurate? I see simpleCsvUtf16leZip

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readDelim.kt
Jolanrensen: No, I can move it for sure :) I just thought it needed to be close to `file`/`path` etc. as it's a direct attribute of it, plus, I would always recommend using named parameters anyways, so I saw no harm. But I can see some people might like to be able to type: `DataFrame.readCsv(source, ';')`, so I'll move it downwards

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentationCsv/DelimParams.kt
Jolanrensen: right,  I copied this statement from Deephaven, but it's only the default case. In our "default" case it can even be an input stream with BOM, making this even less accurate XD I'll just remove it altogether, or just link to `charset`.

Jolanrensen:
> Probably is it worth to mention it on the site documentation?

We could, but it won't be there until the next release, plus it's a useful but niche feature that I think people wil find when they need it.

PR #1666, opened 2026-01-23
Add inplace rename for col(index)

koperagen:
We even had a user request once that goes somewhat like
1. there's CSV with columns
2. it would be nice to select columns from df by index and rename them. why not?

This continues recent development of String API in compiler plugin

Once this functionality implemented there, we'll be able to:
```
val df = DataFrame.readCsv("...")
val df1 = df.convert { col<Int>(0) named "newA" }.with { it * 10 }
val v: Int = df1[0].newA
```

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/rename.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/rename.kt
@@ -11,6 +11,7 @@ import org.jetbrains.kotlinx.dataframe.columns.ColumnAccessor
 import org.jetbrains.kotlinx.dataframe.columns.ColumnReference
 import org.jetbrains.kotlinx.dataframe.columns.ColumnWithPath
 import org.jetbrains.kotlinx.dataframe.columns.FrameColumn
+import org.jetbrains.kotlinx.dataframe.columns.SingleColumn
 import org.jetbrains.kotlinx.dataframe.columns.renamedReference
 import org.jetbrains.kotlinx.dataframe.columns.toColumnSet
 import org.jetbrains.kotlinx.dataframe.documentation.AccessApiLink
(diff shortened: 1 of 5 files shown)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/rename.kt
```diff
@@ -156,6 +156,40 @@ class RenameTests : ColumnsSelectionDslTests() {
...
+
+    @Test
+    fun `col by index named with convert`() {
+        val df = dataFrameOf(
+            "a" to columnOf(1, 4),
+            "b" to columnOf(2, 5),
+            "c" to columnOf(3, 6),
+        )
+
+        val result = df.convert { col<Int>(0) named "newA" }.with { it * 10 }
```
Jolanrensen, line 188: didn't realize this was impossible before :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/rename.kt
```diff
@@ -619,6 +627,15 @@ public interface RenameColumnsSelectionDsl {
...
         toColumnAccessor().named(nameOf.columnName)
 
+    /**
+     * @include [CommonRenameDocs]
+     * @include [CommonRenameDocs.NamedFunctionName]
+     * @include [CommonRenameDocs.SingleColumnReceiver]
+     * @include [CommonRenameDocs.StringParam]
+     */
+    @Interpretable("Named1")
+    public infix fun <C> SingleColumn<C>.named(newName: String): SingleColumn<C> = RenamedSingleColumn(this, newName)
```
Jolanrensen: for consistency with RenamedColumnReference, maybe you could make an `internal fun <C> SingleColumn<C>.renamedColumn(newName: String): SingleColumn<C> =
    RenamedSingleColumn(this, newName)`?

Alternatively, there's also a functional approach: `this.transformSingle { it.rename(...) }.single()`, but using a separate class is also fine :)

Issue #1604, opened 2025-12-03
Compiler plugin generated local types from `@Composable` lambdas don't work


Jolanrensen:
<img width="739" height="848" alt="Image" src="https://github.com/user-attachments/assets/5ea267d7-6151-4eae-8c6b-aa5907408a79" />

The error reads `'val DataRow<Person_59I>.age2: Int' cannot be called in this context with an implicit receiver. Use an explicit receiver if necessary.` Unfortunately, the only way to fix it seems to be moving it outside the lambda body.

To reproduce: 

```kt
@DataSchema
data class Person(val age: Int, val name: String)

@Composable
fun DataFrameScreen(df: DataFrame<Person>) {
    val a = df
        .add("age2") { age }
        .filter { age2 >= 20 }
    a.age
    a.age2

    Column {
        val b = df
            .add("age2") { age }
            .filter { age2 >= 20 }
        b.age
        b.age2
        a.age
        a.age2
    }
}
```

koperagen:
Workaround: put all dataframe code at the top of Composable function outside lambdas. Actually only code that generated local types such as add, convert. filter is ok inside lambda. But i'd say just move everything.

Jolanrensen:
Maybe we can put a note about this in the compiler plugin docs on the website :) and then link to it from our documentation about using dataframe on android

Issue #1610, opened 2025-12-04
Update android example


Jolanrensen:
The android example needs to have some versions bumped and could have some implementations moved to the .toml file

zaleslaw:
Please clarify a task (especially .toml file) with Definition of Done
Is it separate or common file, what kind of dependencies are outdated?

PR #1669, opened 2026-01-26
Describe issue with using compiler plugin in lambdas with DslMarker receivers

koperagen:
Related to https://github.com/Kotlin/dataframe/issues/1604, but i'd like to add a warning to resolve it. For now here's a description with a workaround.
While investigating i found that we don't use `remember` where we should to avoid unnecessary re compositions

And small UI problem. I think it's only applicable to Preview, but worth to fix anyway
image

image

Lastly i couldn't import project in Android studio, so i updated AGP version according to its recommendation

### DIFF
--- a/docs/StardustDocs/topics/compilerPluginLimitationsAndWorkarounds.md
+++ b/docs/StardustDocs/topics/compilerPluginLimitationsAndWorkarounds.md
@@ -1,0 +1,65 @@
+# Known limitations and workarounds
+
+### Compiler plugin in lambdas with receiver marked as DslMarker
+
+Problem: Property calls on a dataframe type created inside @Composable `Column { }` lambda cannot be resolved.
+[Issue 1604](https://github.com/Kotlin/dataframe/issues/1604)
+
+The lambda of `Column` has a receiver parameter `content: @Composable ColumnScope.() -> Unit`.
+
+Here's the declaration from Compose. Receiver parameter types with annotations similar to this one will conflict with the plugin.
+```kotlin
+@LayoutScopeMarker
+interface ColumnScope
+
+@DslMarker
+annotation class LayoutScopeMarker
+```
+
+Repro: The snippet below shows a dataframe variable initialized with a local DataFrame type inside a `Column` lambda. `ageComposableLambdaScope` cannot be resolved.
+
+```kotlin
+@DataSchema
+data class Person(val age: Int, val name: String)
+
+@Composable
+fun DataFrameScreen(df: DataFrame<Person>) {
+    Column {
+        val filteredDf = remember(df) {
+            df
+                .add("ageComposableLambdaScope") { age }
+                .filter { ageComposableLambdaScope >= 20 }
+        }
+        filteredDf.ageComposableLambdaScope // error
+    }
+}
+```
+
+Error message:
+```
+val ColumnsScope<Person_59I>.ageComposableLambdaScope: Int' 
+cannot be called in this context with an implicit receiver. 
+Use an explicit receiver if necessary
+```
+
+Workaround:
+Initialize your dataframe properties outside lambdas with DslMarker receiver parameters.
+
+```kotlin
+@DataSchema
+data class Person(val age: Int, val name: String)
+
+@Composable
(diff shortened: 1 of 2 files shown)


Jolanrensen:
Also related to https://github.com/Kotlin/dataframe/issues/1610 a bit :)

docs/StardustDocs/topics/compilerPluginLimitationsAndWorkarounds.md
```diff
@@ -0,0 +1,54 @@
+# Known limitations and workarounds
+
+### Compiler plugin in lambdas with receiver marked as DslMarker
+
+Problem: Property calls on a dataframe type created inside @Composable `Column { }` lambda cannot be resolved.
+[Issue 1604](https://github.com/Kotlin/dataframe/issues/1604)
+
+Lambda of `Column` has a receiver parameter `content: @Composable ColumnScope.() -> Unit`
```
Jolanrensen: The lambda of `Column` has a receiver parameter: `content: @Composable ColumnScope.() -> Unit`.

docs/StardustDocs/topics/compilerPluginLimitationsAndWorkarounds.md
```diff
@@ -0,0 +1,54 @@
...
+
+```kotlin
+@LayoutScopeMarker
+interface ColumnScope
+
+@DslMarker
+annotation class LayoutScopeMarker
+```
+
+Repro: The snippet below shows how two df variables are initialized with local dataframe types coming from `add` call in different scopes
```
Jolanrensen: *dataframe variables (the concept)
*local DataFrame types (the library)

Jolanrensen:
When writing pages for the website, please check you write full sentences with punctuation. Otherwise it looks good :)
Clever to use `remember() {}` in Compose on Android! It's a good practice indeed :)

docs/StardustDocs/topics/compilerPluginLimitationsAndWorkarounds.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+Repro: The snippet below shows how two dataframe variables are initialized with local DataFrame types coming from `add` call in different scopes.
+Property calls can be resolved for `a`, but cannot be resolved for `b`.
+
+```kotlin
+@DataSchema
+data class Person(val age: Int, val name: String)
+
+@Composable
```
Jolanrensen, line 25: It might be clearer if you split this example because it now contains the bug and the solution interwoven with each other.

I'd first write the reproducer containing the bug, then separately write the workaround, so it's easy to see the difference.

AndreiKingsley (approved):

docs/StardustDocs/topics/compilerPluginLimitationsAndWorkarounds.md
```diff
@@ -0,0 +1,64 @@
+# Known limitations and workarounds
+
+### Compiler plugin in lambdas with receiver marked as DslMarker
+
+Problem: Property calls on a dataframe type created inside @Composable `Column { }` lambda cannot be resolved.
+[Issue 1604](https://github.com/Kotlin/dataframe/issues/1604)
+
+The lambda of `Column` has a receiver parameter `content: @Composable ColumnScope.() -> Unit`.
+
+```kotlin
```
Jolanrensen, line 11: Ah, so it's because Compose has made their ColumnScope use DslMarker

Instead of pasting these two interfaces here, (which I didn't understand came from Compose, at first), it may be clearer to just tell they use DslMarker instead of showing it like this.

So, something like:

"The lambda of `Column {}` uses `@DslMarker` to limit access to outer scopes, which blocks access to generated DataFrame accessors."

koperagen: I'll update to clarify 

> "The lambda of Column {} uses @DslMarker to limit access to outer scopes, which blocks access to generated DataFrame accessors."

This is not accurate, i'm also not 100% what happens yet - need to debug the error message

PR #1670, opened 2026-01-29
Convention plugins part 2 (and disable KSP/Gradle plugin)

Jolanrensen:
Helps part 2 of https://github.com/Kotlin/dataframe/issues/1661: moving all logic from `allProjects` to build conventions

- moved all logic for `kotlin("jvm")` (except for examples) to `dfbuild.kotlinJvm8/11` convention plugins.
    - This makes it easier to see from each module/project which JVM version they use, and we have a central place to define common compiler options, like opt-ins :)
    - This also includes the `instrumentedJars` configuration and task, used by `:samples` and `:dataframe-gradle-plugin` (for now)
- All modules now use `explicitApi()` by default. Fixed instances where this was not the case
- KtLint is enabled by default for all kotlin modules and set up via the version catalog :)
- Moved buildConfig to a build-logic too, allowing us to have a central place to configure compile-time constants like `BuildConfig.VERSION` and `BuildConfig.DEBUG`
- I cherry-picked disabling gradle/ksp plugins from https://github.com/Kotlin/dataframe/pull/1657. This helps building the project successfully.


Jolanrensen:
~~oh, it seems I broke `:core:samplesTest`/explainer dataframes... Need to check what happened there.
...but it always already failed right? that shouldn't be an issue.~~

No...it seems other tests are failing, taking way too long, but it doesn't appear in the logs

Jolanrensen:
I cherry-picked disabling gradle/ksp plugins from https://github.com/Kotlin/dataframe/pull/1657. Maybe this helps building the project successfully.

Jolanrensen:
I also disabled `publishPlugins` in TeamCity, since that task cannot be called anymore, as the plugins are disabled ;P

PR #1677, opened 2026-02-06
Examples separate projects

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1614
Fixes https://github.com/Kotlin/dataframe/issues/1610
Step 3 of https://github.com/Kotlin/dataframe/issues/1661

- Introduces example structure like:
    - /examples/projects/1,2,3: release versions of examples
    - /examples/projects/dev/1,2,3: dev versions of examples
Dev versions of the examples are built using the sources of the project, while release examples use the latest release versions.
Since any example is a completely separate project, you can attach them to the IDE separately only if you need it :).
And most importantly, each example could be extracted to a ZIP and runnable by a user without any additional setup (TODO test this)
- Currently I only moved the android, gradle, and maven example projects into the new folder.

- Running `:promoteDevExamples` copies all /dev examples upwards, making them the new "release" versions. `:syncExampleFolders` is run afterwards.

- Running `:syncAllExampleFolders` (run automatically every `:assemble`) overwrites all relevant versions and properties in the examples from our main libs.versions.toml, gradle-wrapper.properties, .editorconfig, etc. This also works for any maven examples (see build-logic/src/.../buildExampleProjects for the specific implementation).
GH actions was updated to run this command automatically, keeping our examples up-to-date :)
- Running the Test task `:runBuildAllExampleFolders` (run automatically every `:test` in debug mode, so on TC) builds all example projects using `TestBuildingExampleProjects`. Each example project folder gets its own auto-generated Junit test so task dependencies can be set properly and TeamCity can run them in parallel.
    - Thanks to the `:syncAllExampleFolders` task, all dev Gradle Example projects get `includeBuild("../../../..")` in settings.gradle.kts, which substitutes any DataFrame dependeny with our actual sources. This also makes working inside them from the IDE as easy as if it were a subproject :).
    - The Android examples are ignored if the Gradle property `android.sdk.dir` is unset. I modified TeamCity to provide `-Pandroid.sdk.dir=%env.ANDROID_SDK_HOME%` and it seems to work :) This gradle property is passed down to any example containing "android" (in any form) in its name. (This logic can change if needed)
    - The Maven dev project requires a published version of DataFrame. We cannot substitute a Gradle build into Maven in the same way (AFAIK). So for that we use `:publishLocal` to publish the artifacts to /build/maven before running the tests.


Jolanrensen:
I might squash some commits later, as there's been a lot of trial-and-error between me and TeamCity, but all seems to build fine now :)
EDIT: Seems like this is too late... Oops. Should've done that before I merged the master into here...

Carefully read the documentation I wrote. It should explain most of the steps in detail.

examples/projects/android-example/app/build.gradle.kts
```diff
@@ -66,9 +66,17 @@ dependencies {
...
     // https://kotlin.github.io/dataframe/setupcustomgradle.html
-    implementation("org.jetbrains.kotlinx:dataframe-core:1.0.0-Beta4")
-    implementation("org.jetbrains.kotlinx:dataframe-json:1.0.0-Beta4")
-    implementation("org.jetbrains.kotlinx:dataframe-csv:1.0.0-Beta4")
+    //
     // You can add any additional IO modules you like, except for 'dataframe-arrow'.
     // Apache Arrow is not supported well on Android.
+    //
+    // Either use:
+    implementation(libs.dataframe.core)
```
koperagen, line 74: I still believe this notation makes lives of users harder. Impossible to copy paste dependency or even remember it

koperagen: Seems that it's needed for easier sync

koperagen:
Great! Could i ask you to clone fresh clean repo on this branch and import into IDE? I'll do the same 
`git clone -b examples-separate-projects --single-branch https://github.com/Kotlin/dataframe.git`
Let's verify build is ok

minor: promoteExamples already generates some changes:
image

koperagen:
Some invisible changes in mvnw after running runBuildAllExampleFolders:
image

I assume Android examples are disabled because i don't have local.properties android sdk?

build-logic/README.md
```diff
@@ -9,15 +9,50 @@ acting as [Convention Plugins](https://docs.gradle.org/current/userguide/impleme
...
+    - Only apply this plugin to the root project!
+    - Projects inside `/examples/projects` are built using the latest release version of DataFrame.
+      They are meant to be downloadable as separate projects by users.
+    - Projects inside `/examples/projects/dev` are built using sources of DataFrame.
+      They should be kept up to date as sources change.
+    - Available tasks:
+        - `sync<NameOfFolder(Dev)>`: Syncs and overwrites build settings and versions.
+          Copies and overwrites versions from the main `libs.versions.toml`, `gradle.properties`, etc.
+        - `syncAllExampleFolders`: Syncs all example projects. Automatically called on `assemble`.
+        - `promoteExamples`: Promotes the `/examples/projects/dev` example projects to `/examples/project`.
```
koperagen: promoteDevExamples?

gradle/libs.versions.toml
```diff
@@ -151,6 +153,7 @@ swagger = { group = "io.swagger.parser.v3", name = "swagger-parser", version.ref
 
 gradlePlugin-gradle-foojayToolchains = { module = "org.gradle.toolchains:foojay-resolver", version.ref = "gradlePlugin-gradle-foojayToolchains" }
 typesafe-conventions = { module = "dev.panuszewski:typesafe-conventions-gradle-plugin", version.ref = "typesafe-conventions" }
+maven-invoker = { module = "org.apache.maven.shared:maven-invoker", version.ref = "maven-invoker" }
```
zaleslaw, line 159: What does this invoker?

build-logic/src/main/kotlin/dfbuild/buildExampleProjects/gradleTasks.kt
```diff
@@ -0,0 +1,130 @@
...
+                    else ->
+                        settingsGradleKtsContent
+                }
+                    // keep only one newline at the end
+                    .dropLastWhile { it == '\n' }.plus('\n')
+
+            settingsGradleKts.writeText(newSettingsGradleKtsContent)
+
+            // overwrite .editorconfig
+            folder.resolve(".editorconfig").writeText(sourceEditorConfig.readText())
```
koperagen, line 128: Should we bundle editorconfig with our rules to examples projects?

examples/README.md
```diff
@@ -26,6 +26,17 @@ They show how to convert to and from Kotlin DataFrame and their respective table
   * **Multik**: See the [multik folder](./idea-examples/unsupported-data-sources/multik)
     for an example of using Kotlin DataFrame with [Multik](https://github.com/Kotlin/multik).
 
+The example projects in [examples/projects](projects) always target the latest stable version of the library.
+Projects compatible with the latest dev/master version are located in the [examples/projects/dev](projects/dev) folder.
+
+All examples are tested by the main project in debug mode when `gradlew test` is run
```
zaleslaw, line 32: Also could be a useful information for our site https://kotlin.github.io/dataframe/guides-and-examples.html#examples

examples/projects/kotlin-dataframe-plugin-maven-example/pom.xml
```diff
@@ -9,12 +9,25 @@
...
+
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <kotlin.code.style>official</kotlin.code.style>
         <kotlin.compiler.jvmTarget>11</kotlin.compiler.jvmTarget>
     </properties>
 
     <repositories>
+        <repository>
+            <id>mavenLocal</id>
+            <url>file://${user.home}/.m2/repository</url>
```
zaleslaw: Will it work on windows? hmmm

examples/projects/kotlin-dataframe-plugin-gradle-example/.editorconfig
```diff
@@ -0,0 +1,41 @@
+root = true
```
zaleslaw, line 1: do we need this file in each project?

examples/projects/dev/android-example/gradle.properties
```diff
@@ -6,7 +6,7 @@
 # http://www.gradle.org/docs/current/userguide/build_environment.html
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8
```
zaleslaw, line 9: Does it require so much Metasspace?

build-logic/src/testBuildingExamples/kotlin/dfbuild/buildExampleProjects/TestBuildingExampleProjects.kt
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Implementations of this class are auto-generated from 'examples/projects'
+ * into 'build/generated/testBuildingExamples'.
+ *
+ * This class, as well as the generated tests, are automatically registered as
+ * the 'testBuildingExamples' [SourceSet][org.gradle.api.tasks.SourceSet] by the
+ * `dfbuild.buildExampleProjects` convention plugin.
+ */
+@Suppress("unused")
+abstract class TestBuildingExampleProjects {
```
zaleslaw, line 24: I have a feeling that we need to add this to our knowledge library: for example you should demo it for particular version and we will record it

build-logic/src/main/kotlin/dfbuild.buildExampleProjects.gradle.kts
```diff
@@ -0,0 +1,299 @@
...
+
+tasks.named<KotlinCompile>("compile${testBuildingExamples.name.capitalized()}Kotlin") {
+    compilerOptions {
+        jvmTarget = JvmTarget.JVM_21
+        freeCompilerArgs.add("-Xjdk-release=21")
+    }
+}
+
+tasks.named<JavaCompile>("compile${testBuildingExamples.name.capitalized()}Java") {
+    sourceCompatibility = JavaVersion.VERSION_21.toString()
```
zaleslaw: Should we somehow parametrize it to make easier migration on the next Java?

build-logic/src/main/kotlin/dfbuild/buildExampleProjects/generateTestCase.kt
```diff
@@ -0,0 +1,31 @@
+package dfbuild.buildExampleProjects
+
+import org.intellij.lang.annotations.Language
+import java.io.File
+
+typealias Code = String
+
+@Language("kt")
+fun generateTestCase(testClassName: String, folder: File, isDev: Boolean, tags: List<String>): Code =
```
zaleslaw: I have a feeling that it should be a description of the processes, its goal, why do we need this generation

examples/projects/android-example/app/build.gradle.kts
Jolanrensen: I think it's okay, since we will provide them an all-in-one project with libs.versions.toml. This is also the recommended structure by Gradle, so I guess we would just help them this way :)
And indeed, it makes sync way easier!

Jolanrensen:
> I assume Android examples are disabled because i don't have local.properties android sdk?

@koperagen Exactly :) see `gradle.properties` or `/build-logic/README.md` for how to set this up. I thought this was the least invasive way of skipping the tests on platforms that don't (or cannot) have an Android SDK.

build-logic/README.md
Jolanrensen: sure, sounds like a better name :)

build-logic/src/main/kotlin/dfbuild/buildExampleProjects/gradleTasks.kt
Jolanrensen: We don't necessarily have to, but we did tweak the settings to work well with DataFrame. Plus, if we bump KtLint, the version will be synced to the examples as well. If we then need to disable one of the new rules, we will need to do so in each example as well... So yeah

gradle/libs.versions.toml
Jolanrensen: well... maven XD

Jolanrensen: It's added as dependency to the sourceSet running the build tests. Maven invoker can programmatically run a maven build, so we can even test our non-gradle examples :)

build-logic/src/main/kotlin/dfbuild/buildExampleProjects/gradleTasks.kt
koperagen: Doesn't it work for us even without these files, re-using root editorconfig for entire dataframe project? 
People like freedom to choose codestyle in their projects, i don't see why we should enforce anything beyond what IntelliJ already does

examples/README.md
Jolanrensen: I'm not sure. We could, but then it would be mostly boasting about how well we are looking after our examples. I think most users actually already expect we test our examples (I mean, we expected it ourselves too). What is more interesting for users is that they are kept up-to-date :) meaning they can always download a working, up-to-date example from the latest green-built master :)

examples/projects/kotlin-dataframe-plugin-maven-example/pom.xml
Jolanrensen: could you try?

examples/projects/kotlin-dataframe-plugin-gradle-example/.editorconfig
Jolanrensen: yes, definitely. Each project should be individually downloadable as Zip. If you don't supply `root=true`, KtLint will scan the entire system upwards to find all .editorconfig files.

examples/projects/dev/android-example/gradle.properties
Jolanrensen: Not entirely sure, but what I am sure is that we need to limit this somehow, as to not exceed TeamCity's memory. This makes it so the test won't use more than 2.5GB

build-logic/src/main/kotlin/dfbuild.buildExampleProjects.gradle.kts
Jolanrensen: Might be a good idea indeed :) this is the jvm version of `:build-logic` itself, but we can access our version catalog from there as well, so we might just be able to put it there :)

build-logic/src/main/kotlin/dfbuild/buildExampleProjects/gradleTasks.kt
Jolanrensen: that's true, though that might mean that when a user downloads the project and tries to run it, it immediately fails because of KtLint, as it will then use the KtLint defaults, without our exceptions.

And I do want to enable the KtLint plugin in our examples, but mostly for us, so our examples are well-formatted and representable.

Of course they can still change the .editorconfig rules if they wish, it's their project after all once they download it.

examples/projects/kotlin-dataframe-plugin-maven-example/pom.xml
Jolanrensen: actually, I can probably remove this altogether

Jolanrensen:
@zaleslaw @koperagen Okay, final checks. I updated some more docs and other things you mentioned in the review.

Importantly, I updated all links in the documentation and README to point to the new folders and I updated the archive_examples GH actions workflow to also dynamically scan the /examples/projects folder, so we won't need to udpate that ourselves later when we add more examples :)

Check https://github.com/Kotlin/dataframe/tree/example-projects-archives?tab=readme-ov-file for the result of a run

PR #1682, opened 2026-02-13
Support Java Record in toDataFrame conversion

koperagen:


core/build.gradle.kts
```diff
@@ -77,6 +79,20 @@ dependencies {
...
 }
 
+// Configure test sources to use Java 16+ to test record support
+tasks.named<JavaCompile>("compileTestJava") {
+    sourceCompatibility = JavaVersion.VERSION_16.toString()
+    targetCompatibility = JavaVersion.VERSION_16.toString()
+    options.release.set(16)
+}
+
+tasks.named<KotlinCompile>("compileTestKotlin") {
```
Jolanrensen: This may be a slight issue, as this sets áll :core tests to java 16. We claim we support Java 8 for DataFrame, but we may not catch all issues correctly if we don't test with java 8. I found differences/bugs, for instance, with parsing certain Doubles in RTL locales which only occurred in Java 11, but not in 8 and the other way around.

Jolanrensen: is there maybe a mechanism to only run java-record tests with jvm 16 while keeping the others at 8?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/Utils.kt
```diff
@@ -414,6 +415,57 @@ internal fun KCallable<*>.isGetterLike(): Boolean =
...
+ * this will always return `false`.
+ */
+internal val KClass<*>.isJavaRecord: Boolean
+    get() = try {
+        isRecordMethod?.invoke(this.java) as? Boolean ?: false
+    } catch (e: Exception) {
+        false
+    }
+
+private val getRecordComponentsMethod: java.lang.reflect.Method? by lazy {
```
Jolanrensen: You can probably just import `Method` right? Or do you want to emphasize it's from Java? In that case I'd recommend importing it with the alias `JavaMethod` :) I do the same for Dates, Instants etc.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/schema/Utils.kt
```diff
@@ -172,18 +173,24 @@ internal fun createEmptyDataFrameOf(clazz: KClass<*>): AnyFrame =
  * `null` otherwise.
  */
 internal fun getPropertyOrderFromPrimaryConstructor(clazz: KClass<*>): Map<String, Int>? =
-    (clazz.primaryConstructor ?: clazz.constructors.singleOrNull())
-        ?.parameters
-        ?.mapNotNull { it.name }
-        ?.mapIndexed { i, v -> v to i }
-        ?.toMap()
+    if (clazz.isJavaRecord) {
+        // Can fail for Java records with primitive types due to KT-58649
```
Jolanrensen, line 177: interesting... Does it always fail? Otherwise, we could maybe catch the exception so the order is correct if it doesn't fail or the issue is fixed on the reflection side.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/schema/Utils.kt
```diff
@@ -172,18 +173,24 @@ internal fun createEmptyDataFrameOf(clazz: KClass<*>): AnyFrame =
...
+            ?.mapNotNull { it.name }
+            ?.mapIndexed { i, v -> v to i }
+            ?.toMap()
+    }
 
 /**
  * Sorts [this] according to the order of their [columnName] in the primary/single constructor of [klass]
  * if it exists, else, it falls back to lexicographical sorting.
  */
 internal fun <T> Iterable<KCallable<T>>.sortWithConstructor(klass: KClass<*>): List<KCallable<T>> {
```
Jolanrensen, line 193: I assume you removed this check for records? Maybe instead of removing it, you could skip the check only when dealing with a record

Jolanrensen:
Nice!

core/build.gradle.kts
koperagen: This goes both way, isn't it? we claim to support java 21 but don't test against it. I suggest, if we're interested, to specifically run tests compiled with JDK8 compatibility using JDK8, JDK17, JDK 21. Start with :core. It can be done

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/schema/Utils.kt
koperagen: it always fails if there's a primitive type in the "constructor", like `int`. I'd wait until this is fixed on Kotlin side, logic would seem unstable + strange to reproduce it in the plugin (checking if there's a primitive..?)

core/build.gradle.kts
Jolanrensen: yep you're right. Maybe we can run our tests (only in debug mode) with all support LTS java versions :) Would be a nice convention plugin to apply to all modules :)

It's also possible to add a `@Tag` to certain tests so that they can be in/excluded for certain java versions. Maybe after this PR you could give it a try? I'll add the idea to #1661 anyways, because I like it!

koperagen: Sure, can do. In the meantime, reconsidered the approach in the latest commit. Now only 1 test class is compiled and run for jdk 17

core/build.gradle.kts
```diff
@@ -385,6 +419,24 @@ tasks.test {
...
 
+// Test task for Java 16+ language-specific tests
+val testJava16 = tasks.register<Test>("testJava16") {
+    group = LifecycleBasePlugin.VERIFICATION_GROUP
+
+    testClassesDirs = sourceSets["testJava16"].output.classesDirs
+    classpath = sourceSets["testJava16"].runtimeClasspath
+
+    javaLauncher = javaToolchains.launcherFor {
+        languageVersion = JavaLanguageVersion.of(17)
```
Jolanrensen, line 430: 16 or 17? ;P

core/build.gradle.kts
```diff
@@ -385,6 +419,24 @@ tasks.test {
     maxHeapSize = "2048m"
 }
 
+// Test task for Java 16+ language-specific tests
+val testJava16 = tasks.register<Test>("testJava16") {
+    group = LifecycleBasePlugin.VERIFICATION_GROUP
+
+    testClassesDirs = sourceSets["testJava16"].output.classesDirs
```
Jolanrensen: maybe if you use `val testJava16 by sourceSets.creating {}` above, you can refer to them safely here :)

core/build.gradle.kts
Jolanrensen: great! I like it :) Though, I will definitely try to make it a convention plugin later on, as this adds a lot of code to the :core build file.

core/build.gradle.kts
koperagen: we only need Java 16 language features, but it's better to have LTS jdk 17 for running :)

Issue #1570, opened 2025-11-13
Calling `importDataSchema()` without `enableExperimentalOpenApi` provides unhelpful exception


Jolanrensen:
As found [here](https://kotlinlang.slack.com/archives/C4W52CFEZ/p1763042367694659) without `enableExperimentalOpenApi=true`, the `importDataSchema()` function throws a `NotImplementedError`.

This is very unhelpful, as it does not provide any information about what to do.

Jolanrensen:
Well, seems I found the origin of the `NotImplementedError`. It's https://github.com/Kotlin/dataframe/blob/635f57a23c881f5ba83b94750ff3d46eb869bbaf/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/Jdbc.kt#L29 from JDBC...

Our notebook integration tries all `SupportedCodeGenerationFormat` implementors, somehow `Jdbc` accepts any input and fails with the `NotImplementedError` instead of throwing a proper exception.

Issue #1681, opened 2026-02-12
JDBC: Allow reading schemas without querying actual data


Jolanrensen:
When extracting the DataSchema from an SQL query/database (`DataFrameSchema.readSqlTable()`) we currently call `DataFrame.readSqlTable(..., limit = 1).schema()`. This can cause a couple of issues:
- Error when the query or SQL table does not return any rows (actually, maybe this is fine)
- Error when we lack permission to query data; we're only allowed to look at the schema
- Dependent on how/where the data is stored, it may be unexpected and expensive to read a row

There are multiple ways to read just the Schema without reading actual data:
- use `"SELECT * FROM $fullTableName WHERE 1=0"` to read 0 rows: for [example](https://github.com/Kotlin/dataframe/blob/eb32d0b427412de9cf6a2ea6d0d5cf2821d3d73a/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readDataFrameSchema.kt#L475)
    - Downside is that we still need to query the table, albeit with no rows
- use `Connection.metaData.getColumns()` to get all column metadata: for example: [1](https://github.com/Kotlin/dataframe/blob/eb32d0b427412de9cf6a2ea6d0d5cf2821d3d73a/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readDataFrameSchema.kt#L108) and [2](https://github.com/Kotlin/dataframe/blob/eb32d0b427412de9cf6a2ea6d0d5cf2821d3d73a/dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/DbType.kt#L280).
    - This is the most complete solution, but this usually doesn't give us the fully qualified java type for columns, requiring a watertight TYPE_NAME mapping on the implementor side

Jolanrensen:
We also need to properly implement `class Jdbc : SupportedCodeGenerationFormat, SupportedDataFrameFormat`, especially the `SupportedCodeGenerationFormat` part, as that caused https://github.com/Kotlin/dataframe/issues/1570

Issue #450, opened 2023-09-08
Revisit the hierarchy of SupportedCodeGenerationFormat and SupportedDataFrameFormat to make easier addition of new IO sources


zaleslaw:

Jolanrensen:
So, not properly implementing `SupportedCodeGenerationFormat` actually caused https://github.com/Kotlin/dataframe/issues/1570. `TODO` throws an Error, not an Exception, which we don't catch, causing `importDataSchema()` to completely fail in notebooks if `OpenApi` is not present or cannot parse the supplied file.

PR #1714, opened 2026-02-25
`importDataSchema()` notebooks fix

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1570

Made unsupported functions in `Jdbc` throw exceptions instead of errors. Improved error message for `importDataSchema()` in notebooks. Removed unused `importDataSchema: String`


dataframe-jupyter/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
```diff
@@ -87,14 +90,40 @@ internal class Integration(private val notebook: Notebook, private val options:
...
-                execute("""DISPLAY("Data schema successfully imported as ${property.name}: $name")""")
+                display("Data schema successfully imported as ${property.name}: $name", null)
 
                 name
             }
 
             is CodeGenerationReadResult.Error -> {
-                execute(
-                    """DISPLAY("Failed to read data schema from ${importDataSchema.url}: ${codeGenResult.reason}")""",
+                display(
```
koperagen, line 99: Can you please attach a screenshot?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/importDataSchema.kt
```diff
@@ -19,44 +19,3 @@ public fun importDataSchema(path: String): ImportDataSchema = ImportDataSchema(p
 public fun importDataSchema(path: Path): ImportDataSchema = ImportDataSchema(path)
 
 public fun importDataSchema(file: File): ImportDataSchema = ImportDataSchema(file)
-
-@Language("kts")
-internal val importDataSchema =
```
koperagen, line 24: Is this API still available to users? Where is it declared?

Jolanrensen: nope, turns our we've not been using it anymore for a while already. This String used to be invoked `onLoaded {}` a long time ago, but while it was removed there, the String was still present here.

Still, I checked the docs and examples, and they all use the `val name: ImportDataSchema = importDataSchema()` notation, so I think it's fine to remove this here.

dataframe-jupyter/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
Jolanrensen: image

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/importDataSchema.kt
koperagen: Didn't notice these declaration right above
public fun importDataSchema(path: Path): ImportDataSchema = ImportDataSchema(path)
Now it's clear, ok!

dataframe-jupyter/src/main/kotlin/org/jetbrains/kotlinx/dataframe/jupyter/Integration.kt
koperagen: Looks good!
Maybe remove SupportedCodeGenerationFormat interface from Jdbc while it's not implemented? Available formats: ... looks misleading

Jolanrensen: hmm, that might be an idea. @zaleslaw Are you okay with this? It's not possible to generate dataschema interfaces for JDBC without KSP/gradle after all, right? (also in notebooks, it's not possible to generate just types for jdbc, right?)

Jolanrensen:
@zaleslaw I'll merge it in its current state so the bug is solved. We'll decide whether to remove the `SupportedCodeGenerationFormat` supertype from the `Jdbc` class as a part of #450 or #1681

zaleslaw:
@Jolanrensen, I agree with the proposed way

PR #1715, opened 2026-02-27
Added `requireColumn` operation

### LINKED ISSUE #1808: Unblock String API workflows by allowing schema inference from string-based column access
### Problem

String API is intended as a fallback and incremental migration path to the type-safe API.

However, currently the compiler plugin does not learn schema information from string-based column access. As a result:

- code like `"full_name"<String>()` cannot be replaced with `full_name`
- users cannot progressively move from String API to typed API
- String API and compiler plugin workflows are disconnected

This blocks the intended usage pattern where users start with String API and gradually adopt type-safe access.

### Expected

Allow compiler plugin to infer schema information from string-based column access.

### Proposed solution

Introduce an operation (e.g. `require { ... }`) that:

- asserts presence and type of columns accessed via String API
- updates schema information for the compiler plugin
- enables further usage of typed accessors

Example:

df.require { "full_name"<String>() }
df.full_name // becomes available after require

### Acceptance criteria

- Compiler plugin can learn column types from string-based access
- Typed accessors become available after schema is inferred
- String API remains usable as fallback
- Example pipeline using incremental migration compiles successfully
- Documentation explains migration path from String API to typed API

### Motivation

String API is a key fallback and onboarding mechanism.

Without schema inference:
- users are forced to fully define `@DataSchema` upfront
- incremental migration is not possible
- compiler plugin loses a major part of its usability

This is critical for enabling real-world adoption and should be addressed before 1.0.

With compiler plugin, ideally we want to know about schema as early as possible to help transform it. However forcing to generate DataSchema for all columns can create an entry barrier not desirable for simple pipelines. This is the motivation for String API support that was done earlier. With it, our sample pipeline incrementally evolves schema:

```kt
val repos = DataFrame
    .readCsv("https://raw.githubusercontent.com/Kotlin/dataframe/master/data/jetbrains_repositories.csv")

repos
    .add("name") { "full_name"<String>().substringAfterLast("/") }
    .filter { name.lowercase().contains("kotlin") }

val reposUpdated = repos
    .renameToCamelCase()
    .rename { "stargazersCount"<Int>() }.into("stars")
    .filter { stars > 50 }
    .convert { "topics"<String>() }.with {
        val inner = it.removeSurrounding("[", "]")
        if (inner.isEmpty()) emptyList() else inner.split(',').map(String::trim)
    }
    .add("topicCount") { topics.size }
    .add("kind") { getKind("fullName"() , topics) }

reposUpdated.writeCsv("jetbrains_repositories_new.csv")
```

The only thing lacking here is a convenient way to tell plugin that there's full_name: String column, so could further replace

` .add("name") { "full_name"<String>().substringAfterLast("/") }` => `.add("name") { full_name.substringAfterLast("/") }`
` .add("kind") { getKind("fullName"() , topics) }` => `.add("kind") { getKind(fullName , topics) }`

Proposed solution: `require` operation, addition to `cast` and `convertTo`
```kt
val repos = DataFrame
    .readCsv("https://raw.githubusercontent.com/Kotlin/dataframe/master/data/jetbrains_repositories.csv")
    .require { "full_name"<String>() }

repos.full_name // now we can call full_name because otherwise require would've failed
```

Main difference - require will add new information, not substitute it as cast and convertTo

koperagen:
Seems not feasible to have Column*s*Selector because we cannot make interface members inline functions and have reified type parameters. So, no go for our existing String Column Accessors like String.invoke, col, and so on. context parameters are still considered experimental, so it'll have to wait.
In the meantime, multiple require calls in chain :) 
With this new operation we'll further improve workflow where all compile time schema information is derived from operations: *require* (new), operations with selectors using String Column Accessors, and things like add, toDataFrame, map, etc. This can be a good alternative to usual `DataSchema` workflow, with good potential for incremental introduction of type safety and lower entry barrier. Well, you know the idea.

fixes #1808

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/DataFrameReceiver.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/DataFrameReceiver.kt
@@ -8,6 +8,7 @@ import org.jetbrains.kotlinx.dataframe.api.asDataColumn
 import org.jetbrains.kotlinx.dataframe.api.cast
 import org.jetbrains.kotlinx.dataframe.api.isColumnGroup
 import org.jetbrains.kotlinx.dataframe.api.pathOf
+import org.jetbrains.kotlinx.dataframe.api.toPath
 import org.jetbrains.kotlinx.dataframe.columns.ColumnGroup
 import org.jetbrains.kotlinx.dataframe.columns.ColumnPath
 import org.jetbrains.kotlinx.dataframe.columns.ColumnReference
@@ -21,6 +22,7 @@ import org.jetbrains.kotlinx.dataframe.impl.columns.addPath
 import org.jetbrains.kotlinx.dataframe.impl.columns.missing.MissingColumnGroup
 import org.jetbrains.kotlinx.dataframe.impl.columns.missing.MissingDataColumn
 import org.jetbrains.kotlinx.dataframe.nrow
+import kotlin.collections.map
 
 private fun <T> DataFrame<T>.unbox(): DataFrame<T> =
     when (this) {
@@ -47,9 +49,7 @@ internal open class DataFrameReceiver<T>(
                         host = this@DataFrameReceiver,
                     ).asDataColumn().cast()
 
-                UnresolvedColumnsPolicy.Fail -> error(
-                    "Column '${path.joinToString()}' not found among ${df.columnNames()}.",
-                )
+                UnresolvedColumnsPolicy.Fail -> error(formatMissingColumnMessage(path))
             }
 
             is MissingDataColumn -> this
@@ -58,6 +58,43 @@ internal open class DataFrameReceiver<T>(
 
             else -> this
         }
+
+    // Context:
+    // it's strange that we have to reverse-search why the column is missing
+    // would be nice to "fail fast" exactly where resolve failed, knowing the current path and parent.
+    // but it's very unclear what to do with resolveSingle.
+    // at first glance: a lot of changes.
+    @Suppress("FoldInitializerAndIfToElvis")
+    private fun formatMissingColumnMessage(path: ColumnPath): String {
+        val fullPath = path.joinToString()
+
+        for (depth in path.indices) {
+            val currentPath = path.slice(0..depth).toPath()
+            val currentPathString = currentPath.joinToString()
+            val column = df.getColumnOrNull(currentPath)
+            if (column == null) {
+                return if (depth == 0) {
+                    "Column '$currentPathString' not found among ${df.columnNames()}."
+                } else {
+                    val parentPath = currentPath.dropLast()
+                    val parentPathString = parentPath.joinToString()
+                    val parentColumn = df.getColumnOrNull(parentPath)
+                    if (parentColumn != null && parentColumn.isColumnGroup()) {
+                        "Column '$currentPathString' not found among columns of '$parentPathString': ${parentColumn.columnNames()}."
+                    } else {
+                        "Column '$currentPathString' not found among ${df.columnNames()}."
+                    }
+                }
+            }
+
+            if (depth != path.lastIndex) {
+                if (!column.isColumnGroup()) {
+                    return "Column '$fullPath' cannot be resolved: '$currentPathString' is not a column group."
+                }
+            }
+        }
+        return "Column '$fullPath' not found among ${df.columnNames()}."
+    }
 
     override fun getColumnOrNull(name: String) = df.getColumnOrNull(name).check(pathOf(name))
 
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/require.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/require.kt
@@ -1,0 +1,18 @@
+package org.jetbrains.kotlinx.dataframe.impl.api
+
+import org.jetbrains.kotlinx.dataframe.ColumnSelector
+import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.api.getColumnWithPath
+import org.jetbrains.kotlinx.dataframe.api.isSubtypeOf
+import org.jetbrains.kotlinx.dataframe.type
+import kotlin.reflect.KType
+
+@PublishedApi
+internal fun <T, C> DataFrame<T>.requireImpl(column: ColumnSelector<T, C>, type: KType): DataFrame<T> {
+    val resolvedColumn = getColumnWithPath(column)
+    val actualType = resolvedColumn.data.type
+    require(resolvedColumn.data.isSubtypeOf(type)) {
+        "Column '${resolvedColumn.path.joinToString()}' has type '$actualType', which is not subtype of required '$type' type."
+    }
+    return this
+}
--- a/docs/StardustDocs/topics/require.md
+++ b/docs/StardustDocs/topics/require.md
@@ -3,7 +3,7 @@
 
 Throws an exception if the specified column is missing or its type is not subtype of `C`.
 From the compiler plugin perspective, a new column will appear in the compile-time schema as a result of this operation.
-The aim here is to help incrementally migrate workflows to extension properties API.
+The aim here is to help incrementally migrate workflows to [extension properties API](extensionPropertiesApi.md).
 
 Will work in compiler plugin starting from IntelliJ IDEA 2026.2 and Kotlin 2.4.0.
 
@@ -14,6 +14,11 @@ require { column }
 **Related operations**: [](cast.md), [](convertTo)
 
 ```kotlin
+// Before `require` extension property will not be resolved
+// peopleDf.select { name.firstName }
+
+// Require a column with a runtime check
 val df = peopleDf.require { "name"["firstName"]<String>() }
+// Use extension property after `require`
 val v: String = df.name.firstName[0]
 ```
--- a/docs/StardustDocs/topics/require.md
+++ b/docs/StardustDocs/topics/require.md
@@ -1,0 +1,19 @@
+[//]: # (title: require)
+<!---IMPORT org.jetbrains.kotlinx.dataframe.samples.api.Require-->
+
+Throws an exception if the specified column is missing or its type is not subtype of `C`.
+From the compiler plugin perspective, a new column will appear in the compile-time schema as a result of this operation.
+The aim here is to help incrementally migrate workflows to extension properties API.
+
+Will work in compiler plugin starting from IntelliJ IDEA 2026.2 and Kotlin 2.4.0.
+
+```text
+require { column }
+```
+
+**Related operations**: [](cast.md), [](convertTo)
+
+```kotlin
+val df = peopleDf.require { "name"["firstName"]<String>() }
+val v: String = df.name.firstName[0]
+```
--- a/docs/StardustDocs/topics/require.md
+++ b/docs/StardustDocs/topics/require.md
@@ -4,6 +4,7 @@
 Throws an exception if the specified column is missing or its type is not subtype of `C`.
 From the compiler plugin perspective, a new column will appear in the compile-time schema as a result of this operation.
 The aim here is to help incrementally migrate workflows to [extension properties API](extensionPropertiesApi.md).
+We recommend considering declaring a [DataSchema](dataSchema.md) and use [](cast.md) or [](convertTo) if you end up with more than a few `requireColumn` calls.
 
 Will work in compiler plugin starting from IntelliJ IDEA 2026.2 and Kotlin 2.4.0.
 
--- a/docs/StardustDocs/topics/require.md
+++ b/docs/StardustDocs/topics/require.md
@@ -1,4 +1,4 @@
-[//]: # (title: require)
+[//]: # (title: requireColumn)
 <!---IMPORT org.jetbrains.kotlinx.dataframe.samples.api.Require-->
 
 Throws an exception if the specified column is missing or its type is not subtype of `C`.
@@ -8,17 +8,17 @@ The aim here is to help incrementally migrate workflows to [extension properties
 Will work in compiler plugin starting from IntelliJ IDEA 2026.2 and Kotlin 2.4.0.
 
 ```text
-require { column }
+requireColumn { column }
 ```
 
 **Related operations**: [](cast.md), [](convertTo)
 
 ```kotlin
-// Before `require` extension property will not be resolved
+// Before `requireColumn` extension property will not be resolved
 // peopleDf.select { name.firstName }
 
 // Require a column with a runtime check
-val df = peopleDf.require { "name"["firstName"]<String>() }
-// Use extension property after `require`
+val df = peopleDf.requireColumn { "name"["firstName"]<String>() }
+// Use extension property after `requireColumn`
 val v: String = df.name.firstName[0]
 ```
--- a/docs/StardustDocs/topics/require.md
+++ b/docs/StardustDocs/topics/require.md
@@ -23,3 +23,44 @@ val df = peopleDf.requireColumn { "name"["firstName"]<String>() }
 // Use extension property after `requireColumn`
 val v: String = df.name.firstName[0]
 ```
+
+### Advanced example
+
+Let's start with a pipeline that uses only String Column Accessors and String API overloads:
+
+```kotlin
+val repos = DataFrame
+    .readCsv("https://raw.githubusercontent.com/Kotlin/dataframe/master/data/jetbrains_repositories.csv")
+
+repos
+    .filter { "stargazers_count"<Int>() > 100 }
+    .sortByDesc("stargazers_count")
+    .select("full_name", "stargazers_count")
+```
+
+Notice how stargazers_count String is repeated three times. We can refactor this code using `requireColumn`:
+
+```kotlin
+val repos = DataFrame
+    .readCsv("https://raw.githubusercontent.com/Kotlin/dataframe/master/data/jetbrains_repositories.csv")
+    .requireColumn { "stargazers_count"<Int>() }
+
+repos
+    .filter { stargazers_count > 100 }
+    .sortByDesc { stargazers_count }
+    .select { "full_name" and stargazers_count }
+```
+
+This way code becomes a bit more robust. For example, usages of a renamed column will become compile time errors that are easy to spot and update:
+
+```kotlin
+val repos = DataFrame
+    .readCsv("https://raw.githubusercontent.com/Kotlin/dataframe/master/data/jetbrains_repositories.csv")
+    .requireColumn { "stargazers_count"<Int>() }
+    .rename { stargazers_count }.into("stars")
+
+repos
+    .filter { stars > 100 }
+    .sortByDesc { stars }
+    .select { "full_name" and stars }
+```
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -5492,6 +5492,10 @@ public final class org/jetbrains/kotlinx/dataframe/impl/api/MapKt {
 	public static final fun mapNotNullValues (Lorg/jetbrains/kotlinx/dataframe/DataColumn;Lkotlin/jvm/functions/Function1;)Lorg/jetbrains/kotlinx/dataframe/DataColumn;
 }
 
+public final class org/jetbrains/kotlinx/dataframe/impl/api/RequireKt {
+	public static final fun requireImpl (Lorg/jetbrains/kotlinx/dataframe/DataFrame;Lkotlin/jvm/functions/Function2;Lkotlin/reflect/KType;)Lorg/jetbrains/kotlinx/dataframe/DataFrame;
+}
+
 public final class org/jetbrains/kotlinx/dataframe/impl/api/SchemaKt {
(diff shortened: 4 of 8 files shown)


docs/StardustDocs/topics/require.md
```diff
@@ -0,0 +1,19 @@
+[//]: # (title: require)
+<!---IMPORT org.jetbrains.kotlinx.dataframe.samples.api.Require-->
+
+Throws an exception if the specified column is missing or its type is not subtype of `C`.
+From the compiler plugin perspective, a new column will appear in the compile-time schema as a result of this operation.
+The aim here is to help incrementally migrate workflows to extension properties API.
+
+However, this operation is new and will work in compiler plugin starting from IntelliJ IDEA 2026.2 and Kotlin 2.4.0
```
Jolanrensen: I'd leave out "this operation is new", because it will not be new anymore in a year :) simply stating from which version of Intellij/Kotlin it's available is enough

Jolanrensen: and add a period at the end of the sentence please :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/require.kt
```diff
@@ -0,0 +1,18 @@
...
+import org.jetbrains.kotlinx.dataframe.api.getColumnWithPath
+import org.jetbrains.kotlinx.dataframe.api.isSubtypeOf
+import org.jetbrains.kotlinx.dataframe.type
+import kotlin.reflect.KType
+
+@PublishedApi
+internal fun <T, C> DataFrame<T>.requireImpl(column: ColumnSelector<T, C>, type: KType): DataFrame<T> {
+    val resolvedColumn = getColumnWithPath(column)
+    val actualType = resolvedColumn.data.type
+    kotlin.require(resolvedColumn.data.isSubtypeOf(type)) {
```
Jolanrensen: redundant qualifier name

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/DataFrameReceiver.kt
```diff
@@ -59,6 +59,41 @@ internal open class DataFrameReceiver<T>(
             else -> this
         }
 
+    // it's strange that we have to reverse-search why the column is missing
```
Jolanrensen, line 63: What's the purpose of this comment block?
If it's just an explanation for the PR, I'd just leave a comment there. If it's like a TODO, please add "TODO" and create an issue.

Otherwise, keep the comment to what this function does and why I'd call it :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/DataFrameReceiver.kt
```diff
@@ -59,6 +59,41 @@ internal open class DataFrameReceiver<T>(
...
+    // would be nice to "fail fast" exactly where resolve failed, knowing the current path and parent.
+    // but it's very unclear what to do with resolveSingle.
+    // at first glance: a lot of changes.
+    private fun formatMissingColumnMessage(path: ColumnPath): String {
```
Jolanrensen, line 68: nice to have better error messages indeed :)

Jolanrensen:
> Seems not feasible to have ColumnsSelector because we cannot make interface members inline functions and have reified type parameters. So, no go for our existing String Column Accessors like String.invoke, col, and so on. context parameters are still considered experimental, so it'll have to wait.
> In the meantime, multiple require calls in chain :)

Hmm, so this makes our code look something like:

```kt
df.readCsv("something")
    .require { "name"["firstName"]<String>() }
    .require { "name"["lastName"]<String>() }
    .require { "age"<Int>() }
    .require { "address"<String>() }
```

I don't quite understand why you actually need the reified type parameter at all here. Isn't `require {}` just `select {}` where it checks the columns you mention are there and the original (but refined) DF is returned? The rest is then done on the compiler plugin side. So something like:

```kt
@Refine
@Interpretable("Require0")
public fun <T> DataFrame<T>.require(columns: ColumnsSelector<T, *>): DataFrame<T> {
    // attempts to resolve all columns, throw Exception if any column mentioned is not there
    getColumnsWithPaths(UnresolvedColumnsPolicy.Fail, columns)
    // Managed to successfully resolve all columns, compiler plugin can safely assume they are present
    return this
}
```
When executing
```kt
val newDf = df.readCsv("something")
    .require { "name"["firstName"]<String>() and "name"["lastName"]<String>() and "age"<Int>() }
```
the compiler plugin would make sure that you can call:
```kt
newDf.name.firstName
newDf.age
...
```
Slightly different to `.select {}` because that would expose `newDf.firstName` (pulled out of its group), but that's the only difference, right?

Jolanrensen (changes requested):

koperagen:
> I don't quite understand why you actually need the reified type parameter at all here. Isn't require {} just select {} where it checks the columns you mention are there and the original (but refined) DF is returned? The rest is then done on the compiler plugin side. 

Yes and no? We still need to throw an exception if column is not Int, for example. I'd say we cannot do it for an arbitrary selector like `"a"<Int>() and "c"<String>()` because neither Int nor String are available as KType for us

koperagen:
> I don't quite understand why you actually need the reified type parameter at all here. Isn't require {} just select {} where it checks the columns you mention are there and the original (but refined) DF is returned? The rest is then done on the compiler plugin side. So something like:

select is very similar, now i get what you mean. If we select column but provided type is incorrect, select doesn't immediately throws an exception now. This is a limitation, a problem even. But at least we should be more strict for `require` as name implies

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/DataFrameReceiver.kt
koperagen: I'd be confused to see this approach if i wasn't the author (technically i'm not, this check() function was there before). Wanted to provide some context why in my opinion it's done like this for, hopefully, easier understanding of the class

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/DataFrameReceiver.kt
```diff
@@ -59,6 +59,41 @@ internal open class DataFrameReceiver<T>(
...
+        val fullPath = path.joinToString()
+
+        for (index in path.indices) {
```
koperagen: rename index -> depth

Jolanrensen:
> Yes and no? We still need to throw an exception if column is not Int, for example. I'd say we cannot do it for an arbitrary selector like "a"<Int>() and "c"<String>() because neither Int nor String are available as KType for us
>
> select is very similar, now i get what you mean. If we select column but provided type is incorrect, select doesn't immediately throws an exception now. This is a limitation, a problem even. But at least we should be more strict for require as name implies

Ahh, you're right, it indeed doesn't check that. It does check column kind though. For our `colGroup()`, `valueCol()`, and `frameCol()` accessors inside the CS DSL, they call `ensureIsValueColumn()`, etc., which attach a requirement check `.onResolve {}`. But indeed, because we cannot put inline functions with reified types in interfaces, we cannot catch and check the types on resolve...

Plus, doing it completely safely may involve modifying all implementation of `ColumnAccessor` and giving it an expected `KType` or something...

---
But okay, let's look at it from another angle. What if we don't check types at runtime for now, like we already do for `select {}`? 
`df.select { "a"<Int>() }.a` is possible to write already, even if `a` is `String`. It will just fail at the `.a` call instead of `.select {}`.

The String API is already unsafe, and all `require {}` wants to achieve is to make a little bridge from the String API to column accessors from the compiler plugin.

So why runtime checks would be very nice, both for `require {}` as for any other CS DSL function, I don't think it is worth limiting `require {}` just to one column. For now, that is, of course :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/DataFrameReceiver.kt
Jolanrensen: okay, then I'd write it something like:
"Why this function was written this way:  ...your explanantion"
or "Some background about this function: ..."

Just something to prepare the reader for what type of information they will find in this section, because it deviates a bit from what I'd expect in comments :)

(plus, introducing the block of text like that makes it easier to decide whether it's worth reading it at all at when skimming through the code :) )

docs/StardustDocs/topics/require.md
AndreiKingsley: Simply add a note
```
> Experimental; works in compiler plugin starting from Kotlin 2.4.0
> and displayed in IntelliJ IDEA 2026.2 or higher.
> {style="warning"}
```

docs/StardustDocs/topics/require.md
```diff
@@ -0,0 +1,19 @@
+[//]: # (title: require)
+<!---IMPORT org.jetbrains.kotlinx.dataframe.samples.api.Require-->
+
+Throws an exception if the specified column is missing or its type is not subtype of `C`.
+From the compiler plugin perspective, a new column will appear in the compile-time schema as a result of this operation.
+The aim here is to help incrementally migrate workflows to extension properties API.
```
AndreiKingsley: link to EP API topic please.

koperagen:
> df.select { "a"<Int>() }.a is possible to write already, even if a is String. It will just fail at the .a call instead of .select {}.

Exception is only triggered when value is pulled from the column: `df.select { "a"<Int>() }.a[0]`. `Int` in `DataColumn<Int>` itself is an unchecked cast. So error can travel quite a bit before triggering a ClassCast exception. This troubles me
TBF, i'm fine with both approaches. We can choose. My thoughts here:
1. Maybe we should be more strict because `require` is more explicitly about ensuring specific schema than `select`
2. Multiple require being a bit verbose might not be a bad thing, right?
3. Changing ColumnSelector to ColumnsSelector will be a source compatible change in the future 

> The String API is already unsafe, and all require {} wants to achieve is to make a little bridge from the String API to column accessors from the compiler plugin.

I agree

docs/StardustDocs/topics/require.md
```diff
@@ -0,0 +1,19 @@
...
+
+However, this operation is new and will work in compiler plugin starting from IntelliJ IDEA 2026.2 and Kotlin 2.4.0
+
+```text
+require { column }
+```
+
+**Related operations**: [](cast.md), [](convertTo)
+
+```kotlin
```
AndreiKingsley, line 17: 1) Please add a Korro sample in "samples" module
2)Might be a good idea to show this that `peopleDf` originally  doesn't have EPs:
```
// Won't compile
peopleDf.select { name.firstName }
// Declare column with a runtime check
val df = peopleDf.require { "name"["firstName"]<String>() }
// Use extension properties after `require`
df.select { name.firstName }
````

**Review by AndreiKingsley — CHANGES_REQUESTED** [review]

File: core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/DataFrameReceiver.kt (hunk 3 above)
Comment (koperagen) [reply core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/DataFrameReceiver.kt:63]: Ok, makes sense!

File: docs/StardustDocs/topics/require.md (hunk 7 above)
Comment (koperagen) [reply docs/StardustDocs/topics/require.md:17]: I cannot do it yet because require is not supported in compiler plugin :( But i'll do after we update to 2.4.0-RC or something

Comment (koperagen) [reply docs/StardustDocs/topics/require.md:17]: but i'll update the code snippet

**Comment by AndreiKingsley** [comment]
I think we should take the following approach to “require” — it's a great thing if you need to quickly create EPs for 1-3 columns. 
However, I would not recommend using it in production code, and would use the classic `cast` approach instead.
Do you agree? If so, I would put this in the documentation.

File: docs/StardustDocs/topics/require.md (hunk 7 above)
Comment (AndreiKingsley) [reply docs/StardustDocs/topics/require.md:17]: Yeah, I really didn't think about that, sorry 😄 ! 
But please create at least a commented function and korro marks so we don't forget and an issue.

**Comment by koperagen** [comment]
> I think we should take the following approach to “require” — it's a great thing if you need to quickly create EPs for 1-3 columns.
However, I would not recommend using it in production code, and would use the classic cast approach instead.
Do you agree? If so, I would put this in the documentation.

> it's a great thing if you need to quickly create EPs for 1-3 columns.
> would use the classic cast approach instead

I agree, yes

> I would not recommend using it in production code

Why not? :)

**Comment by Jolanrensen** [comment]
> So error can travel quite a bit before triggering a ClassCast exception. This troubles me

Me too, however, I would still put `require` and `select` (and any other selecting multiple columns operation for that matter) on the same level of "safeness".
@zaleslaw @AndreiKingsley what do you think?

> Multiple require being a bit verbose might not be a bad thing, right?

hmm, I mean it's one of the reasons we're looking at https://github.com/Kotlin/dataframe/issues/1168 and why we have this comprehensive `add {}` or even `select {}` DSL.
I fear repetition of the same statement over and over again hurts readability in the API.

> I think we should take the following approach to “require” — it's a great thing if you need to quickly create EPs for 1-3 columns.
However, I would not recommend using it in production code, and would use the classic cast approach instead.
Do you agree? If so, I would put this in the documentation.

On the other hand, if this is one of the goals, limiting `require {}` to one column does force people towards `cast`/`convert`... `cast(verify = true)` is a bit safer, as it fails earlier

**Comment by Jolanrensen** [comment]
As for using it in production. That may be nice :) It's like defining the schema functionally, which can be very versitile:

```kt
df.require {
    "user" {
        "name" {
            "firstName"<String>() and "lastName"<String>()
        } and "age"<Int>()
    } and
        "address" {
            "street"<String>() and "city"<String>()
        }
}
```
however, maybe a different DSL would make more sense as more than half of the CS DSL makes no sense for `require {}` (like `colsOf<T>()`, `except {}` etc. etc.) Plus, we don't care about the return value, unlike `.select {}`

Something more similar to `add {}`:
```kt
df.require {
    "user" {
        "name" {
            "firstName"<String>()
            "lastName"<String>()
        }
        "age"<Int>()
    }
    "address" {
        "street"<String>()
        "city"<String>()
    }
}
```
This would also allow us to build any checks we like into each column definition.

So... Maybe we could follow the `add` pattern; have an overload for a single column and one for multiple columns:

```kt

df
    .require(column<Int>("user")) // implemented now
    .require { // implemented later
        // new DSL for multiple columns, notation TBD
    }

```

AndreiKingsley:
So, why not add 2 functions?
For a single typed column
```
df.requireColumn { singleColumn }
```
For defining schema
```
df.defineColumns {
     "user" {
        "name" {
            "firstName"<String>()
            "lastName"<String>()
        }
        "age"<Int>()
    }
    "address" {
        "street"<String>()
        "city"<String>()
    }
}
```

Jolanrensen:
@AndreiKingsley Yes, I think two would work too, but it's hard for users to tell the difference between the two. Let's say they just want to `require` one column. In your case they could write:

```kt
df.requireColumn { "singleCol"<String>() }
df.defineColumns { "singleCol"<String>() }
```
and there's no way for them to tell the difference or see which benefits them more.

koperagen:
If we go with 2 functions route, one option is `require { singleColumn }` and `cast { DSL here }`

AndreiKingsley:
I'd name it like
`require` and `defineSchema`

koperagen:
Will it be ok to keep the scope of this issue to `require` with single column? Let's decide about name and DSL for second one separately?

Jolanrensen:
> I'd name it like `require` and `defineSchema`

"define" is a bit of an odd name, though. Because it's you, the user, doing the defining, not DataFrame.

We usually have names that are "commands" (imperative) to the library, like "select", "filter", "add", etc. or "end goals" (declarative), like "distinct()", "first()", and or most of the CS DSL: "colsOf", "all", "named"

> Will it be ok to keep the scope of this issue to require with single column? Let's decide about name and DSL for second one separately?

Yes, I think so :) but we should decide the name first, as there's no easy way to make `require {}` allow multiple columns later. So we'll need to decide between `require {}`, `requireCol {}`, `requireColumn {}`, etc.

koperagen:
@Jolanrensen 
`requireColumns`? 
@AndreiKingsley 
For `defineSchema` question is whether we want to discard previous type information or to keep it?

Maybe requireSchema?

Jolanrensen:
@koperagen But then I'd recommend making it `require { multiple columns }`, `requireColumn { column }`.
Because `select {  }`, `add {}`, `convert {}`, `update {}`, etc. also allow multiple columns, and we have `getColumn { singleColumn }` as only named operation for a single column. However, having "require" and another again hurts discoverability and 'not knowing which variant to choose' if we did have both... unless we have two entirely different names:

So:
- `requireColumn {}` for the singular one. It appends "-Column" to not conflict with the other multiple-column operations.
- `cast {}`, `requireColumns {}`, or `requireSchema {}` for the multiple columns one.
- Maybe in the future we could figure out a `require {}` which takes any number of columns and we could drop both variants altogether. However, I would not create `require {}` if it is limited to just one of the cases.

koperagen:
> requireColumn { column } 

Fine by me

> Because select {  }, add {}, convert {}, update {}, etc. also allow multiple columns, and we have getColumn { singleColumn } as only named operation for a single column.

Sorry, i don't agree with this logic. We also have get(ColumnSelector), and at the same time we have getColumns(ColumnsSelector) and operations like after, before, asGroupBy with ColumnSelector.  So type itself is not the only deciding factor

Jolanrensen:
> Sorry, i don't agree with this logic. We also have get(ColumnSelector), and at the same time we have getColumns(ColumnsSelector) and operations like after, before, asGroupBy with ColumnSelector. So type itself is not the only deciding factor

`get` allows both `ColumnSelector` as well as `ColumnsSelector`, (but that trick doesn't seem to work well with `require {}` if it's inlined, I tried :/), so that follows `select` and the other operations in that you can either supply one or multiple columns.

`after`, `before` etc. aren't operations. They are selectors inside the Columns Selection DSL.

That only leaves `asGroupBy {}` as top-level operation asking for a single column. However, I'd say this is more an exception than a rule. You can call `asGroupBy()` as well. It's just that `GroupBy` needs a single `FrameColumn` and if you have multiple, you need to specify which one you mean.

So I still think the logic holds up.

docs/StardustDocs/topics/require.md
koperagen: https://github.com/Kotlin/dataframe/issues/1718

koperagen:
Added a commit with require -> requireColumn rename

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/require.kt
```diff
@@ -0,0 +1,18 @@
...
+import org.jetbrains.kotlinx.dataframe.api.isSubtypeOf
+import org.jetbrains.kotlinx.dataframe.type
+import kotlin.reflect.KType
+
+@PublishedApi
+internal fun <T, C> DataFrame<T>.requireImpl(column: ColumnSelector<T, C>, type: KType): DataFrame<T> {
+    val resolvedColumn = getColumnWithPath(column)
+    val actualType = resolvedColumn.data.type
+    require(resolvedColumn.data.isSubtypeOf(type)) {
+        "Column '${resolvedColumn.path.joinToString()}' has type '$actualType', which is not subtype of required '$type' type."
```
Jolanrensen, line 15: *a subtype of the required '$type' type.

Jolanrensen (approved):
Thanks!

zaleslaw:
@koperagen please don't merge before my approval

Jolanrensen:
For future reference, I made a small proof-of-concept compiler plugin version of `requireColumn {}` that allows columns to be introduced and types to be changed:
https://code.jetbrains.team/p/kt/repositories/kotlin/revision/e8daa124d3081d5525afea69c403e55ac3cb7593

zaleslaw (changes requested):
I like the idea of gradual schema adoption. It gives users a softer entry point instead of pushing them directly into a full schema-based workflow.

At the same time, this feels bigger than a small helper API. It looks like a possible first step toward a real schema-definition DSL, so we should keep that direction in mind.

It is good that this will be documented. I think we should also decide whether we want an example for it. We can either add a small example or create an issue with a rough draft and a clear blocker note, for example the Kotlin version / release target, so we can judge whether it can go into Beta 6.

I also suggest improving the RequireColumn KDoc with examples for both one column and multiple columns. I would explicitly show a 1-column case and a 3-column case, so users do not get the impression that this API is only for a single column. We should also add a note in both the KDoc and the docs that this is not yet the recommended approach for dataframes with dozens or hundreds of columns.

We should also be honest that this adds one more way to work with schema, which means more complexity in the library. Most users will not learn all these paths deeply, so clear website docs and strong KDoc snippets matter a lot.

Action items:

- Expand RequireColumn KDoc with 1-column and 3-column examples
- Add a note in KDoc/docs: good for gradual adoption, not yet recommended for very wide dataframes
- Decide whether to add an example or create an issue for it
- If we create an issue, include a draft and the Kotlin/release blocker for Beta 6 planning
- Add issue for this operation and PR itself, with proper naming and description for Release Notes and other needs

Notify me, when all changes will be applied @koperagen

Jolanrensen:
@zaleslaw it currently only works with 1 column at a time...

koperagen:
@zaleslaw  @Jolanrensen i come up with a more descriptive example for this operation, check it out:
https://github.com/Kotlin/dataframe/pull/1715/commits/22f63bdd21e1b4ff95e3ab4ed8884411658a7f6c

> We should also be honest that this adds one more way to work with schema, which means more complexity in the library. Most users will not learn all these paths deeply, so clear website docs and strong KDoc snippets matter a lot.

Agree

zaleslaw (approved):

Issue #1644, opened 2025-12-17
DataFrame instance in notebooks cannot be cast(verify=true) to itself


koperagen:
This appears to be a regression, works fine in 1.0.0-Beta2 but fails in dev
```
%use dataframe
val test = DataFrame.readJson("test.json")
test
#%%
fun AnyFrame.benchmark() = castTo(test).select { benchmark and primaryMetric.score and primaryMetric.scoreConfidence  }

#%%
test.benchmark()

```
[test.json](https://github.com/user-attachments/files/24212144/test.json)

```
java.lang.IllegalArgumentException: Column has schema:
 primaryMetric:
	score: Float
	scoreError: Float
	scoreConfidence: List<Float>
	scorePercentiles:
		0.0: Float
		50.0: Float
		90.0: Float
		95.0: Float
		99.0: Float
		99.9: Float
		99.99: Float
		99.999: Float
		99.9999: Float
		100.0: Float
	scoreUnit: String
	rawData: List<List<Float>>

 that differs from target schema:
 primaryMetric:
	rawData: List<List<Float>>
	score: Float
	scoreConfidence: List<Float>
	scoreError: Float
	scorePercentiles:
		0.0: Float
		100.0: Float
		50.0: Float
		90.0: Float
		95.0: Float
		99.0: Float
		99.9: Float
		99.99: Float
		99.999: Float
		99.9999: Float
	scoreUnit: String
```

Jolanrensen:
Maybe I missed it when changing schema comparisons?

It should be able to cast if it `MATCHES`, however it might check `EQUALS` now.

PR #1716, opened 2026-02-27
Fix accidental and too strict order check from `cast`

koperagen:
With this PR we make sure df can be cast to a schema that was generated from it. Fixes https://github.com/Kotlin/dataframe/issues/1644

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/schema/DataFrameSchemaImpl.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/schema/DataFrameSchemaImpl.kt
@@ -126,8 +126,6 @@ internal fun Map<String, ColumnSchema>.render(indent: Int, sb: StringBuilder, in
                     sb.appendLine()
                 }
             }
-
-            else -> throw NotImplementedError(columnSchema::class.toString())
         }
     }
     return sb.toString()


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/schema/ColumnSchema.kt
```diff
@@ -80,15 +80,14 @@ public sealed class ColumnSchema {
...
-    /** Checks equality by kind, type, or schema. TODO was matching, check if == works. */
     override fun equals(other: Any?): Boolean {
         val otherType = other as? ColumnSchema ?: return false
         if (otherType.kind != kind) return false
         if (otherType.nullable != nullable) return false
         return when (this) {
             is Value -> type == (otherType as Value).type
-            is Group -> schema == (otherType as Group).schema
-            is Frame -> schema == (otherType as Frame).schema
+            is Group -> schema.compare((otherType as Group).schema).matches()
```
Jolanrensen: This does not follow the logic of `DataFrameSchemaImpl.equals` where `==` means an exact match in the same order. `groupSchema == otherGroupSchema` should thus be `false` too if their order mismatches. (We should probably add that as a note here too)

If `cast<T>(verify = true)` is too strict, it's likely because `convertToImpl()` uses `==` where it's supposed to use `.matches()`

koperagen: Right, fixed it

koperagen:
Updated the fix there, and found another place where i believe matches was intended

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/cast.kt
```diff
@@ -23,4 +23,22 @@ class CastTests {
...
+
+    @DataSchema
+    data class Container(val test: Schema)
+
+    @DataSchema
+    data class Schema(val score: Float, val rawData: String)
+
+    @Test
+    fun `cast should disregard column order`() {
+        // https://github.com/Kotlin/dataframe/issues/1644
```
Jolanrensen, line 35: you can actually link to issues by just writing "Issue #1644" :) No full link required

Jolanrensen:
Thanks!

Issue #679, opened 2024-04-25
`columnOf(arrayOf(1))` inferred as `Array<*>`


Jolanrensen:
Probably related to https://github.com/Kotlin/dataframe/issues/678, 

`columnOf(arrayOf(1)).type shouldBe typeOf<Array<Int>>()` but is `Array<*>` instead.

Other cases, like `columnOf(intArrayOf(1))` or `columnOf(arrayOf(1, null))` work fine.

AndreiKingsley:
Fixed in the latest Kotlin version.

Added a small tests for checking it: #1729.

Issue #678, opened 2024-04-25
Type inference of `Array<primitive>` may be unpredictable


Jolanrensen:
From:
https://youtrack.jetbrains.com/issue/KT-52170/Reflection-typeOfArrayLong-gives-classifier-LongArray

In DataFrame we rely a lot on reflection and `typeOf<>()`, however, a thing to note is that `typeOf<Array<Int>>().classifier == IntArray::class` instead of `Array::class` as might be expected.

This might cause unexpected bugs, so refer to this issue when we find one.

AndreiKingsley:
Fixed in the latest Kotlin version.

Added a small tests for checking it: #1729.

PR #1729, opened 2026-03-09
Added a `columnOf` types test

AndreiKingsley:
Added a small test that checks #678 and #679


AndreiKingsley:
@Jolanrensen I will fix this in this PR then!

Could you please tell me what is supposed to be fixed here?
https://github.com/Kotlin/dataframe/blob/e0f966c0bb6edbb8b0b7d4fc28a16656ec06a62c/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/Rendering.kt#L69

upd: never mind, understand that

AndreiKingsley:
Removed all workarounds related to #678 and #679 and cover those methods with tests

PR #1731, opened 2026-03-09
Prevented escaping of SoftKeywords in code generator 

### LINKED ISSUE #1727: Redundant escaping of SoftKeywords in generated code
`val value = 123` - valid code, no need to escape. 

`dataFrameOf("value" to columnOf(123)).generateDataClasses()` =>

```
@DataSchema
data class DataEntry(
    @ColumnName("value")
    val `value`: Int
)
```

How to fix:
String.needsQuoting checks ModifierKeywords.VALUES.contains(this). We need to substract SoftKeywords from ModifierKeywords, then we'll have a valid set of prohibited identifiers. Bonus points for changing List.contains to Set.contains

Allex-Nik:
Fixes #1727

- Excluded `SoftKeywords` from `ModifierKeywords` in checking if quoting is needed in code generator.
- Changed `List.contains` to `Set.contains` for `HardKeywords` and prohibited `ModifierKeywords` in the `needsQuoting` function.
- Adjusted `Other advanced test` in `OpenApiTests` that expected backticks around soft keywords.
- Added a test for the `generateDataClasses` function used with a soft keyword.


Allex-Nik:
> Please add a test for generateDataClasses function with example provided in the issue

Done :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/generateCode.kt
```diff
@@ -0,0 +1,21 @@
+package org.jetbrains.kotlinx.dataframe.api
+
+import io.kotest.matchers.shouldBe
+import org.junit.Test
+
+class GenerateCodeTests {
+    @Test
+    fun `generateDataClasses with soft keyword`() {
```
koperagen: For historical reasons these tests are here:
core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/CodeGenerationTests.kt

In dataframe it's not always predictable, but in general find usages will likely point you to existing test class for most APIs

Allex-Nik: I see, thank you!

Jolanrensen:
@Allex-Nik I see the maven example build fails... the same happens for Andrei in his PR. I tried to patch it by limiting the logs. Try rebasing on master to see if it helps :)

Allex-Nik:
> @Allex-Nik I see the maven example build fails... the same happens for Andrei in his PR. I tried to patch it by limiting the logs. Try rebasing on master to see if it helps :)

Now it says that Gradle build daemon disappeared unexpectedly, I will try again :)

PR #1739, opened 2026-03-10
Deprecated `insert.under(path)` and fixed `insert.under { path }`

### LINKED ISSUE #1411: Deprecate `insert.under(path)`and fix `insert.under { path }` behavoir
For now, `insert.under { path }` doesn't work as `insert.under(path)` - if path contains non-existing col groups, it tries to retrieve them and throw an error, but it should create them instead (it's a default behavior for operations in KDF with column path specification).

Jolanrensen:
Fixes #1411

exactly like the issue description :)

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/insert.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/insert.kt
@@ -79,9 +79,6 @@ internal interface InsertDocs {
      * __`.`__[**`under`**][InsertClause.under]**`  {  `**`column: `[`ColumnSelector`][ColumnSelector]**` }`**
      *
      * {@include [Indent]}
-     * `| `__`.`__[**`under`**][InsertClause.under]**`(`**` columnPath: `[`ColumnPath`][ColumnPath]**`)`**
-     *
-     * {@include [Indent]}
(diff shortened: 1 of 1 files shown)


AndreiKingsley:
@Jolanrensen Is behavior the same in the Compiler Plugin?

Jolanrensen:
> @Jolanrensen Is behavior the same in the Compiler Plugin?

actually, no!

`insert {}.under { new path }` does not work yet.

Interesting enough, `move {}.under { new path }` does work, so I'll see if I can replicate its logic somewhat.
Also, `move {}.into { new path }` does not work either, so while I'm at it...

koperagen:
> insert {}.under { new path } does not work yet.

it's actually deliberately made to not work (= be identical to runtime)
It'll start working once we update DF dependency there to include this PR
```
receiver.df.asDataFrame()
            .insert(receiver.column.asDataColumn()).under { column.path }
            .toPluginDataFrameSchema()
```

Jolanrensen:
@koperagen actually, just updating the version doesn't seem to work.

This line checks whether the column can actually be resolved:
image
but there's no such check in `move {}.under {}`, so I'll remove it here as well :)

Jolanrensen:
Alright, I'll merge it to master and from the new dev build update the compiler plugin :)

PR #1740, opened 2026-03-11
KoDEx + KtLint fix

Jolanrensen:
GH actions seems to skip the linter after running kodex. This is an attempt to fix that.

It also fixes some linter error that were introduced while the check wasn't working


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/ExpressionsGivenRow.kt
```diff
@@ -34,9 +34,11 @@ import org.jetbrains.kotlinx.dataframe.RowValueExpression as DfRowValueExpressio
...
-    /*
-     * The key for a @set that will define the operation name for the examples below.
-     * Make sure to [alias][your examples].
+    /**
+     * {@comment
+     *    The key for a @set that will define the operation name for the examples below.
+     *    Make sure to [alias][your examples].
+     * }
      */
     typealias OPERATION = Nothing
```
Jolanrensen, line 42: @AndreiKingsley I believe this can remain a normal `/**/` comment but excluded from sources

Jolanrensen: oh, this file is actually already excluded from sources

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/DocumentationUrls.kt
```diff
@@ -2,13 +2,15 @@
 
 package org.jetbrains.kotlinx.dataframe.documentation
 
-/*
- * Documentation URLs KDoc-snippets.
+/**
```
Jolanrensen: @AndreiKingsley this file is already excluded from sources ;P

PR #1735, opened 2026-03-10
Added support for `BIGINT UNSIGNED` type in MySQL and MariaDB DBTypes

AndreiKingsley:
Fixes #762.

PR #1741, opened 2026-03-12
Added custom sqlite types support

### LINKED ISSUE #964: ClassCast exception when reading from specific sqlite db
Firefox places db uses some weird column type that are not standard and not supported by xerial sqlite driver: LONGVARCHAR (https://sqlalchemy.narkive.com/DEL4O0EK/longvarchar). There's also a problem with driver being unable to tell if INTEGER column actually stores Long values
As a result, we have 
1. DataColumn with type = BigDecimal?, and values are all String? values
2. DataColumn with type = Int?, and values are all Long? values
Codegen in notebooks uses type to generate extension properties, and when you use them - you get ClassCast exception

```
class java.lang.String cannot be cast to class java.math.BigDecimal (java.lang.String and java.math.BigDecimal are in module java.base of loader 'bootstrap')
```

AndreiKingsley:
Closes #964


dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -5,21 +5,32 @@ import org.jetbrains.kotlinx.dataframe.io.db.TableColumnMetadata
...
+ * Even for default Sqlite types, you can override the actual [KType] after reading the column
+ * by providing a custom type in [customTypesMap].
  */
-public object Sqlite : DbType("sqlite") {
+public class Sqlite(public val customTypesMap: Map<String, KType> = mapOf()) : DbType("sqlite") {
     override val driverClassName: String
         get() = "org.sqlite.JDBC"
 
+    override fun getExpectedJdbcType(tableColumnMetadata: TableColumnMetadata): KType =
+        customTypesMap[tableColumnMetadata.sqlTypeName] ?: super.getExpectedJdbcType(tableColumnMetadata)
```
Jolanrensen: I'm not sure if `sqlTypeName` is enough to differentiate the right type.

Also, take nullability into account.

AndreiKingsley: For now, I believe `sqlTypeName ` is OK. 
However, we should improve this mechanism in future. This will require us to research use cases in depth.

AndreiKingsley:
Requires #1735

zaleslaw:
The problem is clear, your proposal is also promisable.
I have a couple of question:
- is this behaviour typical only for SQLite database or we could benefit with other SQL databases also
- why it could not be implemented through getExpectedJdbcType, where is a bottleneck?

AndreiKingsley:
@zaleslaw 
1) For now, this issue is most relevant to SQLite. The problem with SQLite is that the type defined by the user and even built-in types may not match the actual values from JDBC.

   In theory, PostgreSQL might have a similar mechanism, since it also allows for the creation of custom types. However,     the rules there are slightly stricter, and our built-in `getExpectedJdbcType` mechanism might be sufficient. With SQLite, on the other hand, there are definitely cases where it won’t be possible to determine the type using JDBC, as described in the issue. 
   
   In fact it needs research, for sure!
   
2) `getExpectedJdbcType` compute type only from column metadata — as and shown in the issue, it can be exactly the same for columns with the different types.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -37,4 +44,10 @@ public object Sqlite : DbType("sqlite") {
         } else {
             DriverManager.getConnection(dbConfig.url, dbConfig.user, dbConfig.password)
         }
+
+    public companion object {
```
Jolanrensen, line 48: A fun trick I just learned. If you make `Sqlite` open, you can make the companion object be the default itself, like `companion object : Sqlite() {}`
This means users can specify both `Sqlite` and `Sqlite.withCustomTypes(customTypes)`. Not sure if you like that :)

Jolanrensen (approved):

zaleslaw (approved):

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
AndreiKingsley: Oh, seems great!

AndreiKingsley: But I am not sure if this is actually good here, need a research of use-cases.

#1797

PR #1750, opened 2026-03-16
Moved `changeType` to `ValueColumn` and adjusted related implementations

### LINKED ISSUE #1675: Remove DataColumn.changeType function
Looks like the only actual usage is very specific detail of convert { }.to < >() implementation: https://github.com/Kotlin/dataframe/blob/a29f5c6f87d7ecdacfa7deb64c3b46568ab5ec38/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convert.kt#L212
Logically, i'd say it shouldn't be a part of DataColumn interface. Maybe move it to ValueColumn

AndreiKingsley:
Closes #1675.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnWithParent.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnWithParent.kt
@@ -25,8 +25,7 @@ internal class ValueColumnWithParent<T>(override val parent: ColumnGroup<*>, ove
 
     override fun forceResolve() = ResolvingValueColumn(this)
 
-    override fun changeType(type: KType) =
-        ValueColumnWithParent(parent, source.changeType(type))
+    override fun changeType(type: KType) = ValueColumnWithParent(parent, source.changeType(type))
 
     override fun addParent(parent: ColumnGroup<*>) = source.addParent(parent)
 }


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnWithParent.kt
```diff
@@ -26,7 +26,7 @@ internal class ValueColumnWithParent<T>(override val parent: ColumnGroup<*>, ove
     override fun forceResolve() = ResolvingValueColumn(this)
 
     override fun changeType(type: KType) =
-        ValueColumnWithParent(parent, source.internal().changeType(type).asValueColumn())
+        ValueColumnWithParent(parent, source.internalValueColumn().changeType(type).asValueColumn())
```
koperagen: is `asValueColumn` still needed?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convert.kt
```diff
@@ -209,12 +210,8 @@ internal fun AnyCol.convertToTypeImpl(to: KType, parserOptions: ParserOptions?):
 
     // catch for ColumnGroup and FrameColumn since they don't have changeType,
     // but user converters can still exist
-    if (from.isSubtypeOf(to)) {
-        try {
-            return (this as DataColumnInternal<*>).changeType(to.withNullability(hasNulls()))
-        } catch (_: UnsupportedOperationException) {
-            //
-        }
+    if (from.isSubtypeOf(to) && this is ValueColumnImpl) {
```
koperagen: why not is ValueColumn?

AndreiKingsley: Mistake

Jolanrensen: you can use `.isValueColumn()`, it has smart cast :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/FrameColumnWithParent.kt
```diff
@@ -24,9 +24,6 @@ internal class FrameColumnWithParent<T>(override val parent: ColumnGroup<*>, ove
 
     override fun forceResolve() = ResolvingFrameColumn(this)
 
-    override fun changeType(type: KType) =
```
Jolanrensen, line 27: LOL this function did have an implementation, but would always call a function with an `UnsupportedOperationException`... good riddance ;P

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnWithParent.kt
Jolanrensen: looks like `internalValueColumn()` is also not needed anymore

koperagen (approved):

Issue #1755, opened 2026-03-17
`Map<String, Any?>` to `DataRow` conversion with nested maps


Jolanrensen:
While `map.toDataRow()` is very helpful, there's no easy way to fully unfold nested maps to to datarows, like

```kt
val map = mapOf(
    "name" to "a",
    "metadata" to
        mapOf(
            "country" to "Philippines",
            "region" to mapOf("name" to "Caraga", "code" to "XIII"),
            "population" to mapOf("value" to "12345", "year" to 2020),
            "wrongMap" to mapOf(1 to 2, 4 to 4),
        ),
)
```
Similarly to `unfold` and `toDataFrame()`, we could add a `maxDepth` parameter that would allow us to specify the maximum depth of nesting we scan for and convert.
`convertKeysToString: Boolean` might be another nice addition because there's no way to tell if keys are strings or not until we encounter a `ClassCastException`.

`Iterable<Map<String, Any?>>.toDataFrame()` should probably gain the same treatment.

And if we're converting keys to strings anyways, let's make it work for `Map<Any, Any?>`, why not

Jolanrensen:
> One important thing: what to do about lists/iterables inside the map? Should we convert those too? A Iterable<Map<String, *>> can be seen and converted to a DataFrame after all, making Iterable<String, List<Map<String, *>>> a DataFrame with FrameColumns.
> 
> Plus, Map<String, Iterable<*>>'s can also be converted to a DataFrame. Should that be done?
> 
> (Also where have I seen this before... oh right JSON objects/arrays -> DataFrame XD)

PR #1756, opened 2026-03-17
Nested Map -> DataRow conversion

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1755

created new overloads for `Map<*, *>.toDataRow()` expanded with `maxDepth` and `convertKeysToString` to allow recursive map -> datarow conversion with key -> string handling

`@JvmOverloads` is added to avoid bytecode incompatibility between versions.

TODO: `Iterable<Map<*, *>>.toDataFrame()`


Jolanrensen:
@AndreiKingsley @koperagen 
One important thing: what to do about lists/iterables inside the map? Should we convert those too? A `Iterable<Map<String, *>>` can be seen and converted to a `DataFrame` after all, making `Iterable<String, List<Map<String, *>>>` a DataFrame with `FrameColumn`s.

Plus, `Map<String, Iterable<*>>`'s can _also_ be converted to a DataFrame. Should that be done?

(Also where have I seen this before... oh right JSON objects/arrays -> DataFrame XD)

AndreiKingsley:
Let's add arguments like `unfoldInnerIterables = false`?

PR #1760, opened 2026-03-18
Added DataFrame.renderToMarkdown(): String function; Improved type rendering in DataFrame.print() and enabled it by default

### LINKED ISSUE #525: Add feature to generate Markdown table
I'd like to have `writeMarkdown()` method that would build [markdown table](https://www.codecademy.com/resources/docs/markdown/tables) from the simple table. BTW, it could even support some cell formatting

koperagen:
fixes #525 
First, i added more tests on original renderToString. Then a little refactoring to extract common logic. +new function

Original rendering has some quirks that can be seen in test asserts. For now i just preserved it as is, but probably let's improve it as well - it will be more obvious with new tests

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/string.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/string.kt
@@ -29,6 +29,10 @@ public fun AnyFrame.renderToString(
     rowIndex: Boolean = true,
 ): String {
     val sb = StringBuilder()
+    val table = prepareTable(rowsLimit, valueLimit, columnTypes, rowIndex)
+    val columnLengths = table.values.mapIndexed { col, vals ->
+        (vals + table.header[col]).maxOf { it.length } + 1
+    }
 
     // title
     if (title) {
@@ -36,40 +40,19 @@ public fun AnyFrame.renderToString(
         sb.appendLine()
     }
 
-    // data
-    val rowsCount = rowsLimit.coerceAtMost(nrow)
-    val cols = if (rowIndex) listOf((0 until rowsCount).toColumn()) + columns() else columns()
-    val header = cols.mapIndexed { colIndex, col ->
-        if (columnTypes && (!rowIndex || colIndex > 0)) {
-            "${col.name()}:${renderType(col)}"
-        } else {
-            col.name()
-        }
-    }
-    val values = cols.map {
-        val top = it.take(rowsLimit)
-        val precision = if (top.isNumber()) top.asNumbers().scale() else 0
-        val decimalFormat =
-            if (precision >= 0) RendererDecimalFormat.fromPrecision(precision) else RendererDecimalFormat.of("%e")
-        top.values().map {
-            renderValueForStdout(it, valueLimit, decimalFormat = decimalFormat).truncatedContent
-        }
-    }
-    val columnLengths = values.mapIndexed { col, vals -> (vals + header[col]).map { it.length }.maxOrNull()!! + 1 }
-
     // top border
     if (borders) {
         sb.append("\u230C")
-        for (i in 1 until columnLengths.sum() + columnLengths.size) sb.append('-')
+        repeat(columnLengths.sum() + columnLengths.size - 1) { sb.append('-') }
         sb.append("\u230D")
         sb.appendLine()
         sb.append("|")
     }
 
     // header
-    for (col in header.indices) {
+    for (col in table.header.indices) {
         val len = columnLengths[col]
-        val str = header[col]
+        val str = table.header[col]
         val padded = if (alignLeft) str.padEnd(len) else str.padStart(len)
         sb.append(padded)
         if (borders) sb.append("|")
@@ -80,18 +63,18 @@ public fun AnyFrame.renderToString(
     if (borders) {
         sb.append("|")
         for (colLength in columnLengths) {
-            for (i in 1..colLength) sb.append('-')
+            repeat(colLength) { sb.append('-') }
             sb.append("|")
         }
         sb.appendLine()
     }
 
     // data
-    for (row in 0 until rowsCount) {
+    for (row in 0 until table.rowsCount) {
         if (borders) sb.append("|")
-        for (col in values.indices) {
+        for (col in table.values.indices) {
             val len = columnLengths[col]
-            val str = values[col][row]
+            val str = table.values[col][row]
             val padded = if (alignLeft) str.padEnd(len) else str.padStart(len)
             sb.append(padded)
             if (borders) sb.append("|")
@@ -100,15 +83,50 @@ public fun AnyFrame.renderToString(
     }
 
     // footer
-    if (nrow > rowsLimit) {
+    if (table.totalRows > rowsLimit) {
         sb.appendLine("...")
     } else if (borders) {
         sb.append("\u230E")
-        for (i in 1 until columnLengths.sum() + columnLengths.size) sb.append('-')
+        repeat(columnLengths.sum() + columnLengths.size - 1) { sb.append('-') }
         sb.append("\u230F")
         sb.appendLine()
     }
     return sb.toString()
+}
+
+private class PreparedTable(
+    val header: List<String>,
+    val values: List<List<String>>,
+    val rowsCount: Int,
+    val totalRows: Int,
+)
+
+private fun AnyFrame.prepareTable(
+    rowsLimit: Int,
+    valueLimit: Int,
+    columnTypes: Boolean,
+    rowIndex: Boolean,
+    escapeValue: (String) -> String = { it },
+): PreparedTable {
+    val rowsCount = rowsLimit.coerceAtMost(nrow)
+    val cols = if (rowIndex) listOf((0 until rowsCount).toColumn()) + columns() else columns()
+    val header = cols.mapIndexed { colIndex, col ->
+        if (columnTypes && (!rowIndex || colIndex > 0)) {
+            "${col.name()}:${renderType(col)}"
+        } else {
+            col.name()
+        }
+    }
+    val values = cols.map { col ->
+        val top = col.take(rowsLimit)
+        val precision = if (top.isNumber()) top.asNumbers().scale() else 0
+        val decimalFormat =
+            if (precision >= 0) RendererDecimalFormat.fromPrecision(precision) else RendererDecimalFormat.of("%e")
+        top.values().map {
+            escapeValue(renderValueForStdout(it, valueLimit, decimalFormat = decimalFormat).truncatedContent)
+        }
+    }
+    return PreparedTable(header, values, rowsCount, nrow)
 }
 
 internal val valueToStringLimitDefault = 1000
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/string.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/string.kt
@@ -100,6 +100,7 @@ public fun AnyFrame.renderToString(
     // footer
     if (table.totalRows > table.rowsCount) {
         sb.appendLine("...")
+        sb.appendLine("${size.ncol} columns x ${size.nrow} rows")
     } else if (borders) {
         sb.append(Borders.BOTTOM_LEFT)
         repeat(columnLengths.sum() + columnLengths.size - 1) { sb.append(Borders.HORIZONTAL) }
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/string.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/string.kt
@@ -26,7 +26,7 @@ public fun AnyFrame.renderToString(
     valueLimit: Int = 40,
     borders: Boolean = false,
     alignLeft: Boolean = false,
-    columnTypes: Boolean = false,
+    columnTypes: Boolean = true,
     title: Boolean = false,
     rowIndex: Boolean = true,
 ): String {
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/string.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/string.kt
@@ -4,6 +4,8 @@ import org.jetbrains.kotlinx.dataframe.AnyFrame
 import org.jetbrains.kotlinx.dataframe.AnyRow
 import org.jetbrains.kotlinx.dataframe.api.asNumbers
 import org.jetbrains.kotlinx.dataframe.api.columnsCount
+import org.jetbrains.kotlinx.dataframe.api.getColumns
+import org.jetbrains.kotlinx.dataframe.api.isColumnGroup
 import org.jetbrains.kotlinx.dataframe.api.isNumber
 import org.jetbrains.kotlinx.dataframe.api.take
 import org.jetbrains.kotlinx.dataframe.api.toColumn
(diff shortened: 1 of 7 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/string.kt
```diff
@@ -100,17 +83,99 @@ public fun AnyFrame.renderToString(
...
+        val decimalFormat =
+            if (precision >= 0) RendererDecimalFormat.fromPrecision(precision) else RendererDecimalFormat.of("%e")
+        top.values().map {
+            escapeValue(renderValueForStdout(it, valueLimit, decimalFormat = decimalFormat).truncatedContent)
+        }
+    }
+    return PreparedTable(header, values, rowsCount, nrow)
+}
+
+public fun AnyFrame.renderToMarkdown(
```
Jolanrensen: I believe this deserves its own file, no? :)

Jolanrensen:
> Original rendering has some quirks that can be seen in test asserts
 
What kind of quirks do you mean?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/rendering/RenderToMarkdownTests.kt
```diff
@@ -0,0 +1,136 @@
...
+        "Charlie", 35, "London",
+    )
+
+    @Test
+    fun `markdown basic structure`() {
+        val result = df.renderToMarkdown()
+        val expected =
+            """
+            |  | name | age | city |
+            |---:|---:|---:|---:|
```
Jolanrensen: what do these colons do? I don't think I've seen them before in MD tables

Jolanrensen: ah! it's alignment, right? Uhm, maybe we should have "no alignment" as an option, as well as right/left alignment. There are some right-to-left languages that could cause issues with a default alignment. (Actually, they probably already break tables like these, they broke our toString() very much too)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/string.kt
```diff
@@ -100,17 +83,99 @@ public fun AnyFrame.renderToString(
...
+    val sb = StringBuilder()
+    if (title) {
+        sb.appendLine("**DataFrame [${size()}]**")
+        sb.appendLine()
+    }
+
+    // header
+    sb.append("|")
+    for (col in table.header) {
+        sb.append(" ${col.replace("|", "\\|")} |")
```
Jolanrensen: isn't this logic already applied in prepareTable?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/string.kt
```diff
@@ -100,17 +83,99 @@ public fun AnyFrame.renderToString(
...
+
+private fun AnyFrame.prepareTable(
+    rowsLimit: Int,
+    valueLimit: Int,
+    columnTypes: Boolean,
+    rowIndex: Boolean,
+    escapeValue: (String) -> String = { it },
+): PreparedTable {
+    val rowsCount = rowsLimit.coerceAtMost(nrow)
+    val cols = if (rowIndex) listOf((0 until rowsCount).toColumn()) + columns() else columns()
```
Jolanrensen: do you know `..<` for `until`? :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/string.kt
Jolanrensen: ah, no it isn't, maybe it should :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/string.kt
```diff
@@ -100,17 +83,99 @@ public fun AnyFrame.renderToString(
...
+            col.name()
+        }
+    }
+    val values = cols.map { col ->
+        val top = col.take(rowsLimit)
+        val precision = if (top.isNumber()) top.asNumbers().scale() else 0
+        val decimalFormat =
+            if (precision >= 0) RendererDecimalFormat.fromPrecision(precision) else RendererDecimalFormat.of("%e")
+        top.values().map {
+            escapeValue(renderValueForStdout(it, valueLimit, decimalFormat = decimalFormat).truncatedContent)
```
Jolanrensen: maybe create a new issue for exploring rendering nested frames as `<details>` like `toStaticHtml`? or at least mention this option in some kdoc :) Speaking of... small kdoc please? :))

Jolanrensen (changes requested):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/string.kt
koperagen: done!

koperagen:
@Jolanrensen please have a look when you have time. I checked default print output on a 7500 rows x 77 columns "real world data" output and from there made a bunch of  changes. One important is always printing column types and on a separate line, which i think makes a lot of sense for String API ease of use / dataschema declarations. 
Went to compare with pandas outputs and stumbled upon this https://pypi.org/project/tabulate/. Thought we could have a few styles as well

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/rendering/RenderToMarkdownTests.kt
Jolanrensen: could we still have a "no-alignment" option?

koperagen: sure! it's still in progress, i'll update once more after review

Jolanrensen:
Wow awesome!

>  One important is always printing column types and on a separate line, which i think makes a lot of sense for String API ease of use / dataschema declarations.

I agree, especially with large columnGroup/frameColumn types!

Small concern about this approach though with `columnTypes = true, borders = false, rowIndex = false` as it becomes quite hard to differentiate between the name/type/values.
```
     name  age    city
   String  Int  String
    Alice   30  Berlin
      Bob   25   Paris
  Charlie   35  London
```

I wonder how other libraries solve this. Maybe we could mark our types somehow too? Something like
```
     name   age     city
  `String` `Int` `String`
    Alice    30   Berlin
      Bob    25    Paris
  Charlie    35   London
```
or

```
     name   age     city
  String<  Int<  String<
    Alice    30   Berlin
      Bob    25    Paris
  Charlie    35   London
```
Okay, my examples are quite ugly, but you get the idea

koperagen:
> Wow awesome!
> 
> > One important is always printing column types and on a separate line, which i think makes a lot of sense for String API ease of use / dataschema declarations.
> 
> I agree, especially with large columnGroup/frameColumn types!
> 
> Small concern about this approach though with `columnTypes = true, borders = false, rowIndex = false` as it becomes quite hard to differentiate between the name/type/values.
> 
> ```
>      name  age    city
>    String  Int  String
>     Alice   30  Berlin
>       Bob   25   Paris
>   Charlie   35  London
> ```
> 
> I wonder how other libraries solve this. Maybe we could mark our types somehow too? Something like
> 
> ```
>      name   age     city
>   `String` `Int` `String`
>     Alice    30   Berlin
>       Bob    25    Paris
>   Charlie    35   London
> ```
> 
> or
> 
> ```
>      name   age     city
>   String<  Int<  String<
>     Alice    30   Berlin
>       Bob    25    Paris
>   Charlie    35   London
> ```
> 
> Okay, my examples are quite ugly, but you get the idea

Without index = true it's indeed a bit confusing
I was looking at R https://dplyr.tidyverse.org/. 

image
pandas simply doesn't render types in print, looks like they instead provide "info" akin to our "describe". 
Maybe backticks are good option for us. Should we wrap always or only when index = false?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/string.kt
```diff
@@ -17,102 +19,202 @@ import org.jetbrains.kotlinx.dataframe.jupyter.RenderedContent
...
     alignLeft: Boolean = false,
-    columnTypes: Boolean = false,
+    columnTypes: Boolean = true,
     title: Boolean = false,
     rowIndex: Boolean = true,
+    borderStyle: StringBorderRenderingStyle = StringBorderRenderingStyle.DottedLineWithRounderCorners,
 ): String {
     val sb = StringBuilder()
+    val table = prepareTable(rowsLimit, valueLimit, columnTypes, rowIndex)
+    val columnLengths = table.values.mapIndexed { col, vals ->
```
Jolanrensen: wait... so this is the... width... of each column in number of characters?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/string.kt
```diff
@@ -17,102 +19,202 @@ import org.jetbrains.kotlinx.dataframe.jupyter.RenderedContent
...
+    rowsLimit: Int = 20,
+    valueLimit: Int = 40,
+    borders: Boolean = false,
+    alignLeft: Boolean = false,
+    columnTypes: Boolean = true,
+    title: Boolean = false,
+    rowIndex: Boolean = true,
+): String = renderToString(rowsLimit, valueLimit, borders, alignLeft, columnTypes, title, rowIndex)
+
+public interface StringBorderRenderingStyle {
```
Jolanrensen: I think you could drop the "Rendering" part. "StringBorderStyle" is already clear enough (probably "BorderStyle" too if we think it doesn't clutter the rest of the library too much)

Jolanrensen:
```
     name    age      city
 <String>  <Int>  <String>
    Alice     30    Berlin
      Bob     25     Paris
  Charlie     35    London
```

Oh the R brackets may also work :) I think I like them better (probably because I'm used to seeing types inside `<>` in Kotlin too :D)

compared to:

```
     name    age      city
 `String`  `Int`  `String`
    Alice     30    Berlin
      Bob     25     Paris
  Charlie     35    London
```

As for when we should wrap them, I think we should just always do it. We had `: ` before to signify we were displaying a type instead of a column name. So I think it would help to have it always:
```
???????????????????????????????????
?   ?      name?    age?      city?
?   ?  <String>?  <Int>?  <String>?
???????????????????????????????????
?  0?     Alice?     30?    Berlin?
?  1?       Bob?     25?     Paris?
?  2?   Charlie?     35?    London?
???????????????????????????????????
```

```
         name    age      city
     <String>  <Int>  <String>
  0     Alice     30    Berlin
  1       Bob     25     Paris
  2   Charlie     35    London
```

May be a bit overboard for frameColumns, but it is still quite clear we're talking about types:

```
            a                                     group
     <Double>  <[a:Double, group:{l:Double, r:Double}]>
  0       0.0  [1 x 2] { a:0.000000, group:{ l:0.000...
```

Jolanrensen:
btw, what about markdown? can we / should we also display the type on a new line? Does md even support mutliple-lines in headers?

Edit:

 |  | name<br>\<String\> | age<br>\<Int\> | city<br>\<String\> |
 |---:|---:|---:|---:|
 | 0 | Alice | 30 | Berlin |
 | 1 | Bob | 25 | Paris |
 | 2 | Charlie | 35 | London |

```
 |  | name<br>\<String\> | age<br>\<Int\> | city<br>\<String\> |
 |---:|---:|---:|---:|
 | 0 | Alice | 30 | Berlin |
 | 1 | Bob | 25 | Paris |
 | 2 | Charlie | 35 | London |
```

koperagen:
I'm not so in favor of mandatory brackets :( 

As for newlines in markdown, i didn't think about `<br>`, interesting it works! Tbh this all is confusing me. When i started i didn't know markdown is a superset of HTML. If this is the case, then why we even need this table renderer when we already have toStaticHtml? :)) But at the same time, this table format is not even "standard" markdown
Maybe it's worth to have it be plain ASCII for max compatibility and look again at `toStaticHtml` for advanced features? It's already more versatile

Jolanrensen:
> I'm not so in favor of mandatory brackets :(

It does provide a clearer picture, no? :) otherwise, we can always add a setting which turns it off. I liked them in the R table :)

As for Markdown... hmm, I think we could still offer it. It's like a hybrid between ascii and html, like, it's readable, but also renderable. I think there's some value in that. But then we should probably keep it as flat and simple as possible, because, as you say, for more advanced cases, we can just turn to HTML.
Maybe we could align it a bit with markdown output from interactive tables?

koperagen:
Another option to always add a small separator between header and data 
```
        name  age    city
      String  Int  String
        ---- ----    ----
  0    Alice   30  Berlin
  1      Bob   25   Paris
  2  Charlie   35  London
```

```
        name  age    city
      String  Int  String
  0    Alice   30  Berlin
  1      Bob   25   Paris
  2  Charlie   35  London
```

```
         name    age      city
     <String>  <Int>  <String>
  0     Alice     30    Berlin
  1       Bob     25     Paris
  2   Charlie     35    London

```

More complex data:
<> rendering
https://gist.github.com/koperagen/bc097d8de3d2b239e82c4f082ac18071

---- rendering
https://gist.github.com/koperagen/ff1c489d9520c2fffe30fee636a22422

just type
https://gist.github.com/koperagen/4ce0d2192f3bb8c5c1d4c8f2fba443a0

Jolanrensen:
@koperagen hmm, that separator could maybe also be nice, but how will it look for column groups and frame columns?

koperagen:
> @koperagen hmm, that separator could maybe also be nice, but how will it look for column groups and frame columns?

Check these:
<> rendering
https://gist.github.com/koperagen/bc097d8de3d2b239e82c4f082ac18071

---- rendering
https://gist.github.com/koperagen/ff1c489d9520c2fffe30fee636a22422

just type
https://gist.github.com/koperagen/4ce0d2192f3bb8c5c1d4c8f2fba443a0

there is comparison of all approaches for complex data with frame and column groups

Jolanrensen:
@koperagen I think the `----` approach is the clearest then :) It makes sure we still have a clear distinction, but no visual (and cognitive) noise with `<{}>` etc. :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/rendering/RenderToStringTests.kt
```diff
@@ -0,0 +1,231 @@
...
+    @Test
+    fun `nullable values`() {
+        val nullDf = dataFrameOf("a", "b")(
+            1, "hello",
+            2, null,
+        )
+        val result = nullDf.renderToString(rowIndex = false, columnTypes = false)
+        val expected = """
+         |  a      b
+         |---   ----
```
Jolanrensen, line 207: is it correct that `a` only has 3 horizontal bars of separation and `b` has 4?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/rendering/RenderToMarkdownTests.kt
```diff
@@ -0,0 +1,152 @@
...
+        "Alice", 30, "Berlin",
+        "Bob", 25, "Paris",
+        "Charlie", 35, "London",
+    )
+
+    @Test
+    fun `markdown basic structure`() {
+        val result = df.renderToMarkdown()
+        val expected =
+            """
```
Jolanrensen, line 21: Image

So the output is valid markdown, that's great :) but one of the advantages of md is that the source code is also quite readable; we throw that away by outputting the tables like this.

On the other hand, we're not the only ones doing that. This is the IntelliJ export to md:

Image

so maybe it's fine like this and give the user the choice to keep it compact or format it for themselves.

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/rendering/RenderToStringTests.kt
koperagen: Sort of? i thought to not make column more wide than it's values

Jolanrensen (approved):

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/rendering/RenderToStringTests.kt
Jolanrensen: ah, that makes sense

Issue #1348, opened 2025-07-25
IntelliJ plugin: "materialize compiler plugin DataFrame type" action


Jolanrensen:
Let's say you have a function that reads a dataframe and performs some conversions on it:

```kt
fun getEnglishPenguins() /* : DataFrame<DanishPenguins_123> */ = 
    DataFrame.read("").cast<DanishPenguins>()
        .rename { arter }.into("species")
        .rename { � }.into("island")
        .rename { n�bl�ngde_mm }.into("bill_length_mm")
        .rename { n�bdybde_mm }.into("bill_depth_mm")
        .rename { luffel�ngde_mm }.into("flipper_length_mm")
        .rename { kropsmasse_g }.into("body_mass_g")
        .rename { k�n }.into("sex")
        .rename { m�ledato }.into("measurement_date")
```

So, imagine your inlay hint tells you you got a `DataFrame<DanishPenguins_123>`. However, you want to define this type somewhere as a data schema (because you want to use it as input to a new function). 
With an IntelliJ plugin, you could make it so that when hovering over this type (or when your cursor is at a variable of that type or touching a function call returning that type), a quick-fix could become available to "Create data schema interface/class". 

What it would do:
- First, it would allow you to define in which scope to put it, like with the "extract to function" action.
- Then, when it is created with the information of the compiler plugin, it has the same name, so `DanishPenguins_123` and the result of the expression gets a `.cast<DanishPenguins_123>()`
- Then a refactoring is started on the name of the interface, so the user can either rename it to something they want or press Enter to leave it like that.

Jolanrensen:
This could also help a ton when creating "magic typed" dataframes as top-level properties, either directly in the file or in a class. It's not possible to write this due to the compiler resolution order:

<img width="1761" height="654" alt="Image" src="https://github.com/user-attachments/assets/67015d5f-7ab3-4d9a-ac37-e7db0c444980" />

Providing a warning/hint when creating `a` to "materialize the type" would solve this issue immediately.

AndreiKingsley:
I was also thinking about schema generation in runtime;
You mark DataFrame (for example with a debug breakpoint) and run a special configuration that creates a data schema for selected DataFrame based on its actual runtime schema.

This could replace gradle/ksp plugins, but it has also an even broader scope — sometimes while working with DF
you can get a DataFrame with a schema that can't be inferred with Compiler Plugin.

AndreiKingsley:
[...]

PR #1762, opened 2026-03-19
Bumping to Kotlin 2.3.20

Jolanrensen:
Bumping to kotlin 2.3.20 and testing the `syncAllExampleFolders` task :) (seems to propagate the version bump well!)

Updating the maven example projects according to: https://kotlinlang.org/docs/whatsnew2320.html#simplified-setup-for-kotlin-projects


samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/concepts/StringApi.kt
```diff
@@ -13,13 +15,25 @@ import org.junit.Test
...
+    }
+
     private val df = dataFrameOf(
         "name" to columnOf("Alice", "Bob"),
         "info" to columnOf(
             "age" to columnOf(23, 27),
             "height" to columnOf(175.5, 160.2),
         ),
-    )
+    ).cast<DfType>()
```
Jolanrensen, line 36: https://github.com/Kotlin/dataframe/issues/1348 would have been really useful to have here

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Modify.kt
```diff
@@ -83,6 +84,7 @@ class Modify : DataFrameSampleHelper("operations", "modify") {
 
     @Test
     fun formatExampleNumbers() {
+        val df2 = getDf2()
```
koperagen, line 87: nice idea for samples

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/RenameToCamelCase.kt
```diff
@@ -11,10 +13,19 @@ import org.jetbrains.kotlinx.dataframe.samples.DataFrameSampleHelper
...
+        val ColumnA: Int
+        val column_b: String
+        val `COLUMN-C`: Boolean
+    }
+
+    private val df =
+        dataFrameOf("ColumnA", "column_b", "COLUMN-C")(
+            1, "a", true,
+            2, "b", false,
+        ).cast<DfType>()
```
koperagen, line 28: i know only db type!

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Modify.kt
Jolanrensen: does seem like a lot of redundant work :/ I'd like to avoid it if possible

Issue #1761, opened 2026-03-19
Create a benchmark to measure reading/joining multiple files into a single DataFrame


zaleslaw:
It is a case where a user can read hundreds of files with the same structure and combine them into a single DataFrame later.
We need to explore the current solution with `concat` operation and compare with other possible approaches

Jolanrensen:
If there is a faster solution than currently exists in the implementation of `Iterable<DataFrame<T>>.concat()`, the `concat` function likely should be updated. It is the main recommended way we currently have of combining multiple dataframes with the same or different schema: https://kotlin.github.io/dataframe/concat.html#schema-unification.

Implementation: https://github.com/Kotlin/dataframe/blob/33e2e48e13c0c22027c626f7e9adb2413a3194ad/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/concat.kt#L86

I have a hunch that it is already optimal, especially when the schemas match, but it never hurts to check.

PR #1764, opened 2026-03-23
Fixed concatKeepingSchema turning ColumnGroups into DataColumns

### LINKED ISSUE #1763: Reading Struct field from an Arrow file with multiple batches leads to incomplete schema
Generate a file with code below, read it `val df = DataFrame.readArrowFeather("test_struct.arrow")` and print schema

Actual: 
```
id: Int
person: DataRow<*>
```

Expected:
```
id: Int
person:
    name: String
    age: Int
```

Problem is in the concatKeepingSchema that treats ColumnGroup as regular DataColumns. Only applicable for files with multiple batches, which is not always the case.

```kt
fun main() {
    val allocator = RootAllocator()

    val structType = FieldType(true, ArrowType.Struct.INSTANCE, null)
    val structField = Field("person", structType, listOf(
        Field("name", FieldType(false, ArrowType.Utf8.INSTANCE, null), emptyList()),
        Field("age", FieldType(false, ArrowType.Int(32, true), null), emptyList()),
    ))
    val schema = Schema(listOf(
        Field("id", FieldType(false, ArrowType.Int(32, true), null), emptyList()),
        structField,
    ))

    VectorSchemaRoot.create(schema, allocator).use { root ->
        val idVec = root.getVector("id") as IntVector
        val structVec = root.getVector("person") as StructVector
        val nameVec = structVec.getChild("name") as VarCharVector
        val ageVec = structVec.getChild("age") as IntVector

        FileOutputStream("test_struct.arrow").use { fos ->
            ArrowFileWriter(root, DictionaryProvider.MapDictionaryProvider(), fos.channel).use { writer ->
                writer.start()

                // batch 1
                idVec.allocateNew(2)
                structVec.allocateNew()
                for (i in 0 until 2) {
                    idVec.set(i, i)
                    nameVec.setSafe(i, "Alice_$i".toByteArray())
                    ageVec.setSafe(i, 25 + i)
                    structVec.setIndexDefined(i)
                }
                root.rowCount = 2
                writer.writeBatch()

                // batch 2
                root.clear()
                idVec.allocateNew(3)
                structVec.allocateNew()
                for (i in 0 until 3) {
                    idVec.set(i, 10 + i)
                    nameVec.setSafe(i, "Bob_$i".toByteArray())
                    ageVec.setSafe(i, 30 + i)
                    structVec.setIndexDefined(i)
                }
                root.rowCount = 3
                writer.writeBatch()

                writer.end()
            }
        }
    }
    allocator.close()
}
```

koperagen:
Fixes #1763

arrow cannot yet read framecolumns, so fix only for ColumnGroup. Maybe in the future we should remove code duplication, but let's wait for benchmark result first. For now i'd like to keep this optimized version


Jolanrensen (approved):
ah, I see, it would probably indeed be better to simply call `.concat()` and not have to fix the logic in two places. But let's await https://github.com/Kotlin/dataframe/issues/1761 indeed. For now it's okay :)

Issue #985, opened 2024-12-04
Simplify KDoc preprocess process


Jolanrensen:
As mentioned [here](https://github.com/Kotlin/dataframe/pull/959#discussion_r1854114787): we can change the publication task to include any sources we want, so in theory, we could just point at the generated sources of `processKDocsMain`.

We'd need to include both the compiled version of the generated sources (so `@ExcludeFromSources` keeps working) as well as a sources.jar built from it.

Jolanrensen:
Relevant KoDEx issue: https://github.com/Jolanrensen/KoDEx/issues/64

PR #1765, opened 2026-03-23
KoDEx convention plugin

Jolanrensen:
created `dfbuild.kodex` convention plugin, migrated :core and csv to use it

Helps #1661 and to a lesser extent #985,

### DIFF
--- a/build-logic/README.md
+++ b/build-logic/README.md
@@ -26,7 +26,7 @@ acting as [Convention Plugins](https://docs.gradle.org/current/userguide/impleme
 - `dfbuild.kodex`: Sets up [KoDEx](https://github.com/Jolanrensen/KoDEx) KDoc preprocessing.
     Requires `dfbuild.kotlinJvm8` or `dfbuild.kotlinJvm11`.
     - This plugin modifies the `Jar` tasks such that its sources contain KoDEx-preprocessed sources.
-    - Available tasks: `preprocessKDocsMain` and `kodex` (alias, does the same thing).
+    - Available tasks: `processKDocsMain` and `kodex` (alias, does the same thing).
     - All tasks and modifications will be skipped if the Gradle property `skipKodex` exists.
     - The plugin can be configured via the `kodexConvention` extension for your needs.
     - See also: https://github.com/Kotlin/dataframe/blob/master/KODEX_KDOC_PREPROCESSING.md


core/build.gradle.kts
```diff
@@ -191,171 +201,6 @@ tasks.withType<KorroTask> {
...
-        logger.lifecycle("generated META-INF/kotlin-jupyter-libraries/libraries.json for :core")
-    }
-}
-
-// If `includeCoreLibrariesJson` is set, modify the processResources task such that it includes
-// a META-INF libraries.json file.
-// This file allows loading dataframe-jupyter when dataframe-core is present on its own in a Kotlin Notebook.
-// This is usually only done when publishing.
-tasks.processResources {
-    if (project.hasProperty("includeCoreLibrariesJson")) {
```
koperagen, line 317: looks like removed by accident

core/build.gradle.kts
```diff
@@ -36,6 +32,20 @@ kotlin.sourceSets {
     }
 }
 
+kodexConvention {
+    // we modified kotlin.sourceSets.(main|test); letting `dfbuild.kodex` know
+    // these will all be processed by KoDEx ánd end up in the final jar.
+    extraSourcesForKodex =
```
koperagen: Why is this needed here but not in csv module? What part of previous code corresponds to this logic?

Jolanrensen: since the convention plugin loads first, it doesn't pick up the additions of "generated-dataschema-accessors" to the kotlin main/test source sets. This is a problem, because it would mean they didn't end up in the final jars... (and csv does not have such an additional sourceset)

Maybe I could simplify the logic somewhat: Instead of "extra" sources, you're simply allowed to override the sourcesets used by KoDEx :) that would be similar to the extension notation of KoDEx we don't use yet (because I didn't get around to finishing it yet):
```kt
kodex {
    preprocess(mySourceSet) { ... }
}
```

core/build.gradle.kts
Jolanrensen: thanks!

core/build.gradle.kts
Jolanrensen: actually... maybe I can already use the `kodex {}` extension notation. It's just the jar creation that needs more work figuring out, but we can leave that as is

Jolanrensen: on second thought... this may need some consideration, as there are some sourcesets that do need to be handled by KoDEx but don't need to be in the final jar (like `src/test`) and the `kodex {}` notation doesn't support that kind of setup yet... Still, this convention plugin could be a good trial run for a design of that :)

Jolanrensen: ah, it seems asking the sourcesets in `afterEvaluate {}` seems to get the most up-to-date version :D

Jolanrensen:
Following up on https://github.com/Kotlin/dataframe/pull/1765#discussion_r2976165864:

I found out we had quite a difficult setup with how we processed our code:

```
generated code --------------> jars
                             /
source code --> KoDEx --> linter
              /       \
tests -------           not in jars
```
Not only is this difficult to understand, but it's also hard to maintain.

I simplified it as follows:
```
(main) generated code
                    \
(main) source code --> KoDEx ---> linter --> jars      (main)
                     /      \
(contextual) tests -           not in jars             (contextual)
```

In the new setup, both the generated code and source code form the "main" sourceset, which is exactly how the module forms it, all is processed by KoDEx, the results are linted, and the jars are created from that.

Besides this, we have a "contextual" sourceset, which by default includes the tests. This makes it so we can refer to tests from the main sourceset from KoDEx and include samples with `@sample`. However, this sourceset will not be included in any jars.

I was able to gather the "complete" sourcesets by getting the directories inside `afterEvaluate {}`, which makes any module-specific code run first before these parts of the convention plugin, and it makes any specific setup obsolete :). I will keep the extension, though, as we may need it in the future.

...

~Now trying to make it build successfully...~

build-logic/README.md
```diff
@@ -23,6 +23,14 @@ acting as [Convention Plugins](https://docs.gradle.org/current/userguide/impleme
 
 - `dfbuild.ktlint`: Sets up our linter plugin. Included by default in `dfbuild.kotlinJvmCommon`.
 
+- `dfbuild.kodex`: Sets up [KoDEx](https://github.com/Jolanrensen/KoDEx) KDoc preprocessing.
+    Requires `dfbuild.kotlinJvm8` or `dfbuild.kotlinJvm11`.
+    - This plugin modifies the `Jar` tasks such that its sources contain KoDEx-preprocessed sources.
+    - Available tasks: `preprocessKDocsMain` and `kodex` (alias, does the same thing).
```
AndreiKingsley: "pre-"?

build-logic/src/main/kotlin/dfbuild.kodex.gradle.kts
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// if `processKDocsMain` runs, the Jar tasks must run after it so the generated-sources are there
+tasks.withType<Jar> {
+    mustRunAfter(changeJarTask, processKDocsMain)
+}
+
+// modify all publishing tasks to depend on `changeJarTask` so the sources are swapped out with generated sources
+tasks.configureEach {
+    if (!project.hasProperty("skipKodex") && name.startsWith("publish")) {
```
AndreiKingsley, line 173: Is there any better way to filter publishing tasks? By type mb?

build-logic/README.md
Jolanrensen: nice catch :)

build-logic/src/main/kotlin/dfbuild.kodex.gradle.kts
Jolanrensen: I mean, this whole part should probably be rewritten. Instead of adjusting the `Jar` task when publishing, we should produce separate jars and publish those instead. However, I didn't have time yet to figure that out in the context of our project.

This is our original logic and has been working well for us thus far :)

PR #1769, opened 2026-03-26
Add documentation and tests for `DataColumnArithmetics.kt`

### LINKED ISSUE #530: Document and test DataColumnArithmetics.kt

Allex-Nik:
Fixes #530

Added:

- KDocs
- Tests
- Website documentation

for `DataColumnArithmetics.kt`

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
@@ -284,7 +284,6 @@ private interface NumberPlusColumnDocs {
     typealias NUMBER_PLUS_COLUMN_NULL_NOTE = Nothing
 }
 
-// Is it possible to refer specifically to `Int.minus(column: DataColumn<Int>)` and not to `Int.minus(other: Byte)`?
 /**
  * @include [NumberPlusColumnDocs]
  * @set [NUMBER_PLUS_COLUMN_SEE_ALSO] [minus][DataColumn.minus], [times][DataColumn.times], [div][DataColumn.div]
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
@@ -3,196 +3,1147 @@ package org.jetbrains.kotlinx.dataframe.api
 import org.jetbrains.kotlinx.dataframe.AnyCol
 import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.Predicate
+import org.jetbrains.kotlinx.dataframe.api.ColumnDivNumberDocs.COLUMN_DIV_NUMBER_COLUMN_TYPE
+import org.jetbrains.kotlinx.dataframe.api.ColumnDivNumberDocs.COLUMN_DIV_NUMBER_EXAMPLE
+import org.jetbrains.kotlinx.dataframe.api.ColumnDivNumberDocs.COLUMN_DIV_NUMBER_INT_NOTE
+import org.jetbrains.kotlinx.dataframe.api.ColumnDivNumberDocs.COLUMN_DIV_NUMBER_NULL_NOTE
+import org.jetbrains.kotlinx.dataframe.api.ColumnDivNumberDocs.COLUMN_DIV_NUMBER_SEE_ALSO
+import org.jetbrains.kotlinx.dataframe.api.ColumnDivNumberDocs.COLUMN_DIV_NUMBER_ZERO_ERROR
+import org.jetbrains.kotlinx.dataframe.api.ColumnMinusNumberDocs.COLUMN_MINUS_NUMBER_COLUMN_TYPE
+import org.jetbrains.kotlinx.dataframe.api.ColumnMinusNumberDocs.COLUMN_MINUS_NUMBER_EXAMPLE
+import org.jetbrains.kotlinx.dataframe.api.ColumnMinusNumberDocs.COLUMN_MINUS_NUMBER_EXAMPLE_NUMBER
+import org.jetbrains.kotlinx.dataframe.api.ColumnMinusNumberDocs.COLUMN_MINUS_NUMBER_NULL_NOTE
+import org.jetbrains.kotlinx.dataframe.api.ColumnMinusNumberDocs.COLUMN_MINUS_NUMBER_SEE_ALSO
+import org.jetbrains.kotlinx.dataframe.api.ColumnPlusNumberDocs.COLUMN_PLUS_NUMBER_COLUMN_TYPE
+import org.jetbrains.kotlinx.dataframe.api.ColumnPlusNumberDocs.COLUMN_PLUS_NUMBER_EXAMPLE
+import org.jetbrains.kotlinx.dataframe.api.ColumnPlusNumberDocs.COLUMN_PLUS_NUMBER_EXAMPLE_NUMBER
+import org.jetbrains.kotlinx.dataframe.api.ColumnPlusNumberDocs.COLUMN_PLUS_NUMBER_NULL_NOTE
+import org.jetbrains.kotlinx.dataframe.api.ColumnPlusNumberDocs.COLUMN_PLUS_NUMBER_SEE_ALSO
+import org.jetbrains.kotlinx.dataframe.api.ColumnPlusStringDocs.COLUMN_PLUS_STRING_RECEIVER
+import org.jetbrains.kotlinx.dataframe.api.ColumnPlusStringDocs.COLUMN_PLUS_STRING_RETURN_TYPE
+import org.jetbrains.kotlinx.dataframe.api.CompareDocs.COMPARE_DESCRIPTION
+import org.jetbrains.kotlinx.dataframe.api.CompareDocs.COMPARE_EXAMPLE
+import org.jetbrains.kotlinx.dataframe.api.CompareDocs.COMPARE_OPERATION
+import org.jetbrains.kotlinx.dataframe.api.CompareDocs.COMPARE_SEE_ALSO
+import org.jetbrains.kotlinx.dataframe.api.NotDocs.NOT_COLUMN_TYPE
+import org.jetbrains.kotlinx.dataframe.api.NotDocs.NOT_DESCRIPTION
+import org.jetbrains.kotlinx.dataframe.api.NotDocs.NOT_NULL_NOTE
+import org.jetbrains.kotlinx.dataframe.api.NotDocs.NOT_RETURN
+import org.jetbrains.kotlinx.dataframe.api.NumberDivColumnDocs.NUMBER_DIV_COLUMN_COLUMN_TYPE
+import org.jetbrains.kotlinx.dataframe.api.NumberDivColumnDocs.NUMBER_DIV_COLUMN_DIVIDEND_TYPE
+import org.jetbrains.kotlinx.dataframe.api.NumberDivColumnDocs.NUMBER_DIV_COLUMN_EXAMPLE
+import org.jetbrains.kotlinx.dataframe.api.NumberDivColumnDocs.NUMBER_DIV_COLUMN_INT_NOTE
+import org.jetbrains.kotlinx.dataframe.api.NumberDivColumnDocs.NUMBER_DIV_COLUMN_NULL_NOTE
+import org.jetbrains.kotlinx.dataframe.api.NumberDivColumnDocs.NUMBER_DIV_COLUMN_SEE_ALSO
+import org.jetbrains.kotlinx.dataframe.api.NumberDivColumnDocs.NUMBER_DIV_COLUMN_ZERO_ERROR
+import org.jetbrains.kotlinx.dataframe.api.NumberMinusColumnDocs.NUMBER_MINUS_COLUMN_COLUMN_TYPE
+import org.jetbrains.kotlinx.dataframe.api.NumberMinusColumnDocs.NUMBER_MINUS_COLUMN_EXAMPLE
+import org.jetbrains.kotlinx.dataframe.api.NumberMinusColumnDocs.NUMBER_MINUS_COLUMN_EXAMPLE_NUMBER
+import org.jetbrains.kotlinx.dataframe.api.NumberMinusColumnDocs.NUMBER_MINUS_COLUMN_NULL_NOTE
+import org.jetbrains.kotlinx.dataframe.api.NumberMinusColumnDocs.NUMBER_MINUS_COLUMN_NUMBER_TYPE
+import org.jetbrains.kotlinx.dataframe.api.NumberMinusColumnDocs.NUMBER_MINUS_COLUMN_SEE_ALSO
+import org.jetbrains.kotlinx.dataframe.api.NumberPlusColumnDocs.NUMBER_PLUS_COLUMN_COLUMN_TYPE
+import org.jetbrains.kotlinx.dataframe.api.NumberPlusColumnDocs.NUMBER_PLUS_COLUMN_EXAMPLE
+import org.jetbrains.kotlinx.dataframe.api.NumberPlusColumnDocs.NUMBER_PLUS_COLUMN_EXAMPLE_NUMBER
+import org.jetbrains.kotlinx.dataframe.api.NumberPlusColumnDocs.NUMBER_PLUS_COLUMN_NULL_NOTE
+import org.jetbrains.kotlinx.dataframe.api.NumberPlusColumnDocs.NUMBER_PLUS_COLUMN_NUMBER_TYPE
+import org.jetbrains.kotlinx.dataframe.api.NumberPlusColumnDocs.NUMBER_PLUS_COLUMN_SEE_ALSO
+import org.jetbrains.kotlinx.dataframe.api.TimesDocs.TIMES_COLUMN_TYPE
+import org.jetbrains.kotlinx.dataframe.api.TimesDocs.TIMES_EXAMPLE
+import org.jetbrains.kotlinx.dataframe.api.TimesDocs.TIMES_EXAMPLE_NUMBER
+import org.jetbrains.kotlinx.dataframe.api.TimesDocs.TIMES_NULL_NOTE
+import org.jetbrains.kotlinx.dataframe.api.TimesDocs.TIMES_SEE_ALSO
+import org.jetbrains.kotlinx.dataframe.api.UnaryMinusDocs.UNARY_MINUS_COLUMN_TYPE
+import org.jetbrains.kotlinx.dataframe.api.UnaryMinusDocs.UNARY_MINUS_NULL_NOTE
 import org.jetbrains.kotlinx.dataframe.columns.ColumnReference
 import java.math.BigDecimal
 import java.math.BigInteger
 
+// region Not
+
+/**
+ * Returns a {@get [NOT_COLUMN_TYPE] [DataColumn]} containing the inverse [Boolean] values
+ * of this {@get [NOT_COLUMN_TYPE] [DataColumn]}. {@get [NOT_NULL_NOTE]}
+ *
+ * {@get [NOT_DESCRIPTION]}
+ *
+ * ### Example
+ * ```kotlin
+ * // Given a DataFrame of financial transactions,
+ * // create a column indicating which transactions failed
+ * val failed = !df.isSuccessful.rename("failed")
+ * ```
+ *
+ * @return {@get [NOT_RETURN]}
+ */
+private interface NotDocs {
+    // type of the column accepted and returned by the function
+    typealias NOT_COLUMN_TYPE = Nothing
+
+    // info about handling `null` values
+    typealias NOT_NULL_NOTE = Nothing
+
+    // description of the function
+    typealias NOT_DESCRIPTION = Nothing
+
+    // info about what the function returns
+    typealias NOT_RETURN = Nothing
+}
+
+/**
+ * @include [NotDocs]
+ * {@set [NOT_DESCRIPTION] Each element in this [column][DataColumn] is transformed
+ * using the logical `not` operation: `true` becomes `false`, and `false` becomes `true`.}
+ * @set [NOT_RETURN] A [DataColumn] containing the negated [Boolean] values of this [column][DataColumn].
+ */
 public operator fun DataColumn<Boolean>.not(): DataColumn<Boolean> = map { !it }
 
+/**
+ * @include [NotDocs]
+ * @set [NOT_NULL_NOTE] Each `null` value in the original [DataColumn] is preserved.
+ * {@set [NOT_DESCRIPTION] Non-null values are transformed using the logical `not` operation:
+ * `true` becomes `false`, and `false` becomes `true`. `null` values remain `null`.}
+ * @set [NOT_RETURN] A [DataColumn] containing the negated [Boolean] values of this [column][DataColumn],
+ * while preserving `null` values.
+ */
 @JvmName("notBooleanNullable")
 public operator fun DataColumn<Boolean?>.not(): DataColumn<Boolean?> = map { it?.not() }
 
+/**
+ * @include [NotDocs]
+ * @set [NOT_COLUMN_TYPE] [ColumnReference]
+ * {@set [NOT_DESCRIPTION] Each value in this [reference][ColumnReference] is transformed
+ * using the logical `not` operation: `true` becomes `false`, and `false` becomes `true`.}
+ * @set [NOT_RETURN] A [ColumnReference] containing the negated [Boolean] values of this [reference][ColumnReference].
+ */
 public operator fun ColumnReference<Boolean>.not(): ColumnReference<Boolean> = map { !it }
 
+/**
+ * @include [NotDocs]
+ * @set [NOT_COLUMN_TYPE] [ColumnReference]
+ * @set [NOT_NULL_NOTE] Each `null` value in the original [ColumnReference] is preserved.
+ * {@set [NOT_DESCRIPTION] Non-null values are transformed using the logical `not` operation:
+ * `true` becomes `false`, and `false` becomes `true`. `null` values remain `null`.}
+ * {@set [NOT_RETURN] A [ColumnReference] containing the negated [Boolean] values of the original [reference][ColumnReference],
+ * while preserving `null` values.}
+ */
 @JvmName("notBooleanNullable")
 public operator fun ColumnReference<Boolean?>.not(): ColumnReference<Boolean?> = map { it?.not() }
 
+// endregion
+
+// region ColumnPlusNumber
+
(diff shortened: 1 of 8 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
```diff
@@ -3,196 +3,1147 @@ package org.jetbrains.kotlinx.dataframe.api
...
 public operator fun DataColumn<Boolean?>.not(): DataColumn<Boolean?> = map { it?.not() }
 
+/**
+ * @include [NotDocs]
+ * @set [NOT_COLUMN_TYPE] [ColumnReference]
+ * {@set [NOT_DESCRIPTION] Each value in this [reference][ColumnReference] is transformed
+ * using the logical `not` operation: `true` becomes `false`, and `false` becomes `true`.}
+ * @set [NOT_RETURN] A [ColumnReference] containing the negated [Boolean] values of this [reference][ColumnReference].
+ */
 public operator fun ColumnReference<Boolean>.not(): ColumnReference<Boolean> = map { !it }
```
Allex-Nik, line 122: Should we reuse the example for `DataColumn` (as one of implementations) here and in other cases when `ColumnReference` is used?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
```diff
@@ -0,0 +1,743 @@
...
+        !columnOf(true, false, true) shouldBe columnOf(false, true, false)
+    }
+
+    @Test
+    fun `not on DataColumn with null`() {
+        !columnOf(false, null, true) shouldBe columnOf(true, null, false)
+    }
+
+    @Test
+    fun `not on ColumnAccessor`() {
```
Allex-Nik, line 24: Does this kind of tests make sense? The function is defined on `ColumnReference`, and I thought I could test it on `DataColumn` and `ColumnAccessor`

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
```diff
@@ -0,0 +1,743 @@
...
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.shouldNotBe
+import org.junit.Test
+import java.math.BigDecimal
+
+class DataColumnArithmeticsTests {
+
+    // region Not
+
+    @Test
```
Allex-Nik, line 13: Is it worth trying to group similar tests together into common methods? I thought it might be good to keep them separate so that the problem is evident from the name of the test already, even though it inflates the amount of code
Also, the grouping might not always be clean because sometimes we would still have to keep in a test several versions of the same column because of different types

docs/StardustDocs/topics/columnArithmetics.md
```diff
@@ -0,0 +1,157 @@
...
+
+<!---IMPORT org.jetbrains.kotlinx.dataframe.samples.api.Modify-->
+
+Kotlin DataFrame provides operators for applying simple arithmetic, logical, string, and comparison operations
+to [`DataColumn`](DataColumn.md) and `ColumnReference` values. These operations include, for example, adding a value to a column,
+multiplying a column by a value, and comparing elements in a column with a value.
+
+## When useful
+
+In most transformations, these column operations are usually not the preferred approach in Kotlin DataFrame,
```
Allex-Nik, line 11: Not sure if I conveyed the idea in sufficient detail. Is there anything that should be added?

docs/StardustDocs/topics/columnArithmetics.md
```diff
@@ -0,0 +1,157 @@
...
+In most transformations, these column operations are usually not the preferred approach in Kotlin DataFrame, 
+because the library provides row-based APIs such as [`add`](add.md), [`update`](update.md), [`map`](map.md), and `expr`, 
+which are usually recommended.
+
+However, column arithmetics might still be useful in some cases. 
+For example, when building temporary plotting expressions. 
+If distance is stored in meters, but you need to [`plot`](https://kotlin.github.io/kandy/welcome.html) it in kilometers, 
+you can convert it directly in the plotting expression without having to create a temporary column in your dataframe just for plotting. 
+
+```kotlin
```
Allex-Nik, line 48: Since the example from the [discussion](https://github.com/Kotlin/dataframe/pull/1617) of that API was not compiling (because we do not support `column1 + column2`), I tried to come up with something close.

Is it valid?

If it is, should it be added to `core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Modify.kt`? It seems that Kandy hasn't been used there.

docs/StardustDocs/topics/columnArithmetics.md
```diff
@@ -0,0 +1,157 @@
...
+}
+```
+
+## Not
+Negates Boolean values in a [`DataColumn`](DataColumn.md) (`ColumnReference`). 
+Returns a result of the same type.
+
+<!---FUN columnArithmetics_not-->
+
+```kotlin
```
Allex-Nik, line 65: For now I have added this one snippet with Korro.

Should I make all the samples on the page like this? If yes, can I use example dataframes different from the dataframe provided by the TestBase? It seems that not all the column arithmetics can be logically illustrated with the dataframe about Alice, Bob, etc

docs/StardustDocs/topics/columnArithmetics.md
```diff
@@ -0,0 +1,157 @@
...
+Nullable Boolean columns preserve `null` values.
+
+## Plus
+Adding a value to a column and concatenation of a column with a `String` are supported. 
+In all cases, a [`DataColumn`](DataColumn.md) (`ColumnReference`) is returned.
+
+### Column plus value or value plus column
+Adds the value to each element of the column. The value can appear on either side of the operator.
+
+```kotlin
```
Allex-Nik, line 84: Do you think any iframes are needed on this page? It seems that the operations are simple enough.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
```diff
@@ -3,196 +3,1147 @@ package org.jetbrains.kotlinx.dataframe.api
...
+ * ### Example
+ * ```kotlin
+ * // Given a DataFrame of financial transactions,
+ * // create a column indicating which transactions failed
+ * val failed = !df.isSuccessful.rename("failed")
+ * ```
+ *
+ * @return {@get [NOT_RETURN]}
+ */
+private interface NotDocs {
```
Allex-Nik, line 82: Do I use the correct interface pattern?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
```diff
@@ -3,196 +3,1146 @@ package org.jetbrains.kotlinx.dataframe.api
...
+ */
+@JvmName("plusNullable")
+public operator fun Int.plus(column: DataColumn<Int?>): DataColumn<Int?> = column.map { it?.plus(this) }
 
-public operator fun Int.div(column: ColumnReference<Int>): ColumnReference<Int> = column.map { this / it }
+/**
+ * @include [NumberPlusColumnDocs]
+ * @set [NUMBER_PLUS_COLUMN_NUMBER_TYPE] [Double]
+ * @set [NUMBER_PLUS_COLUMN_EXAMPLE_NUMBER] 10.0
+ * @set [NUMBER_PLUS_COLUMN_SEE_ALSO] [minus][DataColumn.minus], [times][DataColumn.times], [div][DataColumn.div]
```
Allex-Nik, line 320: This refers to `DataColumn<Int>.minus(value: Int): DataColumn<Int>`. 
Isn't there any way to refer to `Double.minus(column: DataColumn<Int>)` instead?

`[minus][Double.minus]` doesn't work because it refers to `public final operator fun minus(other: Byte): Double` from Kotlin.

docs/StardustDocs/topics/columnArithmetics.md
koperagen: Looks good
Yes, please use different datasets, nice snippets below

docs/StardustDocs/topics/columnArithmetics.md
koperagen: Maybe there's no need, i'm not sure. I suppose they won't be generated well automatically, so it's ok if you skip iframes. Just remove TransformDataFrameExpressions annotation from test fun

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
Jolanrensen: I think yes, but we don't really promote column accessors like this anymore. What you could check, however, is `df.getColumn { !col<Boolean>("isClosed") }`, so the string API inside CS DSL

Jolanrensen: to check the compiler plugin, it's enough to check whether `!dataColumn` works, because those extensions return `DataColumn<>` instances :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
Jolanrensen: I think it does not matter that much how we test it, as long as we do. So that we can catch it when we break something :) The names are certainly very clear!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
```diff
@@ -3,196 +3,1146 @@ package org.jetbrains.kotlinx.dataframe.api
...
+ * Returns a {@get [NOT_COLUMN_TYPE] [DataColumn]} containing the inverse [Boolean] values
+ * of this {@get [NOT_COLUMN_TYPE] [DataColumn]}. {@get [NOT_NULL_NOTE]}
+ *
+ * {@get [NOT_DESCRIPTION]}
+ *
+ * ### Example
+ * ```kotlin
+ * // Given a DataFrame of financial transactions,
+ * // create a column indicating which transactions failed
+ * val failed = !df.isSuccessful.rename("failed")
```
Jolanrensen: hmm, maybe this example could also include an example from inside the selection DSL

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
```diff
@@ -3,196 +3,1146 @@ package org.jetbrains.kotlinx.dataframe.api
...
+ * @include [NotDocs]
+ * {@set [NOT_DESCRIPTION] Each element in this [column][DataColumn] is transformed
+ * using the logical `not` operation: `true` becomes `false`, and `false` becomes `true`.}
+ * @set [NOT_RETURN] A [DataColumn] containing the negated [Boolean] values of this [column][DataColumn].
+ */
 public operator fun DataColumn<Boolean>.not(): DataColumn<Boolean> = map { !it }
 
+/**
+ * @include [NotDocs]
+ * @set [NOT_NULL_NOTE] Each `null` value in the original [DataColumn] is preserved.
```
Jolanrensen, line 106: careful mixing block- and inline tags, you can see when you place your cursor here which text is included by this `@set` tag. Because it doesn't have `{}` around it, the `{@set }` below it becomes part of the `[NOT_NULL_NOTE]`. (okay in practice, the `{@set }` will be read first and removed from the KDOC, so nothing of it actually enters `[NOT_NULL_NOTE]`, but mixing scopes it generally only a good idea if it's intentional)

In this case, I would turn them all into block tags (so without `{}`), then each tag continues until it encounters another or the end of the kdoc is reached. Same as `@param` and `@return` :)

Jolanrensen: the same happens in other places

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
Jolanrensen: yeah, looks good :) Only thing you could add is `@ExcludeFromSources` for each interface that is solely used to be `@include`'d somehwere

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
Jolanrensen: actually, it's not often we work with columns outside a dataframe. we could :) but it's probably best to show it inside a dataframe operation first

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
Jolanrensen: nope... `[Double.minus]` is the closest we can get. If there's multiple matches, the 2026 IDE does give you a list to pick, but there's no way to specify an exact function in KDocs :/ (yet!)

Jolanrensen: is sort of mentioned here https://youtrack.jetbrains.com/issue/KT-69919/DM-2024-07-23-Revisit-KDoc-design-Ambiguous-KDoc-references and related issues

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
```diff
@@ -3,196 +3,1146 @@ package org.jetbrains.kotlinx.dataframe.api
...
+// endregion
+
+// region ColumnPlusString
+
+/**
+ * Returns a {@get [COLUMN_PLUS_STRING_RETURN_TYPE] [DataColumn]} of [String] values
+ * obtained by converting each element of this {@get [COLUMN_PLUS_STRING_RECEIVER] [AnyCol]}
+ * to a [String] and concatenating it with [\str].
+ *
+ * `null` values are converted to the string `"null"`.
```
Jolanrensen, line 375: wait really... I vote to change this XD in all other overloads `null` is ignored. What good is `"€null"`?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
```diff
@@ -3,196 +3,1146 @@ package org.jetbrains.kotlinx.dataframe.api
...
+ *
+ * ### Example
+ * ```kotlin
+ * // Given a DataFrame of expenses to display in a report,
+ * // create a column with a currency symbol appended to each expense amount
+ * val price = df.amount + "€"
```
Jolanrensen: hmm, `"50€"` is a bit weird. Usually the monetary symbol comes before the symbol. Maybe you could turn it into a temperature? like `+ " °C"`?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
```diff
@@ -3,196 +3,1146 @@ package org.jetbrains.kotlinx.dataframe.api
...
+ * @set [COMPARE_DESCRIPTION] using the `<` operator
+ * {@set [COMPARE_EXAMPLE] // Given a DataFrame of orders,
+ * // create a column that indicates which orders cost less than 20 euros
+ * val isCheap = df.orderCost lt 20}
+ * @set [COMPARE_SEE_ALSO] [eq][DataColumn.eq], [neq][DataColumn.neq], [gt][DataColumn.gt]
+ * @set [COMPARE_OPERATION] less than
+ */
 public infix fun <T : Comparable<T>> DataColumn<T>.lt(value: T): DataColumn<Boolean> = isMatching { it < value }
 
 internal fun <T> DataColumn<T>.isMatching(predicate: Predicate<T>): DataColumn<Boolean> = map { predicate(it) }
```
Jolanrensen: maybe we could remove `isMatching` and just use `map` everywhere... This seems unnecessarily pointless

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
```diff
@@ -3,196 +3,1146 @@ package org.jetbrains.kotlinx.dataframe.api
...
+/**
+ * @include [CompareDocs]
+ * @set [COMPARE_DESCRIPTION] using the `>` operator
+ * {@set [COMPARE_EXAMPLE] // Given a DataFrame of orders,
+ * // create a column that indicates which orders cost more than 1,000 euros
+ * val isExpensive = df.orderCost gt 1000}
+ * @set [COMPARE_SEE_ALSO] [eq][DataColumn.eq], [neq][DataColumn.neq], [lt][DataColumn.lt]
+ * @set [COMPARE_OPERATION] greater than
+ */
 public infix fun <T : Comparable<T>> DataColumn<T>.gt(value: T): DataColumn<Boolean> = isMatching { it > value }
```
Jolanrensen: Vote to add `geq` and `leq` (gt or equal). This follows the same name standard as neq, gt, etc. used in LaTeX :)

Jolanrensen: Also, this does not handle nulls :)

`<T : Comparable<T & Any>?>` anyone? ;P

docs/StardustDocs/topics/columnArithmetics.md
```diff
@@ -0,0 +1,157 @@
+[//]: # (title: Column arithmetics)
+
+<!---IMPORT org.jetbrains.kotlinx.dataframe.samples.api.Modify-->
+
+Kotlin DataFrame provides operators for applying simple arithmetic, logical, string, and comparison operations
+to [`DataColumn`](DataColumn.md) and `ColumnReference` values. These operations include, for example, adding a value to a column,
```
Jolanrensen: "adding value to each cell in a column"?

docs/StardustDocs/topics/columnArithmetics.md
Jolanrensen: you can also add it in a new file :) Kandy can be used too I believe. Let me find an example...

Jolanrensen: guides/quickstart.kt has a kandy example :)

docs/StardustDocs/topics/columnArithmetics.md
```diff
@@ -0,0 +1,157 @@
...
+5.0 + df.amount
+```
+
+`null` values are not changed by this operation.
+
+### Column plus String
+Converts each element of the column to a String and concatenates it with the value.
+
+```kotlin
+df.amount + "€"
```
Jolanrensen: I think only `"€" + df.amount` would make sense, but that might be something we cannot overload... right?

docs/StardustDocs/topics/columnArithmetics.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+`null` values are converted to the string `"null"`.
```
Jolanrensen, line 106: vote to change!!

Jolanrensen (changes requested):
thanks! :D

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
Allex-Nik: For some reason, `[Double.minus]` does not suggest me a list in 2026.2 EAP :(
Maybe we should leave it `[minus][DataColumn.minus]` here to refer to `DataColumn<Int>.minus(value: Int): DataColumn<Int>` for now? It seems more helpful than `minus(other: Byte): Double` to me. 
Or is it still better to make it `[Double.minus]`?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
Allex-Nik: I will replace it with 
```kotlin
df.select { !isSuccessful }
// or
!df.isSuccessful
 ```

Comment (Jolanrensen) [reply core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt]: yes, useful! :)

File: core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt (hunk 11 above)
Comment (Allex-Nik) [reply core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt:106]: I've fixed it. Also checked out inline tags, block tags, and the order of their processing - it is clear now. Thank you!

File: core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt (hunk 12 above)
Comment (Allex-Nik) [reply core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt:375]: I agree that keeping `null` values is better. Shall I open an issue for that?

File: docs/StardustDocs/topics/columnArithmetics.md (hunk 17 above)
Comment (Allex-Nik) [reply docs/StardustDocs/topics/columnArithmetics.md]: Seems like we cannot. I have tried to add 
```kotlin
public operator fun String.plus(column: AnyCol): DataColumn<String> = column.map { this + it.toString() }
```
then publish it to mavenLocal and use it, but the `+` operator in `"€" + df.amount` is still resolved to 
```kotlin
@IntrinsicConstEvaluation
public final operator fun plus(
    other: Any?
): String
```

docs/StardustDocs/topics/columnArithmetics.md
Allex-Nik: Agree :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
Allex-Nik: > Vote to add geq and leq (gt or equal)

Sounds good!

> Also, this does not handle nulls :)

Indeed

Should I create an issue for adding `geq` and `leq` and null handling in `gt` and `lt`?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
Allex-Nik: Done :) Is it correct now?

docs/StardustDocs/topics/columnArithmetics.md
Allex-Nik: Agree, changed, but should be "a value", right?

docs/StardustDocs/topics/columnArithmetics.md
Allex-Nik: I replaced `df.amount + "€"` with
```kotlin
weather.select { temperature + " °C" }
// or
weather.temperature + " °C"
```
so that it makes sense :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
Jolanrensen: hmm for me neither... I'd just go with the most helpful one :) Let's hope in the future they make it so we can target specific functions

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
Jolanrensen: yes please :)

docs/StardustDocs/topics/columnArithmetics.md
Jolanrensen: Yeah, I remember; we cannot overload `String` operators unfortunately :/

Jolanrensen (approved):

docs/StardustDocs/topics/columnArithmetics.md
```diff
@@ -0,0 +1,212 @@
...
+
+Kotlin DataFrame provides operators for applying simple arithmetic, logical, string, and comparison operations
+to [`DataColumn`](DataColumn.md) and `ColumnReference` values. These operations include, for example, adding a value 
+to each cell in a column, multiplying a column by a value, and comparing elements in a column with a value.
+
+## When useful
+
+In most transformations, these column operations are usually not the preferred approach in Kotlin DataFrame, 
+because the library provides row-based APIs such as [`add`](add.md), [`update`](update.md), [`map`](map.md), and `expr`, 
+which are usually recommended.
```
koperagen, line 13: Please elaborate on expr { } as a column selector => row expression adapter
for example,
`df.groupBy { city + "mySuffix"  }` is equivalent to 
`df.groupBy { expr { city + "mySuffix" } }`, but in expr city is seen as single String value, not as column

zaleslaw (approved):

AndreiKingsley (approved):
Great KoDEx usage!

docs/StardustDocs/topics/columnArithmetics.md
Allex-Nik: Done :) 
Please see if it is correct now. I tried to explain the `expr` usage and included similar examples. I also specified the name of the column for `expr` (so that the results for both versions look the same), otherwise the column was called "untitled" for the version with `expr`, whereas in the version without `expr` the column had a specific name. Is that fine?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnArithmetics.kt
Allex-Nik: Done :)

zaleslaw (approved):

docs/StardustDocs/topics/columnArithmetics.md
koperagen: Good! Just for you: it's also possible to write expr { } named "city"

koperagen (approved):

AndreiKingsley (approved):

PR #1772, opened 2026-03-26
Substitute type parameters with type arguments in toDataFrame conversion

### LINKED ISSUE #1771: toDataFrame loses types of properties with type parameter return type
easy repro with unfold Pair<Int, String>, toDataFrame under the hood:
`dataFrameOf("a" to columnOf(1 to "str")).unfold { a }`

Actual:
```
a:
 first: Any
 second: Any
```

Expected:
```
a:
 first: Int
 second: String
```

koperagen:
fixes https://github.com/Kotlin/dataframe/issues/1771

Includes little experiment with custom asserters for dataframe shape

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
@@ -22,6 +22,7 @@ import org.jetbrains.kotlinx.dataframe.impl.isGetterLike
 import org.jetbrains.kotlinx.dataframe.impl.isJavaRecord
 import org.jetbrains.kotlinx.dataframe.impl.projectUpTo
 import org.jetbrains.kotlinx.dataframe.impl.recordComponentNames
+import org.jetbrains.kotlinx.dataframe.impl.replace
 import org.jetbrains.kotlinx.dataframe.impl.schema.sortWithConstructor
 import java.lang.reflect.InvocationTargetException
 import java.lang.reflect.Method
@@ -31,13 +32,13 @@ import kotlin.reflect.KCallable
 import kotlin.reflect.KClass
 import kotlin.reflect.KProperty
 import kotlin.reflect.KType
+import kotlin.reflect.KTypeParameter
 import kotlin.reflect.KVisibility
 import kotlin.reflect.full.isSubclassOf
 import kotlin.reflect.full.memberFunctions
 import kotlin.reflect.full.memberProperties
 import kotlin.reflect.full.primaryConstructor
 import kotlin.reflect.full.starProjectedType
-import kotlin.reflect.full.valueParameters
 import kotlin.reflect.full.withNullability
 import kotlin.reflect.jvm.isAccessible
 import kotlin.reflect.jvm.javaField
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
```diff
@@ -429,6 +432,13 @@ internal fun convertToDataFrame(
 
 private fun KType.classifierOrAny(): KClass<*> = classifier as? KClass<*> ?: Any::class
 
+internal fun KType.typeParametersSubstitution(): Map<KTypeParameter, KType> {
+    val klass = classifier as? KClass<*> ?: return emptyMap()
+    return klass.typeParameters.zip(arguments).mapNotNull { (param, projection) ->
+        projection.type?.let { param to it }
```
Jolanrensen, line 438: there seems to be a failing test regarding nullable int arrays not being recognized as nullable now, please check

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/unfold.kt
```diff
@@ -65,4 +71,66 @@ class UnfoldTests {
...
+
+    fun DataFrame<*>.shouldHaveColumnGroup(
+        name: String,
+        block: (ColumnGroup<*>) -> Unit = {
+        },
+    ): ColumnGroup<*> = getColumnOrNull(name).shouldBeInstanceOf<ColumnGroup<*>>(block)
+
+    fun DataFrame<*>.shouldHaveFrameColumn(
+        name: String,
+        block: (FrameColumn<*>) -> Unit = {
```
Jolanrensen: I'd simply write this as ` = {},`, this multiline looks a bit odd

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/unfold.kt
```diff
@@ -65,4 +71,66 @@ class UnfoldTests {
...
+        }
+    }
+
+    fun DataFrame<*>.shouldHaveColumnGroup(
+        name: String,
+        block: (ColumnGroup<*>) -> Unit = {
+        },
+    ): ColumnGroup<*> = getColumnOrNull(name).shouldBeInstanceOf<ColumnGroup<*>>(block)
+
+    fun DataFrame<*>.shouldHaveFrameColumn(
```
Jolanrensen: these are cool! :) maybe put them somewhere in a utils file though, that way we can reuse them

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/unfold.kt
koperagen: fixed. ktlint didn't like {}, but ok with { } apparently

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
koperagen: fixed bug in replace

PR #1775, opened 2026-03-27
Added kdocs for DataSchema and DataRowSchema

koperagen:
I tried to fill in the gap and at the same time provide a new perspective, my Nth attempt to explain dataschema :))

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
@@ -1,4 +1,58 @@
 package org.jetbrains.kotlinx.dataframe.annotations
 
+import org.jetbrains.kotlinx.dataframe.api.cast
+import org.jetbrains.kotlinx.dataframe.api.convertTo
+
+/**
+ *  Annotation to generate extension properties API for a given declaration, according to its properties.
+ *  Annotated declaration should be non-local and non-private interface or a class.
+ *  The aim here is to provide convenient syntax for working with a dataframe instance right after reading from it CSV, JSON, Databases, Arrow, etc.
+ *  After `val df = DataFrame.read*` operation, `df` is a source of truth for the DataSchema.
+ *  One way to look at it, DataSchema "tells" the compiler what's already there. It doesn't affect reading.
+ *  See the list below of code generation methods to simplify the process of getting what we call an initial dataschema.
+ *  Given the initial schema of the data you read, the compiler plugin will provide a typed result for most operations.
+ *
+ * Example:
+ * ```
+ * @DataSchema
+ * data class Group(
+ *     val id: String,
+ *     val participants: List<Person>
+ * )
+ *
+ * @DataSchema
+ * data class Person(
+ *     val name: Name,
+ *     val age: Int,
+ *     val city: String?
+ * )
+ *
+ * @DataSchema
+ * data class Name(
+ *     val firstName: String,
+ *     val lastName: String,
+ * )
+ *
+ * fun main() {
+ *  val url = "https://raw.githubusercontent.com/Kotlin/dataframe/refs/heads/master/data/participants.json"
+ *  val df = DataFrame.readJson(url).cast<Group>()
+ *  val i: Int = df.id[0] // properties style access to columns and values
+ *
+ *  val df1 = df.asGroupBy { participants }.aggregate {
+ *    count() into "groupSize"
+ *    distinct { city } into "cities"
+ *  }
+ *
+ *  // now compiler plugin uses previous knowledge of `Group` combined with its understanding of aggregate operation
+ *  // to help you access new columns
+ *  val l: List<String> = df1.cities[0]
+ * }
+ * ```
+ *
+ * @see [org.jetbrains.kotlinx.dataframe.api.generateDataClasses]
+ * @see [org.jetbrains.kotlinx.dataframe.api.generateInterfaces]
+ * @see [org.jetbrains.kotlinx.dataframe.DataFrame.cast]
+ * @see [org.jetbrains.kotlinx.dataframe.DataFrame.convertTo]
+ */
 @Target(AnnotationTarget.CLASS)
 public annotation class DataSchema(val isOpen: Boolean = true)
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
@@ -5,15 +5,15 @@ import org.jetbrains.kotlinx.dataframe.api.convertTo
 
 /**
  * This annotation marks an interface or data class as a [data schema](https://kotlin.github.io/dataframe/schemas.html).
+ * An annotated declaration should be a non-local and non-private interface or class.
  *
  * This annotation generates an extension properties API for a declaration according to its properties.
- * An annotated declaration should be a non-local and non-private interface or class.
- * The aim is to provide a convenient syntax for working with a dataframe instance right after reading it from CSV,
+ * The aim is to provide a convenient class-like syntax for working with a dataframe instance after reading it from CSV,
  * JSON, databases, Arrow, or other sources.
  *
  * After a `val df = DataFrame.read*` operation, `df` is the source of truth for the data schema.
  * One way to look at it is that a data schema tells the compiler what is already there; it does not affect reading.
- * See the related operations in the See also section.
+ *
  * Given the initial schema of the data you read, the
  * [compiler plugin](https://github.com/JetBrains/kotlin/tree/master/plugins/kotlin-dataframe) provides a typed result
  * for most operations.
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
```diff
@@ -1,4 +1,58 @@
...
+ *  Annotation to generate extension properties API for a given declaration, according to its properties.
+ *  Annotated declaration should be non-local and non-private interface or a class.
+ *  The aim here is to provide convenient syntax for working with a dataframe instance right after reading from it CSV, JSON, Databases, Arrow, etc.
+ *  After `val df = DataFrame.read*` operation, `df` is a source of truth for the DataSchema.
+ *  One way to look at it, DataSchema "tells" the compiler what's already there. It doesn't affect reading.
+ *  See the list below of code generation methods to simplify the process of getting what we call an initial dataschema.
+ *  Given the initial schema of the data you read, the compiler plugin will provide a typed result for most operations.
+ *
+ * Example:
+ * ```
```
Jolanrensen: please annotate any code samples with the right file format to get highlights, anywhere you use Markdown ;P

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
```diff
@@ -1,4 +1,58 @@
...
+ * @DataSchema
+ * data class Group(
+ *     val id: String,
+ *     val participants: List<Person>
```
Jolanrensen: inconsistent trailing commas

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
```diff
@@ -1,4 +1,58 @@
...
+ * )
+ *
+ * @DataSchema
+ * data class Name(
+ *     val firstName: String,
+ *     val lastName: String,
+ * )
+ *
+ * fun main() {
+ *  val url = "https://raw.githubusercontent.com/Kotlin/dataframe/refs/heads/master/data/participants.json"
```
Jolanrensen: inconsistent spacing. I might recommend using the `@sample` KoDEx tag :) similar to Korro, we can include code between `// SampleStart` and `// SampleEnd` comments, making sure it compiles and is formatted correctly.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
```diff
@@ -1,4 +1,58 @@
 package org.jetbrains.kotlinx.dataframe.annotations
 
+import org.jetbrains.kotlinx.dataframe.api.cast
+import org.jetbrains.kotlinx.dataframe.api.convertTo
+
+/**
```
Jolanrensen, line 6: I'd start a bit more generally and introductorily before jumping into exactly what it does.
So: "This annotation marks an interface or data class as 'data schema'" (link to https://kotlin.github.io/dataframe/schemas.html). Then continue with "It's used to generate extension properties, etc.". Gives a bit more context to this key DataFrame component :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
```diff
@@ -1,4 +1,58 @@
...
+ *  Annotation to generate extension properties API for a given declaration, according to its properties.
+ *  Annotated declaration should be non-local and non-private interface or a class.
```
Jolanrensen: *Annotated declaration should be a non-local and non-private interface or class.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
```diff
@@ -1,4 +1,58 @@
...
+ *  The aim here is to provide convenient syntax for working with a dataframe instance right after reading from it CSV, JSON, Databases, Arrow, etc.
```
Jolanrensen: *a convenient syntax

Jolanrensen: *databases

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
```diff
@@ -1,4 +1,58 @@
...
+ *  After `val df = DataFrame.read*` operation, `df` is a source of truth for the DataSchema.
```
Jolanrensen: *After the

Jolanrensen: *the source of truth

Jolanrensen: *data schema (the concept)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
```diff
@@ -1,4 +1,58 @@
...
+ *  One way to look at it, DataSchema "tells" the compiler what's already there. It doesn't affect reading.
```
Jolanrensen: *One way to look at it is that the data schema...

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
```diff
@@ -1,4 +1,58 @@
...
+ *  See the list below of code generation methods to simplify the process of getting what we call an initial dataschema.
```
Jolanrensen: Hmm, maybe "Check out the 'See Also' section?" Depending on how it's rendered, it's not really a list

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
```diff
@@ -1,4 +1,58 @@
 package org.jetbrains.kotlinx.dataframe.annotations
 
+import org.jetbrains.kotlinx.dataframe.api.cast
+import org.jetbrains.kotlinx.dataframe.api.convertTo
+
+/**
+ *  Annotation to generate extension properties API for a given declaration, according to its properties.
```
Jolanrensen: each line in this KDoc starts with two spaces instead of one

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
```diff
@@ -1,4 +1,58 @@
...
+ *  Annotated declaration should be non-local and non-private interface or a class.
+ *  The aim here is to provide convenient syntax for working with a dataframe instance right after reading from it CSV, JSON, Databases, Arrow, etc.
+ *  After `val df = DataFrame.read*` operation, `df` is a source of truth for the DataSchema.
+ *  One way to look at it, DataSchema "tells" the compiler what's already there. It doesn't affect reading.
+ *  See the list below of code generation methods to simplify the process of getting what we call an initial dataschema.
+ *  Given the initial schema of the data you read, the compiler plugin will provide a typed result for most operations.
```
Jolanrensen: link to compiler plugin

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
```diff
@@ -1,4 +1,58 @@
...
+ * @DataSchema
+ * data class Name(
+ *     val firstName: String,
+ *     val lastName: String,
+ * )
+ *
+ * fun main() {
+ *  val url = "https://raw.githubusercontent.com/Kotlin/dataframe/refs/heads/master/data/participants.json"
+ *  val df = DataFrame.readJson(url).cast<Group>()
+ *  val i: Int = df.id[0] // properties style access to columns and values
```
Jolanrensen: you can come up with a better name than `i` and `l` right? ;P I know it's not relevant to the example itself, but I feel we should still set a good example with expressive variable names.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
```diff
@@ -1,4 +1,58 @@
...
+ *
+ *  val df1 = df.asGroupBy { participants }.aggregate {
+ *    count() into "groupSize"
+ *    distinct { city } into "cities"
+ *  }
+ *
+ *  // now compiler plugin uses previous knowledge of `Group` combined with its understanding of aggregate operation
```
Jolanrensen: *the compiler plugin

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
```diff
@@ -1,4 +1,58 @@
...
+ * }
+ * ```
+ *
+ * @see [org.jetbrains.kotlinx.dataframe.api.generateDataClasses]
+ * @see [org.jetbrains.kotlinx.dataframe.api.generateInterfaces]
+ * @see [org.jetbrains.kotlinx.dataframe.DataFrame.cast]
+ * @see [org.jetbrains.kotlinx.dataframe.DataFrame.convertTo]
+ */
 @Target(AnnotationTarget.CLASS)
 public annotation class DataSchema(val isOpen: Boolean = true)
```
Jolanrensen, line 76: We should probably explain what `isOpen` does

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataRowSchemaApi.kt
```diff
@@ -1,9 +1,24 @@
 package org.jetbrains.kotlinx.dataframe.api
 
 import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.annotations.DataSchema
 
+/**
+ * Marker interface that's automatically added to classes annotated with [DataSchema]
```
Jolanrensen: ... why? (because it can help with .append() etc.) Also, we should specify how this is added (with the compiler plugin) and that it's added only to data classes, not to interfaces (right?) and why.

"Added" -> "Added as supertype"

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataRowSchemaApi.kt
```diff
@@ -1,9 +1,24 @@
...
 public interface DataRowSchema
 
+/**
+ * Example:
+ * ```
+ * @DataSchema
+ * data class Person(val name: String, val age: Int)
+ *
+ * fun main() {
+ *  val df = dataFrameOf(Person("Alice", 30), Person("Bob", 25))
```
Jolanrensen: single space indent ;P Same story, try `@sample` :) You could even `@sample` a piece of code and exclude that from sources in this file. For instance:

```kt
@ExcludeFromSources
private interface Sample {

    // to make it compile without the compiler plugin
    fun <T> dataFrameOf(vararg rows: T): DataFrame<T> = TODO()

    // SampleStart
    @DataSchema
    data class Person(val name: String, val age: Int)

    fun main() {
        val df: DataFrame<Person> = dataFrameOf(Person("Alice", 30), Person("Bob", 25))
    }
    // SampleEnd
}

/**
 * Example:
 * @sample [Sample]
 */
public inline fun <reified T : DataRowSchema> dataFrameOf(vararg rows: T): DataFrame<T> =
    rows.asIterable().toDataFrame()
```

Jolanrensen: append could also do with a small sample like this :)

Jolanrensen (changes requested):

Jolanrensen:
Useful! :) Overall I like the extra information, but I feel it lacks some links to other resources. We have a plethora of information on the website we should link to

AndreiKingsley (changes requested):
Please, take #1859  into account.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/annotations/DataSchema.kt
AndreiKingsley: Please also add `isOpen` description to https://kotlin.github.io/dataframe/schemas.html.

PR #1776, opened 2026-03-28
fix: create null-filled columns for remaining names in split into

### LINKED ISSUE #1758: split into(vararg names) should create a null-filled column for remaining names if there's not enough values
```kt
dataFrameOf(
    "a" to columnOf(
        emptyList<Int>(),
        listOf(1, 2),
        listOf(1)
    )
)
    .split { "a"<List<Int>>() }
    .into("a1", "a2", "a3")
```

Result is unexpected:
Where's a3 column? 
Image
Essentially as a result of split, 0 columns can be created even if i explicitly enumerate names. This logic is impossible to be useful in compiler plugin. We need more stable, predicable set of columns as a result

Proposed solution:
Create null-filled columns for remaining names
For `.into("a1", "a2", "a3")` 

At the same time, `split {}.into { ALPHABET[it] }` should continue to work as is

Expected result: 
After the fix 
```kt
dataFrameOf(
    "a" to columnOf(
        emptyList<Int>(),
        listOf(1, 2),
        listOf(1)
    )
)
    .split { "a"<List<Int>>() }
    .into("a1", "a2", "a3")
```
=>
```
a1, a2, a3
null,null, null
1,2, null
1,null, null
```

Nidhi-M21:
Fixes #1758

The fix was made in the `splitImpl` function in `Split.kt`. 
When the number of explicit names exceeds the number of elements 
in the list, null-filled collectors are now created for the remaining columns.

- add null-filled columns when explicit names exceed list elements
- add tests for null filled columns, exact elements, all empty lists and mixed lists

### DIFF
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/split.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/split.kt
@@ -282,4 +282,84 @@ class SplitTests {
         res["a2"][5] shouldBe 7
         res["a3"][5] shouldBe 8
     }
+
+    @Test
+    fun `split into with custom default value`() {
+        val df = dataFrameOf("a")(
+            listOf("A"),
+            listOf("B", "C", "D"),
+        )
+        val res = df.split { "a"<List<String>>() }
+            .default("something")
+            .into("a1", "a2", "a3")
+
+        res["a1"][0] shouldBe "A"
+        res["a2"][0] shouldBe "something"
+        res["a3"][0] shouldBe "something"
+
+        res["a1"][1] shouldBe "B"
+        res["a2"][1] shouldBe "C"
+        res["a3"][1] shouldBe "D"
+    }
+
+    @Test
+    fun `split list with empty list uses custom default value`() {
+        val df = dataFrameOf("a")(
+            emptyList<String>(),
+            listOf("B", "C", "D"),
+        )
+        val res = df.split { "a"<List<String>>() }
+            .default("something")
+            .into("a1", "a2", "a3")
+
+        res.columnNames() shouldBe listOf("a1", "a2", "a3")
+
+        res["a1"][0] shouldBe "something"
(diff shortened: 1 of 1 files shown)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/split.kt
```diff
@@ -173,4 +173,114 @@ class SplitTests {
...
+
+        res["a1"][1] shouldBe null
+        res["a2"][1] shouldBe null
+        res["a3"][1] shouldBe null
+
+        res["a1"][2] shouldBe null
+        res["a2"][2] shouldBe null
+        res["a3"][2] shouldBe null
+    }
+    @Test
```
Jolanrensen, line 246: please run `ktlintFormat`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/split.kt
```diff
@@ -53,6 +53,13 @@ internal fun <T, C, R> splitImpl(
         }
 
         val names = columnNamesGenerator(column, columnCollectors.size)
+
+        repeat(names.size - columnCollectors.size) {
+            val collector = createDataCollector(nrow)
+            repeat(nrow) { collector.add(clause.default) }
```
Jolanrensen, line 59: good job using `.default`! That's even better than simply adding `null`. Could you add a test where this default is changed, though? with `split { ... }.default("something")...`

Jolanrensen: and use both String and lists in your tests :) so we cover them both

Nidhi-M21: Added tests for both lists and strings as requested. Also ran ktlintFormat.

Jolanrensen:
Thanks! I've run our CI and all seems well

PR #1777, opened 2026-03-30
Kotlin datetime format

### LINKED ISSUE #876: Migrate `java.time.format.DateTimeFormatter` to `kotlinx.datetime.format.DateTimeFormat`
Like the title says, as continuation of https://github.com/Kotlin/dataframe/pull/843 we should also update the parser options with the kotlin-first solution.

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/876

Changes parsing behavior for kotlin and java date-time, allowing both to be parsed in a more explicit and predictable way.

API changes:
- `ParserOptions` has changed and gained `DateTimeParserOptions`. `GlobalParserOptions` have also changed.
- `col.convertTo<>()`/`df.convert {}.to<>()` can now always take `ParserOptions`, and we gained a handful of new shortcuts and ones that had java arguments but returned kotlin types were deprecated.

See new kdocs and commit messages for specifics :)

Needs to be followed up by https://github.com/Kotlin/dataframe/issues/1798


Jolanrensen:
> Uhm okay, so the basic new logic is as follows:
> 
> We've got three "modes": `JAVA_TIME` (only parse to java-time types), `KOTLIN_DATETIME` (only parse to kotlin(x)-datetime types), `null`, just try them all, but prefer kotlin types.
> 
> You can add formats (by type) to `customGlobalDateTimeFormats` or `customGlobalJavaFormatters`. If they are detected, they will be attempted first when parsing.
> 
> After both those have been checked, only then do we check the `defaultDateTimeFormats` and `defaultJavaFormatters` when parsing.
> 
> (Locale will now always be taken into account when parsing to java types)
> 
> If the user specifies their own custom format(s) or patterns using `ParserOptions`, all default and global formats will be ignored and only those from the provided `ParserOptions` will be used. This automatically limits it to either `KOTLIN_DATETIME` or `JAVA_TIME` mode.
> 
> Conversions from String to any DateTime type will be Kotlin only for now, allowing you to specify a `DateTimeFormat` with the correct type (or a pattern as shorcut).
> 
> 
> TODO: Find a better way to instantiate a correct `ParserOptions`. I now have about a dozen constructors which is hard to maintain or work with. maybe a subclass? `ParserOptions(dateOptions = KotlinDateOptions())` or `JavaDateOptions()` maybe?

Edit: slightly outdated

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
```diff
@@ -360,11 +374,16 @@ public class Convert<T, out C>(
...
     public fun <R> cast(): Convert<T, R> = Convert(df, columns as ColumnsSelector<T, R>)
 
+    @Deprecated(CONVERT_TO_WITHOUT_PARSER_OPTIONS, level = DeprecationLevel.HIDDEN)
+    @Refine
+    @Interpretable("To0")
+    public inline fun <reified D> to(): DataFrame<T> = to(typeOf<D>())
+
     /**
-     * Converts values in the columns previously selected with [convert] to the specified type [D],
+     * Converts the values values in the columns previously selected with [convert] to the specified [type],
```
AndreiKingsley: values values 😄

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
```diff
@@ -360,11 +374,16 @@ public class Convert<T, out C>(
...
      * preserving their original names and positions within the [DataFrame].
      *
-     * The target type is provided as a reified type argument.
+     * The target type is provided as reified argument.
```
AndreiKingsley: Write here please some info about `ParserOptions` (for `String`/`String?` columns you can specify...)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
```diff
@@ -376,14 +395,27 @@ public class Convert<T, out C>(
...
+     * @return A new [DataFrame] with the values converted to [type].
      */
+    @Suppress("UNCHECKED_CAST")
     @Refine
     @Interpretable("To0")
-    public inline fun <reified D> to(): DataFrame<T> = to(typeOf<D>())
+    public inline fun <reified D> to(parserOptions: ParserOptions? = null): DataFrame<T> =
+        when (parserOptions) {
+            null -> to(typeOf<D>())
+            else -> (this as Convert<T, String?>).to(typeOf<D>(), parserOptions)
```
AndreiKingsley: It this cast safe?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convert.kt
```diff
@@ -152,6 +159,12 @@ internal fun DataColumn<String?>.convertToDoubleImpl(
 internal fun AnyCol.convertToTypeImpl(to: KType, parserOptions: ParserOptions?): AnyCol {
     val from = type
 
+    if (parserOptions != null && from.withNullability(false) != typeOf<String>()) {
```
Jolanrensen: I probably should remove this check... It would make:
```kt
df.convert { javaDateCol and stringCol }.to<LocalDate>(parserOptions)
```
crash

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
```diff
@@ -411,6 +443,30 @@ public class Convert<T, out C>(
  */
 public fun <T> Convert<T, *>.to(type: KType): DataFrame<T> = asColumn { it.convertTo(type) }
 
+/**
+ * Converts the [String] values in the columns previously selected with [convert] to the specified [type],
+ * preserving their original names and positions within the [DataFrame].
+ *
+ * The target type is provided as a [KType].
+ * For the full list of supported types, see [SupportedTypes][ConvertDocs.SupportedTypes].
```
AndreiKingsley: same

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/snippets.kt
```diff
@@ -40,3 +42,13 @@ internal typealias AutoRenamingLink = Nothing
...
  */
 internal typealias ColumnPathCreationSnippet = Nothing
+
+/**
+ * kotlinx-datetime [lacks localization support](https://github.com/Kotlin/kotlinx-datetime/discussions/253).
+ *
+ * If you need to provide a custom [java.util.Locale], we recommend parsing
+ * to a [java.time]-based class first before converting it to [kotlinx.datetime].
+ *
+ * See also: [ParserOptions], [DataFrame.parser.dateTimeLibrary][GlobalParserOptions.dateTimeLibrary]
```
AndreiKingsley, line 52: missing `,`?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/snippets.kt
```diff
@@ -40,3 +42,13 @@ internal typealias AutoRenamingLink = Nothing
  * does not correspond to an existing column or column group — all missing parts will be created automatically.
  */
 internal typealias ColumnPathCreationSnippet = Nothing
+
+/**
+ * kotlinx-datetime [lacks localization support](https://github.com/Kotlin/kotlinx-datetime/discussions/253).
+ *
+ * If you need to provide a custom [java.util.Locale], we recommend parsing
+ * to a [java.time]-based class first before converting it to [kotlinx.datetime].
```
AndreiKingsley: "..by providing [ParserOptions]" may be?

zaleslaw (approved):
I will approve, but want to share my impression from this API design approach

I recognized, that it will be great to have in such places something  like

```

df.parse {
    locale = Locale.CHINA

    dateTime {
        javaFormatter(DateTimeFormatter.ISO_WEEK_DATE)
    }
}

or

df.parse {
    locale = Locale.CHINA

    dateTime {
        parser = kotlinTime(SomeKotlinDateTimeFormat)
    }
}

df.parse {
    locale = Locale.CHINA

    dateTime {
        parser = javaTime(DateTimeFormatter.ISO_WEEK_DATE)
    }
}

based on

fun javaTime(formatter: java.time.format.DateTimeFormatter): DateTimeParser<LocalDateTime>
fun kotlinTime(format: kotlinx.datetime.format.DateTimeFormat<LocalDateTime>): DateTimeParser<LocalDateTime>


DateTimeParserOptions.kotlin { 
    format = LocalDate.Format {
        monthNumber(padding = Padding.SPACE)
        char('/')
        day()
        char(' ')
        year()
    }
    strict = true
}

DateTimeParserOptions.java {
    formatter = DateTimeFormatter.ISO_WEEK_DATE
    locale = Locale.CHINA
}

```

But probably it's not possible for us for now, and could be added later or by user request

From other hand, the current approach give the end user ability to easier navigate and  autocomplete the code

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convert.kt
```diff
@@ -152,6 +159,12 @@ internal fun DataColumn<String?>.convertToDoubleImpl(
...
+        error(
```
AndreiKingsley: Would be nice to catch it using runtime types.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/parse.kt
```diff
@@ -147,8 +269,42 @@ internal object Parsers : GlobalParserOptions {
...
+
+        // since we don't know what type of date-time this pattern is meant for, let's add it to all types
+        if (formatType == null) {
+            for (type in supportedJavaFormatterTypes) {
+                customGlobalJavaFormatters[type]!! += formatter
+            }
+        } else {
+            require(formatType in supportedJavaFormatterTypes) {
+                // requires #962
+                "Format type $formatType is not supported for Java date-time parsing, we only support parsing to: ${supportedJavaFormatterTypes.toList()}"
```
AndreiKingsley: not supported **yet** then?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/LazyMap.kt
```diff
@@ -0,0 +1,52 @@
+package org.jetbrains.kotlinx.dataframe.impl
+
+import kotlin.collections.plusAssign
+
+/**
+ * Map that resolves values lazily.
+ */
+internal class LazyMap<K, out V>(val lazyMap: Map<K, Lazy<V>>) : Map<K, V> {
```
AndreiKingsley: `lazyMap` inside `LazyMap` looks weird.
 It's more like "mapOfLazyValues"

AndreiKingsley (changes requested):
Please add a short snippet about ParserOptions (what is it and when applicable) and it to convert/parse methods KDocs.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convert.kt
AndreiKingsley: Hmm that's interesting indeed. I suggested compile-time check, but indeed there are cases like this one.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
AndreiKingsley: Oh, and maybe use `cast` here?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/snippets.kt
Jolanrensen: well, or adjust global parser options, which is what the "see also" points to

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/parse.kt
Jolanrensen: sure :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/LazyMap.kt
Jolanrensen: fair

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Jolanrensen: actually, no it isn't :) Same reason as `df.convert { javaDateCol and stringCol }.to<LocalDate>(parserOptions)`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convert.kt
Jolanrensen: actually for single columns we already do this sorta. `anyCol.convertTo<X>()` does not allow you to provide ParserOptions. It's just a problem if you mix types and you want to provide parseroptions. But well, in that case we can just ignore them

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Jolanrensen: I'll simplify and call convertToTypeImpl directly

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Jolanrensen, line 1: seems I forgot to update the Grammar

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Jolanrensen: KoDEx incoming ;P

Jolanrensen:
Well, it turns out the whole `DataColumn<String?>.convertToDoubleImpl()` function is not needed anymore! It existed purely to try both available double parsers. Well, with this PR we support multiple parsers out of the box :) 
I added it here: https://github.com/Kotlin/dataframe/pull/1777/changes/31d862b6496f29b5f4c6d1d82a475b0f36f16c9f with some small changes to the POSIX-fallback double parser, as that should not run if the user explicitly provides a Locale.
Turns out some tests now succeed where they failed before too :)

Jolanrensen:
Two more things:

`JavaInstants` can now also get custom DateTimeFormatters. Missed that before :) It's not possible for Kotlin as Instants are in the stdlib but DateTimeFormat is not.

Added KDocs to global parser options where it was missing and added an `availableParserTypes` getter :) should provide some better transparency

AndreiKingsley (approved):
Nice, thank you!

PR #1790, opened 2026-04-03
Added more targets for DisableInterpretation to help combine use of plugin together with library code

koperagen:
We need a way to make this code compile and work with plugin enabled in the project. Now it doesn't because declaring local classes is impossible in inline

```
inline fun <reified T> DataFrame<T>.doSomething(): DataFrame<T> { 
 ...
}
```

Let's add more targets to DisableInterpretation:
```
@DisableInterpretation
inline fun <reified T> DataFrame<T>.doSomething(): DataFrame<T> { 
 ...
}
```


Jolanrensen (approved):

Jolanrensen:
seems useful indeed :)

Jolanrensen:
related to https://code.jetbrains.team/p/kt/repositories/kotlin/reviews/26658/timeline

PR #1793, opened 2026-04-08
Removed redundant move overloads where insideGroup argument doesn't affect the result

koperagen:
vararg columns: String can only be top-level columns, so insideGroup true/false argument effectively always give the same result as overload of the function without it:

```
public fun <T> DataFrame<T>.moveToStart(vararg columns: String)
public fun <T> DataFrame<T>.moveToEnd(vararg columns: String)
public fun <T> DataFrame<T>.moveTo(newColumnIndex: Int, vararg columns: String)
```

follow up on https://github.com/Kotlin/dataframe/issues/1255


Jolanrensen:
No clue how I missed this checking https://github.com/Kotlin/dataframe/pull/1489 XD thanks!

Jolanrensen (approved):

PR #1807, opened 2026-04-15
Added reading ListVector of primitives and StructVector values

### LINKED ISSUE #1804: Regression in dataframe-arrow: `reading from org.apache.arrow.vector.complex.ListVector is not implemented`
Found a new bug running the spark-parquet-dataframe example https://github.com/Kotlin/dataframe/tree/master/examples/idea-examples/spark-parquet-dataframe

```
Exception in thread "main" kotlin.NotImplementedError: reading from org.apache.arrow.vector.complex.ListVector is not implemented
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingImplKt.readField(arrowReadingImpl.kt:377)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingImplKt.readField(arrowReadingImpl.kt:309)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingImplKt.readField(arrowReadingImpl.kt:387)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingImplKt.readArrowImpl(arrowReadingImpl.kt:431)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingImplKt.readArrowDatasetImpl(arrowReadingImpl.kt:482)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingKt.readParquet(arrowReading.kt:245)
	at org.jetbrains.kotlinx.dataframe.io.ArrowReadingKt.readParquet$default(arrowReading.kt:240)
	at org.jetbrains.kotlinx.dataframe.examples.spark.parquet.SparkParquetDataframeKt.main(SparkParquetDataframe.kt:229)
```

Since https://github.com/Kotlin/dataframe/pull/1803 has a release-version and dev-version I could conclude this problem did not exist in 1.0.0-Beta4, but was introduced in the meantime. I suspect in https://github.com/Kotlin/dataframe/pull/1621

koperagen:
Related to
https://github.com/Kotlin/dataframe/issues/1804 and https://github.com/Kotlin/dataframe/issues/271 (still no writing though)


dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
```diff
@@ -301,15 +308,62 @@ private fun List<Nothing?>.withTypeNullable(
     return this to nothingType(nullable)
 }
 
-private fun readField(vector: FieldVector, field: Field, nullability: NullabilityOptions): AnyBaseCol {
+private fun readField(
+    vector: FieldVector,
+    field: Field,
+    nullability: NullabilityOptions,
+    range1: IntRange? = null,
```
Jolanrensen: I believe this can simply be `range: IntRange = 0 until vector.valueCount,` right?

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
```diff
@@ -301,15 +308,62 @@ private fun List<Nothing?>.withTypeNullable(
...
-        val range = 0 until vector.valueCount
+        val range = range1 ?: (0 until vector.valueCount)
         if (vector is StructVector) {
             val columns = field.children.map { childField ->
-                readField(vector.getChild(childField.name), childField, nullability)
+                readField(vector.getChild(childField.name), childField, nullability, range)
             }
             return DataColumn.createColumnGroup(field.name, columns.toDataFrame())
         }
+        if (vector is ListVector) {
```
Jolanrensen, line 330: maybe these two cases should be part of the when-statement and/or extracted to a function. It becomes a bit hard to understand the structure of this function since it has grown so much

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowTypesMatching.kt
```diff
@@ -37,89 +38,70 @@ public fun AnyCol.toArrowField(mismatchSubscriber: (ConvertingMismatch) -> Unit
...
-        columnType.isSubtypeOf(typeOf<Int?>()) ->
-            Field(
-                column.name(),
-                FieldType(nullable, ArrowType.Int(32, true), null),
-                emptyList(),
-            )
+internal fun KType.toArrowField(name: String, mismatchSubscriber: (ConvertingMismatch) -> Unit): Field {
+    val nullable = isMarkedNullable
+    return when {
+        isSubtypeOf(typeOf<String?>()) ->
```
Jolanrensen, line 65: I'm not sure if it occurs, but this branch also enters when the type is `Nothing?`, like with a null-filled column. If this is intentionally turned into a String? column, we should probably add this branch explicitly

Jolanrensen:
`read parquet with two batches and nulls in second batch` seems to fail in debug mode

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowTypesMatching.kt
Jolanrensen: may potentially be related to the failing test? maybe? not sure ;P

Jolanrensen:
The issue from https://github.com/Kotlin/dataframe/issues/1804 seems to be fixed though :) the spark-parquet example now works well

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowTypesMatching.kt
koperagen: Added writing Nothing? columns as NullVectors

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
```diff
@@ -383,6 +403,89 @@ private fun readField(vector: FieldVector, field: Field, nullability: Nullabilit
...
+
+        override fun isNull(index: Int) = this@asAccessor.isNull(index)
+    }
+
+// Arrow in Java doesn't support allocating 64-bit-indexed vectors itself
+private fun LargeListVector.asAccessor() =
+    object : ListVectorAccessor {
+        override val dataVector: FieldVector get() = this@asAccessor.dataVector
+
+        override fun getElementStartIndex(index: Int) = Math.toIntExact(this@asAccessor.getElementStartIndex(index))
```
Jolanrensen, line 480: won't this cause issues if the index is larger than Int.MAX? (which is usually the reason someone chooses to use Longs as list indices instead of Ints)

Jolanrensen: Aah! but DataFrame uses `Int`, so this is not a problem. Still, if an overflow does happen, maybe provide an error explaining that this is a DataFrame limitation instead of just `ArithmeticException("integer overflow")`

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowTypesMatching.kt
```diff
@@ -27,8 +28,8 @@ public fun AnyCol.toArrowField(mismatchSubscriber: (ConvertingMismatch) -> Unit
     val column = this
     val columnType = column.type()
     val nullable = columnType.isMarkedNullable
-    return when {
-        column is ColumnGroup<*> -> {
+    return when (column) {
+        is ColumnGroup<*> -> {
```
Jolanrensen, line 32: isn't this duplicate logic with below?

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowTypesMatching.kt
```diff
@@ -37,89 +38,72 @@ public fun AnyCol.toArrowField(mismatchSubscriber: (ConvertingMismatch) -> Unit
...
-        columnType.isSubtypeOf(typeOf<Int?>()) ->
-            Field(
-                column.name(),
-                FieldType(nullable, ArrowType.Int(32, true), null),
-                emptyList(),
-            )
+internal fun KType.toArrowField(name: String, mismatchSubscriber: (ConvertingMismatch) -> Unit): Field {
+    val nullable = isMarkedNullable
+    return when {
+        this == nullableNothingType -> Field(name, FieldType(true, ArrowType.Null(), null), emptyList())
```
Jolanrensen, line 63: do you have a test with an empty column? because I have a hunch that will be of type `Nothing` and that will then still turn into a Utf8 column

Jolanrensen (approved):

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
koperagen: For now Arrow itself cannot allocate more than 32 bit array: JVM limitation according to javadocs. I'd investigate later

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowTypesMatching.kt
koperagen: Sorry, not related to reading ListVector - feel free to create an issue. This change here is just a small refactoring "extract method" to prepare for writing FrameColumns

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowTypesMatching.kt
koperagen: very similar, unified it

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
Jolanrensen: oh really :o, okay, then it's fine as is :)

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowTypesMatching.kt
Jolanrensen: https://github.com/Kotlin/dataframe/issues/1812

PR #1816, opened 2026-04-20
Refactoring the expressions-converter plugin for docs samples with easier API

koperagen:


plugins/expressions-converter/src/org/jetbrains/kotlinx/dataframe/ExplainerIrTransformer.kt
```diff
@@ -166,134 +164,61 @@ public class ExplainerIrTransformer(public val pluginContext: IrPluginContext) :
                 val transformedExtensionReceiver = expression.dispatchReceiver?.transform(this, data)
                 expression.dispatchReceiver = transformedExtensionReceiver
             }
-
-            return transformDataFrameExpression(expression, expression.symbol.owner.name, receiver = receiver, data)
+            val builder = pluginContext.irBuiltIns.createIrBuilder(expression.symbol)
+            return builder.transformDataFrameExpression(
+                expression,
+                expression.symbol.owner.name,
+                receiver = receiver,
```
Jolanrensen, line 171: this being the only argument with a name is a bit odd. I'd but names everywhere :)

Jolanrensen (approved):
looks simpler indeed :)

PR #1820, opened 2026-04-21
Fixed shapefile read from remote url

AndreiKingsley:


koperagen (approved):

dataframe-geo/src/main/kotlin/org/jetbrains/kotlinx/dataframe/geo/io/read.kt
```diff
@@ -135,6 +137,16 @@ private fun readShapeFileImpl(url: URL): GeoDataFrame<*> {
     }
 }
 
+private fun resolveRemoteShapefileUrl(url: URL): URL {
+    val urlPath = url.path.trimEnd('/')
+    return if (urlPath.endsWith(".shp") || urlPath.endsWith(".shp.gz")) {
+        url
+    } else {
+        val dirName = urlPath.substringAfterLast('/')
+        URI("${url.toString().trimEnd('/')}/$dirName.shp").toURL()
```
Jolanrensen, line 146: so... if the user supplies another url, something like "/path/to/shapeFile.zip", we will get "shapeFile.zip.shp" and break? I would write it a bit safer with a helpful exception message

AndreiKingsley: I just followed Nikita's logic; we should research usecases and improve it in the future

PR #1826, opened 2026-04-23
Removed remaining data modifiers from intermediate operations classes: Corr, SplitWithTransform for cleaner public API

koperagen:
Minor but still an API  change, we already did this for most other classes. 

copy, component* functions will be removed with this change:
Image

Image

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/split.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/split.kt
@@ -50,14 +50,17 @@ public class Split<T, C>(
     override fun toString(): String = "Split(df=$df, columns=$columns)"
 }
 
-public data class SplitWithTransform<T, C, R>(
+public class SplitWithTransform<T, C, R>(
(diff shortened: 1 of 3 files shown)


Jolanrensen (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/split.kt
```diff
@@ -50,14 +50,17 @@ public class Split<T, C>(
...
     internal val df: DataFrame<T>,
     internal val columns: ColumnsSelector<T, C?>,
     internal val inward: Boolean,
     internal val tartypeOf: KType,
     internal val default: R? = null,
     internal val transform: DataRow<T>.(C) -> Iterable<R>,
-)
+) {
+    override fun toString(): String =
+        "SplitWithTransform(df=$df, columns=$columns, inward=$inward, tartypeOf=$tartypeOf, default=$default, transform=$transform)"
```
Jolanrensen, line 62: these are appreciated :)

PR #1828, opened 2026-04-24
Added kdocs for DataFrame.concat, updated its signature for better plugin support, annotated DataFrameBuilder APIs for plugin support

koperagen:
dataFrameOf().randomDouble and others:
Since we're unlikely to deprecate these soon, i think why not support them in compiler plugin

For concat i think it'll be clearer both with and without plugin, and easier for implementation

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/concat.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/concat.kt
@@ -31,7 +31,24 @@ public fun <T> DataRow<T>.concat(vararg rows: DataRow<T>): DataFrame<T> = (listO
 
 public fun <T> DataFrame<T>.concat(vararg frames: DataFrame<T>): DataFrame<T> = concatImpl(listOf(this) + frames)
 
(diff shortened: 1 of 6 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/concat.kt
```diff
@@ -31,7 +31,24 @@ public fun <T> DataRow<T>.concat(vararg rows: DataRow<T>): DataFrame<T> = (listO
...
+/**
+ * Vertically concat rows from two DataFrames with different schemas.
+ * Use [Iterable.concat] to concatenate DataFrames with identical schemas.
+ *
+ * #### Schema unification
+ * If input DataFrame objects have different schemas, every column in the resulting DataFrame will get the lowest common type of the original columns with the same name.
+ *
+ * Rows of columns missing in either DataFrame will be filled with `null` values.
+ *
+ * Example: `[A: Int, B: String]` concat `[A: Double, C: Float]` yields `[A: Number, B: String?, C: Float?]`.
```
Jolanrensen, line 43: helpful!

Jolanrensen (approved):

PR #1832, opened 2026-04-29
moving titanic example to examples/projects and enabling compiler plugin

Jolanrensen:
Helps #1614 
moving titanic example to examples/projects and enabling compiler plugin


examples/projects/dev/titanic/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/titanic/ml/Passenger.kt
```diff
@@ -0,0 +1,41 @@
...
+import org.jetbrains.kotlinx.dataframe.api.convertTo
+import org.jetbrains.kotlinx.dataframe.api.generateInterfaces
+import org.jetbrains.kotlinx.dataframe.api.print
+import org.jetbrains.kotlinx.dataframe.io.readCsv
+
+/**
+ * This data schema is generated by reading the CSV once.
+ * We then call [generateInterfaces] on its schema to generate the Kotlin interface.
+ *
+ * Afterward, we can relatively safely [cast] or [convertTo] any search call to [Passenger] and
```
koperagen: "Any search call" - likely copy paste mistake

I'd expand, wdyt?
We're working with a single file and thus generated DataSchema accurately represents our DataFrame instance.
Working with multiple files of similar shape, generated schema provides initial estimation to help get started. It can happen that types of columns or set of columns slightly vary from file to file, in which case we recommend to fallback to manually editing DataSchema declarations

koperagen: another idea 
Complete DataSchema represents fail fast approach to ensure correctness. But DataSchema can declare only a subset of columns that you want to work with. Extra columns can be preserved (default for cast, convertTo) or dropped  via convertTo(...).

examples/projects/dev/titanic/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/titanic/ml/Passenger.kt
```diff
@@ -0,0 +1,41 @@
...
+ * the compiler plugin can handle the rest.
+ */
+fun main() {
```
zaleslaw, line 30: Seems like a new pattern to keep schema generation and declaration separately

zaleslaw: Instead of writing Gradle schema + running gradle task

zaleslaw (approved):

examples/projects/dev/titanic/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/titanic/ml/Passenger.kt
Jolanrensen: indeed :) also used it in the youtube example.

examples/projects/dev/titanic/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/titanic/ml/Passenger.kt
Jolanrensen: I think I can mention a small part of this but refer to the website for the full explanation

koperagen (approved):

Issue #987, opened 2024-12-04
Use Korro for SQL custom database example


Jolanrensen:
Follow up of https://github.com/Kotlin/dataframe/pull/986.

It has a lot of code in text that's not checked for correctness at the moment. If we cannot connect to the database, that's fine, we can still check the code compiles with Korro. This also makes it survive refactorings.

zaleslaw:
Need to clarify what are the exactly examples, which should be covered with Korro

let's say, that

1. https://kotlin.github.io/dataframe/sql.html - and all subpages
2. https://kotlin.github.io/dataframe/readsqldatabases.html - this page
3. https://kotlin.github.io/dataframe/readsqlfromcustomdatabase.html - but there is an example 
4. https://kotlinlang.org/docs/data-analysis-connect-to-db.html

Allex-Nik:
It's done except for https://kotlinlang.org/docs/data-analysis-connect-to-db.html, for which I've created a separate issue.

PR #1833, opened 2026-04-29
Apply `korro` to samples in data sources

Allex-Nik:
Adds usage of `korro` to the samples in the following webpages:
- https://kotlin.github.io/dataframe/data-sources.html - and all subpages
- https://kotlin.github.io/dataframe/readsqldatabases.html
- https://kotlin.github.io/dataframe/read.html

Related to #987, but does not change https://kotlinlang.org/docs/data-analysis-connect-to-db.html (as it will be done separately later). In addition, https://kotlin.github.io/dataframe/readsqlfromcustomdatabase.html was already done by @Jolanrensen


Allex-Nik:
Is the package correct for the samples? Should `ReadSQLDatabases` be in `api` instead of `io`?

docs/StardustDocs/topics/readSqlDatabases.md
```diff
@@ -337,32 +351,35 @@ val dataSource = HikariDataSource(config)
...
-```kotlin
-import java.sql.Connection
-import java.sql.DriverManager
+<!---FUN readSqlQueryExtension-->
 
+```kotlin
 val connection = DriverManager.getConnection("URL_TO_CONNECT_DATABASE")
 
-val df = connection.readDataFrame(dbConfig, "SELECT * FROM Users WHERE age > 35", 10)
+val df = connection.readDataFrame("SELECT * FROM Users WHERE age > 35")
```
Allex-Nik, line 365: It seemed to me that `dbConfig` and 10 shouldn't be here. Is it correct?

docs/StardustDocs/topics/dataSources/Parquet.md
```diff
@@ -12,6 +12,8 @@ Use Kotlin DataFrame to read Parquet datasets using Apache Arrow for fast, typed
 Kotlin DataFrame can read Parquet files through Apache Arrow’s Dataset API. Learn how and when to use it.
 </link-summary>
 
+<!---IMPORT org.jetbrains.kotlinx.dataframe.samples.io.Parquet-->
```
Allex-Nik, line 15: I added the import and applied `korro`, so some samples below were changed by the functions in `Parquet.kt` that had already existed

docs/StardustDocs/topics/dataSources/Parquet.md
```diff
@@ -61,21 +65,25 @@ public fun DataFrame.Companion.readParquet(
...
 // Read from file paths (as strings)
 val df = DataFrame.readParquet("data/sales.parquet")
 ```
 
+<!---END-->
+
 <!---FUN readParquetFilePath-->
 
 ```kotlin
-// Read from Path objects
```
Allex-Nik: Shall we adjust the functions in `Parquet.kt` to include the lines that were removed here and below in this file? Or do you think one `val df = ...` line is enough?

docs/StardustDocs/topics/readSqlDatabases.md
Jolanrensen: I think you're correct. That's why we use Korro, haha. I do think you removed the `limit = 10` argument which used to be there before

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Read.kt
```diff
@@ -1,38 +1,81 @@
...
 import org.jetbrains.kotlinx.dataframe.testJson
 import org.junit.Ignore
 import org.junit.Test
-import java.time.format.DateTimeFormatter
+import java.io.File
+import java.net.URI
 import java.util.Locale
 import kotlin.reflect.typeOf
 
 class Read {
```
Jolanrensen: wow, you also did these! Nice ;P

docs/StardustDocs/topics/readSqlDatabases.md
```diff
@@ -142,12 +143,14 @@ val df = DataFrame.readSqlTable(dbConfig, tableName, 100)
...
 
+<!---END-->
+
 You can find a full example project [here](https://github.com/zaleslaw/KotlinDataFrame-SQL-Examples/).
 
 ## Getting Started with Notebooks
 
-To use the latest version of the Kotlin DataFrame library 
-and a specific version of the JDBC driver for your database (MariaDB is used as an example below) in your Notebook,
+To use the latest version of the `Kotlin DataFrame` library
```
Jolanrensen: this needs no backticks, it's the name of the library https://kotlin.github.io/dataframe/spellingconventions.html

docs/StardustDocs/topics/dataSources/Parquet.md
```diff
@@ -61,21 +65,25 @@ public fun DataFrame.Companion.readParquet(
...
-val path = Paths.get("data/sales.parquet")
 val df = DataFrame.readParquet(path)
```
Jolanrensen, line 88: this is the only one without a comment now

docs/StardustDocs/topics/dataSources/Parquet.md
Jolanrensen: hmm, maybe it would help to add val path = Path("data/sales.parquet") and val file = File("data/sales.parquet")
to the examples, otherwise they're all the same except for the name of the argument and the comment

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/io/Parquet.kt
```diff
@@ -1,15 +1,65 @@
...
 import java.io.File
+import java.net.URL
+import java.nio.file.Path
 import java.nio.file.Paths
 
+private const val ARROW_PARQUET_DEFAULT_BATCH_SIZE: Long = 1000
+
 class Parquet {
+
+    interface ReadParquetOverloads {
```
Jolanrensen, line 22: the funny thing is that Korro does nothing useful for us here, because if we change the API, this example won't break XD, but okay, at least we can check the code compiles and is formatted neatly

Jolanrensen (approved):

docs/StardustDocs/topics/read.md
```diff
@@ -335,10 +347,14 @@ val df = DataFrame.readJson(file)
 
 To read a JSON file from a URL:
 
+<!---FUN readJsonFromUrl-->
+
 ```kotlin
 DataFrame.readJson("https://covid.ourworldindata.org/data/owid-covid-data.json")
```
koperagen: Change this to, for example,  https://raw.githubusercontent.com/Kotlin/dataframe/refs/heads/master/data/participants.json

docs/StardustDocs/topics/read.md
```diff
@@ -377,12 +393,16 @@ Let's take a look at the following JSON:
 ]
 ```
 
-We can read it from file:
+We can read it from the file:
+
+<!---FUN readJson-->
 
 ```kotlin
 val df = DataFrame.readJson(file)
```
koperagen: Maybe worth to turn code snipper with text above into 


val text = """[
    {
        "A": "1",
        "B": 1,
        "C": 1.0,
        "D": true
    },
    {
        "A": "2",
        "B": 2,
        "C": 1.1,
        "D": null
    },
    {
        "A": "3",
        "B": 3,
        "C": 1,
        "D": false
    },
    {
        "A": "4",
        "B": 4,
        "C": 1.3,
        "D": true
    }
]
"""
val df = DataFrame.readJsonStr(text)

koperagen:
Tables in `Manage type clashes﻿` and `Column type inference from CSV﻿` could be replaced with iframes

docs/StardustDocs/topics/readSqlDatabases.md
Allex-Nik: I removed it because this example is meant to rewrite with the extension function the original example with Connection (the `readSqlQueryConnection` sample above), and this original example didn't have this argument

As for `dbConfig`, it just didn't compile with it :)

docs/StardustDocs/topics/dataSources/Parquet.md
Allex-Nik: Ah, but these tests are not `@Ignore` now (so they really work as tests), and read from `resources`. So it seems we can either really create the `data` directory in `samples` and put `sales.parquet` there, or add `@Ignore` to these tests and remove all the `shouldBe` (or not include these `val path`, `val file` to the samples) :(
Maybe it's better to include in the sample just `val df = ...` and a comment (like "Read from File objects") because it's clear what `path` and `file` are, and it would save the test and let us avoid adding/moving data in the module

docs/StardustDocs/topics/readSqlDatabases.md
Allex-Nik: Got it, thank you!

docs/StardustDocs/topics/dataSources/Parquet.md
Allex-Nik: Added the comment

docs/StardustDocs/topics/read.md
Allex-Nik: Done

docs/StardustDocs/topics/read.md
Allex-Nik: Done

Allex-Nik:
> Tables in `Manage type clashes﻿` and `Column type inference from CSV﻿` could be replaced with iframes

I have done it in these sections (and in some others for consistency) and also added iframes for the results where it seemed useful

zaleslaw (approved):

docs/StardustDocs/topics/dataSources/Parquet.md
Jolanrensen: sure :) I wouldn't let these documentation tests read data if we don't have to indeed. But an `@Ignore` is usually fine, because the most important thins is that they compile correctly

koperagen (approved):

PR #1837, opened 2026-04-30
API change: Rename to

### LINKED ISSUE #555: Rename operation ~into~ to
`df.rename { col }.into("newName")` is not very common English. I always want to type `to` instead of `into`, so my suggestion is to rename this function.

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/555

I made `into` deprecated by warning indefinitely, so people that are familiar with `into` (or, AI) can still find it. And while I found no clashes with `to` for `Pair`-creation, I can see someone might want to avoid that confusion, so they can still use `into`.

The compiler plugin should also keep working, as I used the same interpretable annotation.

All usage, (dev) examples, and docs of rename `into` have been updated to `to`.


AndreiKingsley (approved):
Nice! Please also add it to the Migration Guide:
https://kotlin.github.io/dataframe/migrationto-1-0.html#removed-functions-and-classes

zaleslaw (approved):

PR #1838, opened 2026-05-01
fixed ExcludeFromSources in DslGrammarTemplateColumnsSelectionDsl.kt to exclude only the right elements

### LINKED ISSUE #1835: Documentation bug: `DslGrammarTemplateColumnsSelectionDsl` should not be excluded from sources
Doing so breaks a lot of KDocs:
Image

Probably a few sub-elements can be excluded, but no documentation element the user can target should be excluded.

Jolanrensen:
Fixes #1835

Also fixed `col` / `cols` appearing in rendered CS DSL grammar


Jolanrensen:
image
Now looks like this :)

koperagen (approved):

PR #1839, opened 2026-05-04
To type overloads with diff nullability

### LINKED ISSUE #1695: Remove redundant `toType` overloads with different nullability in convert API
So this causes an exception, as expected:
```kt
dataFrameOf("a" to columnOf(123, 322), "b" to columnOf(1.0, null)).convert { a and b }.to<Double>()
```
Method threw 'org.jetbrains.kotlinx.dataframe.exceptions.CellConversionException' exception.
I think this exception is the reason why two overloads for each operation were introduced:

```kt
public fun <T> Convert<T, Any>.toDouble(): DataFrame<T> = to<Double>()
public fun <T> Convert<T, Any?>.toDouble(): DataFrame<T> = to<Double?>()
```
But i want to argue it's actually redundant, `public fun <T> Convert<T, Any?>.toDouble(): DataFrame<T> = to<Double?>()` is enough:

If mixed nullability columns are selected, `to<Double?>()` is resolved
```kt
dataFrameOf("a" to columnOf(123, 322), "b" to columnOf(1.0, null)).convert { a and b }.toDouble()
```
And its result is accurate, a still has not-null Double type, just as originally it was Int.
```
a: Double
b: Double?
```
So in terms of type accuracy `public fun <T> Convert<T, Any>.toDouble(): DataFrame<T> = to<Double>()` is not needed.

Now let's return to the exception. If we deceive the compiler and resolve wrong overload, that normally wouldn't be resolved, then we get exception (that we might expect, since all our columns are Int! (not really))
```kt
dataFrameOf("a" to columnOf(123, 322), "b" to columnOf(1, null)).convert { (a and b).cast<Int>() }.toDouble()
```

Let's check other column selectors. all() -> ColumnSelector<Any?>, so again "nullable" overload is resolved.
```kt
dataFrameOf("a" to columnOf(123, 322), "b" to columnOf(1, null)).convert { all() }.toDouble()
```

Looks like we can remove ~27 overloads from convert.kt and everything will work same as before

### LINKED ISSUE #1903: Fix `NoClassDefFoundError` with `toUrl` and `toURL` called on `Convert`
Given a dataframe 
```kotlin
val df = dataFrameOf(
        "a" to columnOf("https://kotlinlang.org", "https://kotlinlang.org"),
        "b" to columnOf("https://github.com/Kotlin/dataframe", null)
    )
```
running the code 
```kotlin
val convertedToUrl = df.convert { a and b }.toUrl()
``` 
results in an error:

```
Exception in thread "main" java.lang.NoClassDefFoundError: org/jetbrains/kotlinx/dataframe/api/ConvertKt$toUrl$$inlined$asColumn$1
```

and running the code
```kotlin
@Suppress("DEPRECATION")
val convertedToUrl = df.convert { a and b }.toURL()
```
results in an error as well:

```
Exception in thread "main" java.lang.NoClassDefFoundError: org/jetbrains/kotlinx/dataframe/api/ConvertKt$toURL$$inlined$asColumn$1 (wrong name: org/jetbrains/kotlinx/dataframe/api/ConvertKt$toUrl$$inlined$asColumn$1)
```

The operating system is MacOS.

The problem does not occur if one of the functions is renamed.

Allex-Nik:
Fixes #1695
Fixes #1903


samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Convert.kt
```diff
@@ -0,0 +1,687 @@
...
+            "a" to columnOf(URL("https://example.com/photo1.png"), URL("https://example.com/photo1.png")),
+            "b" to columnOf(URL("https://example.com/photo3.jpg"), null),
+        )
+
+        val converted = df.convert { a and b }.toImg<_, URL?>()
+
+        assertNullabilityPreserved(converted, typeOf<IMG>())
+    }
+
+//    @Test
```
Allex-Nik: `toUrl`/`toURL` throw NoClassDefFoundError. Seems like they have the same signature which causes a conflict in case insensitive environment

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
```diff
@@ -1710,34 +1487,13 @@ public fun <T> Convert<T, Long>.toLocalDate(zone: TimeZone = defaultTimeZone): D
  * @param zone The [TimeZone] used to interpret the [Int] timestamp as a date. Defaults to the system current time zone.
  * @return A new [DataFrame] with the values converted to [LocalDate].
  */
-@JvmName("toLocalDateFromTInt")
+@JvmName("toLocalDateFromTIntNullable")
```
Allex-Nik: Should it be corrected like that?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
```diff
@@ -1004,28 +978,6 @@ public fun <T> Convert<T, URL>.toIFrame(
...
- *
- * @param width Optional width of the image in pixels.
- * @param height Optional height of the image in pixels.
- * @return A new [DataFrame] with the values converted to an [IMG].
- */
-@JvmName("toImgFromUrl")
-@Refine
-@Converter(IMG::class, nullable = false)
-@Interpretable("ToSpecificType")
-public fun <T, R : URL?> Convert<T, URL>.toImg(width: Int? = null, height: Int? = null): DataFrame<T> =
```
Allex-Nik, line 1024: Do we need the `R`?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
```diff
@@ -4253,16 +3304,6 @@ public fun <T> Convert<T, String?>.toLocalDate(pattern: String? = null, locale:
     toJavaLocalDate(formatter = pattern?.let { DateTimeFormatter.ofPattern(pattern) }, locale = locale)
         .convert(this.columns).toLocalDate()
 
-@Deprecated(
```
Allex-Nik, line 4254: I couldn't test these last 3 deprecated functions: when I use it in a test, the compile time schema of the converted dataframe contains `String` and `String?`. I didn't investigate it further because first I wanted to ask if we need to test them at all :)

koperagen: Hmm, they seem to miss compiler plugin annotations, not sure why

koperagen: it's ok to not test them

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Jolanrensen: I think R should have been the second argument in `Convert<>`

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/Convert.kt
Allex-Nik: Created #1903 about it

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
```diff
@@ -1004,28 +978,6 @@ public fun <T> Convert<T, URL>.toIFrame(
 public fun <T, R : URL?> Convert<T, URL?>.toImg(width: Int? = null, height: Int? = null): DataFrame<T> =
```
Allex-Nik: Should we make `R` the second argument in `Convert`?
```kotlin
public fun <T, R : URL?> Convert<T, R>.toImg...
```

Would it be wrong to remove `R` instead? 
```kotlin
public fun <T> Convert<T, URL?>.toImg...
```

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Allex-Nik: I continued the topic [here](https://github.com/Kotlin/dataframe/pull/1839#discussion_r3436241509) :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Jolanrensen: `R` is unused, no? It can probably be removed, it might be a mistake :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Jolanrensen: probably it should be removed indeed

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Jolanrensen: for correctness, yes, but that breaks binary compatibility :/ on the other hand, removing all these overloads also breaks it, so it's probably fine.

Since we're breaking binary compatibility anyway, maybe you could try removing as many `@JvmName`s as possible? Many won't be needed anymore now

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Allex-Nik: Yes, it is unused. I removed it

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Allex-Nik: Ah, I see. Now, after removal of some overloads, we can delete some `@JvmName`s without getting name clashes. I have done it :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
```diff
@@ -1051,7 +1001,6 @@ public fun DataColumn<String?>.convertToURL(): DataColumn<URL?> = convertToUrl()
 public fun DataColumn<String?>.convertToUrl(): DataColumn<URL?> = map { it?.let { URI(it).toURL() } }
 
 @Deprecated(TO_URL, ReplaceWith(TO_URL_REPLACE), DeprecationLevel.ERROR)
-@JvmName("toURLFromStringNullable")
 @Refine
```
Jolanrensen: Does this remaining toURL() function not cause issues on macos?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Jolanrensen: nice :) so this one can then also be reverted or removed right?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
```diff
@@ -2003,7 +1686,7 @@ public fun DataColumn<DateTimeComponents>.convertToLocalTime(): DataColumn<Local
 public fun DataColumn<DateTimeComponents?>.convertToLocalTime(): DataColumn<LocalTime?> = convertTo<LocalTime?>()
 
 /**
- * Converts values in the [Long] columns previously selected with [convert] to the [LocalDate],
+ * Converts values in the [Long] columns previously selected with [convert] to the [LocalTime],
```
Jolanrensen: in all these kdocs the final `the` in this line should be removed. Maybe out of scope for this issue, but we should probably do that at some point

Jolanrensen (approved):
nice cleanup :D

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Allex-Nik: There are 
```kotlin
public fun <T> Convert<T, Long?>.toLocalDate(zone: TimeZone = defaultTimeZone): DataFrame<T>
```
and
```kotlin
public fun <T> Convert<T, Int?>.toLocalDate(zone: TimeZone = defaultTimeZone): DataFrame<T>
```
Both of them have a `@JvmName`. I can delete a `@JvmName` in any one of them. What stopped me from doing that is that their JVM names would become asymmetric: now they are "toLocalDateFromTLongNullable" and "toLocalDateFromTIntNullable", but after the deletion one of them would become "toLocalDate", and another one would remain as it is. It seems to me that it could make naming a bit less clear. 

And there are a few similar cases in this file. 

What do you think? Do you think we should delete one `@JvmName` in each such pair anyway?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Allex-Nik: It does, but I created a separate [issue](https://github.com/Kotlin/dataframe/issues/1903) for it, and it is in the 1.1 milestone now. Do you think we should remove the deprecated `toURL` in this PR?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Allex-Nik: Done

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Jolanrensen: Oh, then I mean remove the "Nullable" part :) since they all accept nullable arguments now. I think it only adds confusion if someone were to use it from another jvm language

Jolanrensen (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Allex-Nik: Done

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/convert.kt
Jolanrensen: If it's impossible to call `toUrl()` on MacOS, then yes, it's a critical bug IMO, so it should be fixed asap :)

Allex-Nik: I have removed the deprecated `toURL()` function, there is no conflict anymore :)

PR #1416, opened 2025-08-28
(internal MVP) Prototype new iteration of ImportDataSchema annotation

koperagen:
This PR enables basic idea of a new workflow with imported schemas:
1. KSP processor finds "import schema" declarations (now CLASSES, not KT FILES), matches declaration to reader, reads data, writes schemas to directory. Can be triggered as usual Gradle task whenever needed
2. Compiler plugin handles all codegen 

The goal for now is MVP for internal testing.

How this approach different:
1. Schema declaration is now class, not file annotation. More convenient syntax for declarations
2. Schemas are stored under version control system, more transparent compared to hidden generated code directory

What's new:
Approach to custom formats and generic schema preprocessing
We explore idea of service loader in KSP processor. Anyone can provide SchemaReader implementation in another module of their project and generate schemas for arbitrary data, for example 

KSP plugin now has two new parameters:
1. disables old annotation processing for DataSchema and ImportDataSchema so they don't conflict with the compiler plugin

2. output directory where json files will be generated, serves as input of compiler plugin

```
ksp {
  arg("dataframe.experimentalImportSchema", "true") 
  arg("dataframe.importedSchemasOutput", path)
}
```

Example setup:
```
plugins {
    kotlin("jvm") version "2.3.255-SNAPSHOT"
    kotlin("plugin.dataframe") version "2.3.255-SNAPSHOT"
    id("com.google.devtools.ksp") version "2.2.0-2.0.2"
}

repositories {
    mavenLocal()
    maven("https://packages.jetbrains.team/maven/p/kt/dev/")
    mavenCentral()
}

dependencies {
    val version = "1.0.0-dev"
    implementation("org.jetbrains.kotlinx:dataframe-core:$version")
    implementation("org.jetbrains.kotlinx:dataframe-json:$version")
    implementation("org.jetbrains.kotlinx:dataframe-csv:$version")
    ksp("org.jetbrains.kotlinx.dataframe:symbol-processor-all:$version")

    // Module with custom readers
    ksp(project(":reader"))
    implementation(project(":reader"))
    testImplementation(kotlin("test"))
}

tasks.test {
    useJUnitPlatform()
}

val schemasDir = layout.projectDirectory.dir("src/schemas")!!
ksp {
    arg("dataframe.importedSchemasOutput", schemasDir.toString())
    arg("dataframe.experimentalImportSchema", "true")
    arg("dataframe.resolutionDir", layout.projectDirectory.asFile.absolutePath)
}

kotlin {
    jvmToolchain(11)
[...]

PR #1843, opened 2026-05-04
Docs warnings update

Jolanrensen:
Updated deprecation message for the gradle/ksp plugin

Updated OpenAPI warnings


AndreiKingsley (approved):
Nice!
Should we mention somehow https://github.com/Kotlin/dataframe/pull/1416
and generally the fact that we still want to add this feature (generation of the initial schema)?

Jolanrensen:
@AndreiKingsley hmm maybe, but probably only if we can point to an issue for it, where people can drop ideas or feedback

Issue #898, opened 2024-10-01
Merge sample/api folders for Korro


Jolanrensen:
Code samples for the documentation website reside in [core/.../test/.../samples/api](./core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api),
and [tests/.../samples/api](./tests/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api)
and they are copied over to Markdown files in [docs/StardustDocs/topics](./docs/StardustDocs/topics)
by [Korro](https://github.com/devcrocod/korro).

What I do not know is why we have two separate folders for these samples. It's a bit confusing, so unless we have a reason not to, we should probably merge the two.

AndreiKingsley:
So. I tried different things. 
1) "tests" module can no be moved into the "core/test" - it depends on Kandy which requires jvm >=11.
2) "samples" from "core/test" are really hard to move to the "tests" module - there're a lot of internal staff.

So my solution is simple.
1) Keep "tests" module (maybe rename it to "samples") and use it for creating DF and Kandy plot samples for docs.
2) After migration to new HTML dataframes in docs, we can remove old "samples" tests (except for those used by Kodex in KDocs.).

Jolanrensen:
> "tests" module can no be moved into the "core/test" - it depends on Kandy which requires jvm >=11.

Agree, not only that, but it depends on other modules too.

> "samples" from "core/test" are really hard to move to the "tests" module - there're a lot of internal stuff.

Friend module? :))

But yes, sounds like a plan!

AndreiKingsley:
> Friend module? :))

I tried, but there're a lot of issues with cross-dependencies, so it would take too much time to resolve all of them.

AndreiKingsley:
I would reformulate what needs to be done within this issue as follows:

1) Remove from `:core` all code samples that are inserted into Writerside sources using Korro.
    * we can then make special samples for KDocs in future
2) Deprecate `DFExplainer` in favor of `SampleHelper` and rewrite all existing samples with SampleHelper.

`:samples` (former `:tests`) module now serves as unified WS samples source and kandy-plot/df-iframe samples handler.
`:core` tests should not contain WS samples (but might be for Kdocs)

@Jolanrensen do you agree?

Jolanrensen:
@AndreiKingsley yes, that sounds like a good plan! Hopefully using `SampleHelper` we can drop the need for internal access. This was one of the things holding us back from extracting the samples from :core

[...]

PR #1846, opened 2026-05-06
Migrated to Korro 0.2.2

### LINKED ISSUE #1814: Migrate korro to 0.2.1
Update korro version to [0.2.0](https://github.com/devcrocod/korro/releases/tag/0.2.0) and [migrate](https://github.com/devcrocod/korro/blob/master/MIGRATION.md).

AndreiKingsley:
Closes #1814.


build.gradle.kts
```diff
@@ -71,6 +71,14 @@ println("Current DataFrame version: $version")
 
 subprojects {
     this.version = rootProject.version
+
+    tasks.matching {
```
Jolanrensen: we should not use `subprojects {}` anymore. Instead, this logic should be in the relevant build-logic convention plugins

Jolanrensen: or only in the build files where korro is applied

Jolanrensen (changes requested):

build.gradle.kts
Jolanrensen: I mean, eventually we will probably have a korro convention plugin https://github.com/Kotlin/dataframe/issues/1661 where this logic could live

AndreiKingsley: @Jolanrensen we will use korro only in :samples after migration (#898), so there's no need in korro convention plugin.

Can we just keep it for now?

Jolanrensen: Hmm. I see you did write mustRunAfter logic in :core but not in :samples, which looks a bit odd. It's okay to do it without convention plugin, but then the logic should be in :samples, not in the root project

Jolanrensen (approved):

PR #1847, opened 2026-05-07
Added column kind and type checks in `explode` and `ungroup`

### LINKED ISSUE #295: Inconsistent behaviour for sending invalid input to `ungroup`
Example: I have two columns `x: ColumnGroup, y: String`

Observed behavior:

1. Call `dataframe.ungroup("x")`... x is ungrouped as expected.
2. Call `dataframe.ungroup("y")... nothing happens 
3. Call `dataframe.ungroup("z")... an `IllegalStateException` is thrown.

Especially 2. and 3. being different is a bit confusing. As a minimum, I would expect invalid input to be treated the same, ie. either invalid input is always ignored or some kind of exception is always thrown.

I don't know which behavior is the correct one, and that choice also depends on your vision of the API (although I would lean towards throwing).

I noticed that `explode` has the same odd behavior, so I assume this is a general problem in the API, but I haven't checked other functions.

AndreiKingsley:
Fixes #295.

Add column kind check in `ungroup` and `explode`.

Add  a `verify` flag in `explode` that allows to keep old behavior (`false` by default). It's used internally in `gather` implementation. Update KDocs.

`split` in fact don't need this verification — we can actually split values of **any** type by providing a custom splitter!

### DIFF
--- a/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
+++ b/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
@@ -3,6 +3,8 @@
 package org.jetbrains.kotlinx.dataframe.api
 
 import org.jetbrains.kotlinx.dataframe.AnyCol
+import org.jetbrains.kotlinx.dataframe.AnyFrame
+import org.jetbrains.kotlinx.dataframe.DataFrame
 import org.jetbrains.kotlinx.dataframe.columns.ColumnGroup
 import org.jetbrains.kotlinx.dataframe.columns.ColumnKind
 import org.jetbrains.kotlinx.dataframe.columns.FrameColumn
@@ -20,6 +22,7 @@ import kotlin.contracts.ExperimentalContracts
 import kotlin.contracts.contract
 import kotlin.reflect.KType
 import kotlin.reflect.full.isSubtypeOf
+import kotlin.reflect.full.withNullability
 import kotlin.reflect.typeOf
 
 public fun AnyCol.isColumnGroup(): Boolean {
@@ -31,6 +34,9 @@ public fun AnyCol.isFrameColumn(): Boolean {
     contract { returns(true) implies (this@isFrameColumn is FrameColumn<*>) }
     return kind() == ColumnKind.Frame
 }
+
+public fun AnyCol.isFrameColumnOrValueColumnOfDataFrame(): Boolean =
+    isFrameColumn() || (isValueColumn() && type().isSubtypeOf(typeOf<AnyFrame>().withNullability(true)))
 
 public fun AnyCol.isValueColumn(): Boolean {
     contract { returns(true) implies (this@isValueColumn is ValueColumn<*>) }
--- a/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
+++ b/core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
@@ -3,8 +3,6 @@
 package org.jetbrains.kotlinx.dataframe.api
 
 import org.jetbrains.kotlinx.dataframe.AnyCol
-import org.jetbrains.kotlinx.dataframe.AnyFrame
-import org.jetbrains.kotlinx.dataframe.DataFrame
 import org.jetbrains.kotlinx.dataframe.columns.ColumnGroup
 import org.jetbrains.kotlinx.dataframe.columns.ColumnKind
 import org.jetbrains.kotlinx.dataframe.columns.FrameColumn
@@ -22,7 +20,6 @@ import kotlin.contracts.ExperimentalContracts
 import kotlin.contracts.contract
 import kotlin.reflect.KType
 import kotlin.reflect.full.isSubtypeOf
-import kotlin.reflect.full.withNullability
 import kotlin.reflect.typeOf
 
 public fun AnyCol.isColumnGroup(): Boolean {
@@ -34,9 +31,6 @@ public fun AnyCol.isFrameColumn(): Boolean {
     contract { returns(true) implies (this@isFrameColumn is FrameColumn<*>) }
     return kind() == ColumnKind.Frame
 }
-
-public fun AnyCol.isFrameColumnOrValueColumnOfDataFrame(): Boolean =
-    isFrameColumn() || (isValueColumn() && type().isSubtypeOf(typeOf<AnyFrame>().withNullability(true)))
 
 public fun AnyCol.isValueColumn(): Boolean {
     contract { returns(true) implies (this@isValueColumn is ValueColumn<*>) }
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
@@ -73,6 +73,8 @@ public fun AnyCol.isPrimitiveOrMixedNumber(): Boolean = type().isPrimitiveOrMixe
 
 public fun AnyCol.isList(): Boolean = typeClass == List::class
 
+public fun AnyCol.isPair(): Boolean = typeClass == Pair::class
+
 public fun AnyCol.isString(): Boolean = typeClass == String::class
 
 /** @include [valuesAreComparable] */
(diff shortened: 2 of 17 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
```diff
@@ -32,6 +35,9 @@ public fun AnyCol.isFrameColumn(): Boolean {
     return kind() == ColumnKind.Frame
 }
 
+public fun AnyCol.isFrameColumnOrValueColumnOfDataFrame(): Boolean =
```
Jolanrensen: does this need to be a public function?

Jolanrensen: also this returns true for `DataColumn<Nothing?>`. You should not use subtype

Jolanrensen: and `typeOf<AnyFrame>().withNullability(true) == typeOf<AnyFrame?>()`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/explode.kt
```diff
@@ -67,18 +98,30 @@ internal typealias ExplodeDocs = Nothing
  * df.explode { colsOf<List<Double>>() }
  * ```
  *
- * @param dropEmpty If `true`, removes rows with empty lists or DataFrames.
- *                  If `false`, such rows will be exploded into `null` values.
+ * {@include [ExplodeDocs.DropEmptySnippet]}
```
Jolanrensen: no brackets needed, tags end at the next `@`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/explode.kt
```diff
@@ -67,18 +98,30 @@ internal typealias ExplodeDocs = Nothing
...
- * @return A new [DataFrame] with exploded columns.
+ * {@include [ExplodeDocs.DefaultExplodeColumnsSnippet]}
+ * {@include [ExplodeDocs.ReturnSnippet]}
+ * {@include [ExplodeDocs.ThrowsSnippet]}
  */
 @Refine
 @Interpretable("Explode0")
 public fun <T> DataFrame<T>.explode(
     dropEmpty: Boolean = true,
+    verify: Boolean = true,
```
Jolanrensen: the compiler plugin will break if `verify` is supplied. This will need to be fixed on the compiler plugin side. Maybe it's easier to make `gather` simply call `explodeImpl`, avoiding the verification check.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/explode.kt
```diff
@@ -138,8 +191,14 @@ public fun <T, C> DataFrame<T>.explode(vararg columns: KProperty<C>, dropEmpty:
  *
  * ### This `explode` overload
  */
-@ExcludeFromSources
-internal typealias ExplodeDataRowDocs = Nothing
+internal interface ExplodeDataRowDocs {
```
Jolanrensen, line 180: this whole interface can be excluded from sources

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/explode.kt
```diff
@@ -47,8 +47,39 @@ private val defaultExplodeColumns: ColumnsSelector<*, *> = {
  *
  * ### This `explode` overload
  */
-@ExcludeFromSources
-internal typealias ExplodeDocs = Nothing
+internal interface ExplodeDocs {
```
Jolanrensen, line 61: the whole interface can be excluded

Jolanrensen (changes requested):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
koperagen: i'm not sure that we should explode "data column of dataframe?". We don't treat such columns as FrameColumns for example in code generation

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/explode.kt
koperagen: ++
I think it should not be a public parameter

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
AndreiKingsley: We had a test for `split` with such column
https://github.com/Kotlin/dataframe/blob/aba878c03f19057cdfe359353eee169d7c512e0f/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/split.kt#L54
and `explode` implementation works correctly with such columns.
@Jolanrensen @koperagen should we ask to convert it to FrameColumn manually?
Then should we add `ValueColumn<AnyFrame>` -> `FrameColumn` predefined conversion?

Jolanrensen: @AndreiKingsley the split example is actually quite interesting. Column `g` is a `FrameColumn` up until `.withNull()`, which turns it into a `ValueColumn` (yes, `update` changes its kind :o), because a `FrameColumn` cannot have null values.

So, we have two options:
- keep it as is. Users might be oblivious to the fact there now suddenly is a `ValueColumn<DataFrame<*>?>` instead of a `FrameColumn`, so they might be surprised that the usual functions don't work anymore
- make it only take `FrameColumns`, will require users to `.fillNulls { colsOf<AnyFrame?>() }.with { DataFrame.empty() }` before they can use split or explode

AndreiKingsley:
@Jolanrensen ok, now `DataColumn<AnyFrame?>` should be converted to `FrameColumn` manually before `explode`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
```diff
@@ -3,6 +3,8 @@
 package org.jetbrains.kotlinx.dataframe.api
 
 import org.jetbrains.kotlinx.dataframe.AnyCol
+import org.jetbrains.kotlinx.dataframe.AnyFrame
+import org.jetbrains.kotlinx.dataframe.DataFrame
```
Jolanrensen: unused import

Jolanrensen (approved):

core/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/DataColumnType.kt
```diff
@@ -32,6 +35,9 @@ public fun AnyCol.isFrameColumn(): Boolean {
     return kind() == ColumnKind.Frame
 }
 
+public fun AnyCol.isFrameColumnOrValueColumnOfDataFrame(): Boolean =
```
Jolanrensen: these generated-sources are out of date. They don't need to be part of the PR, they will be auto-regenerated by gh actions

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/explode.kt
```diff
@@ -67,18 +88,26 @@ internal typealias ExplodeDocs = Nothing
...
  */
 @Refine
 @Interpretable("Explode0")
 public fun <T> DataFrame<T>.explode(
     dropEmpty: Boolean = true,
     selector: ColumnsSelector<T, *> = defaultExplodeColumns,
-): DataFrame<T> = explodeImpl(dropEmpty, selector)
+): DataFrame<T> {
+    getColumnsWithPaths(selector).forEach { col ->
+        require(col.isFrameColumn() || col.isList()) {
```
koperagen: Wdyt about some custom exceptions? 
it might be nice to be able to set "exception breakpoints" to see what goes wrong
ExplodeWrongColumnKindException(df, column, kind) : IllegalArgumentException

AndreiKingsley: Added!

koperagen (approved):

Issue #1855, opened 2026-05-11
New API function to convert `ValueColumn<DataFrame<*>?>` -> `FrameColumn`


Jolanrensen:
Idea from: https://github.com/Kotlin/dataframe/pull/1847#issuecomment-4406279928

As some functions only work with `FrameColumn` and `ValueColumn<DataFrame<*>>` should probably not be allowed (https://github.com/Kotlin/dataframe/issues/1852, https://github.com/Kotlin/dataframe/issues/1854#issuecomment-4420250555),
we might need some easy-to-discover functions to convert between them besides `fillNulls {}`.

- `Convert<T, AnyFrame?>.toFrameColumn(fillNullsWith: DataRow<T>.() -> AnyFrame = DataFrame.empty())`
- `convertToFrameColumn(fillNullsWith: DataRow<T>.() -> AnyFrame = DataFrame.empty())`
or maybe even more direct:
- `toFrameColumn(fillNullsWith: DataRow<T>.() -> AnyFrame = DataFrame.empty())`

something like that

Issue #1854, opened 2026-05-11
Compiler plugin sometimes renders `ValueColumn` as frame column


Jolanrensen:
In operations like `update` `.withNull()`
<img width="1442" height="403" alt="Image" src="https://github.com/user-attachments/assets/8321496a-f79d-4de8-8e57-b07f0c456f12" />

a `FrameColumn` can turn into a `ValueColumn<DataFrame<*>?>` (because nullable framecolumns are not allowed). This is not reflected in the schema produced by the compiler plugin, even though the column is correctly recognized as a nullable dataframe-value column:

<img width="1442" height="403" alt="Image" src="https://github.com/user-attachments/assets/1f1df6c8-964a-4586-97f1-71bf6c8a7c4d" />

A similar problem can occur for `ValueColumn<DataFrame<>>`, such as with `convert` `asColumn`, `DataColumn.createValueColumn(dataframes)`, or [`toDataFrame { preserve(DataFrame::class) }`](https://github.com/Kotlin/dataframe/issues/1852).

<img width="1736" height="387" alt="Image" src="https://github.com/user-attachments/assets/bfe935f3-1c1d-4e1f-ad9f-57e075f4e8c9" />

(Though, I might like the idea of blocking any creation of `ValueColumn<DataFrame<*>>`, pushing users to `FrameColumn`)

There may be more cases where this happens, let's find and gather them here

Jolanrensen:
I believe the best way forward is to block creation of `ValueColumn<DataFrame<*>>` and force creating of a `FrameColumn` instead.

That way

- `DataColumn<DataFrame<*>>` will always be a `FrameColumn` and should be expanded
- `DataColumn<DataFrame<*>?>` will always be a `ValueColumn` and should not be expanded

Relevant https://github.com/Kotlin/dataframe/issues/1852

PR #1856, opened 2026-05-11
Disallowing `ValueColumn<AnyFrame>` because FrameColumn should be used

### LINKED ISSUE #1852: `toDataFrame { preserve(DataFrame::class) }` can create `ValueColumn<DataFrame<*>>`
As seen here:
https://github.com/Kotlin/dataframe/blob/4e150057d438039a825f8655478cf5873d9648c5/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt#L165

`toDataFrame { preserve(DataFrame::class) }` can create a `ValueColumn<DataFrame<*>>`. Besides the obvious `DataColumn.createValueColumn()`, this is the only place in the API (I could find) where a `ValueColumn<DataFrame<*>>` (non-nullable) is created instead of a more useful `FrameColumn`.

Having to deal with this type makes it harder to reason about the possibilities for `explode`, `split`, and the compiler plugin and it could be a source of bugs.

I would probably go as far as to add a check to `ValueColumnImpl` to fail always 
```kt
if (type.isSubtypeOf(typeOf<AnyFrame>()) && type != nothingType)
```
and then in DEBUG mode:
```kt
if (values.isNotEmpty() && values.all { it is AnyFrame })
```

Jolanrensen:
Disallowing `ValueColumn<AnyFrame>` from being created because FrameColumn should be used instead.

Fixes https://github.com/Kotlin/dataframe/issues/1852
Unblocks https://github.com/Kotlin/dataframe/issues/1854
and may be followed up by https://github.com/Kotlin/dataframe/issues/1855

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
@@ -334,6 +334,11 @@ internal fun convertToDataFrame(
         val shouldCreateFrameCol = kClass == DataFrame::class && !nullable
         val shouldCreateColumnGroup = kClass == DataRow::class
 
+        if (shouldCreateFrameCol && shouldCreateValueCol) {
+            throw IllegalArgumentException(
+                "You cannot create a `ValueColumn<DataFrame<*>>`, a `FrameColumn` should be used instead. If you used `preserve(DataFrame::class)`, please remove it.",
+            )
+        }
         when {
             hasExceptions -> DataColumn.createByInference(it.columnName, values, nullable = nullable)
 
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnImpl.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/columns/ValueColumnImpl.kt
@@ -1,12 +1,17 @@
 package org.jetbrains.kotlinx.dataframe.impl.columns
 
+import org.jetbrains.kotlinx.dataframe.AnyFrame
 import org.jetbrains.kotlinx.dataframe.AnyRow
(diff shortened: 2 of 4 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/toDataFrame.kt
```diff
@@ -334,6 +334,11 @@ internal fun convertToDataFrame(
         val shouldCreateFrameCol = kClass == DataFrame::class && !nullable
         val shouldCreateColumnGroup = kClass == DataRow::class
 
+        if (shouldCreateFrameCol && shouldCreateValueCol) {
+            throw IllegalArgumentException(
```
koperagen, line 338: Could we add a "deprecated" overload of preserve?

Jolanrensen: not really, because if it instantiates a `ValueColumn` with dataframes, it will fail there. This is just a more helpful early error message

Jolanrensen: oh you mean `preserve(KClass<DataFrame>)`? and provide a Deprecation ERROR there instead of at runtime? 

Yes, probably! I'll have a look

Jolanrensen: ah, that's probably not very useful. `preserve` are all vararg overloads, so while you could make `preserve(DataFrame::class)` or `preserve(Obj::dataFrame)` throw an error, the moment you add 2 different types, it will be allowed again... And then there are variables where the type of KClass could have been erased, of course...
Maybe if this was a common use-case before, it may have been worth considering, but I think in this case a helpful exception is plenty.

AndreiKingsley (approved):

PR #1858, opened 2026-05-12
PoC:  DF without `SupportedDataFrameFormat`

Jolanrensen:
#450 @zaleslaw 

Simple proof-of-concept to see what DF without `DataFrame.read()` and `SupportedDataFrameFormat` would look like. We don't really need it after removing the KSP/Gradle. The only thing we would lose is the `DataFrame.read()` function.

Creating the PR, I did see a lot of examples where this function was used, so maybe entirely removing it is not the best approach.
`DataFrame.read()` may still have a place in the library to read data instance. It's common practice in other DataFrame-like libraries in other ecosystems. But I do believe we should lean more into the dynamically typed approach if we do want to keep it. So, accepting `Any`, accepting custom properties, etc.

If we do remove it, we should probably also refactor `SupportedCodeGenerationFormat`, because it's tightly interwoven between OpenAPI() and `importDataSchema()` in notebooks.


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -78,7 +79,7 @@ class SortDataColumn {
 
     @Test
     fun `sort by nested column`() {
-        val df = testResource("ds_salaries.csv").readDataFrame().cast<DsSalaries>()
+        val df = DataFrame.readCsv(testResource("ds_salaries.csv")).cast<DsSalaries>()
```
zaleslaw, line 82: Looks more better:)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/sortGroupedDataframe.kt
```diff
@@ -4,13 +4,14 @@ import io.kotest.matchers.shouldBe
...
 import org.jetbrains.kotlinx.dataframe.io.read
+import org.jetbrains.kotlinx.dataframe.io.readCsv
 import org.junit.Test
 
 class SortGroupedDataframeTests {
 
     @Test
     fun `Sorted grouped iris dataset`() {
-        val irisData = DataFrame.read("src/test/resources/irisDataset.csv")
+        val irisData = DataFrame.readCsv("src/test/resources/irisDataset.csv")
```
zaleslaw, line 14: readCsv is better to me rather than read(....csv)

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/Jdbc.kt
```diff
@@ -1,53 +0,0 @@
...
-import org.jetbrains.kotlinx.dataframe.DataFrame
-import org.jetbrains.kotlinx.dataframe.codeGen.AbstractDefaultReadMethod
-import org.jetbrains.kotlinx.dataframe.codeGen.Code
-import org.jetbrains.kotlinx.dataframe.codeGen.DefaultReadDfMethod
-import java.io.File
-import java.io.InputStream
-import java.nio.file.Path
-
-// TODO: https://github.com/Kotlin/dataframe/issues/450
-public class Jdbc :
```
zaleslaw, line 13: Make no sense for JDBC, it's readFROMFILE

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/sortGroupedDataframe.kt
Jolanrensen: it is safer, for sure, but I do like the idea of just pointing dataframe to a file and letting it "figure it out"

PR #1859, opened 2026-05-13
Updated data schema documentation

### LINKED ISSUE #309: `convertTo()` requires values for computed properties
When a class contains computed properties, `convertTo()` still expects values for those properties to be defined:

```
@DataSchema
data class Foo (val first : Int, val second : Int) {
    val total : Int get() = first.plus(second)
    val highest : Int get() = listOf<Int>(first, second).max()
}

val df : DataFrame<Foo> = dataFrameOf("first", "second") (
    1, 2,
    25, 75,
    500, 499
).convertTo<Foo>()
```

`java.lang.IllegalArgumentException: The following columns were not found in DataFrame: [highest, total], and their type was not nullable. Use `fill` to initialize these columns`

This affects post-processing after `.readCSV()`, making it tricky to properly create/convert to a DataFrame reflecting known classes. Instead it's much easier to read the CSV externally, convert rows to collection of classes, then use `.toDataFrame()`.

Trying to use `fill` results in Unresolved Reference and Type Mismatch (it's certainly possible I have the wrong syntax):
```
val df : DataFrame<Foo> = dataFrameOf("first", "second") (
    1, 2,
    25, 75,
    500, 499
).convertTo<Foo> {
    fill { total }.with { first.plus(second) }
}
```

AndreiKingsley:
Closes #309.

We now explicitly recommend against using “complex” classes for data schemas in the documentation.

I also think it's better use interface for describing data schemas, especially for beginners, so I put interfaces on the first place and updated example with them.


docs/StardustDocs/topics/schemas/DataSchemaGenerationMethods.md
```diff
@@ -143,60 +159,43 @@ in cases where the schema cannot be inferred automatically from the source.
...
+    val amount: Double,
+    val orderId: Int
+)
 
 @DataSchema
-interface _DataFrameType1 {
-    val orders: List<_DataFrameType11>
-    val user: kotlin.String
-}
+data class Customer(
```
koperagen, line 182: re-generate classes please, it'll have nested structure and domain name Orders instead of Customer1

koperagen:
Btw, as for closing #309:

`val DataRow<DataSchemaType>.myComputedProperty get() = ...`

Can be used on type directly:

```
DataFrame.readCsv().cast<DataSchemaType().aggregate { 
  last().myComputedProperty into "last"
  first().myComputedProperty into "first"
}
```

Keeping in mind that compiler plugin operations change initial type

```
DataFrame.readCsv().cast<DataSchemaType()
    .add("example") { 1 }.filter { myComputedProperty } // unresolved because of add 
```

Workaround, use cast in DataRow context:
`df.filter { cast<DataSchemaType>().myComputedProperty }`

AndreiKingsley:
@koperagen should we add this trick to the documentation?

docs/StardustDocs/topics/Home.topic
```diff
@@ -19,16 +19,16 @@
 
         <primary>
             <title>First steps</title>
-            <a href="SetupKotlinNotebook.md"/>
+            <a href="SetupGradle.md"/>
```
Jolanrensen, line 22: hmm maybe we should keep the setup kotlin notebook for a while but just move it downward

docs/StardustDocs/topics/extensionPropertiesApi.md
```diff
@@ -119,24 +119,27 @@ You can do it quickly with [`generate..()` methods](DataSchemaGenerationMethods.
...
+}
+```
 ```
 
 Read the [`DataFrame`](DataFrame.md) from the CSV file and specify the schema with 
 [`.convertTo()`](convertTo.md) or [`cast()`](cast.md):
 
 ```kotlin
-val df = DataFrame.readCsv("example.csv").convertTo<Person>()
+val df = DataFrame.readCsv("example.csv").cast<Person>()
```
Jolanrensen, line 149: this is an interesting example... CSV is inherently flat, yet we have a nested type here XD This can only occur if there's json inside the csv, which is not that common

Jolanrensen: Maybe specify that it's not "just a CSV file", but that it contains a JSON column

docs/StardustDocs/topics/schemas/schemas.md
```diff
@@ -1,20 +1,36 @@
...
 A *schema* of a [`DataFrame`](DataFrame.md) is a mapping from column names to column types.  
-This data schema can be expressed as a Kotlin class or interface.  
-If the DataFrame is hierarchical — contains a [column group](DataColumn.md#columngroup) or a
+This data schema can be expressed as a Kotlin interface or data class by annotating it with `@DataSchema`.  
+If the dataframe is hierarchical — contains a [column group](DataColumn.md#columngroup) or a
 [column of dataframes](DataColumn.md#framecolumn) — the data schema reflects this structure,
 with a separate class representing the schema of each column group or nested `DataFrame`.
 
-For example, consider a simple hierarchical DataFrame from
+For example, consider a simple hierarchical dataframe from
```
Jolanrensen, line 30: yes, here we should also mention that this is a csv which has a json column, which is why it has a hierarchical structure

docs/StardustDocs/topics/schemas/schemas.md
```diff
@@ -83,8 +99,31 @@ df.filter { age >= 18 }
 
 See [](extensionPropertiesApi.md) for more information.
 
+## `@DataSchema` annotation
+
+`@DataSchema` is a Kotlin annotation that marks a data class or interface as a data schema.
+Compiler plugin generates [extension properties](extensionPropertiesApi.md) for the `DataFrame`
```
Jolanrensen: The compiler plugin (and link to it)

docs/StardustDocs/topics/schemas/schemas.md
```diff
@@ -83,8 +99,31 @@ df.filter { age >= 18 }
 
 See [](extensionPropertiesApi.md) for more information.
 
+## `@DataSchema` annotation
+
+`@DataSchema` is a Kotlin annotation that marks a data class or interface as a data schema.
+Compiler plugin generates [extension properties](extensionPropertiesApi.md) for the `DataFrame` 
+(or `DataRow`, `ColumnGroup`, etc.)
```
Jolanrensen: links?

docs/StardustDocs/topics/schemas/schemas.md
```diff
@@ -83,8 +99,31 @@ df.filter { age >= 18 }
...
-## Schema Retrieving
+Each property of an annotated class or interface corresponds to a column in the `DataFrame` 
+(or `DataRow`, `ColumnGroup`, etc.).
+The property name is the column name, and the property type is the column type.
+
+> While you can annotate any Kotlin class or object with a `@DataSchema`,
+> we highly recommend using it only on interfaces and data classes specially made
+> for representing the data schema of a `DataFrame`.
+> 
+> Use only trivial properties, avoiding computed, `lateinit` or delegated properties.
```
Jolanrensen: ", or"

Jolanrensen (approved):

koperagen:
> @koperagen should we add this trick to the documentation?

Yes please, somewhere about dataschemas. As an alternative to having computed properties right in the classes

docs/StardustDocs/topics/Home.topic
AndreiKingsley: We already have a Quickstart guide reference at the first place, so I believe it's ok.

docs/StardustDocs/topics/extensionPropertiesApi.md
```diff
@@ -229,3 +239,54 @@ interface Info {
...
+df.add("name") { "branchName" }
+    // unresolved because of `add`
+    .filter { it.profit > 0 }
+```
+
+However, you can work around this by casting back to the original schema:
+
+```kotlin
+df.add("name") { "branchName" }
+    // unresolved because of `add`
```
koperagen: comment no longer applicable here

Jolanrensen (approved):

koperagen (approved):

PR #1863, opened 2026-05-18
Inline `AnyFrame`, `AnyRow`, and `AnyBaseCol` type aliases in public API

### LINKED ISSUE #906: Inline some type aliases in public API
What is done here: https://github.com/Kotlin/dataframe/pull/888 but entire `kotlin/org/jetbrains/kotlinx/dataframe/api` package
`AnyFrame`, `AnyRow`, `AnyBaseCol`
Change should not include internal usages, and declarations itself should not be removed. Main idea is to make code completion in public API cleaner

Allex-Nik:
Fixes #906

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
@@ -1,11 +1,8 @@
 package org.jetbrains.kotlinx.dataframe.api
 
-import org.jetbrains.kotlinx.dataframe.AnyBaseCol
 import org.jetbrains.kotlinx.dataframe.AnyCol
 import org.jetbrains.kotlinx.dataframe.AnyColumnGroupAccessor
 import org.jetbrains.kotlinx.dataframe.AnyColumnReference
-import org.jetbrains.kotlinx.dataframe.AnyFrame
-import org.jetbrains.kotlinx.dataframe.AnyRow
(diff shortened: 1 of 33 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
```diff
@@ -1,11 +1,8 @@
 package org.jetbrains.kotlinx.dataframe.api
```
Allex-Nik, line 1: I've also encountered:

- AnyCol
- Predicate
- ColumnFilter
- AnyColumnReference
- AnyColumnGroupAccessor
- StringCol
- ColumnGroupReference
- ColumnGroupAccessor

Should we inline them/some of them as well?

`ColumnFilter`, `ColumnGroupAccessor`, and maybe `Predicate` are probably fine as they are, but it seems logical to me to inline `AnyCol`, `AnyColumnReference`, `AnyColumnGroupAccessor`, `StringCol`, `ColumnGroupReference` for the same reasons as `AnyFrame`, `AnyRow`, and `AnyBaseCol`.

koperagen (approved):
Thank you!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/add.kt
koperagen: Sure, feel free to create an issue to continue this task. I only choose these 3 because they're have most usages

PR #1869, opened 2026-05-20
Added Json instance overload to `readJson(inputStream, ...)`

### LINKED ISSUE #459: Feature: JSON parse options
Currently we just use Klaxon to read JSON using the default settings. We should allow users to change Klaxon settings.

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/459

In the same vein as `readCsv/readTsv/readDelim`, the overload that takes an `InputStream` now gains an underlying-library-specific parameter to customize how it parses from the stream. This is an advanced use-case, but it can prevent users having to first manually parse their JSON manually before passing it onto DataFrame.
Since it's an advanced feature, only adding it to the `InputStream` overload makes sense. If we ever swap out the underlying library, only this function would break.

New example:

```kt
DataFrame.readJson(
    stream = myJson.byteInputStream(),
    jsonInstance = Json {
        isLenient = true
        allowTrailingComma = true
        allowComments = true
    },
)
```


dataframe-json/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
```diff
@@ -274,17 +334,46 @@ public fun DataRow.Companion.readJson(
...
  * @return [DataFrame] from the given [stream].
  */
 @OptIn(ExperimentalSerializationApi::class)
+public fun DataFrame.Companion.readJson(
+    stream: InputStream,
+    header: List<String> = emptyList(),
+    keyValuePaths: List<JsonPath> = emptyList(),
+    typeClashTactic: TypeClashTactic = ARRAY_AND_VALUE_COLUMNS,
+    unifyNumbers: Boolean = true,
+    jsonInstance: Json = Json.Default,
```
koperagen: Hmm, not sure i really like this name. configuration maybe? Maybe reference from other libraries (kotlinx.serialization probably only ever uses it as receiver?)

koperagen (approved):
Nice!

dataframe-json/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/json.kt
Jolanrensen: I agree, I'll have a look. They named the class just "Json", which is a bit misleading. Especially because `JsonConfiguration` is an argument inside this `Json` class.

Jolanrensen: lol, it's indeed called "json" or it's used as a receiver almost everywhere XD. But, I see it being called "format" in like 5% of the cases. Let's go with that :)

Issue #1742, opened 2026-03-12
Add detailed error message for type mismatches in generated extension properties


AndreiKingsley:
Inspired by #964.

Sometimes, after schema/property generation, the generated properties have incorrect types.
This results in a `ClassCast` runtime exception, but only when the values are actually used.

As was discussed in #964, we need to improve error message. 
To do that we need to add `try/catch` section to the generated properties (both in notebooks and compiler plugin).

Also since we can't specify reason of incorrect type generation, we need to add a topic to the website and refer to it from 
in the exception message.

Proposed exception message:
```
Column 'column' has expected type 'CompilerType' but the value at index 'index' was actually of type 'ActualType'.
This may occur due to incorrect column generation, which can be caused by a number of different issues. 
See (link) for more information about the possible reason.
```

AndreiKingsley:
@koperagen what do you think?

koperagen:
> [@koperagen](https://github.com/koperagen) what do you think?

Good idea, let's do it. Need some time to implement because codegen in compiler plugin is a bit nuanced, don't have time right now for it. 
I think we should add a function that encapsulates this try/catch with error message somehow to avoid generating complex code. This function needs to be public API, any ideas for the name?
But most importantly all such mismatches should be fixed, even with dedicated documentation page and helpful exception it's very bothersome

AndreiKingsley:
We can't completely fix them, as in #964 - we can't catch all places with wrong generated types. 
But for sure, we can ask users to report cases when the wrong type is caused by some bug.

koperagen:
hmmm, maybe we can add some safeguards in the codegen

Jolanrensen:
Instead of casting inside the generated code, we could simply add it as an (extension) function on `DataRow` itself.
So instead of:
```kt
public val DataRow<ColumnDescription?>.freq: Int? @JvmName("NullableColumnDescription_freq") get() = this["freq"] as Int?
```
we could call:

```kt
public val DataRow<ColumnDescription?>.freq: Int? @JvmName("NullableColumnDescription_freq") get() = this.getAs<Int?>("freq")
```

which would be an inline function with try/catch and a proper error. wdyt?

AndreiKingsley:
Merged into Kotlin master!

PR #1870, opened 2026-05-22
Added troubleshooting guide for Data Schemas and Extension Properties

AndreiKingsley:
Helps #1742.

This is a new page for the referencing from EP errors.


docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,83 @@
+# Data Schemas and Extension Properties Troubleshooting
+
+Sometimes you can get an exception with a message containing
+
+```plain text
+..exception in generated DataFrame extension property..
+```
+
+This means there is a runtime error while accessing a [`DataFrame` extension property](extensionPropertiesApi.md),
+generated by the [Compiler Plugin](Compiler-Plugin.md) or in the [Kotlin Notebook](SetupKotlinNotebook.md).
```
Jolanrensen: *in Kotlin Notebook (the name of the product), or *in a (Kotlin) notebook (the concept)

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,83 @@
...
+
+Such errors are caused by generating extension properties for data schemas that are not compatible with the
```
Jolanrensen, line 14: maybe "mismatch" is a better fit here

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,83 @@
...
+```plain text
+..exception in generated DataFrame extension property..
+```
+
+This means there is a runtime error while accessing a [`DataFrame` extension property](extensionPropertiesApi.md),
+generated by the [Compiler Plugin](Compiler-Plugin.md) or in the [Kotlin Notebook](SetupKotlinNotebook.md).
+
+Such errors are caused by generating extension properties for data schemas that are not compatible with the 
+[`DataFrame`](DataFrame.md), [`DataRow`](DataRow.md), etc. 
+In most cases, the schema contains columns of the wrong names or types.
```
Jolanrensen: columns with an incorrect name or type

Jolanrensen: For example:

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,83 @@
...
+
+// Compiles correctly but fails on runtime
+df.filter { age > 20 }
+```
+
+## Possible reasons
+
+### Incompatible manually defined data schema
+
+If you define initad data schema manually,
```
Jolanrensen: the initial

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,83 @@
...
+make sure your data schema is compatible with the [`DataFrame`](DataFrame.md).
```
Jolanrensen: I think the concept "the dataframe" is better here

Jolanrensen: and you're describing just a solution here, not the reason it fails. I'd make a clear distinction for first the reason it fails and then second how to solve it

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,83 @@
...
+
+* Use [special methods](DataSchemaGenerationMethods.md) for generating a data schema code 
+instead of defining data schema manually.
+* Use [`.cast<Schema>()`](cast.md) with `verify=true` for verifying the `Schema` compatibility.
+
+### Incorrect column types after `DataFrame` creation
+
+Sometimes the runtime schema is wrong itself,
```
Jolanrensen: *schema itself is wrong

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,83 @@
...
+### Incorrect column types after `DataFrame` creation
+
+Sometimes the runtime schema is wrong itself, 
+because column types differ from the actual column value types.
+This can happen when reading a DataFrame from files or databases.
+
+> Such cases are most probably bugs! Please report them on [GitHub Issues](https://github.com/Kotlin/dataframe/issues).
+{style="warning"}
+
+Possible workarounds:
```
Jolanrensen: here it's better :) first you describe the problem/cause and then the workaround

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,83 @@
...
+
+* Specify the correct type using [`.replace {}`](replace.md) and `ValueColumn.changeType()`:
```
Jolanrensen: may be important to mention the two different ways a column can have a "type". There's the internal KType, used by runtime functions, and there's the compile-time type, visible in the IDE, which is also sometimes used in runtime when you refer to a column in an inline reified function, but usually it's visual only

Jolanrensen: oh and this workaround only works if you're working with a value column, obviously ;P

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+* Use [`.inferType { columns }`](inferType.md) to infer the correct types 
+for the selected columns from the actual values.
+**It can take a long time and use up a lot of resources for large dataframes!**
+
+#### Problems with type affinity in SQLite
+
+Because of [SQLite type affinity](https://sqlite.org/datatype3.html), 
+the column typed defined by JDBC may differ from the actual values in the column.
```
Jolanrensen: can you link to JDBC docs?

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,83 @@
...
+This problem often occurs when reading data from an SQLite database with column of custom types.
+
+You can provide types for such columns manually:
+
+```Kotlin
```
Jolanrensen: korro :)

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,83 @@
...
+This can happen when reading a DataFrame from files or databases.
+
+> Such cases are most probably bugs! Please report them on [GitHub Issues](https://github.com/Kotlin/dataframe/issues).
+{style="warning"}
+
+Possible workarounds:
+
+* Specify the correct type using [`.replace {}`](replace.md) and `ValueColumn.changeType()`:
+
+```kotlin
```
Jolanrensen: korro?

Jolanrensen (changes requested):

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
AndreiKingsley: We already introduced "compatible" term here:
https://kotlin.github.io/dataframe/schemas.html#dataschema-annotation

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,83 @@
...
+interface Schema {
+    val age: String
+}
+```
+
+```kotlin
+val df = dataFrameOf("age" to columnOf(17, 32, 26)).cast<Schema>()
+
+// Compiles correctly but fails on runtime
+df.filter { age > 20 }
```
koperagen: Now it won't compile
It should be the opposite
age: Int, but String values in runtime.

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
AndreiKingsley: Because user defined incorrect data schema, as in the example above. Isn't it a reason?

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
AndreiKingsley: Ok, but what to say exactly about that here?

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
AndreiKingsley: Ok, I will add this sample with korro!

AndreiKingsley: image
Oh, there's even a compile time check!

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,83 @@
...
+### Incompatible manually defined data schema
+
+If you define initad data schema manually,
+make sure your data schema is compatible with the [`DataFrame`](DataFrame.md).
+
+* Use [special methods](DataSchemaGenerationMethods.md) for generating a data schema code 
+instead of defining data schema manually.
+* Use [`.cast<Schema>()`](cast.md) with `verify=true` for verifying the `Schema` compatibility.
+
+### Incorrect column types after `DataFrame` creation
```
koperagen: ### Bug in the DataFrame reader
Internally a column was created with incorrect KType that doesn't represent actual runtime values. As a result, schema().print(), generateDataClasses, print(columnTypes = true) show misleading type.

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,83 @@
...
+
+* Specify the correct type using [`.replace {}`](replace.md) and `ValueColumn.changeType()`:
+
+```kotlin
+df.replace { wrongTypeCol }.with { it.asValueColumn().changeType(typeOf<ActualType>) }
+```
+
+* Use [`.inferType { columns }`](inferType.md) to infer the correct types 
+for the selected columns from the actual values.
+**It can take a long time and use up a lot of resources for large dataframes!**
```
koperagen: citation needed

AndreiKingsley: What do you mean?

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,83 @@
...
+
+Possible workarounds:
+
+* Specify the correct type using [`.replace {}`](replace.md) and `ValueColumn.changeType()`:
+
+```kotlin
+df.replace { wrongTypeCol }.with { it.asValueColumn().changeType(typeOf<ActualType>) }
+```
+
+* Use [`.inferType { columns }`](inferType.md) to infer the correct types
```
koperagen: Maybe worth to mention that it won't be able to infer generic arguments such as Map

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
koperagen: Neither will be enough to fix initial problem

1. Use one of methods to fix KType of affected column
2. Align property type of the DataSchema with now-correct type as shown by print(), generate* or

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,83 @@
...
+## Possible reasons
+
+### Incompatible manually defined data schema
+
+If you define initad data schema manually,
+make sure your data schema is compatible with the [`DataFrame`](DataFrame.md).
+
+* Use [special methods](DataSchemaGenerationMethods.md) for generating a data schema code 
+instead of defining data schema manually.
+* Use [`.cast<Schema>()`](cast.md) with `verify=true` for verifying the `Schema` compatibility.
```
koperagen, line 58: I'd put this on the first place.
Use cast(verify = true) to catch incompatibility early (actually we should rename cast(verify = false) to uncheckedCast and have cast(verify = true) by default)

> Use [special methods](DataSchemaGenerationMethods.md) for generating a data schema code

Might not be enough. I faced it a few times and solution was to manually edit the schema, add nullability, because sample of data is not representative.

koperagen:
I'd like to see screenshots of errors as they appear in the console + repros. Might be even useful to have in the guide to illustrate the problem

AndreiKingsley:
> I'd like to see screenshots of errors as they appear in the console + repros. Might be even useful to have in the guide to illustrate the problem

I will add them after adjusting exception and messages in EP.

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
AndreiKingsley: Can't add imports with korro :(

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
koperagen: Since we recommend it as a workaround, let's create an issue to see how is performance and see how it can be improved if needed

> **It can take a long time and use up a lot of resources for large dataframes!**

It can take a long time and use up a lot of resources for large dataframes, so we recommend to use it once to figure out actual type and change it to changeType for production code

Maybe like this?

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -20,49 +26,81 @@ interface Schema {
...
-make sure your data schema is compatible with the [`DataFrame`](DataFrame.md).
+If you define the initial data schema manually,
+make sure your data schema is compatible with the dataframe.
 
-* Use [special methods](DataSchemaGenerationMethods.md) for generating a data schema code 
-instead of defining data schema manually.
 * Use [`.cast<Schema>()`](cast.md) with `verify=true` for verifying the `Schema` compatibility.
+* Use [special methods](DataSchemaGenerationMethods.md) for generating a data schema code 
+instead of defining data schema manually. However, you may still need to edit a generated schema —
+adjust a type nullability or the whole type, since sometimes type can be inferred incorrect.
```
koperagen: incorrect if data sample (dataframe?) used for generation is not representative (of all similar data?) or if there's a bug in DataFrame

Not sure about whole sentence, but i'd like to use "representative" :) I think it's a nice concept to understand about our data schema generation

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
Jolanrensen: are the imports necessary?

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,121 @@
...
+
+   * Use [`.inferType { columns }`](inferType.md) to infer the correct types 
+   for the selected columns from the actual values. Doesn't work for generic types like `Map`.
+    > `inferType()` can take a long time and use up a lot of resources for large dataframes, so we recommend 
+    > using it once to figure out the actual type and change it to `changeType()` for production code.
+    {style="warning"}
+
+2) Change the compiler-time schema of dataframe
+
+You need to edit (you can view a correct one with [`.schema().print()`](schema.md)) or regenerate the correct data schema (with [generate..()](DataSchemaGenerationMethods.md) methods)
```
Jolanrensen: *the correct one

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
```diff
@@ -0,0 +1,121 @@
...
+instead of defining data schema manually. However, you may still need to edit a generated schema —
+adjust a type nullability or the whole type, since sometimes type can be inferred incorrect.
+
+### Bug in the DataFrame reader
+
+Sometimes a column is created internally with an incorrect `KType` that doesn't represent actual runtime values.
+As a result, [`schema().print()`](schema.md), 
+[`generateInterfaces`](DataSchemaGenerationMethods.md#generateinterfaces), 
+`print(columnTypes = true)` show misleading type.
+
```
Jolanrensen, line 70: or the compiler plugin shows an incorrect type (on hover), for that you need solution 2... 

Uhm.. maybe you could split the runtime/ktype mismatch and runtime/compile-time mismatch in a separate section. Because you're now describing the runtime/ktype issue and solution 2 solves it for the compiler plugin but not for ktype (if you use `cast`). It's all a bit mixed up.

Maybe it could be one section, but then you clearly need to state there are 2 places a column can have a type:
- KType in the column instance
- Compile-time type as specified by the user or the compiler plugin

and how to solve mismatches with runtime values in both cases

koperagen: My understanding of the structure, we have 1 problem (exception in column accessor generated by plugin) and 2 reasons for that:
1. DataSchema is incorrect
  - in which case solution is to align with runtime schema
2. There's a bug in dataframe with KType. Now advice to align is not applicable because what print() or schema() shows is also wrong
 - in which case we further breakdown solution in two steps
  1. fix runtime first by changeType or inferType
  2. and now fix original problem (exception is column accessor generated by plugin) - align with runtime

docs/StardustDocs/topics/schemas/Data-Schemas-And-Extension-Properties-Troubleshooting.md
AndreiKingsley: I believe it's a good practice, isn't it? 
`Sqlite` is non-trivial import.

Jolanrensen: hmm maybe yes, but on the other hand, the IDE is usually clever enough to find the right one. I believe using Korro is more important than being able to show imports. Maybe Pasha can add a feature to show imports at some point?

Jolanrensen: if anything, you can allways add `// org.jetbrains.kotlinx.dataframe.io.db.Sqlite` :)

Issue #1872, opened 2026-05-25
Add a custom type for extension properties exception


AndreiKingsley:
Nice to have it for implementing/improvement extension properties exceptions (#1742).

zaleslaw:
Could you please clarify the description/the header, add motivation (why?), and the idea of naming/draft, if you have something

Issue #1871, opened 2026-05-25
Add / improve custom exceptions


AndreiKingsley:
Now we use mostly `error`, `require` or throw common exceptions. We should introduce custom exceptions.

We have some:
https://github.com/Kotlin/dataframe/tree/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/exceptions
but most cases are not covered / not used. 
We have  https://github.com/Kotlin/dataframe/blob/master/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/exceptions/ColumnNotFoundException.kt
that are not used anywhere!

It can also significantly improve DF debugging — exception should contain df/row/column field for that purpose.

Jolanrensen:
Even better would be Rich Errors https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0441-rich-errors-motivation.md once they arrive. But until then, yes, DataFrameException all the way ;P

zaleslaw:
Please collect more cases. I believe it's an important topic, on the other hand, Kotlin leads us to avoiding the custom Exceptions which differ only in messages

koperagen:
As a part of this we could proofread exceptions, make rendering of types / classes consistently formatted:

<img width="2126" height="294" alt="Image" src="https://github.com/user-attachments/assets/2f540d18-ea4d-4f64-8f6a-c46b77ea58d5" />

PR #1873, opened 2026-05-25
Added try-catch for column access with a detailed error messages

AndreiKingsley:
Library codegen implementation of #1742.

I want to add the same codegen to Compiler Plugin.

Can be improved with #1871 and #1872, but for now I added simple different exception handling `error`. 

Requires new doc page from #1870.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
@@ -299,7 +299,7 @@ internal open class ExtensionsCodeGeneratorImpl(private val typeRendering: TypeR
         propertyType: String,
         getter: String,
         visibility: String,
-        columnName: String
+        columnName: String,
     ): String {
         // jvm name is required to prevent signature clash like this:
         // val DataRow<Type>.name: String
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
@@ -277,19 +277,29 @@ internal open class ExtensionsCodeGeneratorImpl(private val typeRendering: TypeR
 
     private val troubleshootingLink = ""
 
-    // TODO: improve with custom errors: #1871, #1872
-    private fun generateTryCatchColumnAccess(columnAccessCode: String, columnName: String): String =
-        """
-        try {
-            $columnAccessCode
-        } catch (e: Exception) {
-             when {
-                e is IllegalArgumentException -> error(message = "Column not found exception in the generated DataFrame extension property '$columnName': " + e.getLocalizedMessage() + ". See $troubleshootingLink for more information.")
-                e is ClassCastException -> error(message = "Incorrect column type exception in generated DataFrame extension property '$columnName': " + e.getLocalizedMessage() + " See $troubleshootingLink for more information.")
-                else -> error(message = "Unexpected exception in generated DataFrame extension property '$columnName'. Please report it to https://github.com/Kotlin/dataframe/issues." + "Exception message: " + e.toString())
+    /**
+     * Generate a try-catch block for column accessor with custom error messages.
+     *
+     * Catch any exception. Returns an `error()` with a special message. Can be improved with custom exception (#1872).
+     *
+     * 1) IllegalArgumentException -> most probably column not found. Should be replaced with a custom exception (#1871)
+     * 2) ClassCastException -> most probably incorrect column type. Should be replaced with a custom exception (#1871)
+     * 3) else -> unexpected exception.
+     */
+    private fun generateTryCatchColumnAccess(columnAccessCode: String, columnName: String): String {
+        val renderedColumnName = renderStringLiteral(columnName)
+        return """
+            try {
+                $columnAccessCode
+            } catch (e: Exception) {
+                 when (e) {
+                    is IllegalArgumentException -> error(message = "Column not found exception in the generated DataFrame extension property '$renderedColumnName': " + e.getLocalizedMessage() + ". See $troubleshootingLink for more information.")
+                    is ClassCastException -> error(message = "Incorrect column type exception in generated DataFrame extension property '$renderedColumnName': " + e.getLocalizedMessage() + " See $troubleshootingLink for more information.")
+                    else -> error(message = "Unexpected exception in generated DataFrame extension property '$renderedColumnName'. Please report it to https://github.com/Kotlin/dataframe/issues." + "Exception message: " + e.toString())
+                }
             }
-        }
-        """.trimIndent()
+            """.trimIndent()
+    }
 
     private fun generatePropertyCode(
         marker: IsolatedMarker,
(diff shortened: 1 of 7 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
```diff
@@ -274,6 +275,32 @@ internal open class ExtensionsCodeGeneratorImpl(private val typeRendering: TypeR
...
+     * 3) else -> unexpected exception.
+     */
+    private fun generateTryCatchColumnAccess(columnAccessCode: String, columnName: String): String {
+        val renderedColumnName = renderStringLiteral(columnName)
+        return """
+            try {
+                $columnAccessCode
+            } catch (e: Exception) {
+                 when (e) {
+                    is IllegalArgumentException -> error(message = "Column not found exception in the generated DataFrame extension property '$renderedColumnName': " + e.getLocalizedMessage() + ". See $troubleshootingLink for more information.")
```
Jolanrensen: these may require imports, you could put the FQ name here

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
```diff
@@ -274,6 +275,32 @@ internal open class ExtensionsCodeGeneratorImpl(private val typeRendering: TypeR
 
     private fun String.removeQuotes() = this.removeSurrounding("`")
 
+    private val troubleshootingLink = ""
+
+    /**
+     * Generate a try-catch block for column accessor with custom error messages.
+     *
+     * Catch any exception. Returns an `error()` with a special message. Can be improved with custom exception (#1872).
```
Jolanrensen: an *IllegalStateException, `error()` is just the function that triggers it

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
```diff
@@ -274,6 +275,32 @@ internal open class ExtensionsCodeGeneratorImpl(private val typeRendering: TypeR
...
+    private fun generateTryCatchColumnAccess(columnAccessCode: String, columnName: String): String {
+        val renderedColumnName = renderStringLiteral(columnName)
+        return """
+            try {
+                $columnAccessCode
+            } catch (e: Exception) {
+                 when (e) {
+                    is IllegalArgumentException -> error(message = "Column not found exception in the generated DataFrame extension property '$renderedColumnName': " + e.getLocalizedMessage() + ". See $troubleshootingLink for more information.")
+                    is ClassCastException -> error(message = "Incorrect column type exception in generated DataFrame extension property '$renderedColumnName': " + e.getLocalizedMessage() + " See $troubleshootingLink for more information.")
+                    else -> error(message = "Unexpected exception in generated DataFrame extension property '$renderedColumnName'. Please report it to https://github.com/Kotlin/dataframe/issues." + "Exception message: " + e.toString())
```
Jolanrensen: it may be better to manually throw the IllegalStateException, because then you can add the cause there

Jolanrensen (changes requested):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
```diff
@@ -274,6 +275,34 @@ internal open class ExtensionsCodeGeneratorImpl(private val typeRendering: TypeR
...
+     * 3) else -> unexpected exception.
+     */
+    private fun generateTryCatchColumnAccess(columnAccessCode: String, columnName: String): String {
+        val renderedColumnName = renderStringLiteral(columnName)
+        return """
+            try {
+                $columnAccessCode
+            } catch (e: kotlin.Exception) {
+                 val msg = when (e) {
+                    is kotlin.IllegalArgumentException -> "Column not found exception in the generated DataFrame extension property '$renderedColumnName': " + e.getLocalizedMessage() + ". See $troubleshootingLink for more information."
```
Jolanrensen: Is there a reason you mix string templates and `+` here, actually? Just curious :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
```diff
@@ -274,6 +275,34 @@ internal open class ExtensionsCodeGeneratorImpl(private val typeRendering: TypeR
...
+                    is kotlin.ClassCastException -> "Incorrect column type exception in generated DataFrame extension property '$renderedColumnName': " + e.getLocalizedMessage() + " See $troubleshootingLink for more information."
+                    else -> "Unexpected exception in generated DataFrame extension property '$renderedColumnName'. Please report it to https://github.com/Kotlin/dataframe/issues." + "Exception message: " + e.toString()
```
Jolanrensen: The `+` here is unnecessary

Jolanrensen:
you've got some failing tests :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
AndreiKingsley: For me just better readability. But let's be consistent and idiomatic, ok.

Jolanrensen: you can also use a string builder if that looks more readable :)

AndreiKingsley:
@Jolanrensen this PR is not supposed to me merged (most probably) — @koperagen doesn't want to change EP logic on library side, but I need you review on it because it's much easier to review it here rather then in Compiler Plugin.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
AndreiKingsley: I rewrote with better templates!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/CodeGeneratorImpl.kt
```diff
@@ -274,6 +274,39 @@ internal open class ExtensionsCodeGeneratorImpl(private val typeRendering: TypeR
...
+     * Catch any exception. Throws an [IllegalStateException] with a special message and original exception.
+     * Can be improved with custom exception (#1872).
+     *
+     * 1) IllegalArgumentException -> most probably column not found. Should be replaced with a custom exception (#1871)
+     * 2) ClassCastException -> most probably incorrect column type. Should be replaced with a custom exception (#1871)
+     * 3) else -> unexpected exception.
+     */
+    private fun generateTryCatchColumnAccess(columnAccessCode: String, columnName: String): String {
+        val renderedColumnName = renderStringLiteral(columnName)
+        return $$"""
```
Jolanrensen, line 291: nice `$$` usage :) but can you indent the code a bit further to the right? It's currently left of the `return`. Also if you assign it to a variable first, you can annotate it with `@Language("kotlin")` which might be nice

PR #1875, opened 2026-05-27
`countDistinct` overloads on `GroupBy`

### LINKED ISSUE #533: add countDistinct on GroupBy
I would like to have a function `countDistinct()` on grouped dataframe for counting distinct values by group

```kotlin
df.groupBy { column_0 }.countDistinct()
```

```kotlin
df.groupBy { column_0 }.countDistinct { column_1 }
```

Allex-Nik:
Fixes #533.

The `countDistinct` function is currently defined only on `DataFrame`. In this PR, I add two `countDistinct` aggregations on `GroupBy`:
- without a columns selector (all columns are considered)
- with a columns selector.

KDocs and website docs are adjusted accordingly.


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/countDistinct.kt
```diff
@@ -94,3 +102,81 @@ public fun <T> DataFrame<T>.countDistinct(vararg columns: AnyColumnReference): I
...
+ * that will be considered for evaluating whether the rows are distinct.
+ */
+@Refine
+@Interpretable("GroupByCountDistinct0")
+public fun <T, C> Grouped<T>.countDistinct(
+    resultName: String = "countDistinct",
+    columns: ColumnsSelector<T, C>,
+): DataFrame<T> =
+    aggregateValue(resultName) {
+        countDistinct(columns) default 0
```
Allex-Nik, line 178: `default 0` is used in `count` on `Grouped`. I see that it can be useful on `PivotGroupBy`, but is there any case when it is applied with `Grouped`?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/countDistinct.kt
```diff
@@ -0,0 +1,139 @@
...
+        )
+        val result = df.groupBy("group").countDistinct()
+        val expected = dataFrameOf(
+            "group" to columnOf(1),
+            "countDistinct" to columnOf(1),
+        )
+        result shouldBe expected
+    }
+
+    // TODO: check columns as well when #1531 is fixed
```
Allex-Nik, line 50: I saw in the repo TODOs left as reminders. Is it fine to leave it?

docs/StardustDocs/topics/groupBy.md
Allex-Nik, line 884: `count` was mentioned in both `Aggregation` and `Aggregation statistics` (example only in `Aggregation`).
Do you think `countDistinct` should be mentioned in both these sections as well (as a common statistical aggregation), or is it enough to only mention it in the `Aggregation` section?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/countDistinct.kt
Jolanrensen: probably yes, if one of the groups becomes empty for some reason. Better to be safe than sorry

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/countDistinct.kt
Jolanrensen: sure, as long as the issue is linked like this :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/countDistinct.kt
```diff
@@ -0,0 +1,139 @@
...
+    @Test
+    fun `countDistinct on empty GroupBy`() {
+        df.drop(df.nrow).groupBy("group").countDistinct().count() shouldBe 0
```
Jolanrensen: purely cosmetic, but I'd recommend formatting DF operations across multiple lines, per operation, to understand what follows what, and it reads best like a sentence:

```kt
df
    .drop(df.nrow)
    .groupBy("group").countDistinct()
    .count() shouldBe 0
```
I've disabled the respective ktlint rule to allow this notation, otherwise it always puts everthing on the next line.

Jolanrensen: I think we follow this principle in examples, not as important in tests, but I thought it would be informative nonetheless :)

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/CountDistinctSamples.kt
```diff
@@ -0,0 +1,139 @@
...
+        // SampleStart
+        df.countDistinct { "name"["firstName"] and "city" } // the result is 9
+        // SampleEnd
+    }
+
+    @Test
+    fun countDistinctColumnsDf() {
+        df.format().perRowCol { row, _ ->
+            val lastName = df[row.index()].name.lastName
+            background(lastNameToColor(lastName)) and textColor(black)
```
Jolanrensen: hmm. It took me a bit longer than expected to figure out what the colors meant exactly. I think the problem mainly is that your colors mean two things: "green" means "these rows are considered the same and counted as 1", while "white" means "these rows are all distinct". Could I recommend using a distinct color for each distinct row and using the same color for rows considered similar? (maybe use fewer rows if you run out of color XD)

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/CountDistinctSamples.kt
```diff
@@ -0,0 +1,139 @@
...
+        df.groupBy { isHappy }
+            .toDataFrame()
+            .convert { group }.with { group ->
+                val firstNameCol = group["name"]["firstName"]
+                group.format().perRowCol { row, _ ->
+                    val firstName = firstNameCol[row.index()] as String
+                    background(firstNameToColor(firstName)) and textColor(black)
+                }
+            }
+            .saveDfHtmlSample()
```
Jolanrensen, line 135: candidate for all nested frames to be expanded by default :D I recommended Andrei a way to achieve this not that long ago... maybe you can find and apply it here :) otherwise we can do it later

Jolanrensen:
good job! :D

docs/StardustDocs/topics/groupBy.md
Jolanrensen: hmm, probably mention it everywhere `count` is mentioned :D

Allex-Nik: Done

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/countDistinct.kt
Allex-Nik: > I'd recommend formatting DF operations across multiple lines, per operation

It is only for the cases when we do many (for example, 3+) DF operations, right? 
If we do 2 operations: `df.groupBy { isHappy }.countDistinct { name.firstName }`, would you keep them in one line?

Jolanrensen: yes. However, I count groupBy+countDistinct still as 1 operation, so I'd keep them in one line/attached regardless.

(okay, for groupBy the line is a bit more of a gray area, but for, say, `df.update { cols }.where {}.with {}`, it's clearer to see they belong together on one line, especially if there are other operations after it.)

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/CountDistinctSamples.kt
Allex-Nik: What if we color only the two rows that will be considered the same? I think this way we would focus the reader's attention on the most important thing: these two rows will be counted as one.
rows

Jolanrensen: @Allex-Nik hmm, slightly better, but could you see what it looks like with different colors for a dataframe with 4 rows max? I still believe that may show off the idea more clearly. This doesn't work well if the dataframe is larger than that, I believe

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/CountDistinctSamples.kt
Allex-Nik: I found the script, but I was thinking: since this expanding should be applied to many different samples, would it make sense to create a separate issue for this and implement it in a separate PR? Maybe within this issue we could also investigate if we need to just add a helper method with this script to `DataFrameSampleHelper` (and maybe even to update the existing samples to be expanded where needed) or we can change the main rendering script (`init.js`, `html.kt`) so that tables are expanded by default.

Or do you still think it's better to try expanding on this sample in this PR for now? If yes, I think I can add this script as a method in `DataFrameSampleHelper` and apply it only for `countDistinct`

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/countDistinct.kt
Allex-Nik: I see, thank you! I have formatted operations in the tests

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/CountDistinctSamples.kt
Allex-Nik: I have updated the dataframe in the `countDistinct` samples so that it only has 4 rows. I have also adjusted samples so that the new shorted dataframe still illustrates the ideas in the documentation. In the `groupBy` docs, I have made a separate sample for using `countDistinct` on `GroupBy` because the dataframe in the `countDistinct` docs is now different, so the sample cannot be reused in the `groupBy` docs.

It looks better indeed!

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/CountDistinctSamples.kt
Jolanrensen: We can do both maybe :) creating an issue is a good idea; then we can add the script here temporarily with a TODO pointing to the issue. That way the website looks good and we will fix it properly in the future. wdyt?

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/CountDistinctSamples.kt
Jolanrensen: image
indeed! :D thanks!

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/CountDistinctSamples.kt
Allex-Nik: I added the script and applied it to expand two samples in `countDistinct`.
The initial script did not expand column groups in nested frames, so I (with Claude) adjusted it to process them too. Checked on `countDistinct` that it works

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/DataFrameSampleHelper.kt
```diff
@@ -44,4 +49,49 @@ abstract class DataFrameSampleHelper(sampleName: String, subFolder: String = "sa
...
+                        let rootDf = table.df;
+                        expandNestedFrames(rootDf, rootDf);
+                        DataFrame.renderTable(rootDf.id);
+                    }
+                });
+            })();
+            """.trimIndent(),
+    )
+
+    fun DataFrame<*>.toExpandedHtml() =
```
Jolanrensen: Hmm, this name may be confusing. It's not like the html is expanded, right? It's just the nested dataframes that get opened. Maybe verbosity is better than compactness here, so: `toHtmlWithOpenedNestedDfs` or something like that ;P.

Allex-Nik: I agree :)
Renamed `toExpandedHtml` to `toHtmlWithOpenedNestedDfs`

Jolanrensen (approved):

Issue #1880, opened 2026-05-29
Avoid breaking existing code with Iterable<DataSchema>.toDataFrame() calls when enabling the plugin


koperagen:
Problem from example being converted to use plugin found by @Jolanrensen 
```kt
@DataSchema
data class MySchema(val i: Int)

fun main() {
  val df: DataFrame<MySchema> = listOf(MySchema(1)).toDataFrame() // breaks when plugin is enabled, type is now MySchema_98

  df.toList() // breaks when plugin is enabled
}
```

Proposed change:
Add toDataFrame overload. If T is already DataSchema type, no need to introduce plugin-generated local type.
```kt
public inline fun <reified T : DataRowSchema> Iterable<T>.toDataFrame(): DataFrame<T> =
    @DisableInterpretation toDataFrame {
        properties()
    }
```

Jolanrensen:
Additionally `DataFrame<X_123>.toList()` should probably cause a compiler plugin error and recommend users to rewrite to `.toListOf<ActualDataClassType>()` :)

Jolanrensen:
@koperagen Hmm, specifically for Spark, this change spells trouble for the `!` types introduced by Java:

No DataSchema:
<img width="1353" height="1080" alt="Image" src="https://github.com/user-attachments/assets/e8da70d1-4747-4645-a8cf-5023439a3c1e" />

With DataSchema:
<img width="1353" height="1127" alt="Image" src="https://github.com/user-attachments/assets/7b7c9f23-e830-4173-8219-732ff1f58ad9" />

But this is related to #1908

PR #1886, opened 2026-06-03
Moving final examples to /projects

### LINKED ISSUE #1614: Make `:examples` be separate projects and use the compiler plugin

Jolanrensen:
Fixes #1614 

Done:
- multik
- exposed
- hibernate


examples/projects/dev/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/main.kt
```diff
@@ -110,14 +109,16 @@ private fun SessionFactory.loadCustomersAsDataFrame(): DataFrame<DfCustomers> {
                 )
             }
             .toDataFrame()
+            // TODO This cast is a workaround for two bugs in DataFrame:
+            //   1) This `cast` should not be needed. `List<DfCustomers>` should become `DataFrame<DfCustomers>`
```
koperagen: I wouldn't say it's a bug :) #1880 is a new feature

Jolanrensen: fair ;P

Jolanrensen:
oh actually, let me also remove ksp! I don't think it's used anywhere anymore now :D

koperagen (approved):

docs/StardustDocs/topics/concepts/concepts.md
```diff
@@ -43,8 +43,8 @@ This is why it was designed to be hierarchical and allows nesting of columns and
...
 * [**Interoperable**](collectionsInterop.md) — convertable with Kotlin data classes and collections.
   This also means conversion to/from other libraries' data structures is usually quite straightforward!
-  See our [examples](https://github.com/Kotlin/dataframe/tree/master/examples/idea-examples/unsupported-data-sources) 
-  for some conversions between DataFrame and [Apache Spark](https://github.com/Kotlin/dataframe/tree/master/examples/projects/kotlin-spark), [Multik](https://github.com/Kotlin/dataframe/tree/master/examples/idea-examples/unsupported-data-sources/multik), and [JetBrains Exposed](https://github.com/Kotlin/dataframe/tree/master/examples/idea-examples/unsupported-data-sources/exposed).
+  See our [examples](https://github.com/Kotlin/dataframe/tree/master/examples/projects) 
+  for some conversions between DataFrame and [Apache Spark](https://github.com/Kotlin/dataframe/tree/master/examples/projects/kotlin-spark), [Multik](https://github.com/Kotlin/dataframe/tree/master/examples/projects/multik), and [JetBrains Exposed](https://github.com/Kotlin/dataframe/tree/master/examples/projects/exposed).
```
Allex-Nik, line 47: Would it make sense to mention the `hibernate` example here as well? It also contains conversions

Issue #1888, opened 2026-06-05
`format` + `where` fails on a `DataFrame` with a `ColumnGroup`


Allex-Nik:
When `format` with a `where` clause is applied to a `DataFrame` containing a `ColumnGroup`, if fails with an exception saying that the column mentioned in the `where` clause is not found. 
The problem does not occur if the `where` clause is removed or if the `ColumnGroup` is ungrouped.

Here is an example:
```kotlin
val df = dataFrameOf("firstName", "lastName", "age")(
    "Alice", "Cooper", 15,
    "Bob", "Dylan", 45,
    "Charlie", "Daniels", 20,
)

val groupedName = df.group { firstName and lastName }.into("name")

// this works
df.format { all() }.where { age < 18 }.with { background(RgbColor(242, 210, 189)) and textColor(black) }

// this also works
groupedName.format { all() }.with { background(RgbColor(242, 210, 189)) and textColor(black) }

// this fails
groupedName.format { all() }.where { age < 18 }.with { background(RgbColor(242, 210, 189)) and textColor(black) }
```

The exception from the example is the following: `java.lang.IllegalArgumentException: Column not found: 'age'`.

A notebook with the example is attached below.

[formatIssue.ipynb](https://github.com/user-attachments/files/28635152/formatIssue.ipynb)

PR #1889, opened 2026-06-05
Update the documentation of Row condition 

### LINKED ISSUE #1706: Update "Row condition" documentation
Missing information that row condition can be either `RowFilter` or `RowValueFilter` with different signatures. 
Add description about the difference and examples.
 
https://kotlin.github.io/dataframe/datarow.html#row-conditions

Allex-Nik:
Fixes #1706
Helps #898

- Add `RowFilter` and `RowValueFilter` as kinds of row conditions
- Move `DataRow` samples from `core` to `samples`
- Minor fixes


docs/StardustDocs/topics/concepts/DataRow.md
```diff
@@ -1,82 +1,324 @@
...
+* `String.invoke<T>(): T` — cell value of type `T` by this row and the given `this` column name;
+* `ColumnPath.invoke<T>(): T` — cell value of type `T` by this row and the given `this` column path;
+* `ColumnReference.invoke(): T` — cell value of type `T` by this row and the given `this` column;
+* `df()` — [`DataFrame`](DataFrame.md) that the current row belongs to.
 
 </snippet>
 
+The following dataframe will be used in the examples below:
+
+<!---FUN dfDataRow-->
```
Allex-Nik, line 46: I show it once above all the samples to avoid putting `df` into every section with samples. To make comparison of samples with one `df` still convenient, I collapse each sample separately so that the user expands one sample at a time and it is easy for them to compare the sample with the `df` without scrolling much.

docs/StardustDocs/topics/concepts/DataRow.md
```diff
@@ -1,82 +1,324 @@
...
+
+#### update (expression) {collapsible="true"}
+
+<!---FUN updateWithExpression-->
+<tabs>
+<tab title="Properties">
+
+```kotlin
 // Row expression computes updated values
 df.update { weight }.at(1, 3, 4).with { prev()?.weight }
```
Allex-Nik: This is the original example. Do you think we should replace it with something more plausible?

Something like `df.update { age }.at(1).with { it + 1 }`?

docs/StardustDocs/topics/concepts/DataRow.md
```diff
@@ -1,82 +1,324 @@
...
+#### Row filter examples
+
+##### filter {collapsible="true"}
+
+<!---FUN filterWithCondition-->
 
 ```kotlin
-// Row condition is used to filter rows by index
+// Row filter is used to filter rows by index
 df.filter { index() % 5 == 0 }
```
Allex-Nik: Also an original example. Do you think it's fine?

docs/StardustDocs/topics/concepts/DataRow.md
```diff
@@ -1,82 +1,324 @@
...
+```
+
+</tab></tabs>
+<!---END-->
+
+### Row value filter
+`RowValueFilter` is used via the `where` clause after selecting columns in functions
+such as [`update`](update.md), [`gather`](gather.md), and [`format`](format.md).
+Like `RowFilter`, it returns a `Boolean` indicating whether the row should be included in the result.
+However, unlike `RowFilter`, where both `this` and `it` refer to the current [`DataRow`](DataRow.md),
```
Allex-Nik, line 289: Is it correct? As far as I understood, it's not that `RowFilter` works with entire rows and `RowValueFilter` works only with parts of rows. `RowValueFilter` can also access the whole row, it just additionally accesses the selected part of the row as `it`, right?

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/DataRowApiSamples.kt
```diff
@@ -0,0 +1,228 @@
...
+        // SampleEnd
+    }
+
+    // TODO: format the whole rows when #1888 is fixed
+    @Test
+    fun formatWithCondition_properties() {
+        // SampleStart
+        // Row value filter is used to format only rows with minors
+        df
+            .format { age }
```
Allex-Nik: I think it would be more natural to format the entire rows, but if we include `name` (which is a `ColumnGroup`) into `format`, it fails. The related issue is #1888. As a temporary measure, I formatted only `age`.
Another solution would be to ungroup `name` in the dataset for this `DataRow.md` page for now instead.

docs/StardustDocs/topics/concepts/DataRow.md
```diff
@@ -1,82 +1,324 @@
...
+<!---FUN filterWithCondition-->
 
 ```kotlin
-// Row condition is used to filter rows by index
+// Row filter is used to filter rows by index
 df.filter { index() % 5 == 0 }
+```
+
+<!---END-->
+<inline-frame src="resources/filter.html" width="100%"/>
```
AndreiKingsley: Not found:
Image

docs/StardustDocs/topics/concepts/DataRow.md
```diff
@@ -1,82 +1,324 @@
...
+<!---END-->
+<inline-frame src="resources/dfDataRow.html" width="100%"/>
+
 ## Row expressions
-Row expressions provide a value for every row of [`DataFrame`](DataFrame.md) and are used in [add](add.md), [filter](filter.md), [forEach](iterate.md), [update](update.md) and other operations.
+Row expressions provide a value for every row of [`DataFrame`](DataFrame.md) 
+and are used in [add](add.md), [filter](filter.md), [forEach](iterate.md), [update](update.md), and other operations.
 
-<!---FUN expressions-->
+Row expression signature: ```DataRow.(DataRow) -> T```. Row values can be accessed with or without ```it``` keyword.
```
AndreiKingsley: But below you show and example of `update.with` which has a different signature (it has a s type of selected column values)!
I believe we miss here Row-Value Expression definition https://github.com/Kotlin/dataframe/blob/5d17a73265bef8b8ef6064261e1faa058f59fd7c/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/ExpressionsGivenRow.kt#L83
Could you please review it and rewrite it the same way as Row-Value Condition.

docs/StardustDocs/topics/concepts/DataRow.md
AndreiKingsley: I'd expand the original one - use both `prev()` and `it` and write about it in the comment

docs/StardustDocs/topics/concepts/DataRow.md
AndreiKingsley: I'd replace with something more meaningful — `name ==  && age > ...`

docs/StardustDocs/topics/concepts/DataRow.md
AndreiKingsley: > RowValueFilter works only with parts of rows

Why?

AndreiKingsley (changes requested):

docs/StardustDocs/topics/concepts/DataRow.md
AndreiKingsley: And also, just here, I'd show the original `df` first with highlighted rows meeting the predicate.

AndreiKingsley:
@Allex-Nik let me fix #1888 so you can make examples prettier!

docs/StardustDocs/topics/concepts/DataRow.md
Allex-Nik: I do not think that `RowValueFilter` works only with parts of rows. In the second sentence I mean that as far as I understood, the distinction between `RowFilter` and `RowValueFilter` is not "one works only with entire rows and another works only with parts of rows" :)

AndreiKingsley: Sorry, missread. 
So, yes, totally correct! But not sure if  `RowValueFilter` is used only via the `where`.

docs/StardustDocs/topics/concepts/DataRow.md
Allex-Nik: Done

docs/StardustDocs/topics/concepts/DataRow.md
Allex-Nik: Done

docs/StardustDocs/topics/concepts/DataRow.md
Allex-Nik: Done

docs/StardustDocs/topics/concepts/DataRow.md
Allex-Nik: Removed mentioning `where`.

docs/StardustDocs/topics/concepts/DataRow.md
Allex-Nik: Done. I followed the logic from `ExpressionsGivenRow`, although KDocs in this file referred to the selected cell as "previous" while explaining `RowValueExpression`:
`Providing a new value for every selected cell given the row of that cell and its previous value`
and I thought that maybe "current" would be more clear in this case. Please have a look at the `RowValueExpression` explanation in `DataRow.md`

AndreiKingsley (approved):
Great! 
I'd also add links to these topic to KDoc helpers.

PR #1890, opened 2026-06-08
Configure `KorroGenerateTask` to run after test tasks

AndreiKingsley:
Korro must run after test, since tests generate korro outputs.
Now generated source should be correct!

### DIFF
--- a/samples/build.gradle.kts
+++ b/samples/build.gradle.kts
@@ -75,9 +75,9 @@ dependencies {
     testImplementation(libs.hikaricp)
 }
 
-// Tests create
+// Running tests creates korro output lines!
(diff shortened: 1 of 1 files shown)


samples/build.gradle.kts
```diff
@@ -74,6 +75,11 @@ dependencies {
     testImplementation(libs.hikaricp)
 }
 
+// Tests create
+tasks.withType<KorroGenerateTask>().configureEach {
+    mustRunAfter(tasks.test)
```
Jolanrensen: I think we should use "dependsOn". Otherwise it's still possible to run korro without running the tests.

Jolanrensen (changes requested):

Jolanrensen (approved):

PR #1896, opened 2026-06-12
Added a method for handling extension property exceptions

AndreiKingsley:
Helps #1742.

This method should be used in the extension property generation logic inside the Compiler Plugin.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/exceptions/extensionPropertyException.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/exceptions/extensionPropertyException.kt
@@ -21,13 +21,13 @@ private const val TROUBLESHOOTING_LINK =
 public fun handleExtensionPropertyException(e: Exception, columnName: String) {
     val msg = when (e) {
         is IllegalArgumentException ->
-            "Column not found exception in the generated DataFrame extension property '$columnName': ${e.localizedMessage}. See $$TROUBLESHOOTING_LINK for more information."
+            "Column not found exception in the generated DataFrame extension property '$columnName': ${e.localizedMessage}. See $TROUBLESHOOTING_LINK for more information."
 
         is ClassCastException ->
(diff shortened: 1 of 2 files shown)


AndreiKingsley:
@Jolanrensen @koperagen  should we mark such method (that shouldn't be used by a regular user but should be `public`) with some annotation?

koperagen (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/exceptions/extensionPropertyException.kt
```diff
@@ -0,0 +1,33 @@
...
+ * 3) else -> unexpected exception.
+ *
+ * @param [e] any [Exception] caught inside `catch(e: Exception)`.
+ * @param [columnName] rendered column name used in the resulting exception message.
+ * @throws [IllegalStateException] with a custom message depends on type of [e] and contating [e] as cause.
+ */
+public fun handleExtensionPropertyException(e: Exception, columnName: String) {
+    val msg = when (e) {
+        is IllegalArgumentException ->
+            "Column not found exception in the generated DataFrame extension property '$columnName': ${e.localizedMessage}. See $$TROUBLESHOOTING_LINK for more information."
```
Jolanrensen: check the `$` signs again ;P

Jolanrensen:
@AndreiKingsley hmm, we have `@RequiredByIntellijPlugin`, but that is only used for public functions too. For the compiler plugin we usually just have obscure public functions.

I'm not sure whether we already tried something like:
- `internal` with suppressing all warnings (but we know how that can break stuff now)
- Deprecated HIDDEN and suppressing warnings?
- OptIn required?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/exceptions/extensionPropertyException.kt
AndreiKingsley: Thanks!

AndreiKingsley:
Ok, let's then first finalize the standard approach; for now, I’ll just add this without any annotation.

PR #1898, opened 2026-06-15
Fixed `format` for column group

### LINKED ISSUE #1888: `format` + `where` fails on a `DataFrame` with a `ColumnGroup`

AndreiKingsley:
Fixes #1888.
 
 @Allex-Nik please check

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
@@ -173,7 +173,6 @@ internal fun AnyFrame.toHtmlData(
         col: ColumnWithPath<*>,
         rowsLimit: Int?,
         configuration: DisplayConfiguration,
-        renderRootDf: AnyFrame = this,
     ): ColumnDataForJs {
         val values = if (rowsLimit != null) rows().take(rowsLimit) else rows()
         val scale = if (col.isNumber()) col.asNumbers().scale() else 1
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
```diff
@@ -265,7 +265,7 @@ internal fun AnyFrame.toHtmlData(
         }
         val nested = if (col is ColumnGroup<*>) {
             col.columns().map {
-                col.columnToJs(it.addParentPath(col.path), rowsLimit, configuration, renderRootDf)
+                renderRootDf.columnToJs(it.addParentPath(col.path), rowsLimit, configuration, renderRootDf)
```
Jolanrensen: in both places `columnToJs` is called, both the receiver and `renderRootDf` are the same value. I believe this can be simplified

Allex-Nik (commented):
It works now! Thanks :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/html.kt
AndreiKingsley: Indeed, removed it.

Jolanrensen (approved):

PR #1901, opened 2026-06-16
Ensure scoped table retrieval for MySQL/MariaDB connections

### LINKED ISSUE #1746: `readAllSqlTables` doesn't work for MySQL with multiple databases
Looks like for MySQL, `readAllSqlTables` catches table lists for all active mysql dbs (not only specified by url) 
if `catalogue` is not specified.

To reproduce.

https://gist.github.com/AndreiKingsley/b9fff2950a94eae7308c2da72150e05e

Seems like it's a MySQL specific feature, but need to check it in all other DBs.

And don't forget to add tests for it 😄 !

It could require API changes

zaleslaw:
Closes #1746 

# Fix #1746: readAllSqlTables leaks tables from other databases on MySQL/MariaDB

## 1. Problem

`readAllSqlTables` and `DataFrameSchema.readAllSqlTables` called without an explicit `catalogue`
incorrectly return tables from **all accessible databases** when connected to MySQL or MariaDB.

Root cause: both functions passed `null` directly to `DatabaseMetaData.getTables(null, ...)`.
On MySQL/MariaDB, `null` means *all catalogs*, so a connection scoped to
`jdbc:mysql://host/mydb` would enumerate tables from every database visible to that user.
Subsequent `SELECT * FROM table_from_another_db` then fails with "table doesn't exist"
because the query runs in the wrong database context.

This is a MySQL/MariaDB-specific behavior: PostgreSQL and SQLite connections are inherently
single-database, so `getTables(null, ...)` there already returns only the current DB.

## 2. Solution

When `catalogue` is not provided by the caller, fall back to `connection.catalog`
(standard JDBC API — returns the database name from the JDBC URL).
The same pattern is already used in `DbType.getTableColumnsMetadata`.

```kotlin
val effectiveCatalogue = catalogue ?: try {
    connection.catalog.takeUnless { it.isNullOrBlank() }
} catch (_: Exception) { null }
```

If the URL has no database (`jdbc:mysql://host` without `/dbname`), `connection.catalog`
returns `null`/`""`, which falls through to `null` — preserving the existing
"all catalogs" behavior for that case.

Fixed in two places:
- `readJdbc.kt` — `DataFrame.readAllSqlTables(connection, ...)`
- `readDataFrameSchema.kt` — `DataFrameSchema.readAllSqlTables(connection, ...)` (same bug, hardcoded `null`)

## 3. Situation by Database

| Database     | Affected   | Reason                                                                          |
|--------------|------------|---------------------------------------------------------------------------------|
| MySQL        | Fixed      | `getTables(null,...)` returns all databases; `connection.catalog` = DB from URL |
| MariaDB      | Fixed      | Same behavior as MySQL                                                          |
| SQL Server   | Fixed      | Same catalog/database model; same fix applies                                   |
| PostgreSQL   | No change  | Cross-database access impossible; `getTables` already scoped to current DB      |
| SQLite       | No change  | Single-file = single catalog; `connection.catalog` returns `null`               |
| H2           | No change  | In-process, single DB per connection; not affected                              |
| DuckDB       | No change  | `connection.catalog` returns `null` for default in-memory catalog               |

Tests added (all `@Ignore`, require local server) in:
- `local/mysqlTest.kt`
- `local/mariadbTest.kt`
- `local/mssqlTest.kt`

Each test creates a second database with a unique table, connects to DB1 by URL,
calls `readAllSqlTables` without `catalogue`, and asserts the second DB's table is absent.


dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readDataFrameSchema.kt
```diff
@@ -432,7 +432,12 @@ public fun DataFrameSchema.Companion.readAllSqlTables(
 
     // exclude system- and other tables without data
     val tableTypes = determinedDbType.tableTypes?.toTypedArray()
-    val tables = metaData.getTables(null, null, null, tableTypes)
+    // Same catalog-scoping fix as in readJdbc.kt: use connection.catalog to avoid reading
+    // tables from all MySQL/MariaDB databases when no specific catalog is given.
+    val effectiveCatalog = try {
+        connection.catalog.takeUnless { it.isNullOrBlank() }
+    } catch (_: Exception) { null }
```
Jolanrensen: if you're not interested in the exception, maybe you could try `runCatching { ... }.getOrNull()`. Just an alternative way to write it, but both are valid of course :)

AndreiKingsley (approved):

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/readDataFrameSchema.kt
zaleslaw: Agree, it could be better for reading or handling

PR #1904, opened 2026-06-18
Aligned column name clashing handling across different IO

### LINKED ISSUE #387: Add column name repairing for all IO sources as a common step for dataframe creation
The solution could be based on the PR #386

AndreiKingsley:
Closes #387.

Now `ColumnNameGenerator` is used in all IO implementations, behavior is aligned.

Deprecated Excel approach. Added a snippet about behavior in IO KDocs.

Added kodex to IO modules.


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/snippets.kt
```diff
@@ -6,31 +6,17 @@ import org.jetbrains.kotlinx.dataframe.DataFrame
...
- * the first time the name of the column is encountered, no number is appended.
- *
- * It is recommended to [rename][org.jetbrains.kotlinx.dataframe.api.rename] them
- * to maintain clarity and improve code readability.
  */
-internal typealias AutoRenamingColumnsInDataFrame = Nothing
-
-/** [Auto-renaming columns in DataFrame][AutoRenamingColumnsInDataFrame] */
-internal typealias AutoRenamingLink = Nothing
+internal typealias AutoRenameInputSnippet = Nothing
```
Jolanrensen, line 19: is this one used?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/topics.kt
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * {@comment
+ *    Auto-renaming in [DataFrame] KDoc-topic.
+ *    Link to it with `@include [AutoRenamingLink]`.
+ * }
+ *
+ * ## Auto-renaming columns in [DataFrame]
+ *
+ * [DataFrame] can not contain columns with duplicate names.
```
Jolanrensen: cannot

Jolanrensen: no clue why, English is weird

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/topics.kt
```diff
@@ -0,0 +1,28 @@
...
+ * However, sometimes after reading dataframes from sources or
+ * after some operations, columns with duplicate names may appear in the result.
+ *
+ * In such cases, columns with duplicate names are automatically renamed in the resulting [DataFrame]
+ * using the pattern `"\$name\$n"`, where `name` is the original column name
+ * and `n` is a unique index (1, 2, 3, and so on);
+ * the first time the name of the column is encountered, no number is appended.
+ *
```
Jolanrensen, line 36: maybe it's clearer to move the sample here. `For example: "name", "name1", "name2", etc.` That may be clearer :)

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/documentation/utils.kt
```diff
@@ -0,0 +1,66 @@
...
+import kotlin.annotation.AnnotationTarget.FUNCTION
+import kotlin.annotation.AnnotationTarget.LOCAL_VARIABLE
+import kotlin.annotation.AnnotationTarget.PROPERTY
+import kotlin.annotation.AnnotationTarget.PROPERTY_GETTER
+import kotlin.annotation.AnnotationTarget.PROPERTY_SETTER
+import kotlin.annotation.AnnotationTarget.TYPE
+import kotlin.annotation.AnnotationTarget.TYPEALIAS
+import kotlin.annotation.AnnotationTarget.VALUE_PARAMETER
+
+// TODO use KDoc-helpers from :core
```
Jolanrensen: Try this in the build.gradle.kts. I remembered I built some system for adding contextual sources for KoDEx :P

```
kodexConvention {
    contextualSourcesDirectories =
        project(projects.core.path)
            .sourceSets
            .main.get()
            .allSource
}

tasks.processKDocsMain {
    dependsOn(project(projects.core.path).tasks.assemble)
}
```

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -66,6 +68,8 @@ public class Excel : SupportedDataFrameFormat {
 
 private const val MESSAGE_REMOVE_1_1 = "Will be removed in 1.1."
 internal const val READ_EXCEL_OLD = "This function is only here for binary compatibility. $MESSAGE_REMOVE_1_1"
+internal const val NAME_REPAIR_STRATEGY: String =
```
Jolanrensen: We don't have a system yet for tracking deprecations in other modules than core, but maybe we could mirror the same structure?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/snippets.kt
AndreiKingsley: Should be used in IO modules, after kodex fix

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
AndreiKingsley: How exactly?

Jolanrensen: in `util/deprecationMessages.kt` I mean

zaleslaw (approved):
Yeah, I missed earlier that in the JDBC module it was fixed in February by @Jolanrensen 

Could be merged!

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/documentation/utils.kt
AndreiKingsley: Can it be added to convention plugin?

Jolanrensen: well, this is the convention plugin in action ;P `contextualSourcesDirectories` is an argument for the `kodexConvention` extension. If we add this setting by default, then it will also be applied to the `:core` module, which shouldn't happen

Jolanrensen: maybe we could do something clever and traverse the project dependencies of any module the kodex convention plugin is applied and add those to the context... hmmm.... I'll give it a try :)

Jolanrensen:
@AndreiKingsley btw, don't forget to add the new generated-sources folders to both GH actions workflows

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentationCsv/CommonReadDelimDocs.kt
```diff
@@ -17,6 +17,8 @@ import java.net.URL
  *
  * Reads any $[FILE_TYPE] $[DATA] to a [DataFrame][DataFrame].
  *
+ * @include [AutoRenameInputSnippet]
```
Jolanrensen: Unresolved reference.

you need to specify the fully qualified name to reference an internal snippet. I believe it can't be imported due to Kotlin limitations, but it can be referenced

dataframe-excel/build.gradle.kts
```diff
@@ -38,3 +38,7 @@ kotlinPublications {
         packageName = artifactId
     }
 }
+
+tasks.processKDocsMain {
```
Jolanrensen, line 43: can be omitted, right?

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
```diff
@@ -32,7 +32,10 @@ import org.jetbrains.kotlinx.dataframe.api.forEach
 import org.jetbrains.kotlinx.dataframe.api.select
 import org.jetbrains.kotlinx.dataframe.codeGen.AbstractDefaultReadMethod
 import org.jetbrains.kotlinx.dataframe.codeGen.DefaultReadDfMethod
-import org.jetbrains.kotlinx.dataframe.exceptions.DuplicateColumnNamesException
+import org.jetbrains.kotlinx.dataframe.documentation.AutoRenameInputSnippet
```
Jolanrensen: wait, here you can just import an internal declaration?

dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentationCsv/CommonReadDelimDocs.kt
Jolanrensen: Oh I see you can just import it from the other two usages... interesting... well, then please add the import :)

AndreiKingsley: hmm, import is valid but `proccessKDocsMain` hadn't included it until I used FQN

dataframe-excel/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt
AndreiKingsley: Looks like Kotlin Compiler allows it, but Kodex can't handle it for some reason

Jolanrensen: I'll take a look. All paths, including contextual ones, are put into one giant Map, so this shouldn't be an issue

Jolanrensen: For me it seems to work fine:
image

Jolanrensen: https://github.com/Kotlin/dataframe/blob/ab224e866965f91fa277ccd1c479f01ccb6e7599/dataframe-excel/generated-sources/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/xlsx.kt#L257

AndreiKingsley: Weird. For me, generated sources simply don't have this line!

Jolanrensen: might be a local cache issue then, because it works fine on the CI

Jolanrensen (approved):

PR #1905, opened 2026-06-22
Group by values exceptions

### LINKED ISSUE #1569: Exception when calling `updateGroups { it }.first().values()` on `GroupBy`
Assume we have the following dataframe:
```kt
val df = dataFrameOf(
    "name" to columnOf("Alice", "Bob", "Charlie"),
    "age" to columnOf(15, 20, 25),
)
```

Calling
```kt
df.groupBy { age }.updateGroups { it }.first().values()
```
either in notebooks or outside them causes an exception. In notebooks this exception is formulated in the following way: 

> java.lang.IllegalStateException: Can not insert column `age` because column with this path already exists in DataFrame 

If we remove `age` from every group:
```kt
df.groupBy { age }.updateGroups { it.remove { age } }.first().values()
```
the exception does not occur.

### Notebooks
In notebooks, a similar problem occurs even without using `values()`. That is, calling:
```kt
df.groupBy { age }.updateGroups { it }.first()
```
causes:

> The problem is found in one of the loaded libraries: check library renderers
java.lang.IllegalStateException: Can not insert column `age` because column with this path already exists in DataFrame
org.jetbrains.kotlinx.jupyter.exceptions.ReplLibraryException: The problem is found in one of the loaded libraries: check library renderers

If we add `concat()`, the exception does not occur:
```kt
df.groupBy { age }.updateGroups { it }.first().concat()
```

Jolanrensen:
Fixes https://github.com/Kotlin/dataframe/issues/1569 by means of adding a better explanation in exceptions and kdocs

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
@@ -545,6 +545,10 @@ public interface GroupBy<out T, out G> : Grouped<G> {
      * Creates a new [GroupBy] by transforming each group’s [DataFrame]
      * using the provided [transform] function.
      *
+     * __NOTE:__ This operation removes key-column status from each column in the group.
+     * In other words, each column in the group is treated as a new column,
+     * and not omitted when [`.values()`][Grouped.values] or other aggregations are called.
+     *
      * @param [transform] A lambda that takes each group as a [DataFrame]
      * (available both as a receiver and as a parameter) and returns a transformed [DataFrame].
      * @return A new [GroupBy] instance containing the transformed groups.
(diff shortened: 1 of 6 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/insert.kt
```diff
@@ -423,3 +423,9 @@ public fun <T> InsertClause<T>.at(position: Int): DataFrame<T> = df.add(column).
 // endregion
 
 // endregion
+
+public sealed class InsertException(message: String) : IllegalStateException(message)
```
AndreiKingsley, line 432: KDocs please!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/groupBy.kt
```diff
@@ -564,6 +568,10 @@ public interface GroupBy<out T, out G> : Grouped<G> {
      * gb.filter { category == "Engineer" || group.rowsCount() >= 5 }
      * ```
      *
+     * __NOTE:__ This operation removes key-column status from each column in the group.
+     * In other words, each column in the group is treated as a new column,
+     * and not omitted when [`.values()`][Grouped.values] or other aggregations called.
```
AndreiKingsley: are called

AndreiKingsley (approved):

PR #1907, opened 2026-06-22
Adds `Iterable<DataRowSchema>.toDataFrame()` overload

### LINKED ISSUE #1880: Avoid breaking existing code with Iterable<DataSchema>.toDataFrame() calls when enabling the plugin

Jolanrensen:
Fixes #1880

Added Iterable<T>.toDataFrame() overload for DataRowSchema types, 
compiler plugin test in :samples, because the compiler plugin is live there.

This fixes the compiler plugin creating a new type immediately when turning a list of `@DataSchema` annotated classes into a dataframe.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/toDataFrame.kt
@@ -6,6 +6,7 @@ import org.jetbrains.kotlinx.dataframe.DataFrame
 import org.jetbrains.kotlinx.dataframe.DataRow
 import org.jetbrains.kotlinx.dataframe.annotations.AccessApiOverload
 import org.jetbrains.kotlinx.dataframe.annotations.DataSchema
+import org.jetbrains.kotlinx.dataframe.annotations.DisableInterpretation
 import org.jetbrains.kotlinx.dataframe.annotations.Interpretable
 import org.jetbrains.kotlinx.dataframe.annotations.Refine
 import org.jetbrains.kotlinx.dataframe.columns.BaseColumn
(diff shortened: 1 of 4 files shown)


Jolanrensen:
image
As expected, this directly touches the kotlin-spark dev example :D (where this limitation was detected)
I'll fix the example right away

Jolanrensen:
A workaround for #1908 could be to write the function like this:

```kt
@JvmName("dataRowSchemaIterableToDataFrame")
@Suppress("INCORRECT_LEFT_COMPONENT_OF_INTERSECTION")
public inline fun <reified T : DataRowSchema> Iterable<T>.toDataFrame(): DataFrame<T & Any> =
    @DisableInterpretation
    toDataFrame {
        properties()
    }
```

But notebooks don't like `T & Any` and it could better be solved from the compiler plugin side I believe.

AndreiKingsley (approved):

koperagen:
Let's probably not merge since this needs a workaround if flexible type appears. A few things need to be fixed on compiler plugin side to properly work with such types. Let's wait these fixes in stable release of Kotlin, postpone feature to 1.1

Even this is  not ideal because it suppresses INCORRECT_LEFT_COMPONENT_OF_INTERSECTION, which is not guaranteed to work in the future
```kt
public inline fun <reified T : DataRowSchema> Iterable<T>.toDataFrame(): DataFrame<T & Any> =
    @DisableInterpretation
    toDataFrame {
        properties()
    }
```

PR #1910, opened 2026-06-25
Enable Module dependencies in KoDEx context 

Jolanrensen:
Helps https://github.com/Kotlin/dataframe/pull/1904 by automatically adding module dependencies in the KoDEx scope. Removes tests from the KoDEx scope, because we didn't use it anymore.

Note that this might slow down the build. KoDEx has no notion of multi-module projects, meaning `:core` will be reprocessed each time another module includes it...


AndreiKingsley (approved):
Nice! 
The only concern about kodex `@sample` source logic,
but for now we haven't integrated in our KDoc pipeline.
Maybe we need to create an issue and leave there a link to this PR?

AndreiKingsley:
The only thing https://github.com/Kotlin/dataframe/blob/master/dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentationCsv/utils.kt
is not needed anymore after this PR, is it?

AndreiKingsley:
And https://github.com/Kotlin/dataframe/blob/master/dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentationCsv/DocumentationUrls.kt
can be merged into core

Jolanrensen:
> The only thing https://github.com/Kotlin/dataframe/blob/master/dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentationCsv/utils.kt is not needed anymore after this PR, is it?

> And https://github.com/Kotlin/dataframe/blob/master/dataframe-csv/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentationCsv/DocumentationUrls.kt can be merged into core

Uhm, well, I suppose yes indeed, however, they are `internal` declarations, so `@ExportAsHtml` and `@ExcludeFromSources` still need to be there.

Also, the URLs don't like being referenced...

image

but maybe they can be public, they're excluded from sources anyway... Same as `@ExcludeFromSources`, it will be removed by definition.
`@ExportAsHtml` cannot be made public, but then again, we don't use it outside core for now.

Jolanrensen:
> Nice! The only concern about kodex `@sample` source logic, but for now we haven't integrated in our KDoc pipeline. Maybe we need to create an issue and leave there a link to this PR?

I think we will figure that out later once we need it. We'll need to modify the convention plugin anyway if we want to `@sample` some function from the :samples module in another module, plus it might be tricky, because the IDE won't get such references, just KoDEx, it's something for later if we think we need it :) But for now just handwritten samples in KDocs are fine, we don't have that many after all

Jolanrensen:
I added support for this in KoDEx itself https://github.com/Jolanrensen/KoDEx/pull/102, will publish that and use it in this PR before merging.

AndreiKingsley (approved):

PR #1914, opened 2026-07-01
Made emptyDataFrame reified so resulting "null object" dataframes are suitable for aggregation and other column access operations like select

### LINKED ISSUE #463: Streamline Empty DF creation functions
Currently there's a discrepancy between
```kotlin
emptyDataFrame<Person>() // non inline/reified function -> just a cast DF without columns/rows
DataFrame.emptyOf<Person>() // inline/reified function -> DF with columns but 0 rows
```
Without reading the KDocs this is not clear to the user.

koperagen:


Jolanrensen (approved):
nice!

PR #1916, opened 2026-07-03
Deprecating `DataFrame.read()` and `SupportedDataFrameFormat`

### LINKED ISSUE #450: Revisit the hierarchy of SupportedCodeGenerationFormat and SupportedDataFrameFormat to make easier addition of new IO sources

Jolanrensen:
Fixes #450

Closes #1858, as this surpasses it.


dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReading.kt
```diff
@@ -19,10 +19,15 @@ import java.nio.channels.SeekableByteChannel
 import java.nio.file.Files
 import java.nio.file.Path
 
+@Deprecated("SupportedDataFrameFormat is deprecated. Will be ERROR in 1.1.", level = DeprecationLevel.ERROR)
```
koperagen: "Will be ERROR" contradicts level = ERROR

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/codeGen/SchemaReader.kt
```diff
@@ -7,26 +7,28 @@ import org.jetbrains.kotlinx.dataframe.codeGen.Code
...
+@Suppress("DEPRECATION_ERROR")
 public val CodeGenerator.Companion.urlDfReader: (url: URL, formats: List<SupportedFormat>) -> DfReadResult
     get() = { url, formats ->
         try {
             val (format, df) = url.openStream().use {
                 DataFrame.read(
                     stream = it,
-                    format = guessFormat(url, formats) as? SupportedDataFrameFormat?,
-                    formats = formats.filterIsInstance<SupportedDataFrameFormat>(),
+                    format = guessFormat(url, formats) as? org.jetbrains.kotlinx.dataframe.io.SupportedDataFrameFormat?,
```
koperagen: Why FQN? :o

Jolanrensen: It started to complain about the import :/ But let me check if that's just a fluke of the IDE, because it should indeed not be necessary

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReading.kt
Jolanrensen: you're right, they should say "1.0"... I think I made an error copying them over

Jolanrensen:
also need to update the example projects and notebooks, of course...

zaleslaw (approved):

PR #1919, opened 2026-07-06
Replacing dependency updates Gradle plugin with Caupain

### LINKED ISSUE #1917: Explore alternative dependency updates checkers
- Our current one (Ben Manes dependency-updates) checks all loaded dependencies in the available sourcesets, meaning I had to add this [ugly solution](https://github.com/Kotlin/dataframe/pull/1893) to get proper version tracking of our example projects
- It doesn't track plugin updates
- It cannot auto-bump versions

For instance: https://github.com/deezer/caupain/tree/main 
- Caupain only looks at the versions toml file directly, meaning this workaround won't be needed
- It can report actual version reference names
- It respects `#ignoreUpdates`
- It can do pretty output like MD and html, as well as the usual console output, or JSON if we want to make a pretty print DF out of it like before
- Cherry on top? It can auto-bump versions (Though this likely needs some strict rules etc.)

Jolanrensen:
Fixes #1917
(Updated after #1924)

We gain two commands:
- `checkDependencyUpdates`, which checks our `libs.versions.toml` for outdated dependencies and outputs the results in the console and `/build/reports/dependency-updates.md|html`.
  - This report can also run from GH actions, manually or triggered on a time interval. This includes links to release notes if there are any! See here an example of what Caupain can output as Github action: https://github.com/Kotlin/dataframe/actions/runs/28795771151
- `replaceOutdatedDependencies`, which actually updates the versions in `libs.versions.toml` to the latest versions! This should be relatively safe since the convention plugin can be configured to filter out certain versions, and `# ignoreUpdates` can be used to exclude specific libraries/versions. However, each version bump should still be checked carefully. I'd recommend running `replaceOutdatedDependencies`, copy the new contents of `libs.versions.toml`, undo the changes, then compare the file with your clipboard. That way you can one-by-one bump and check :)


Jolanrensen:
See here an example of what Caupain can output as Github action: https://github.com/Kotlin/dataframe/actions/runs/28795771151

Allex-Nik (approved):
Great!
I've also run the new commands locally - everything works

Jolanrensen:
merging because master starts failing, as `com.github.ben-manes.versions` no longer exists (it migrated to `io.`)

PR #1922, opened 2026-07-07
Disabled renaming to camelcase for `generateInterfaces/DataClasses` functions

### LINKED ISSUE #1921: `NameNormalizer` in `generateDataClasses`/`generateInterfaces` should not call `renameToCamelCase`
`NameNormalizer` in generateDataClasses/generateInterfaces should not call `renameToCamelCase`

Jolanrensen:
Fixes #1921 

Changed `NameNormalizer.default` to `id()` instead of renameToCamelCase by delimiter.

I also tried to change this behavior in `openapi-generator`, since it also automatically renames accessors to camel case. This uncovered a bug with nullability, but that was also fixable.


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/generateCode.kt
```diff
@@ -285,7 +287,18 @@ internal inline fun <reified T> markerName(): String =
...
  * Converts delimited 'my_name', 'my name', etc., String to camelCase 'myName'
  */
-public val NameNormalizer.Companion.default: NameNormalizer get() = NameNormalizer.from(setOf('\t', ' ', '_'))
+public val NameNormalizer.Companion.toCamelCaseByDelimiter: NameNormalizer
+    get() = NameNormalizer.toCamelCaseFrom(setOf('\t', ' ', '_'))
+
+/**
+ * Performs no name normalization.
+ *
+ * (Inside the actual code generation, [ValidFieldName] is still used to prevent incompilable code from being generated)
```
koperagen: I think "actual" is a bit confusing, please try to reformulate without it

dataframe-openapi-generator/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/utils.kt
```diff
@@ -17,14 +18,27 @@ internal fun String.withoutTopInterfaceName(topInterfaceName: ValidFieldName): S
...
         .replaceFirstChar { it.uppercaseChar() }
 
 internal fun String.toNullable() = if (this.last() == '?') this else "$this?"
 
+/**
+ * Returns whether this [FieldType] refers to a [Marker][OpenApiMarker] that is itself nullable
+ * (such as a nullable enum, whose nullability is defined by the type and cannot be removed by
+ * making a property required). Returns `false` for primitives and non-nullable markers.
+ */
+internal fun FieldType.refersToNullableMarker(getRefMarker: GetRefMarker, topInterfaceName: ValidFieldName): Boolean {
```
koperagen, line 32: What caused tests to start failing? Is it because of changes in name normalization?

Jolanrensen: No, because I removed `snakeToLowerCamelCase()` inside openapi-generator. I'm not exactly sure why, but it suddenly generated `override val eye_color: EyeColor` in `Cat` when I removed `@ColumnName`, which is incorrect because in the supertype `eye_color` is nullable. Adding this fix avoids that issue. `Cat` and ` Pet` now have the same nullability for `eye_color` and since `Cat : Pet`, the property is omitted once again in the generation.

Allex-Nik (approved):

PR #1923, opened 2026-07-08
Updated the Hibernate example

### LINKED ISSUE #1680: Modernize the Hibernate example if it's possible on the current version
After our example migration to a separate project and splitting into separate projects, we probably could obtain the possibility to have a modern and up-to-date Hibernate example

- [ ] check the possibility of migration to 7.+ Hibernate version in the particular project
- [ ] learn the best modern practices and share some references
- [ ] move from XML configs to the code configs
- [ ] research possibility to improve DataFrame construction from ```@Entity``` object
- [ ] don't forget to update documentation around this sample (could use LLM)

Allex-Nik:
Fixes #1680 

- Bump hibernate to 7.4.4.Final
- Add JPA plugin
- Replace xml config with code config
- Some other fixes


examples/projects/dev/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/entities.kt
Allex-Nik, line 17: The JPA plugin (applies no-arg and all-open compiler plugins) makes entities open classes and generates a constructor with no parameters for each of them. 
Making the entities open classes allows Hibernate to apply some features like lazy loading.
The no-arg constructor allows us to remove default values where they are not actually needed and are not intended. 

Usage of the JPA plugin is recommended [here](https://blog.jetbrains.com/idea/2026/01/how-to-avoid-common-pitfalls-with-jpa-and-kotlin/#non-final-classes-and-no-argument-constructors). The JPA plugin [applies](https://youtrack.jetbrains.com/issue/KT-79389/Add-allopen-plugin-JPA-preset-to-kotlin.plugin.jpa?_gl=1*1vi3x2l*_gcl_au*MTg0ODQ1Mzk4Ny4xNzgzNDMxNTMz*FPAU*MTg0ODQ1Mzk4Ny4xNzgzNDMxNTMz*_ga*MjAzMjM2ODk3MS4xNzc0NjM4OTcw*_ga_9J976DJZ68*czE3ODM1MjE0NTkkbzE3JGcxJHQxNzgzNTIxNDkxJGoyOCRsMCRoMA..&_cl=MTsxOzE7bzFvNktGNE1DRWtIMXJDRGQ2MHN6N210SzZaeTZiTW5LOVNZMEVlYmlvUWNFTEo1QnZ0ZDhrREhKQVMybGFiUjs=) all-open since Kotlin 2.3.20.

examples/projects/dev/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/entities.kt
```diff
@@ -17,9 +20,10 @@ class AlbumsEntity(
     @Column(name = "AlbumId")
     var albumId: Int? = null,
     @Column(name = "Title", length = 160, nullable = false)
-    var title: String = "",
-    @Column(name = "ArtistId", nullable = false)
-    var artistId: Int = 0,
+    var title: String,
+    @ManyToOne(fetch = FetchType.LAZY, optional = false)
```
Allex-Nik, line 24: Artist is loaded only if it is needed, and an album must have an artist

examples/projects/dev/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/entities.kt
```diff
@@ -17,9 +20,10 @@ class AlbumsEntity(
...
+    @JoinColumn(name = "ArtistId", nullable = false)
+    var artist: ArtistsEntity,
```
Allex-Nik, line 26: This way Hibernate knows and tracks the connection (FK) between entities: we don't need to manually flush ids into the DB and call them with `!!`.

examples/projects/dev/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/main.kt
```diff
@@ -43,17 +46,16 @@ fun main() {
...
 
 private fun SessionFactory.insertSampleData() {
-    withTransaction { session ->
+    inTransaction { session ->
         // a few artists and albums (minimal, not used further; just demo schema)
         val artist1 = ArtistsEntity(name = "AC/DC")
         val artist2 = ArtistsEntity(name = "Queen")
         session.persist(artist1)
         session.persist(artist2)
-        session.flush()
```
Allex-Nik, line 52: The Albums entity now refers to the Artists entity instead of id in the DB

examples/projects/dev/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/main.kt
```diff
@@ -89,15 +91,15 @@ private fun SessionFactory.loadCustomersAsDataFrame(): DataFrame<DfCustomers> =
         val root: Root<CustomersEntity> = criteriaQuery.from(CustomersEntity::class.java)
         criteriaQuery.select(root)
 
-        session.createQuery(criteriaQuery)
+        session.createSelectionQuery(criteriaQuery)
```
Allex-Nik, line 94: `createSelectionQuery` returns `SelectionQuery` instead of `Query`, [exposing](https://docs.hibernate.org/stable/orm/userguide/html_single/#hql-SelectionQuery) only methods which are relevant to selection queries.
When used with HQL, `createSelectionQuery` also [throws](https://docs.hibernate.org/orm/7.4/introduction/html_single/#hql-queries) an exception if passed a query string that begins with "insert", "delete", or "update".

examples/projects/dev/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/main.kt
```diff
@@ -89,15 +91,15 @@ private fun SessionFactory.loadCustomersAsDataFrame(): DataFrame<DfCustomers> =
...
             .resultList
             .map { c ->
                 DfCustomers(
                     address = c.address,
                     city = c.city,
                     company = c.company,
                     country = c.country,
-                    customerId = c.customerId ?: -1,
+                    customerId = c.customerId ?: error("Loaded customer must have a generated id"),
```
Allex-Nik, line 102: It seems better to fail soon in case of null id, even though this situation is highly unlikely

examples/projects/dev/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/main.kt
```diff
@@ -118,7 +120,7 @@ private fun SessionFactory.loadCustomersAsDataFrame(): DataFrame<DfCustomers> =
     }
 
 /** DTO used for aggregation projection. */
-private data class CountryCountDto(val country: String, val customerCount: Long)
+private data class CountryCountDto(val country: String?, val customerCount: Long)
```
Allex-Nik, line 123: `country` is nullable in `CustomersEntity`. Without this change, if some of the customers in `insertSampleData` is created without a `country`, the example would fail with NPE.

examples/projects/dev/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/main.kt
```diff
@@ -132,9 +134,8 @@ private fun SessionFactory.countCustomersPerCountryWithHibernate() {
         val cb = session.criteriaBuilder
         val cq: CriteriaQuery<CountryCountDto> = cb.createQuery(CountryCountDto::class.java)
         val root: Root<CustomersEntity> = cq.from(CustomersEntity::class.java)
-
-        val countryPath = root.get<String>("country")
```
Allex-Nik, line 136: I found a cool thing here. 

Hibernate has so-called [Static Metamodel Generator](https://docs.hibernate.org/stable/orm/userguide/html_single/#tooling-modelgen) which allows users to construct Criteria queries in a strongly-typed way. For each entity, it generates static metamodel classes, and their properties can be used for type-safe path references instead of strings.

Using it in this example would mean that this generator would create, among others, a `CustomersEntity_` class (and put it into the `build` folder), and we could write, for example, 
```kotlin
root.get(CustomersEntity_.country)
```
instead of
```kotlin
root.get<String?>("country")
```

I didn't use it here because even though it allows type-safe access, it
1) requires adding two more dependencies: `hibernate-processor` and `kapt`, so it seemed unjustified to add this to replace just two strings;
2) might introduce unnecessary complexity and distract users from the main thing we want to show them: how to use DataFrame :) 

What do you think?

examples/projects/dev/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/main.kt
```diff
@@ -132,9 +134,8 @@ private fun SessionFactory.countCustomersPerCountryWithHibernate() {
...
-        val idPath = root.get<Long>("customerId")
+        val countryPath = root.get<String?>("country")
+        val idPath = root.get<Int?>("customerId")
```
Allex-Nik: It's `Int?` in the `CustomersEntity`. `countExpr` will still have type `JpaExpression<Long!>!` because of the `count` function, so `customerCount` in `CountryCountDto` should remain `Long`.

examples/projects/dev/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/main.kt
```diff
@@ -175,22 +176,18 @@ private fun DataFrame<DfCustomers>.analyzeAndPrintResults() {
...
 
-private fun SessionFactory.replaceCustomersFromDataFrame(df: DataFrame<DfCustomers>) {
-    withTransaction { session ->
+private fun SessionFactory.replaceCustomersFromDataFrame(df: DataFrame<DfCustomers>) =
+    inTransaction { session ->
         val criteriaBuilder: CriteriaBuilder = session.criteriaBuilder
         val criteriaDelete: CriteriaDelete<CustomersEntity> =
             criteriaBuilder.createCriteriaDelete(CustomersEntity::class.java)
         criteriaDelete.from(CustomersEntity::class.java)
-
```
Allex-Nik, line 184: I think it should be a single transaction. Otherwise, if the deletion query succeeds and the second part fails, the transaction will revert to the state when the DB is empty.

examples/projects/dev/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/main.kt
```diff
@@ -209,40 +206,22 @@ private fun DataRow<DfCustomers>.toCustomersEntity(): CustomersEntity =
         supportRepId = this.supportRepId,
     )
 
-private inline fun <T> SessionFactory.withSession(block: (session: org.hibernate.Session) -> T): T =
```
Allex-Nik, line 212: I think we do not need `withSession` and `withTransaction` because their logic is roughly the same as in the built-in `inTransaction` method. The difference is that the exception handling block in rollback in the built-in version applies `addSuppressed` to the exception (instead of throwing it) without replacing the original exception.

examples/projects/dev/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/main.kt
```diff
@@ -209,40 +206,22 @@ private fun DataRow<DfCustomers>.toCustomersEntity(): CustomersEntity =
...
-}
-
 /** Read-only transaction helper for SELECT queries to minimize overhead. */
-private inline fun <T> SessionFactory.withReadOnlyTransaction(block: (session: org.hibernate.Session) -> T): T =
-    withSession { session ->
-        session.beginTransaction()
+private inline fun <T> SessionFactory.withReadOnlyTransaction(crossinline block: (session: Session) -> T): T =
+    fromTransaction { session ->
         // Minimize overhead for read operations
         session.isDefaultReadOnly = true
```
Allex-Nik, line 213: It was already there, just cool things to find out about optimizations for reading. 

```kotlin
session.isDefaultReadOnly = true
```
changes the default mode for entities and proxies in a session from modifiable to [read-only](https://docs.hibernate.org/orm/7.4/javadocs/org/hibernate/Session.html#setDefaultReadOnly(boolean)). This eliminates the need for Hibernate to dirty-check such objects. Dirty-checking is tracking if an entity was changed and synchronizing these changes with the database. Dirty-checking simplifies development, but introduces overhead: Hibernate has to keep two copies of each object (the snapshot and the live object), and Hibernate has to compare entities with their copies during a flush. With read-only entities, Hibernate doesn't do any of these.

```kotlin
session.hibernateFlushMode = FlushMode.MANUAL
```
[Recommended](https://docs.hibernate.org/orm/7.4/javadocs/org/hibernate/FlushMode.html#MANUAL) for read-only transactions. With `FlushMode.AUTO`, Hibernate synchronizes entities with the database before any query. When we can guarantee that the whole transaction is read-only, this synchronization is not needed.

examples/projects/dev/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/main.kt
```diff
@@ -209,40 +206,22 @@ private fun DataRow<DfCustomers>.toCustomersEntity(): CustomersEntity =
...
         session.hibernateFlushMode = FlushMode.MANUAL
-        try {
```
Allex-Nik, line 235: This is done by the built-in `fromTransaction` method

examples/projects/hibernate/gradle/libs.versions.toml
```diff
@@ -22,3 +22,5 @@ ktlint-gradle = { id = "org.jlleitschuh.gradle.ktlint", version.ref = "ktlint-gr
 
 # The Kotlin DataFrame Compiler plugin is the same version as the Kotlin plugin.
 kotlin-dataframe = { id = "org.jetbrains.kotlin.plugin.dataframe", version.ref = "kotlin" }
+
+kotlin-jpa = { id = "org.jetbrains.kotlin.plugin.jpa", version.ref = "kotlin" }
```
Allex-Nik, line 27: Should it be in the root `libs.versions.toml` as well? I saw that the Hibernate version was also specified in the root `libs.versions.toml`.

Jolanrensen: yep, please add all dependencies both in the examples as well as in the root toml file. This allows our dependency checker to see if there are any updates even if the specific example is not loaded :)

If you introduce new versions that need to be copied (you don't in your PR, but for the sake of explanation), they should also be added in the `dfbuild.buildExampleProjects` convention plugin

examples/projects/dev/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/main.kt
Jolanrensen: kapt? hmm, maybe if it were KSP it might be nice, but kapt is quite Java oriented. That said, if the "hibernate get started" guides use it, we should probably copy it. If it's just an optional addition, we can leave it out :)

examples/projects/dev/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/main.kt
Jolanrensen: indeed, best not reinvent the wheel

Jolanrensen (approved):
cool!

examples/projects/dev/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/main.kt
Allex-Nik: It seems that `hibernate-processor` doesn't support KSP :(

The guide says: you can use strings, but type-safety is a compelling advantage :)

Jolanrensen: hmmm, then it might be worth adding. We also boast about type-safety after all, plus it's a good test to see if our plugin and their don't interfere :)

Allex-Nik: Done, it works :)

examples/projects/hibernate/src/main/kotlin/org/jetbrains/kotlinx/dataframe/examples/hibernate/main.kt
```diff
@@ -132,9 +134,8 @@ private fun SessionFactory.countCustomersPerCountryWithHibernate() {
         val cb = session.criteriaBuilder
         val cq: CriteriaQuery<CountryCountDto> = cb.createQuery(CountryCountDto::class.java)
         val root: Root<CustomersEntity> = cq.from(CustomersEntity::class.java)
-
-        val countryPath = root.get<String>("country")
-        val idPath = root.get<Long>("customerId")
+        val countryPath = root.get(CustomersEntity_.country)
```
Jolanrensen, line 137: neat!

PR #1925, opened 2026-07-09
Clarify that schema of property with `DataFrame<T>?` type is not a FrameColumn

### LINKED ISSUE #1498: FieldKind.of logic is wrong for FrameColumn/ColumnGroup
For example expected type of `a` is `DataColumn<DataFrame<LocalType>?>`, but schema - `FrameColumn`.

```kt
val df = dataFrameOf("a" to columnOf(List(4) {
    dataFrameOf(
        "c" to columnOf(2)
    )
}))

val res = df.convert { a }.with { it.takeIf { it.rowsCount() % 2 == 0 } }
res[0].a?.c


// FieldKind.of is wrong =(
println(res.compileTimeSchema())
```

For now this logic is used in toDataFrame (library), compileTimeSchema (library), and extract schema from DataSchema declarations (compiler plugin, maybe in Jupyter integration)

Potential problems:
1. Incorrect compileTimeSchema
2. Extension properties for res are correct, but after another operation plugin will perform extract schema and probably some kind of runtime exception could happen while accessing `a/c` in df1
`val df1 = res.add("addMe") { 123 }`

koperagen:
This makes extracted schema represent runtime more accurately
With compiler plugin enabled, check this:
```kt
val df = dataFrameOf("a" to columnOf(List(4) {
    dataFrameOf(
        "c" to columnOf(2)
    )
}))
    .addId()

val res = df
    .convert { a }.with { it.takeIf { index() % 2 == 0 } }
    .add("f1") { a }

println(res.compileTimeSchema())
```

Reproduced by `extractNullableDataFrameSchema` test

Before, schema would lie to us that `a` and `f1` are FrameColumns (and has no nulls):

```
id: Int
a: *
    c: Int?
f1: *
    c: Int?
```
    

After:

```
id: Int
a: DataFrame<`FieldkindKt$main$df$2$A_651`>?
f1: DataFrame<`FieldkindKt$main$df$2$A_651`>?
```

Result is arguably not much better, because now suddenly we have not expanded structural dataframe type in the schema. In this case it is also anonymous and cannot be "materialized" by either generateCode or recent compiler warnings utilities. 
Resolving this usability problem will be our next step. We can either, idk, re-define operations in a way that nulls automatically replaced with empty frames, or fix codegen/schema/model to reflect possible DataFrame values better

Regarding changes made in this PR:
1. toDataFrame is expected to behave exactly as before, but it requires additional condition
2. convertTo used to be able to convert `FrameColumn<Schema>` to a `ValueColumn<DataFrame<Schema>?>` (which is valid operation) due to MarkersExtractor.get wrongly interpreting `DataFrame<Schema>?` as FrameColumn. I added branch that explicitly allows such pre-existing convertTo 

fixes #1498

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MarkersExtractor.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MarkersExtractor.kt
@@ -22,7 +22,7 @@ import kotlin.reflect.typeOf
 internal fun KType.getFieldKind(): FieldKind =
     FieldKind.of(
         this,
-        isDataFrame = { jvmErasure == DataFrame::class },
+        isDataFrame = { jvmErasure == DataFrame::class && !isMarkedNullable },
         isListToFrame = {
             jvmErasure == List::class && (arguments[0].type?.jvmErasure?.hasAnnotation<DataSchema>() == true)
         },
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convertTo.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convertTo.kt
@@ -193,7 +193,14 @@ internal fun AnyFrame.convertToImpl(
 
                         when (targetSchema.kind) {
                             ColumnKind.Value ->
-                                convertedColumn ?: originalColumn.convertTo(to)
+                                when {
+                                    convertedColumn != null -> convertedColumn
+
+                                    originalColumn.kind == ColumnKind.Frame && to.jvmErasure == DataFrame::class ->
+                                        originalColumn
+
+                                    else -> originalColumn.convertTo(to)
+                                }
 
                             ColumnKind.Group -> {
                                 val column = when {
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/convertTo.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/convertTo.kt
@@ -184,6 +184,21 @@ class ConvertToTests {
     @DataSchema
     data class DataSchemaWithAnyFrame(val dfs: AnyFrame?)
 
+    private fun locationsFrame(): DataFrame<Location?> =
+        listOf(
+            Location("Home", Gps(0.0, 0.0)),
+            Location("Away", null),
+            null,
+        ).toDataFrame()
+            .alsoDebug("locations:")
+
+    private fun gpsFrame(): DataFrame<Gps?> =
+        listOf(
+            Gps(0.0, 0.0),
+            null,
+        ).toDataFrame()
+            .alsoDebug("gps:")
+
     @Test
     fun test() {
         val df1 = dataFrameOf("a")(1, 2, 3)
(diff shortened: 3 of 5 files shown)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/convertTo.kt
```diff
@@ -394,4 +412,11 @@ class ConvertToTests {
         converted["a"].type() shouldBe typeOf<SimpleEnum?>()
         converted shouldBe dataFrameOf("a")(SimpleEnum.A, SimpleEnum.B, null)
     }
+
```
Jolanrensen: what about this?

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/convertTo.kt
```diff
@@ -184,6 +184,21 @@ class ConvertToTests {
     @DataSchema
     data class DataSchemaWithAnyFrame(val dfs: AnyFrame?)
 
+    private fun locationsFrame(): DataFrame<Location?> =
```
Jolanrensen, line 187: What does a `DataFrame` with a nullable type mean? `DataFrame<Location>` makes sense: you have two columns "name" and "gps". But I don't know how to interpret `DataFrame<Location?>`.

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/convertTo.kt
koperagen: leftovers, will remove

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/convertTo.kt
Jolanrensen: should we prohibit such types? Or explicitly allow them (changing "name" and "gps" to nullable versions and fill them with nulls, like in our json reader)

koperagen: `Iterable<T>.toDataFrame(): DataFrame<T>`
In this case T just happens to be Location?
Let's say, with compiler plugin enabled this will not happen

Jolanrensen: Oh, that's indeed exactly what happens, I see.

Jolanrensen: Indeed, the compiler plugin will come up with a new type. But it might be worth to define this behavior as a rule (in docs, tests, etc).

- `DataFrame<T>`: `T` is expected to be not nullable. If it is nullable, all value-columns will become nullable and all frame-columns can contain empty dataframes.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MarkersExtractor.kt
```diff
@@ -22,7 +22,7 @@ import kotlin.reflect.typeOf
 internal fun KType.getFieldKind(): FieldKind =
     FieldKind.of(
```
Jolanrensen, line 23: `FieldKind.of` is a very strange factory function... Is it this way because it's used elsewhere? like in the compiler plugin? :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convertTo.kt
```diff
@@ -193,7 +193,14 @@ internal fun AnyFrame.convertToImpl(
 
                         when (targetSchema.kind) {
                             ColumnKind.Value ->
-                                convertedColumn ?: originalColumn.convertTo(to)
+                                when {
+                                    convertedColumn != null -> convertedColumn
+
+                                    originalColumn.kind == ColumnKind.Frame && to.jvmErasure == DataFrame::class ->
+                                        originalColumn
```
Jolanrensen, line 200: if you return `originalColumn` the `kind` won't have changed to `Value`, right?

Jolanrensen: Plus, if it does change the kind, this might enable converting `FrameColumn -> ValueColumn<AnyFrame>` which is not allowed

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MarkersExtractor.kt
koperagen: indeed, it is used there

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/convertTo.kt
koperagen: If originalColumn has no nulls in runtime then creating ValueColumn is a violation and our checks fail
So yes.. 
It seems to create misalignment one way or another.

Jolanrensen: So isn't this step then gonna break regardless? It converts FrameColumn -> ValueColumn
But FrameColumn cannot contain nulls, so it will always need to create a ValueColumn<DataFrame<>> which is not allowed. 
I don't think there's conversion possible from FrameColumn -> ValueColumn. Maybe the other way around but only if the nulls are turned into empty dataframes

koperagen: Approach that i chose now is keeping runtime column "narrower" than compile time - it's usually a valid approach. So, even if target type is `DataFrame<T>?` we just keep runtime FrameColumn where possible
let's say convertTo<...>.dropNulls { nullableDataFrameCol into "frameCol" } => valid frameCol
Problem is that it's turns out not only narrow, but also of a different kind ...

Jolanrensen: I think logic like that will only complicate things more... Can we try to keep the rules the same everywhere?
- `DataColumn<DataFrame<>?>` must always be a `ValueColumn`, in compile time as well as runtime
- `DataColumn<DataFrame<>>` must always be a `FrameColumn`, in both cases too

This will make it much easier for us and to explain to users. So if the target type is `DataColumn<DataFrame<>?>`, we turn it into a value column, if it's`DataColumn<DataFrame<>>`, it becomes a frame column.

> `convertTo<...>.dropNulls { nullableDataFrameCol into "frameCol" } => valid frameCol`

This will then also work correctly, right? because `nullableDataFrameCol: DataFrame<>?` so it must have been a `ValueColumn`, then after `dropNulls`, it automatically turns into a `FrameColumn`
(If this doesn't work for `dropNulls` yet, that should probably be fixed. It works correctly for `update`)

koperagen: `DataColumn<DataFrame<>?> must always be a ValueColumn, in compile time as well as runtime` - existing check prevents that
If field is DataFrame<>? and in runtime there were no nulls, according to this logic `toDataFrame` (changing topic from convertTo because problem originates there) should create a ValueColumn - but we forbid that

Jolanrensen: I see now... that's indeed a problem. In that case `DataColumn<DataFrame<>?>` can indeed be a `FrameColumn` in runtime too... Alright:
- `DataColumn<DataFrame<>>` must always be a `FrameColumn` in runtime
- `DataColumn<DataFrame<>?>` is likely a `ValueColumn` but can also be a `FrameColumn` in runtime because it may not contain nulls.

I suppose from the `convertToImpl` this means we should treat ColumnKind a bit more loosely when dealing with _nullable DataFrame_ types. But only for those. If we're explicit enough about this case it should be fine I think :)

koperagen: Yes, this matches my understanding of the situation.

> I suppose from the convertToImpl this means we should treat ColumnKind a bit more loosely when dealing with nullable DataFrame types. But only for those. If we're explicit enough about this case it should be fine I think :)

Yeah, this should be resolved.
I see situation like this: we had a bug that used to hide a few design problems. Now the bugs is fixed, but design problems become more apparent. Same as before having DataFrame<T>? is error prone - before it could lead to NPE, now it leads to inconsistency if specific column selectors are used, for example, or when code for schema is generated.
So i'd say `DataFrame<T>?` in `@DataSchema` is not recommended, `DataFrame<T>?` value in add() { } can appear if needed but has usability issues around schema printing and codegen

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/convertTo.kt
koperagen: ok, but as a different issue

PR #1928, opened 2026-07-10
Fixes slicing `ValueColumn<AnyFrame?> -> FrameColumn`

### LINKED ISSUE #1926: Slicing operations on `DataColumn<AnyFrame?>` can throw incorrect exceptions
This should not throw an exception, instead, a `FrameColumn` should be created.

To reproduce:
```kt
val df = dataFrameOf("col")(0)
val col = DataColumn.createValueColumn("col", listOf(df, null))
col.dropNulls() // IllegalArgumentException: ValueColumnImpl cannot contain AnyFrame values. Use FrameColumn instead.
```
This applies to any filtering function, or more specifically, every function that calls `fun get(indices: Iterable<Int>): DataColumn<T>`  or `fun get(range: IntRange): ValueColumn<T>` on `ValueColumnImpl`, including `filter {}`, `dropNa {}`, `take()`, `getRows()` etc.

Functions like `update` treat this correctly and turn the column into a frame column:

```kt
val df = dataFrameOf(
    DataColumn.createValueColumn("col", listOf(df, null)),
)
df["col"].let { // checking we actually have a value column here
    it.kind() shouldBe ColumnKind.Value
     it.type().isSubtypeOf(typeOf<DataFrame<*>?>()) shouldBe true
}
val updatedDf = df.update("col").where { it == null }.with { emptyDataFrame<Person>() }
updatedDf["col"].let {
    // correct
    it.kind() shouldBe ColumnKind.Frame
    it.type().isSubtypeOf(typeOf<DataFrame<*>>()) shouldBe true
}

val droppedDf = df.dropNulls("col") // ValueColumnImpl cannot contain AnyFrame values. Use FrameColumn instead.
droppedDf["col"].let {
    it.kind() shouldBe ColumnKind.Frame
    it.type().isSubtypeOf(typeOf<DataFrame<*>>()) shouldBe true
}
```

Jolanrensen:
Fixes #1926 by letting slicing operations on ValueColumn potentially return a FrameColumn.

Like #1925 this points at a deeper issue with the way we handle dataframes inside columns, but at least does this make it consistent with `update`, `convert`, `replace` etc.

### DIFF
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/drop.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/drop.kt
@@ -1,11 +1,48 @@
 package org.jetbrains.kotlinx.dataframe.api
 
+import io.kotest.matchers.shouldBe
+import org.jetbrains.kotlinx.dataframe.DataColumn
+import org.jetbrains.kotlinx.dataframe.DataFrame
+import org.jetbrains.kotlinx.dataframe.columns.ColumnKind
 import org.jetbrains.kotlinx.dataframe.samples.api.firstName
 import org.jetbrains.kotlinx.dataframe.samples.api.lastName
 import org.jetbrains.kotlinx.dataframe.samples.api.name
 import org.junit.Test
+import kotlin.reflect.full.isSubtypeOf
+import kotlin.reflect.typeOf
 
 class DropTests : ColumnsSelectionDslTests() {
+
+    // Issue #1926
+    @Test
+    fun `removing nulls from ValueCol of DF should turn into FrameCol`() {
+        val df = dataFrameOf(
+            DataColumn.createValueColumn("col", listOf(df, null)),
+        )
+        df["col"].let {
+            it.kind() shouldBe ColumnKind.Value
+            it.type() shouldBe typeOf<DataFrame<Person>?>()
+        }
+        val updatedDf = df.update("col").where { it == null }.with { emptyDataFrame<Person>() }
+        updatedDf["col"].let {
+            // correct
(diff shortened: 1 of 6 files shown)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/drop.kt
```diff
@@ -1,12 +1,49 @@
...
+            it.type() shouldBe typeOf<DataFrame<Person>?>()
+        }
+        val updatedDf = df.update("col").where { it == null }.with { emptyDataFrame<Person>() }
+        updatedDf["col"].let {
+            // correct
+            it.kind() shouldBe ColumnKind.Frame
+            it.type().isSubtypeOf(typeOf<DataFrame<*>>()) shouldBe true
+        }
+
+        df["col"].dropNulls() // should work correctly
```
koperagen: wrap with shouldNotThrow

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/drop.kt
```diff
@@ -1,12 +1,49 @@
...
+        val droppedDf = df.dropNulls("col")
+        droppedDf["col"].let {
```
koperagen: please split into multiple tests

koperagen: nested class if you prefer to have some common "subject under test"

Jolanrensen: Actually, I think this test(s) should probably be moved elsewhere anyways, since we don't necessarily test `drop`, but some lower level functionality

PR #1936, opened 2026-07-20
Added `GroupBy` transformation KDocs

### LINKED ISSUE #1526: Add `groupBy` and `pivot` transformations KDocs
### Problem

`groupBy` and `pivot` transformation operations currently lack KDocs.

These are core APIs with non-trivial behavior and DSL usage, but users have to rely on external documentation or guess semantics.

### Expected

Add KDocs for all transformation-related `groupBy` and `pivot` operations.

### Acceptance criteria

- KDocs are added for:
  - `groupBy` transformation APIs
  - `pivot` transformation APIs
- Each KDoc includes:
  - short description of behavior
  - at least one usage example
  - explanation of how schema is affected (if applicable)
- KDocs are consistent with existing style in other DataFrame APIs
- Links to website documentation are added where relevant

### Why it matters

- These are core operations used in most pipelines
- Missing KDocs makes API hard to discover and use correctly
- Public API documentation should be complete before 1.0

AndreiKingsley:
Closes #1526 + KDocs for `DataFrame.sort..()` methods + renamed `.desc()` to `.reversed()`


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/SelectingColumns.kt
```diff
@@ -81,6 +81,9 @@ internal interface SelectingColumns {
      * This is an entity formed by calling any (combination) of the functions
      * in the DSL that is or can be resolved into one or more columns.
      *
+     * Columns Selection DSL allows using [Extension Properties][AccessApis.ExtensionPropertiesApi]
```
Jolanrensen: *The Columns Selection DSL

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/SelectingColumns.kt
```diff
@@ -134,6 +137,8 @@ internal interface SelectingColumns {
      * This is an entity formed by calling any (combination) of the functions
      * in the DSL that is or can be resolved into a single column.
      *
+     * Column Selection DSL allows using [Extension Properties][AccessApis.ExtensionPropertiesApi]
```
Jolanrensen: *The Columns Selection DSL

Jolanrensen:
please update the description to include the `.reversed()` function :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -15,36 +17,130 @@ import org.jetbrains.kotlinx.dataframe.columns.SingleColumn
...
 import org.jetbrains.kotlinx.dataframe.index
+import org.jetbrains.kotlinx.dataframe.name
 import org.jetbrains.kotlinx.dataframe.nrow
 import org.jetbrains.kotlinx.dataframe.type
 import org.jetbrains.kotlinx.dataframe.util.DEPRECATED_ACCESS_API
+import org.jetbrains.kotlinx.dataframe.util.DESC_TO_REVERSED
 import kotlin.reflect.KProperty
 
+/**
+ * [SortDsl] allows selecting columns to sort rows by (the order in which columns are selected
```
Jolanrensen: `The [Sort DSL][SortDsl]`

Jolanrensen: "by" what?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -15,36 +17,130 @@ import org.jetbrains.kotlinx.dataframe.columns.SingleColumn
...
+ * determines the sort priority).
+ * It also allows reversing the sort order for individual columns or column sets,
+ * and controlling the position of `null` values.
+ *
+ * The order in which columns are selected determines the sort priority.
```
Jolanrensen, line 45: duplicate of above. Maybe the first occurence should be removed?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -15,36 +17,130 @@ import org.jetbrains.kotlinx.dataframe.columns.SingleColumn
...
+ * Use [\reversed] to impose the reverse ordering for a column or column set.
+ *
+ * By default, `null` values are considered the smallest values when sorting.
+ * Use [nullsLast] to treat them as the largest values.
+ */
+@ExcludeFromSources
+private typealias SortDslSnippet = Nothing
+
+/**
+ * A specialized [ColumnsSelectionDsl] for selecting columns to sort rows by
```
Jolanrensen: use `{@include [ColumnsSelectionDslLink]}`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -15,36 +17,130 @@ import org.jetbrains.kotlinx.dataframe.columns.SingleColumn
...
+ * // Sort rows by the product of "volume" and "quantity" descending
+ * df.sortBy { expr { volume * quantity }.reversed() }
+ * ```
+ */
 public interface SortDsl<out T> : ColumnsSelectionDsl<T> {
+
+    /**
+     * Marks the selected columns to be sorted in reversed order.
+     */
+    public fun <C> ColumnSet<C>.reversed(): ColumnSet<C> = addFlag(SortFlag.Reversed)
```
Jolanrensen, line 85: lol, the sort-flag was already named "reversed". Yeah, that's definitely a better name :)

Jolanrensen: Should we use `C : Comparable<*>`? or is that too pushy for String API usage? (`cols("a", "b").reversed()` would fail unless the types are provided explicitly) We do have these restrictions on `median()` etc.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -15,36 +17,130 @@ import org.jetbrains.kotlinx.dataframe.columns.SingleColumn
...
+
+    /**
+     * Marks the selected column to be sorted in reversed order.
+     */
+    public fun <C> SingleColumn<C>.reversed(): SingleColumn<C> = addFlag(SortFlag.Reversed).single()
```
Jolanrensen, line 90: Same here, `C` could be marked self-comparable, so `<C : Comparable<C & Any>?>` unless we want to allow `col("a").reversed()`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -15,36 +17,130 @@ import org.jetbrains.kotlinx.dataframe.columns.SingleColumn
...
+
+    /**
+     * Marks the selected column to be sorted in reversed order.
+     */
+    public fun String.reversed(): SingleColumn<Comparable<*>?> = invoke<Comparable<*>>().desc()
```
Jolanrensen: `invoke` should probably also have a nullable type

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -15,36 +17,130 @@ import org.jetbrains.kotlinx.dataframe.columns.SingleColumn
...
+     *
+     * When [flag] is `false`, the column remains unchanged.
+     *
+     * @param flag whether `null` values should be placed last.
+     */
     public fun <C> SingleColumn<C?>.nullsLast(flag: Boolean = true): SingleColumn<C?> =
         if (flag) addFlag(SortFlag.NullsLast).single() else this
 
+    /**
+     * Places `null` values after non-null values when sorting by this column.
```
Jolanrensen: there's mixed usage of smaller/bigger, before/after, ascending/descending making it a bit confusing. Maybe we should stick to "less"/"greater", as that's what "Comparable" itself uses. We should also mention "index" somewhere, as in "for ascending, we mean the greater the index, the greater the value". Hopefully this avoids confusion about our DataFrames expanding downwards, meaning when "ascending" the smallest values are shown at the top and largest at the bottom of the DataFrame

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -54,28 +150,48 @@ public interface SortDsl<out T> : ColumnsSelectionDsl<T> {
...
 /**
- * [SortColumnsSelector] is used to express or select multiple columns to sort by, represented by [ColumnsResolver]`<C>`,
- * using the context of [SortDsl]`<T>` as `this` and `it`.
+ * A specialized [ColumnsSelector] used for selecting columns for sorting.
  *
- * So:
- * ```kotlin
- * SortDsl<T>.(it: SortDsl<T>) -> ColumnSet<C>
- * ```
+ * Provides [SortDsl] both as the receiver (`this`) and the lambda parameter (`it`), and expects
```
Jolanrensen, line 167: `Opens the [Sort DSL][SortDsl] scope`? may be clearer phrased

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -54,28 +150,48 @@ public interface SortDsl<out T> : ColumnsSelectionDsl<T> {
...
+ * a [ColumnsResolver] as the return value.
+ *
+ * Enables defining the descending ordering of sort columns using [desc][SortDsl.desc]
```
Jolanrensen: *reversed

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -54,28 +150,48 @@ public interface SortDsl<out T> : ColumnsSelectionDsl<T> {
...
+ *
+ * Accepts only [Comparable] values.
+ *
+ * See also
+ *   - [sortDesc][DataColumn.sortDesc] for sorting in descending order;
+ *   - [sortWith][DataColumn.sortWith] for sorting by providing a custom comparator.
+ *
+ * @return A new [ValueColumn] containing the sorted values from the original column.
+ */
 public fun <T : Comparable<T>> DataColumn<T>.sort(): ValueColumn<T> =
```
Jolanrensen, line 190: :o this function breaks on `col<Int?>("a").sort()` because `Int` isn't `Comparable<Int?>`. We should have `<T : Comparable<T & Any>?>`. Maybe we can fix that already?

Jolanrensen: otherwise we should open an issue

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -54,28 +150,48 @@ public interface SortDsl<out T> : ColumnsSelectionDsl<T> {
...
     DataColumn.createValueColumn(name, values().sorted(), type, defaultValue = defaultValue())
 
+/**
+ * Sorts the values in this [column][DataColumn] in descengind order.
```
Jolanrensen: typo

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
Jolanrensen: this applies to other functions below as well

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -96,40 +212,188 @@ public fun <T : Comparable<T>> DataColumn<T>.sortDesc(): ValueColumn<T> =
...
+/**
+ * Sorts this [DataFrame] rows by the specified [columns] in ascending or descending (default) order.
+ *
+ * Returns a new [DataFrame] containing the same rows, sorted according to the selected columns.
+ *
+ * Select columns to sort by, adjust sorting order and `null`s position using [SortDsl].
+ *
+ * @include [SortDslSnippet]
+ *
+ * See {@include [CSDslLink]}.
```
Jolanrensen, line 386: ah it seems we have two of these links

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -162,8 +484,77 @@ public fun <T, G> GroupBy<T, G>.sortBy(vararg cols: ColumnReference<*>): GroupBy
...
+ *
+ * ### Examples
+ * ```kotlin
+ * gb.sortByDesc("age")
+ * // Sort group rows by "age" column values descending
+ * // and then by the ("name") column values descending
+ * gb.sortByDesc("age", "name")
+ * ```
+ *
+ * @param cols The [Column Names][String] that defines which columns are used for sorting.
```
Jolanrensen: *column names that define

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -190,6 +620,37 @@ private fun <T, G, C> GroupBy<T, G>.createColumnFromGroupExpression(
...
+ *
+ * For more information: {@include [DocumentationUrls.GroupByTransformation]}
+ *
+ * ### Example
+ * ```kotlin
+ * // Sorts groups by the product of their size and the maximum value of the "quantity" column ascending
+ * gb.sortByGroup { group -> group.rowsCount() * group.quantity.max() }
+ * ```
+ *
+ * @param [nullsLast] Whether `null` values should be placed after non-null values.
```
Jolanrensen: again "after"? better: "considered greater than non-null values"

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
Jolanrensen: Hmm actually, how  does `nullsLast` work? is it always "last" regardless of whether we're ascending or descending? This should be explained clearly

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
Jolanrensen: actually... if "last" always means "last", like I commented before, it's probably fine. I'm just a bit confused about ascending/descending + nullsFirst/nullsLast

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -206,15 +698,93 @@ public fun <T, G, C> GroupBy<T, G>.sortByGroupDesc(
...
+ */
 public fun <T, G> GroupBy<T, G>.sortByCountAsc(): GroupBy<T, G> = sortByGroup { nrow }
 
+/**
+ * Sorts the key-group pairs of this [GroupBy] by the numer of rows in groups descending.
+ *
+ * Returns a new [GroupBy] containing the same keys and groups in the resulting order.
+ * The contents of the groups remain unchanged.
+ *
+ * Don't confuse this with [sortBy][GroupBy.sortBy] that sorts the group rows.
```
Jolanrensen, line 751: nice

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -206,15 +698,93 @@ public fun <T, G, C> GroupBy<T, G>.sortByGroupDesc(
...
+ * Check out [`GroupBy grammar`][Grammar].
+ *
+ * For more information: {@include [DocumentationUrls.GroupByTransformation]}
+ *
+ * @param [nullsLast] Whether `null` values should be placed after non-null values.
+ * @return A new [GroupBy] with its key-group pairs sorted by the keys.
+ */
 public fun <T, G> GroupBy<T, G>.sortByKeyDesc(nullsLast: Boolean = false): GroupBy<T, G> =
     toDataFrame()
         .sortBy { keys.columns().toColumnSet().desc().nullsLast(nullsLast) }
```
Jolanrensen: deprecated function being used

Jolanrensen (changes requested):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
AndreiKingsley: ?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
AndreiKingsley: Let's open a general issue for all sort related methods

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
AndreiKingsley: Why don't you like it? 
It actually does what it said.

Jolanrensen: It may be a bit verbose, but sure we can keep it

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
Jolanrensen: `{@include [ColumnsSelectionDslLink]}` and `{@include [CSDslLink]}` do the same thing

AndreiKingsley: I will fix it in another PR!

AndreiKingsley:
@Jolanrensen updated `nullsLast` description

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sort.kt
```diff
@@ -115,7 +115,11 @@ public interface SortDsl<out T> : ColumnsSelectionDsl<T> {
     public fun <C> KProperty<C>.desc(): SingleColumn<C> = toColumnAccessor().desc()
 
     /**
-     * Places `null` values after non-null values when sorting by the selected columns.
+     * Places `null` values after non-null values when sorting in ascending order,
+     * or before non-null values when sorting in descending order by these columns.
+     *
+     * By default, `null` values are considered the smallest values when sorting.
+     * Use this method to treat them as the largest values.
```
Jolanrensen, line 122: Clear! Thanks :D

Jolanrensen (approved):

PR #1942, opened 2026-07-22
Updated `print` result in quickstart.md

koperagen:


AndreiKingsley (approved):

Jolanrensen:
We should probably migrate this to a text output with Korro

AndreiKingsley:
@Jolanrensen oops, I missed this when approved:
This output needs a custom indent because by default spaces before the first chat are ignored and columns name are not aligned with columns:
image

AndreiKingsley:
So it can't be just done via korro text output, needs to be modified manually

PR #1946, opened 2026-07-23
Handle SQLite BOOLEAN columns correctly and add tests

### LINKED ISSUE #1935: Verify SQLite Boolean column handling in Gradle and Maven projects
## Problem

Issue #1013 reports a SQLite Boolean-related problem in Kotlin Notebook:

- SQLite stores Boolean-like values as integers (`0` / `1`);
- generated property access expects `Boolean`;
- direct property access in Notebook can fail with:

```text
ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Boolean
```

The same report also mentions that in a regular IntelliJ/Gradle class, direct access fails earlier with:

```text
Unresolved reference: is_active
```

That part may be a separate schema/accessor-generation issue rather than the same SQLite Boolean mapping problem.

We need to verify whether the actual SQLite Boolean conversion/casting issue is reproducible in regular Gradle and Maven projects when schema/accessor setup is configured correctly.

## Goal

Check whether SQLite columns declared or inferred as Boolean are handled consistently outside Kotlin Notebook.

### Gradle

Create or reuse a Gradle sample with:

- SQLite database containing a Boolean-like column stored as `0` / `1`;
- JDBC/DataFrame read path;
- explicit schema or generated accessors;
- direct typed access to the Boolean column.

Verify whether:

```kotlin
df.is_active
```

or equivalent typed access works, fails with `ClassCastException`, or requires explicit conversion.

### Maven

Repeat the same verification in a Maven project, if Maven support is expected for the same workflow.

## Expected Behavior

One of the following should be clearly defined:

- SQLite `0` / `1` values are converted to `Boolean` when schema expects `Boolean`; or
- SQLite Boolean-like columns are read as `Int`, and users must convert them explicitly; or
- a clear diagnostic explains that SQLite does not have a native Boolean type and that `0` / `1` values cannot be accessed as `Boolean` without conversion.

The behavior should be consistent across Notebook, Gradle, and Maven where the same schema/accessor setup is available.

## Definition of Done

- Verify the issue in a regular Gradle project.
- Verify the issue in a Maven project, if applicable.
- Document whether the Notebook failure from #1013 reproduces outside Notebook.
- If reproducible:
  - identify whether the problem is in JDBC type inference, schema/accessor generation, or row/property access;
  - add a minimal regression test or sample;
  - fix conversion or improve diagnostics.
- If not reproducible:
  - document why #1013 is Notebook-specific;
  - close or narrow #1013 accordingly.
- Link findings back to #1013.

## Related

- #1013
- Stack Overflow report: https://stackoverflow.com/questions/79333504/kotlin-dataframe-sqlite-integer-cannot-be-cast-to-boolean-in-notebook
- SQLite type documentation: https://www.sqlite.org/datatype3.html

AndreiKingsley:
Fixes #1935

### DIFF
--- a/dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/sqliteTest.kt
+++ b/dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/sqliteTest.kt
@@ -14,6 +14,7 @@ import java.io.File
 import java.nio.file.Files
 import java.sql.Connection
 import java.sql.DriverManager
+import java.sql.Types
 import kotlin.reflect.typeOf
 
 @DataSchema
@@ -45,6 +46,13 @@ interface CustomerOrderSQLite {
     val orderDate: String?
     val totalAmount: Double
     val orderDetails: ByteArray?
+}
+
+@DataSchema
+interface FlagSQLite {
+    val id: Int?
+    val enabled: Boolean
+    val optional: Boolean?
 }
 
 class SqliteTest {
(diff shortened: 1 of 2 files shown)


dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/sqliteTest.kt
```diff
@@ -247,18 +278,34 @@ class SqliteTest {
...
+        val flagsTableName = "Flags"
+        val df = DataFrame.readSqlTable(connection, flagsTableName).cast<FlagSQLite>()
+
+        df.rowsCount() shouldBe 2
+        df["enabled"][0] shouldBe true
+        df["enabled"][1] shouldBe false
+        df["optional"][0] shouldBe false
+        df["optional"][1] shouldBe null
+
+        val schema = DataFrameSchema.readSqlTable(connection, flagsTableName)
```
koperagen, line 307: Please add check for df["optional"].type() too, to make sure same logic applies

PR #1947, opened 2026-07-23
Hide TODO page

koperagen:
No idea what content to put there. All other root pages seem to have additional content about mix of operations, but for values we only have.. values? Would be strange to have Get values (some info about values()) and then values (some other info about values())
image


Jolanrensen (approved):
strange page indeed, best to remove it altogether.

PR #1949, opened 2026-07-24
Restructured setup guides to have compiler plugin closer to library dependency setup

koperagen:
1. Joined compiler plugin setup with library dependency setup as we discussed
2. Updated maven setup, took inspiration from https://kotlinlang.org/docs/power-assert.html#maven
3. Compacted Android setup. I believe all these JDK/minSdk were there only because we used to have a lot of extras for Apache Arrow setup. Turns out it doesn't work at all on Android. Our library doesn't impose any additional restrictions on JDK, i believe. Showing setup of 1.8 everywhere seems redundant. Maaybe we should leave minSdk? Let me know what you think.

### DIFF
--- a/docs/StardustDocs/topics/setup/SetupAndroid.md
+++ b/docs/StardustDocs/topics/setup/SetupAndroid.md
@@ -31,23 +31,6 @@ dependencies {
     // You can add any additional IO modules you like, except for 'dataframe-arrow'.
     // Apache Arrow is not supported well on Android.
 }
-
-android {
-    defaultConfig {
-        minSdk = 21
-    }
-    // Requires Java 8 or higher
-    compileOptions {
-        sourceCompatibility = JavaVersion.VERSION_1_8
-        targetCompatibility = JavaVersion.VERSION_1_8
-    }
-    kotlinOptions {
-        jvmTarget = "1.8"
-    }
-}
-tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> { 
-    kotlinOptions.jvmTarget = "1.8" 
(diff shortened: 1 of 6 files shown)


docs/StardustDocs/topics/setup/SetupAndroid.md
```diff
@@ -55,61 +40,38 @@ tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
...
-    kotlin("plugin.dataframe") version "%compilerPluginKotlinVersion%"
+dependencies {
+    // Core Kotlin DataFrame API, CSV and JSON IO.
+    // See custom Gradle setup:
+    // https://kotlin.github.io/dataframe/setupcustomgradle.html
+    implementation("org.jetbrains.kotlinx:dataframe-core:%dataFrameVersion%")
+    implementation("org.jetbrains.kotlinx:dataframe-json:%dataFrameVersion%")
+    implementation("org.jetbrains.kotlinx:dataframe-csv:%dataFrameVersion%")
+    // You can add any additional IO modules you like, except for 'dataframe-arrow'.
+    // Apache Arrow is not supported well on Android.
```
Jolanrensen: *not well supported

Jolanrensen:
We can leave out `minSdk` I believe. Recently I had to bump it to 23 in our example, but this was due to AGP restrictions, not due to DataFrame. Targeting a higher JVM should also be fine, I think, so leaving it out too is probably best

AndreiKingsley:
I will take a look!

AndreiKingsley (approved):

Issue #1797, opened 2026-04-12
Possible `Sqlite` enhancement


AndreiKingsley:
See https://github.com/Kotlin/dataframe/pull/1741/changes/BASE..7ee5de69722473434af15238cbaac320a77b4add#r3064970981

PR #2001, opened 2026-07-27
Sqlite types fix

### LINKED ISSUE #1013: Kotlin Dataframe SQLite Integer cannot be cast to Boolean in Notebook
I make this issue following a request from @zaleslaw on [Stack Overflow](https://stackoverflow.com/questions/79333504/kotlin-dataframe-sqlite-integer-cannot-be-cast-to-boolean-in-notebook).

### What happens : 

Given a SQLite database, with a Boolean (nullable does not change the behavior) column, directly trying to access the value throws a `java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Boolean (java.lang.Integer and java.lang.Boolean are in module java.base of loader 'bootstrap')`.

Using indexing `[""]` or `.get()` works as expected.

Additionally, when running the same code outside of a Notebook (in a normal IntelliJ class), using direct access throws `"Unresolved reference: is_active"` error.

### What I expect : 

I expect at least to see the same (semantically understandable) type of error than in Main.kt. 
Additionally, the fact that IntelliJ offers completion for the direct access makes it confusing (and is the reason I ran into the error). If this isn't possible, I'd like to not see this completion. (The completion isn't present in a non Notebook environment for me).

image

I also found property access [in your documentation](https://kotlin.github.io/dataframe/indexing.html), so I am expecting it is possible to use it this way. 

### How to reproduce : 

I have created a small reproduceable repository [here](https://github.com/jlengrand/kotlin-sqlite-dataframe.git).

Running the `mainNotebook.ipynb` notebook should reproduce the issue. I have also created a `Main.kt` file for reference. 

Note : SQLite [does not have a Boolean type](https://www.sqlite.org/datatype3.html), and instead relies on 1s and 0s, this seems to be related.

Let me know if I can help more!

### LINKED ISSUE #1747: Improve SQLite types handling
Orginal problem in #964.

In order to support custom types not handled correctly by [SQLite type affinity](https://www.sqlite.org/datatype3.html#type_affinity), custom type mapping possibility was added in #1741.

However that might not be enough. Now we only support type definitions based on `sqlTypeName`. However, this may not be the single determining factor. 
So it seems that there are a lot of edge cases of determining the JDBC type in SQLite, and several column type determination mechanisms can be particularly relevant for it:
* column name -> type (like `colTypes: Map<String, ColType>` in CSV IO)
* may be column metadata (not only `sqlTypeName`) -> type
* inferring from values?

It's also important not only for custom types, but for simple `INTEGER` nullable columns .
Xerial SQLite JDBC driver seems to give identical metadata for them, however actually types of values can be both `Int?` and `Long?` (looks like a separate issue perhaps).
```
TableColumnMetadata(name='int_col', sqlTypeName='INTEGER', jdbcType=4, size=2147483647, javaClassName='java.lang.Object', isNullable=true)
TableColumnMetadata(name='long_col', sqlTypeName='INTEGER', jdbcType=4, size=2147483647, javaClassName='java.lang.Object', isNullable=true)
```

Need to investigate different usecases and and practices and improve handling of SQLite types.

### LINKED ISSUE #1935: Verify SQLite Boolean column handling in Gradle and Maven projects

AndreiKingsley:
Fixes #1013.
Fixes #1935. 
Fixes #1747.

Helps #1797.

Significantly improved different types support for SQLite `DbType`.

* Add out-of-box support for `Boolean` and date-time types. They don't have a their own storage classes inside SQLite, but they have an official specification, so we can detect them from column metadata name and try to extract from possible SQLite storage classes (primitives like `Int`, `Long`, `String`, etc., see full list in code).

* Add a special DSL for specifying any other custom SQLite type , by providing converter from storage classes to expected Kotlin types. Column `KType`s are infered automatically!


dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,516 @@
...
+        expectedJdbcType: KType,
+        expectedPreprocessedValueType: KType,
+    ): D {
+        customMappingFor(tableColumnMetadata)?.let { (_, convert) ->
+            @Suppress("UNCHECKED_CAST")
+            return convert(value) as D
+        }
+        val target = expectedPreprocessedValueType.classifier
+        @Suppress("UNCHECKED_CAST")
+        return when (target) {
```
Jolanrensen: best to stick to KTypes, they're cheaper than KClasses AFAIK

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,516 @@
...
+
+        /**
+         * Register a converter for a specific column by name. Column-name overrides take
+         * precedence over type-name overrides registered via [forType].
+         *
+         * @param T the storage class of the raw stored value.
+         * @param R the target Kotlin type for the resulting DataFrame column.
+         */
+        public inline fun <T, reified R> forColumn(columnName: String, crossinline convert: (T) -> R) {
+            val mapping: SqliteCustomTypeConverter<T> = typeOf<R>() to { raw -> convert(raw) }
```
Jolanrensen: You only store type `R`, making `getExpectedJdbcType` return the wrong type. `getExpectedJdbcType` should return the type `getObject` returns. Then, if preprocessing is done, `getPreprocessedValueType` should return the preprocessed type, aka `R` here.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,516 @@
...
+                }
+        }
+
+        return super.getExpectedJdbcType(tableColumnMetadata)
+    }
+
+    // For DECIMAL/NUMERIC we already resolved the DataFrame type from the storage class in
+    // getExpectedJdbcType, so we keep that as-is. For other types we let the base decide
+    // (base maps TIMESTAMP → Instant, BINARY(UUID) → Uuid, etc.).
+    override fun getPreprocessedValueType(tableColumnMetadata: TableColumnMetadata, expectedJdbcType: KType): KType =
```
Jolanrensen: This function should return the type the values have after they are preprocessed. However, it doesn't seem to do anything with the type mappings

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
Jolanrensen: Same for `forType(""") {}`

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,516 @@
...
+ * ### Example
+ * ```
+ * val format = LocalDateTime.Format {
+ *     year(); char('-'); monthNumber(); char('-'); day()
+ *     char(' ')
+ *     hour(); char(':'); minute(); char(':'); second()
+ *     chars(" UTC")
+ * }
+ *
+ * val sqliteCustom = Sqlite.withCustomConverters {
```
Jolanrensen, line 61: that's a nice looking dsl btw :)

zaleslaw:
@AndreiKingsley please fix Copilot and Jolan review comments
After that I'll do a last round

Jolanrensen:
Seems to have a lot of duplication with AdvancedDbType. Maybe the necessary changes (like converters based on column name) could be added there instead. No need to reinvent the wheel :)

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -220,15 +247,40 @@ public class SqliteCustomConvertersBuilder
  *  4. The base `DbType` mapping.
  *
  * Both maps take a [SqliteCustomTypeConverter] lambda that returns both the target [KType] and the
- * converted value.
+ * converteing lamda.
```
zaleslaw: I don't understand

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/sqliteCustomTypesTest.kt
```diff
@@ -292,4 +289,53 @@ class SqliteTestCustomTypes {
...
+                try {
+                    connection.close()
+                } catch (e: Exception) {
+                    // Log, but not fail
+                    println("Warning: Could not clean up test database file: ${e.message}")
+                }
+            }
+        }
+
+        // @Ignore
```
zaleslaw: What's the future of this test

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,568 @@
...
+         *   `typeOf<R>()` — must be a non-abstract type reachable by reflection.
+         *
+         * @param [sqlTypeName] name of the declared SQL type (as written in `CREATE TABLE`).
+         * @param [convert] lambda to convert the raw stored value to the target Kotlin type.
+         */
+        public inline fun <reified T, reified R> forType(sqlTypeName: String, crossinline convert: (T) -> R) {
+            val mapping: SqliteCustomTypeConverter<T, R> = SqliteCustomTypeConverter(
+                typeOf<T>(),
+                typeOf<R>(),
+                { raw -> convert(raw) },
```
Jolanrensen: lambda can be moved out. I'd also recommend using named arguments for SqliteCustomTypeConverter

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,568 @@
...
+    ): AnyJdbcToDataFrameConverter = readerConverter(expectedKType, typeOf<R>().withNullability(nullable), reader)
+
+    private inline fun <R> readerConverter(
+        expectedKType: KType,
+        resultKType: KType,
+        crossinline reader: (ResultSet, Int) -> R,
+    ): AnyJdbcToDataFrameConverter =
+        JdbcToDataFrameConverter<Any?, Any?, Any?>(
+            expectedJdbcType = expectedKType,
+            resultSetReader = { rs, i -> reader(rs, i) },
```
Jolanrensen: I don't entirely understand why you would add a custom convert opportunity on the reader stage instead of on the preprocess stage which is what it's for.

There's 3 steps:
- How the values are read from the ResultSet. This defaults to `getObject(i + 1)` and `expectedJdbcType` should tell the type the object is
- Then they can be preprocessed. This defaults to `{ it }`, no preprocessing, but can be used to convert values individually. `preprocessedValueType` should tell the type the value is after preprocessing
- Finally there's columnBuilder, which turns a List of values into an actual column. This defaults to something like `columnOf(...)` but can be customized. `targetSchema` should tell the schema of that column.

Jolanrensen: So it could be something like
```kt
jdbcToDfConverterFor(jdbcSourceType = expectedKType)
    .withPreprocessor(preprocessValueType = resultKType) { ... }
```

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,568 @@
...
+            customColumnsConverterCache.getOrPut(columnName) {
+                generateConverter(tableColumnMetadata)
+            }
+        } else {
+            super.getConverter(tableColumnMetadata)
+        }
+    }
+
+    /**
+     * Delegate all unexpected column type handlings to defaulat [DbType]
```
Jolanrensen: defaulat

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,568 @@
...
+     */
+    private fun fallbackConverter(meta: TableColumnMetadata): AnyJdbcToDataFrameConverter {
+        val expectedJdbcTypeBase = fallback.getExpectedJdbcType(meta)
+        val preprocessedValueType = fallback.getPreprocessedValueType(meta, expectedJdbcTypeBase)
+        val targetSchema = fallback.getTargetColumnSchema(meta, preprocessedValueType)
+        return JdbcToDataFrameConverter<Any?, Any?, Any?>(
+            expectedJdbcType = preprocessedValueType,
+            resultSetReader = { rs, i ->
+                val raw = fallback.getValueFromResultSet<Any?>(rs, i, meta, expectedJdbcTypeBase)
+                fallback.preprocessValue<Any?, Any?>(raw, meta, expectedJdbcTypeBase, preprocessedValueType)
```
Jolanrensen: this second step should be in valuePreprocessor

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,568 @@
...
+                }
+        }
+
+        // 4) DECIMAL / NUMERIC — trust the driver-reported class of the stored value.
+        when (tableColumnMetadata.jdbcType) {
+            Types.DECIMAL, Types.NUMERIC -> return jdbcToDfConverterFor<Any>(expectedKType.withNullability(nullable))
+        }
+
+        // 5) Fallback — delegate to the base [DbType] end-to-end pipeline.
+        return fallbackConverter(tableColumnMetadata)
```
Jolanrensen, line 361: Hmm, a fallback could be useful indeed, but not having this fallback in AdvancedDb was done on purpose. It forces AdvancedDbTypes to provide a complete type mapping/converting picture and keeps the logic in one place. This keeps it independent and better testable.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,568 @@
...
+            "TIME" in declaredUpper ->
+                return readerConverter<LocalTime?>(expectedKType, nullable) { rs, i ->
+                    convertToLocalTime(rs.getObject(i + 1), tableColumnMetadata)
+                }
+        }
+
+        // 3) BOOLEAN / BIT — SQLite has no boolean storage class. Xerial reports `Types.BOOLEAN`
+        //    metadata but `getObject` returns Integer (0/1). Convert explicitly.
+        when (tableColumnMetadata.jdbcType) {
+            Types.BOOLEAN, Types.BIT ->
```
zaleslaw: This branch keys off jdbcType, but Xerial derives jdbcType from the stored storage class, not the declared type — the same reason you match dates by name just above. Verified empirically: a BOOLEAN column stored as TEXT 'true' reports getColumnType == VARCHAR(12), not BOOLEAN(16), so it never reaches this branch and is read as String; the String case in convertToBoolean is then unreachable. Suggest matching by declared name for consistency: "BOOL" in declaredUpper || "BIT" in declaredUpper.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,568 @@
...
+        nullable: Boolean,
+        crossinline reader: (ResultSet, Int) -> R,
+    ): AnyJdbcToDataFrameConverter = readerConverter(expectedKType, typeOf<R>().withNullability(nullable), reader)
+
+    private inline fun <R> readerConverter(
+        expectedKType: KType,
+        resultKType: KType,
+        crossinline reader: (ResultSet, Int) -> R,
+    ): AnyJdbcToDataFrameConverter =
+        JdbcToDataFrameConverter<Any?, Any?, Any?>(
```
zaleslaw: AdvancedDbType already ships a builder DSL for this — see `DuckDb.parseDuckDbType: jdbcToDfConverterFor<T>(nullable).withPreprocessor { … }.` Hand-building the converter with valuePreprocessor = null and converting inside resultSetReader skips the not-null propagation withPreprocessor/withColumnBuilder do for you, and sets expectedJdbcType to the raw storage type while the reader already returns the converted value (so J no longer matches what's read). Prefer reading the raw value and converting via withPreprocessor.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
zaleslaw: Yeah!

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,568 @@
...
+            }
+        } else {
+            super.getConverter(tableColumnMetadata)
+        }
+    }
+
+    /**
+     * Delegate all unexpected column type handlings to defaulat [DbType]
+     */
+    private val fallback: DbType = object : DbType("sqlite") {
```
zaleslaw, line 274: This anonymous DbType exists only to reach the default mapping, since AdvancedDbType makes the pipeline methods final. It duplicates driverClassName and is fragile. If most columns just need the plain mapping, reconsider whether Sqlite should extend AdvancedDbType, or extract the default mapping into a shared helper.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,568 @@
...
+        when (value) {
+            null -> null
+            is Int -> instantToLocalDateTime(Instant.fromEpochSeconds(value.toLong()))
+            is Long -> instantToLocalDateTime(Instant.fromEpochSeconds(value))
+            is Double -> instantToLocalDateTime(julianDayToInstant(value))
+            is String -> parseStringAsLocalDateTime(value, meta)
+            else -> unsupportedConversion(value, "kotlinx.datetime.LocalDateTime", meta)
+        }
+
+    private fun convertToLocalTime(value: Any?, meta: TableColumnMetadata): LocalTime? =
```
zaleslaw, line 469: MAYBE: DATE/DATETIME/TIMESTAMP handle a REAL (Double) value (Julian day), but there's no is Double branch here, so a TIME stored as REAL (fraction of a day) throws. Add the branch or document the omission.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,568 @@
...
+ *     // The "ratio" column overrides its declared type — read as Double.
+ *     forColumn("ratio") { raw: String -> raw.toDouble() }
+ * }
+ *
+ * val df = DataFrame.readSqlTable(connection, "events", dbType = sqliteCustom)
+ * ```
+ */
+public data class SqliteCustomTypeConverter<T, R>(
+    val expectedType: KType,
+    val resultingType: KType,
```
zaleslaw, line 78: MAYBE: resultingType is used as-is (typeOf<R>()), without withNullability(isNullable) — the test pins this as intentional. Worth a KDoc warning: forType<String>("X") on a column that contains NULLs yields a non-null String schema type and relies on downstream Infer.Nulls.

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/sqliteCustomTypesTest.kt
```diff
@@ -38,17 +47,295 @@ class SqliteTestCustomTypes {
...
+            eventsDf["occurredAt"][1] shouldBe Instant.parse("2023-08-15T18:45:30Z")
+            eventsDf["note"][0] shouldBe Instant.parse("2023-07-21T10:30:00Z")
+            eventsDf["note"][1] shouldBe null
+        }
+    }
+
+    /**
+     * Xerial SQLite JDBC driver seems to give identical metadata for `Int?` and `Long?` columns
+     */
+    class DynamicTypes {
```
zaleslaw, line 297: Good that forColumn<Long?>() is a manual escape hatch, but Int-vs-Long is still unresolved (Xerial reports identical metadata; the converter is generated once per column while getObject may return Long per row). A doc note that this is a workaround, not auto-detection, and that #1747 stays open would set expectations.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,568 @@
...
+            val convert = converter.convert as (Any?) -> Any?
+            return readerConverter(converter.expectedType, converter.resultingType)
+            { rs, i -> convert(rs.getObject(i + 1)) }
+        }
+
+        // 2) Date/time affinity by declared type name substring (matters more than jdbcType —
+        //    Xerial derives that from the actual storage class and flips it to FLOAT/INTEGER
+        //    when the stored value is a Julian day / Unix seconds).
+        when {
+            "DATETIME" in declaredUpper ->
```
zaleslaw, line 316: substring match also catches MODIFICATION_DATE/UPDATE_TIME; fine as heuristic, worth a KDoc note.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,568 @@
...
+        when (tableColumnMetadata.jdbcType) {
+            Types.BOOLEAN, Types.BIT ->
+                return readerConverter<Boolean?>(expectedKType, nullable) { rs, i ->
+                    convertToBoolean(rs.getObject(i + 1), tableColumnMetadata)
+                }
+        }
+
+        // 4) DECIMAL / NUMERIC — trust the driver-reported class of the stored value.
+        when (tableColumnMetadata.jdbcType) {
+            Types.DECIMAL, Types.NUMERIC -> return jdbcToDfConverterFor<Any>(expectedKType.withNullability(nullable))
```
zaleslaw: MAYBE, NEED TO CHECK: withNullability redundant (javaClassNameToKType already applied it).

zaleslaw (changes requested):
Mostly agree with the main thing which @Jolanrensen is mentioned plus a couple of minor notes and well documented workarounds and problems

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/sqliteCustomTypesTest.kt
AndreiKingsley: The only way to auto-detect this is reflection, which is too expensive, so we decided not to use it (globally across the Kotlin DataFrame).
This is a bug in the JDBC driver (which I believe will not be fixed), so for the time being, the official recommendation is to explicitly specify the column types.

AndreiKingsley: @Jolanrensen do you agree here?

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/sqliteCustomTypesTest.kt
AndreiKingsley: It should be here. The only problem with db connection on TC, don't know why.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
AndreiKingsley: We trust user specified type OVER jdbc column metadata nullability here.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -244,10 +244,8 @@ public class SqliteCustomConvertersBuilder
  *     Use this to override every column that shares a declared type.
  *  3. The built-in SQLite conversion for BOOLEAN, DATE, DATETIME, TIME, TIMESTAMP, DECIMAL,
  *     NUMERIC.
+ *     **This conversion is also applied for types that have substring of these type names**.
```
Jolanrensen, line 247: good note :) maybe add the example of "LONGVARCHAR" -> "LONG" to show how stupid this rule is and everyone takes extra care about this rule

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/sqliteTest.kt
```diff
@@ -47,6 +48,13 @@ interface CustomerOrderSQLite {
     val orderDetails: ByteArray?
 }
 
+@DataSchema
+interface FlagSQLite {
```
Jolanrensen: `FlagSqlite` seems better CamelCased. (though our own `renameToCamelCase()` makes it `FlagSqLite` XD)

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,547 @@
...
+                convert(it)
+            }
+        }
+
+        // 2) Date/time affinity by declared type name substring (matters more than jdbcType —
+        //    Xerial derives that from the actual storage class and flips it to FLOAT/INTEGER
+        //    when the stored value is a Julian day / Unix seconds).
+        when {
+            "DATETIME" in declaredUpper ->
+                return jdbcToDfConverterFor<Any?>(expectedKType).withPreprocessor {
```
Jolanrensen: I think putting all `.withPreprocessor` calls on a new line helps with readability

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,547 @@
...
+        // 5) Fallback — delegate to the base [DbType] end-to-end pipeline.
+        return fallbackConverter(tableColumnMetadata)
+    }
+
+    /**
+     * Builds a converter that mirrors the base [DbType] pipeline end-to-end for this column using [fallback].
+     */
+    private fun fallbackConverter(meta: TableColumnMetadata): AnyJdbcToDataFrameConverter {
+        val expectedJdbcTypeBase = fallback.getExpectedJdbcType(meta)
+        val preprocessedValueType = fallback.getPreprocessedValueType(meta, expectedJdbcTypeBase)
```
Jolanrensen, line 369: using the `preprocessedValueType` from `fallback` only makes sense if you're also using `fallback.preprocessedValue()` as valuePreprocessor. Otherwise you can get type mismatches

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
AndreiKingsley: The note about conversion on the DataFrame side for common SQL types (logic I've added in `generateConverter` implementation)

"LONGVARCHAR" -> "LONG"  is a type affinity rule, and performed by SQLite. There's example for custom converter with LONGVARCHAR and note about that.

dataframe-jdbc/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/sqliteTest.kt
AndreiKingsley: driver entities contains `SQLite`, but ok

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,557 @@
...
+        val nullable = tableColumnMetadata.isNullable
+        val declaredUpper = tableColumnMetadata.sqlTypeName.uppercase()
+
+        val expectedKType = javaClassNameToKType(tableColumnMetadata.javaClassName, nullable)
+
+        // 1) Custom mapping. Column-name override beats type-name override.
+        customMappingFor(tableColumnMetadata)?.let { converter ->
+            @Suppress("UNCHECKED_CAST")
+            val convert = converter.convert as (Any?) -> Any?
+            return jdbcToDfConverterFor<Any?>(
```
Jolanrensen, line 305: oh, come to think of it. You can leave them as is, but it's not necessary to specify `?` in `jdbcToDfConverterFor<>(kType)`. The converter requires you to handle nulls in each step anyway. Actual nullability information is carried using the ktype.

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
```diff
@@ -1,31 +1,557 @@
...
+
+    /**
+     * Builds a converter that mirrors the base [DbType] pipeline end-to-end for this column using [fallback].
+     */
+    private fun fallbackConverter(meta: TableColumnMetadata): AnyJdbcToDataFrameConverter {
+        val expectedJdbcTypeBase = fallback.getExpectedJdbcType(meta)
+        val preprocessedValueType = fallback.getPreprocessedValueType(meta, expectedJdbcTypeBase)
+        return jdbcToDfConverterFor<Any?>(
+            expectedJdbcTypeBase,
+        ).withPreprocessor<Any?, Any?>(
```
Jolanrensen, line 372: Can be shorter:
```kt
jdbcToDfConverterFor<Any>(expectedJdbcTypeBase)
    .withPreprocessor<Any, Any>(preprocessedValueType) {
        fallback.preprocessValue(it, meta, expectedJdbcTypeBase, preprocessedValueType)
    }
```

Jolanrensen (approved):

dataframe-jdbc/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/db/Sqlite.kt
Jolanrensen: just so you know :)

zaleslaw (approved):

PR #2009, opened 2026-07-28
Migrated delegated property usages in build files for Gradle 10

koperagen:
Suddenly it all became deprecated. Will be error in Gradle 10, apparently. 

registering:
[DEPRECATION] 'fun <T> ExistingDomainObjectDelegate<out T>.getValue(receiver: Any?, property: KProperty<*>): T' is deprecated. Use the named() API instead. 
See the Gradle 9.6 upgrading guide

getting:
[DEPRECATION] 'fun provideDelegate(thisRef: Any?, property: KProperty<*>): ExistingDomainObjectDelegate<Configuration>' is deprecated. Use 'val element = getByName(name)' instead. See the Gradle 9.6 upgrading guide

creating:
[DEPRECATION] 'fun provideDelegate(thisRef: Any?, property: KProperty<*>): ExistingDomainObjectDelegate<KotlinSourceSet>' is deprecated. Use 'val element = create(name)' instead. See the Gradle 9.6 upgrading guide.

### DIFF
--- a/build-logic/src/main/kotlin/dfbuild.buildConfig.gradle.kts
+++ b/build-logic/src/main/kotlin/dfbuild.buildConfig.gradle.kts
@@ -19,7 +19,7 @@ buildConfig {
 }
 
 // combining buildconfig with ktlint
-val buildConfigSources by kotlin.sourceSets.creating {
+val buildConfigSources = kotlin.sourceSets.create("buildConfigSources") {
     kotlin.srcDir("build/generated/sources/buildConfig/main")
 }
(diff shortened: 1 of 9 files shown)


Jolanrensen (approved):
Yep, it's a shame :/ I always prefer less duplication and fewer strings, but I guess it simplifies Gradle a bit by removing it.

More information can be found here https://docs.google.com/document/d/140dOuBcCEDCKLGTiY80NNx2bCQ7dUv8EVFsn4ph86fU/edit?tab=t.0#heading=h.wbxzemi759rh

PR #2012, opened 2026-07-29
KDocs for min/max

### LINKED ISSUE #1971: Add KDocs for non-deprecated `max` APIs
## Add KDocs for non-deprecated `max` APIs

`core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt` contains many public `max` APIs without KDocs.

The coverage sheet also marks this API group as missing KDocs:
`max/maxBy/maxByOrNull/maxOf/maxOfOrNull/maxFor`.

Please add concise KDocs for the current non-deprecated APIs only. Deprecated overloads and binary-compatibility overloads do not need new KDocs.

### Scope

Add KDocs for non-deprecated overloads in these groups:

- `DataColumn.max`
- `DataColumn.maxOrNull`
- `DataColumn.maxBy`
- `DataColumn.maxByOrNull`
- `DataColumn.maxOf`
- `DataColumn.maxOfOrNull`
- `DataRow.rowMaxOf`
- `DataRow.rowMaxOfOrNull`
- `DataFrame.max`
- `DataFrame.maxOrNull`
- `DataFrame.maxBy`
- `DataFrame.maxByOrNull`
- `DataFrame.maxOf`
- `DataFrame.maxOfOrNull`
- `DataFrame.maxFor`
- `Grouped.max`
- `Grouped.maxFor`
- `Grouped.maxOf`
- `GroupBy.maxBy`
- `Pivot.max`
- `Pivot.maxFor`
- `Pivot.maxOf`
- `Pivot.maxBy`
- `PivotGroupBy.max`
- `PivotGroupBy.maxFor`
- `PivotGroupBy.maxOf`
- `PivotGroupBy.maxBy`

### KDocs should explain

  - maximum value
  - row with maximum selected/computed value
  - dataframe/data row with maxima per selected columns
- Difference between:
  - `max`
  - `maxOrNull`
  - `maxBy`
  - `maxByOrNull`
  - `maxOf`
  - `maxOfOrNull`
  - `maxFor`
- Behavior on empty inputs or all-null inputs:
  - non-null variants throw/suggest failure through `suggestIfNull`
  - `*OrNull` variants return `null`
- How `skipNaN` affects floating-point values.
- How selected comparable columns are handled.
- For grouped/pivot variants, that the operation is applied per group or per pivot cell.

### Acceptance criteria

- Non-deprecated public `max*` APIs have concise KDocs.
- Deprecated and hidden binary-compatibility overloads are not documented.
- KDocs are consistent with the existing site docs:
  https://kotlin.github.io/dataframe/maxby.html
- KDocs clearly distinguish value-returning, row-returning, and per-column aggregation variants.

### LINKED ISSUE #1974: Add KDocs for non-deprecated `min` APIs
## Add KDocs for non-deprecated `min` APIs

`core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/min.kt` contains many public `min` APIs without KDocs.

The coverage sheet marks this API group as missing KDocs:
`min/*`.

Please add concise KDocs for the current non-deprecated APIs only. Deprecated overloads and binary-compatibility overloads do not need new KDocs.

### Scope

Add KDocs for non-deprecated overloads in these groups:

- `DataColumn.min`
- `DataColumn.minOrNull`
- `DataColumn.minBy`
- `DataColumn.minByOrNull`
- `DataColumn.minOf`
- `DataColumn.minOfOrNull`
- `DataRow.rowMinOf`
- `DataRow.rowMinOfOrNull`
- `DataFrame.min`
- `DataFrame.minOrNull`
- `DataFrame.minFor`
- `DataFrame.minOf`
- `DataFrame.minOfOrNull`
- `DataFrame.minBy`
- `DataFrame.minByOrNull`
- `Grouped.min`
- `Grouped.minFor`
- `Grouped.minOf`
- `GroupBy.minBy`
- `Pivot.min`
- `Pivot.minFor`
- `Pivot.minOf`
- `Pivot.minBy`
- `PivotGroupBy.min`
- `PivotGroupBy.minFor`
- `PivotGroupBy.minOf`
- `PivotGroupBy.minBy`

### KDocs should explain

  - minimum value
  - row with minimum selected/computed value
  - row/dataframe with per-column minima
  - grouped or pivoted minimum values
- Difference between:
  - `min`
  - `minOrNull`
  - `minBy`
  - `minByOrNull`
  - `minOf`
  - `minOfOrNull`
  - `minFor`
- Behavior on empty inputs or all-null inputs:
  - non-null variants fail through `suggestIfNull`
  - `*OrNull` variants return `null`
- How `skipNaN` affects floating-point values.
- How selected comparable columns are handled.
- For grouped/pivot variants, that the operation is applied per group or per pivot cell.

### Acceptance criteria

- Non-deprecated public `min*` APIs have concise KDocs.
- Deprecated and hidden binary-compatibility overloads are not documented.
- KDocs are consistent with the existing site docs:
  https://kotlin.github.io/dataframe/minmax.html
- KDocs clearly distinguish value-returning, row-returning, and per-column aggregation variants.

Jolanrensen:
Fixes #1971, #1974 

I let Claude attempt write it using all our KDoc guides. It's quite impressive!
Only made some small fixes and maintainability improvements, but it looks good to me. All questions I had are answered in the KDocs and it seemed to understand and use KoDEx quite well.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
@@ -36,7 +36,7 @@ import kotlin.reflect.KProperty
 /**
  * {@comment
  *    The Max Operation KDoc-topic; it also holds all common `max` KDoc-snippets.
- *    Link to it with `[The Max Operation][MaxDocs]`.
+ *    Link to it with `{@include [MaxDocsLink]}`.
  * }
  *
  * ## The Max Operation
@@ -86,7 +86,7 @@ internal interface MaxDocs {
      *
      * `null` values in the input are always ignored.
      *
-     * If the input contains [`NaN`][NaN] values, the result will be `NaN` as well,
+     * If the input contains [`NaN`][NaN] values, the result will be `NaN`,
      * unless `skipNaN` is set to `true`.
      */
     @ExcludeFromSources
@@ -214,6 +214,10 @@ internal interface MaxDocs {
     typealias MaxForSelectingOptions = Nothing
 }
 
+/** [The Max Operation][MaxDocs] */
+@ExcludeFromSources
+private typealias MaxDocsLink = Nothing
+
 /** {@set [SelectingColumns.OPERATION] [max][max]} */
 @ExcludeFromSources
 private typealias SetMaxOperationArg = Nothing
@@ -244,7 +248,7 @@ private typealias SetMaxOrNullOperationArg = Nothing
  * - [maxOf][DataColumn.maxOf] — the maximum of the values a selector returns for each element.
  * - [maxBy][DataColumn.maxBy] — the element for which a selector returns the maximum value.
  * - [min][DataColumn.min] — the mirror operation.
- * - [The Max Operation][MaxDocs] — an overview of all `max` modes.
+ * - {@include [MaxDocsLink]} — an overview of all `max` modes.
  *
  * For more information: {@include [DocumentationUrls.MinMax]}
  *
@@ -279,7 +283,7 @@ public fun <T : Comparable<T>> DataColumn<T?>.max(skipNaN: Boolean = skipNaNDefa
  * - [maxByOrNull][DataColumn.maxByOrNull] — the element for which a selector returns
  *   the maximum value.
  * - [minOrNull][DataColumn.minOrNull] — the mirror operation.
- * - [The Max Operation][MaxDocs] — an overview of all `max` modes.
+ * - {@include [MaxDocsLink]} — an overview of all `max` modes.
  *
  * For more information: {@include [DocumentationUrls.MinMax]}
  *
@@ -311,13 +315,13 @@ public fun <T : Comparable<T>> DataColumn<T?>.maxOrNull(skipNaN: Boolean = skipN
  * - [maxOf][DataColumn.maxOf] — returns the maximum [selector] value itself
  *   instead of the element it belongs to.
  * - [minBy][DataColumn.minBy] — the mirror operation.
- * - [The Max Operation][MaxDocs] — an overview of all `max` modes.
+ * - {@include [MaxDocsLink]} — an overview of all `max` modes.
  *
  * For more information: {@include [DocumentationUrls.MaxBy]}
  *
  * ### Example
  * ```kotlin
- * // The longest first name in the "name/firstName" column
+ * // The longest first name in the "name"/"firstName" column
  * df.name.firstName.maxBy { it.length }
  * ```
  *
@@ -346,13 +350,13 @@ public inline fun <T, reified R : Comparable<R & Any>?> DataColumn<T>.maxBy(
  * - [maxOfOrNull][DataColumn.maxOfOrNull] — returns the maximum [selector] value itself
  *   instead of the element it belongs to.
  * - [minByOrNull][DataColumn.minByOrNull] — the mirror operation.
- * - [The Max Operation][MaxDocs] — an overview of all `max` modes.
+ * - {@include [MaxDocsLink]} — an overview of all `max` modes.
  *
  * For more information: {@include [DocumentationUrls.MaxBy]}
  *
  * ### Example
  * ```kotlin
- * // The longest first name in the "name/firstName" column,
+ * // The longest first name in the "name"/"firstName" column,
  * // or `null` if the column is empty
  * df.name.firstName.maxByOrNull { it.length }
  * ```
@@ -382,13 +386,13 @@ public inline fun <T, reified R : Comparable<R & Any>?> DataColumn<T>.maxByOrNul
  * - [maxBy][DataColumn.maxBy] — returns the element the maximum [selector] value belongs to
  *   instead of that value.
  * - [minOf][DataColumn.minOf] — the mirror operation.
- * - [The Max Operation][MaxDocs] — an overview of all `max` modes.
+ * - {@include [MaxDocsLink]} — an overview of all `max` modes.
  *
  * For more information: {@include [DocumentationUrls.MinMax]}
  *
  * ### Example
  * ```kotlin
- * // The length of the longest first name in the "name/firstName" column
+ * // The length of the longest first name in the "name"/"firstName" column
  * df.name.firstName.maxOf { it.length }
  * ```
  *
@@ -417,13 +421,13 @@ public inline fun <T, reified R : Comparable<R & Any>?> DataColumn<T>.maxOf(
  * - [maxByOrNull][DataColumn.maxByOrNull] — returns the element the maximum [selector] value
  *   belongs to instead of that value.
  * - [minOfOrNull][DataColumn.minOfOrNull] — the mirror operation.
- * - [The Max Operation][MaxDocs] — an overview of all `max` modes.
+ * - {@include [MaxDocsLink]} — an overview of all `max` modes.
  *
  * For more information: {@include [DocumentationUrls.MinMax]}
  *
  * ### Example
  * ```kotlin
- * // The length of the longest first name in the "name/firstName" column,
+ * // The length of the longest first name in the "name"/"firstName" column,
  * // or `null` if the column is empty
  * df.name.firstName.maxOfOrNull { it.length }
  * ```
@@ -465,7 +469,7 @@ public fun DataRow<*>.rowMax(): Nothing = error(ROW_MAX)
  * - [rowMaxOf][DataRow.rowMaxOf] — throws instead of returning `null` when there's nothing to compare.
  * - [rowMinOfOrNull][DataRow.rowMinOfOrNull] — the mirror operation.
  * - [maxOrNull][DataFrame.maxOrNull] — the maximum of the values in specific columns of a [DataFrame].
- * - [The Max Operation][MaxDocs] — an overview of all `max` modes.
+ * - {@include [MaxDocsLink]} — an overview of all `max` modes.
  *
  * For more information: {@include [DocumentationUrls.RowStatistics]}
  *
@@ -500,7 +504,7 @@ public inline fun <reified T : Comparable<T>> DataRow<*>.rowMaxOfOrNull(skipNaN:
  *   when there's nothing to compare.
  * - [rowMinOf][DataRow.rowMinOf] — the mirror operation.
  * - [max][DataFrame.max] — the maximum of the values in specific columns of a [DataFrame].
- * - [The Max Operation][MaxDocs] — an overview of all `max` modes.
+ * - {@include [MaxDocsLink]} — an overview of all `max` modes.
  *
  * For more information: {@include [DocumentationUrls.RowStatistics]}
  *
@@ -539,14 +543,14 @@ public inline fun <reified T : Comparable<T>> DataRow<*>.rowMaxOf(skipNaN: Boole
  * - [maxFor][DataFrame.maxFor] — the same, but for an explicit selection of columns.
  * - [max][DataFrame.max]` { columns }` — a single maximum of all values in the selected columns.
  * - [min][DataFrame.min] — the mirror operation.
- * - [The Max Operation][MaxDocs] — an overview of all `max` modes.
+ * - {@include [MaxDocsLink]} — an overview of all `max` modes.
  *
  * For more information: {@include [DocumentationUrls.MinMax]}
  *
  * ### Example
  * ```kotlin
  * // A single row with the largest value of each comparable column
- * // ("name/firstName", "name/lastName", "age", "city", "weight", and "isHappy")
+ * // ("name"/"firstName", "name"/"lastName", "age", "city", "weight", and "isHappy")
  * df.max()
  * ```
  *
@@ -577,7 +581,7 @@ public fun <T> DataFrame<T>.max(skipNaN: Boolean = skipNaNDefault): DataRow<T> =
  * - [max][DataFrame.max]`()` — the same, but for all suitable columns at once.
  * - [max][DataFrame.max]` { columns }` — a single maximum of all values in the selected columns.
  * - [minFor][DataFrame.minFor] — the mirror operation.
- * - [The Max Operation][MaxDocs] — an overview of all `max` modes.
+ * - {@include [MaxDocsLink]} — an overview of all `max` modes.
  *
  * For more information: {@include [DocumentationUrls.MinMax]}
  *
@@ -618,7 +622,7 @@ public fun <T, C : Comparable<*>?> DataFrame<T>.maxFor(
  * - [max][DataFrame.max]`()` — the same, but for all suitable columns at once.
  * - [max][DataFrame.max]` { columns }` — a single maximum of all values in the selected columns.
  * - [minFor][DataFrame.minFor] — the mirror operation.
- * - [The Max Operation][MaxDocs] — an overview of all `max` modes.
+ * - {@include [MaxDocsLink]} — an overview of all `max` modes.
  *
  * For more information: {@include [DocumentationUrls.MinMax]}
  *
@@ -666,7 +670,7 @@ public fun <T, C : Comparable<*>?> DataFrame<T>.maxFor(
  * - [maxFor][DataFrame.maxFor] — the maximum of each selected column separately.
  * - [maxOf][DataFrame.maxOf] — the maximum of the values a row expression returns for each row.
  * - [min][DataFrame.min] — the mirror operation.
- * - [The Max Operation][MaxDocs] — an overview of all `max` modes.
+ * - {@include [MaxDocsLink]} — an overview of all `max` modes.
  *
  * For more information: {@include [DocumentationUrls.MinMax]}
  *
@@ -706,7 +710,7 @@ public fun <T, C : Comparable<C & Any>?> DataFrame<T>.max(
  * - [maxFor][DataFrame.maxFor] — the maximum of each selected column separately.
  * - [maxOf][DataFrame.maxOf] — the maximum of the values a row expression returns for each row.
  * - [min][DataFrame.min] — the mirror operation.
- * - [The Max Operation][MaxDocs] — an overview of all `max` modes.
+ * - {@include [MaxDocsLink]} — an overview of all `max` modes.
  *
  * For more information: {@include [DocumentationUrls.MinMax]}
  *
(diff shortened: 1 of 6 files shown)


Jolanrensen:
@copilot+claude-opus-4.6 want to review? You can see the preprocessed kdocs here https://github.com/Kotlin/dataframe/commit/9c75023ec5ce4b9158fb6ffcac0af25c8bea204d

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
```diff
@@ -26,29 +31,268 @@ import org.jetbrains.kotlinx.dataframe.util.ROW_MAX
...
 
+// region docs
+
+/**
+ * {@comment
+ *    The Max Operation KDoc-topic; it also holds all common `max` KDoc-snippets.
+ *    Link to it with `{@include [MaxDocsLink]}`.
+ * }
+ *
+ * ## The Max Operation
```
AndreiKingsley, line 42: Please, remove all headers

Could you take a look, why Claude decided to bring header back? Because of examples or why?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
```diff
@@ -26,29 +31,268 @@ import org.jetbrains.kotlinx.dataframe.util.ROW_MAX
...
+ *
+ * Computes the [maximum](https://en.wikipedia.org/wiki/Maximum_and_minimum) of values.
```
AndreiKingsley, line 44: Do we need this link here actually?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/CommonMinMaxDocs.kt
```diff
@@ -0,0 +1,150 @@
...
+     *    are treated. KDoc-snippet.}
+     *
+     * Only self-comparable values are supported: values of a type `T : Comparable<T>`
+     * that are mutually comparable (like strings, primitive numbers, or dates).
+     * This includes all primitive number types, but no mix of different number types.
+     *
+     * `null` values in the input are always ignored.
+     *
+     * If the input contains [`NaN`][NaN] values, the result will be `NaN`,
+     * unless `skipNaN` is set to `true`.
```
AndreiKingsley: can be linked to an argument

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
AndreiKingsley: Sorry, now I’ve realised what this is topic, nevermind

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
```diff
@@ -26,29 +31,268 @@ import org.jetbrains.kotlinx.dataframe.util.ROW_MAX
...
 
+/**
+ * Returns the maximum of the values in this [DataColumn].
+ *
+ * {@include [MaxDocs.InputValuesSnippet]}
+ *
+ * {@include [MaxDocs.ThrowsOnEmptySnippet]}
+ *
+ * See also:
+ * - [maxOrNull][DataColumn.maxOrNull] — returns `null` instead of throwing for an empty column.
```
AndreiKingsley: or column of `null`s

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
```diff
@@ -26,29 +31,268 @@ import org.jetbrains.kotlinx.dataframe.util.ROW_MAX
...
+ *
+ * @include [MaxDocs.SkipNaNParam]
+ * @return The largest value in this column.
+ * @throws NoSuchElementException if there are no values to compare.
+ */
 public fun <T : Comparable<T>> DataColumn<T?>.max(skipNaN: Boolean = skipNaNDefault): T =
     maxOrNull(skipNaN).suggestIfNull("max")
 
+/**
+ * Returns the maximum of the values in this [DataColumn], or `null` if there is nothing to compare.
```
AndreiKingsley, line 137: I'd add
```
(i.e., column is empty or only `null` values are compared)
```

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
```diff
@@ -26,29 +31,268 @@ import org.jetbrains.kotlinx.dataframe.util.ROW_MAX
...
+ * Returns the first element of this [DataColumn] for which the given [selector]
+ * returns the maximum value.
+ *
+ * {@include [MaxDocs.InputValuesSnippet]}
+ *
+ * {@include [MaxDocs.ThrowsOnEmptySnippet]}
+ *
+ * See also:
+ * - [maxByOrNull][DataColumn.maxByOrNull] — returns `null` instead of throwing for an empty column.
+ * - [maxOf][DataColumn.maxOf] — returns the maximum [selector] value itself
```
AndreiKingsley: I'd place `maxOf` above see also section, with a message

"Don't confuse it with [maxOf] that returns maximum value computed with [selector] instead of the original element"

And the same in [maxOf] KDoc.

(and in `maxByOrNull`, `min..()`s, etc.)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
```diff
@@ -329,11 +1434,81 @@ public fun <T, R : Comparable<*>?> Pivot<T>.maxFor(
...
+ * value among all the values in the selected columns of the corresponding group.
+ *
+ * {@include [MaxDocs.InputValuesSnippet]}
+ *
+ * {@include [MaxDocs.NullCellOnEmptySnippet]}
+ *
+ * {@include [SelectingColumns.ColumnGroupsAndNestedColumnsSnippet]}
+ *
+ * See [Selecting Columns][MaxDocs.MaxSelectingOptions].
+ *
```
AndreiKingsley, line 1521: All link to pivot grammar

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
AndreiKingsley: Ok, there's a comprehensive snippet below, it's ok.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
```diff
@@ -64,27 +308,180 @@ public fun DataRow<*>.rowMaxOrNull(): Nothing? = error(ROW_MAX_OR_NULL)
...
+ * - [rowMinOfOrNull][DataRow.rowMinOfOrNull] — the mirror operation.
+ * - [maxOrNull][DataFrame.maxOrNull] — the maximum of the values in specific columns of a [DataFrame].
+ * - {@include [MaxDocsLink]} — an overview of all `max` modes.
+ *
+ * For more information: {@include [DocumentationUrls.RowStatistics]}
+ *
+ * ### Example
+ * ```kotlin
+ * // The largest of all `Int` values in the first row
+ * // (so, in the "age" and "weight" columns), or `null` if there are none
```
AndreiKingsley: > (so, in the "age" and "weight" columns)

Add more context —
"Assuming there's two int columns in DF — age and weight,..."

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
```diff
@@ -204,18 +920,123 @@ public inline fun <T, reified C : Comparable<C & Any>?> DataFrame<T>.maxByOrNull
...
+ *   per group.
+ * - [min][Grouped.min] — the mirror operation.
+ * - [aggregate][Grouped.aggregate] — the general way to aggregate groups.
+ * - {@include [MaxDocsLink]} — an overview of all `max` modes.
+ *
+ * For more information: {@include [DocumentationUrls.GroupByStatistics]}
+ *
+ * ### Example
+ * ```kotlin
+ * // For each city, the largest value of each comparable column
```
AndreiKingsley, line 961: We need to write a more detailed comment: "city" is a column; we group by it and take the maximum value from each group, and a new column called ‘max’ will appear with this value.

AndreiKingsley (changes requested):
I am not sure about `maxBy`/`maxFor` descriptions — they look complicated (especially for `maxByOrNull`), and hard to understand difference. May be slightly rephrase/restructurize them.

Also it's hard to get confused about `max()`/`maxFor { }`/`max { } `, I'd emphasise differences/commons between them.

Please add links to pivot/groupBy grammars.

Make comments in examples code more detailed.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
Jolanrensen: indeed, I checked the others were gone, I think it's fine to introduce this concept, right?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
Jolanrensen: not necessarily, but it was on the website, so for completeness it might be nice. Just in case people are confusing maximal and maximum or something

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
```diff
@@ -26,29 +31,268 @@ import org.jetbrains.kotlinx.dataframe.util.ROW_MAX
...
+/**
+ * Returns the first element of this [DataColumn] for which the given [selector]
+ * returns the maximum value, or `null` if there is nothing to compare.
+ *
+ * {@include [MaxDocs.InputValuesSnippet]}
+ *
+ * {@include [MaxDocs.NullOnEmptySnippet]}
+ *
+ * See also:
+ * - [maxBy][DataColumn.maxBy] — throws instead of returning `null` for an empty column.
```
Allex-Nik: Maybe we should remove "for an empty column" here or replace it with something like "for a column with nothing to compare"?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
```diff
@@ -26,29 +31,268 @@ import org.jetbrains.kotlinx.dataframe.util.ROW_MAX
...
     crossinline selector: (T) -> R,
 ): T? = Aggregators.max<R>(skipNaN).aggregateByOrNull(this, selector)
 
+/**
+ * Returns the maximum of the values that the given [selector] returns
+ * for each element of this [DataColumn].
+ *
+ * {@include [MaxDocs.InputValuesSnippet]}
+ *
+ * {@include [MaxDocs.ThrowsOnEmptySnippet]}
```
Allex-Nik, line 242: Would it make sense to merge it with the `@throws` section?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
```diff
@@ -204,18 +920,123 @@ public inline fun <T, reified C : Comparable<C & Any>?> DataFrame<T>.maxByOrNull
...
+ *
+ * See also:
+ * - [max][Grouped.max]`()` — the same, but for all suitable columns at once.
+ * - [max][Grouped.max]` { columns }` — a single maximum of all values in the selected columns,
+ *   per group.
+ * - [minFor][Grouped.minFor] — the mirror operation.
+ * - [aggregate][Grouped.aggregate] — the general way to aggregate groups.
+ * - {@include [MaxDocsLink]} — an overview of all `max` modes.
+ *
+ * For more information: {@include [DocumentationUrls.GroupByStatistics]}
```
Allex-Nik: We also have some examples and explanations here :)
https://kotlin.github.io/dataframe/groupby.html#aggregation-statistics 
And for pivot as well :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
```diff
@@ -351,11 +1526,75 @@ public fun <T, R : Comparable<R & Any>?> Pivot<T>.max(
...
+ * returns for each row, per group.
+ *
+ * Returns a single [DataRow] with the [pivot] keys as (nested) columns, containing the maximum
+ * of the expression's results for the rows of the corresponding group.
+ *
+ * {@include [MaxDocs.RowExpressionSnippet]}
+ *
+ * {@include [MaxDocs.InputValuesSnippet]}
+ *
+ * {@include [MaxDocs.NullCellOnEmptySnippet]}
```
Allex-Nik, line 1571: When the input contains only `null`s, it seems that it throws an exception instead of returning `null` in the result cell.

For example, 
```kotlin
val df = dataFrameOf("firstName", "lastName", "age", "city", "weight", "isHappy")(
        "Alice", "Cooper", 15, "London", 54, true,
        "Bob", "Dylan", 45, "Dubai", 87, true,
        "Charlie", "Daniels", 20, "Moscow", null, false,
)

val res = df.pivot { isHappy }.maxOf { weight?.times(2) }
```
returns 
```Exception in thread "main" java.util.NoSuchElementException: No elements for `maxOf` operation. Use `maxOfOrNull` instead.```

In this case, `df.pivot { isHappy }` contains only one row in the `false` key, and the `weight` in this row is `null`.

Another thing is that `maxOfOrNull` suggested in the exception is not defined on `Pivot`.

The same goes for `PivotGroupBy.maxOf` (maybe it's also the same in `minOf`, need to check).

File: core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt [hunk 14]
```diff
@@ -351,11 +1526,75 @@ public fun <T, R : Comparable<R & Any>?> Pivot<T>.max(
...
+ *
+ * ### Example
+ * ```kotlin
+ * // For each city, the "name" of the person with the largest weight-to-age ratio
+ * df.pivot { city }.maxBy { (weight ?: 0) / age }.with { name }
+ * ```
+ *
+ * @include [MaxDocs.SkipNaNParam]
+ * @param [rowExpression] The [RowExpression] to evaluate for each row.
+ * @return A [ReducedPivot] holding, per group, the first row with the maximum expression result.
```
Allex-Nik: Maybe "the first row for which the [rowExpression] produced the maximum value" would be easier to understand

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
Jolanrensen: this one links to DataColumn.maxOrNull, so no

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
Jolanrensen: I can probably just remove the mention of age and weight. It doesn't add anything useful

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
Jolanrensen: Hmm, I think I prefer having them both. The first one is more towards the top and extensive; people are more likely to read it. However, the `@throws` one gets formatted nicely among `@return` etc., though it looks better smaller

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
Jolanrensen: that's incorrect, `df.groupBy { city }.max()` does not create a column 'max', it's the same as `maxFor { all self comparable cols }`, so it works per-column

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/min.kt
```diff
@@ -433,11 +1927,85 @@ public fun <T, R : Comparable<R & Any>?> PivotGroupBy<T>.min(
...
+ *
+ * ### Example
+ * ```kotlin
+ * // Per city and last name, the "firstName" of the person with the smallest weight-to-age ratio
+ * df.pivot { city }.groupBy { name.lastName }.minBy { (weight ?: 0) / age }.with { name.firstName }
+ * ```
+ *
+ * @include [MinDocs.SkipNaNParam]
+ * @param [rowExpression] The [RowExpression] to evaluate for each row.
+ * @return A [ReducedPivotGroupBy] holding,
```
Allex-Nik: "per group,"?

Allex-Nik (commented):
Left a couple of comments, otherwise it looks good to me :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
AndreiKingsley: Since we use EPs here, it's better to specify explicitly that's these are columns

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/min.kt
Jolanrensen: nice catch

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/max.kt
Jolanrensen: it seems you're right indeed. If there's nothing left to compare, maxOf fails... I'll check everywhere NullCellOnEmptySnippet is used

Jolanrensen: `maxBy` does fill it in with nulls, same as `maxFor`, it's just `maxOf` acting differently

Jolanrensen: I can just replace maxOf with maxOfOrNull in `PivotGroupBy<T>.maxOf()`. That way they function as expected again :) Seems like this was an oversight, actually

AndreiKingsley (approved):
Nice!

Allex-Nik (approved):
Great!

PR #2014, opened 2026-07-30
Migrate NamedValue .copy visibility according to future Kotlin 2.5 requirement

koperagen:
As things now: private constructor with implicitly public constructor, this code will become error in 2.5 release. 
Looks like we deed to decide what ABI we want for it
I propose changing visibility to internal for constructor and for .copy because we use it in implementation, but it doesn't make much sense to have it in public API


Jolanrensen:
For context: https://youtrack.jetbrains.com/issue/KTLC-22/LC-Change-default-visibility-of-generated-copy-method-to-match-primary-constructor-visibility

`@ConsistentCopyVisibility` can be seen as an early opt-in to the behavior where the `copy` method gets the same visibility as the primary constructor. So, `internal` after these changes.

Jolanrensen (approved):

PR #2015, opened 2026-07-30
Remove unused simplegit that required JDK 21+

koperagen:
Seems like it's unused, right? If i'm not missing something. I'd remove it altogether because now by simply existing it requires JDK 21+ for build to work

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -12,8 +12,6 @@ plugins {
         alias(publisher)
         alias(serialization) apply false
         alias(dokka)
-
-        alias(simpleGit) apply false
     }
 }
 
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -56,7 +56,6 @@ kotestAsserions = "6.2.2"
 jsoup = "1.22.2"
 arrow = "19.0.0"
 kodex = "0.6.0"
-simpleGit = "2.3.1"
 caupain = "1.9.1"
 plugin-publish = "2.1.1"
 shadow = "9.5.1"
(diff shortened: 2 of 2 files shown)


Jolanrensen (approved):
you're right! This is now solved by GH actions. It can be removed :)

Though the JDK requirement is just for Gradle itself, right? It's recommended to keep that as up-to-date as possible, while our published library should be compatible with a JVM as old as possible. We currently set gradle-jdk to 21 for the entire project, but we could probably bump that to the latest LTS JDK version (25?) soon

koperagen:
> you're right! This is now solved by GH actions. It can be removed :)
> 
> Though the JDK requirement is just for Gradle itself, right? It's recommended to keep that as up-to-date as possible, while our published library should be compatible with a JVM as old as possible. We currently set gradle-jdk to 21 for the entire project, but we could probably bump that to the latest LTS JDK version (25?) soon

Yes, makes sense to run on latest JDK. At the same time good if it's not strictly required, just a default for us :) I tried building the project on windows and only had JDK 17 there. Had to install 21 additionaly because simple git required it

Jolanrensen:
> Yes, makes sense to run on latest JDK. At the same time good if it's not strictly required, just a default for us :) I tried building the project on windows and only had JDK 17 there. Had to install 21 additionaly because simple git required it

Not just simple-git ;P
https://github.com/Kotlin/dataframe/blob/master/CONTRIBUTING.md#environment-requirements

You can indeed lower the version to 17 by adjusting gradle-jdk in libs.versions.toml, this adjusts it project-wide,
but not sure if something else might break. Was it just simple-git? or did something else break too?

koperagen:
> > Yes, makes sense to run on latest JDK. At the same time good if it's not strictly required, just a default for us :) I tried building the project on windows and only had JDK 17 there. Had to install 21 additionaly because simple git required it
> 
> Not just simple-git ;P https://github.com/Kotlin/dataframe/blob/master/CONTRIBUTING.md#environment-requirements
> 
> You can indeed lower the version to 17 by adjusting gradle-jdk in libs.versions.toml, this adjusts it project-wide, but not sure if something else might break. Was it just simple-git? or did something else break too?

I don't know, let me check

> > Yes, makes sense to run on latest JDK. At the same time good if it's not strictly required, just a default for us :) I tried building the project on windows and only had JDK 17 there. Had to install 21 additionaly because simple git required it
> 
> Not just simple-git ;P https://github.com/Kotlin/dataframe/blob/master/CONTRIBUTING.md#environment-requirements
> 
> You can indeed lower the version to 17 by adjusting gradle-jdk in libs.versions.toml, this adjusts it project-wide, but not sure if something else might break. Was it just simple-git? or did something else break too?

Wait, so this i target JDK of compiled build logic / convention plugins. Didn't realize that. If it's like this i don't mind, for me it failed on simple git probably because it's somehow earlier in resolution. 
kotlin {
    jvmToolchain(libs.versions.gradle.jdk.get().toInt())
}

Jolanrensen:
@koperagen indeed `jvmToolchain(libs.versions.gradle.jdk.get().toInt())` will fail if you don't have the right version installed. Maybe we could provide a more helpful error there ;P (though I thought Gradle might say something more helpful...)

Anyway, we should still remove simple-git :)

PR #2017, opened 2026-07-31
Configure Gradle daemon JVM toolchain

### LINKED ISSUE #2016: Setup gradle-daemon-jvm.properties
Right now to run our build one needs JDK 21+ installed on PC as a prerequisite. Gradle seems to provide a way to automatically install a toolchain JDK specifically for running a gradle daemon
https://docs.gradle.org/current/userguide/gradle_daemon.html#sec:configure_provision

koperagen:
## Summary

- add the generated `gradle/gradle-daemon-jvm.properties` file
- require Java 21 for the Gradle daemon, matching the existing `gradle-jdk` version
- provide Foojay download URLs for supported x86-64 and AArch64 platforms so Gradle can provision the daemon JDK when no compatible local JDK is installed

## Validation

- confirmed the root settings convention already applies `org.gradle.toolchains.foojay-resolver-convention` 1.0.0
- confirmed the project pins the Gradle JDK to Java 21
- verified the branch is one commit ahead of `master` and changes only the generated daemon JVM properties file

Full build not run; this is a generated build-configuration-only change.

Closes #2016


koperagen:
Updated the implementation after review of the maintenance workflow.

The generated properties file is no longer the only change. The root `updateDaemonJvm` task now reads its language version from `libs.versions.gradle.jdk`, making that catalog entry the single source of truth for both project toolchains and the Gradle daemon.

Upgrade workflow, including a future move from JDK 21 to JDK 25:

1. Change `gradle-jdk` in `gradle/libs.versions.toml`.
2. Run `./gradlew updateDaemonJvm`.
3. Commit the regenerated `gradle/gradle-daemon-jvm.properties`.

The literal platform URLs in that file are generated task output, analogous to Gradle Wrapper metadata; they are not intended to be edited manually. `CONTRIBUTING.md` now documents the same workflow and no longer lists a locally installed JDK 21 as a prerequisite.

This comment supersedes the original PR summary, which described the initial generated-file-only version of the change.

CONTRIBUTING.md
```diff
@@ -85,10 +85,12 @@ so do familiarize yourself with the following guidelines.
...
 ## Environment requirements
 
-* JDK >= 21 referred to by the `JAVA_HOME` environment variable.
+* A Java runtime available through `JAVA_HOME` or `PATH` is required to launch the Gradle Wrapper.
+  The build itself runs on the daemon JDK provisioned from `gradle/gradle-daemon-jvm.properties`.
 
-  * Note, any version above 21 should work in theory, but JDK 21 is the only version we test with,
-  so it is the recommended version.
+  * The daemon JDK version is controlled by `gradle-jdk` in `gradle/libs.versions.toml`.
+    After changing it (for example, from `21` to `25`), run `./gradlew updateDaemonJvm`
```
Jolanrensen, line 92: could you mention this `updateDaemonJvm` task also in comments in the .toml file? Just so when we bump it there, we don't forget to run the task

build.gradle.kts
```diff
@@ -15,6 +17,13 @@ plugins {
     }
 }
 
+// `gradle-jdk` is the single source of truth for both build toolchains and the Gradle daemon.
+// To upgrade (for example, to JDK 25), change it in gradle/libs.versions.toml, run
+// `./gradlew updateDaemonJvm`, and commit the regenerated gradle-daemon-jvm.properties file.
+tasks.named<UpdateDaemonJvm>("updateDaemonJvm") {
+    languageVersion.set(libs.versions.gradle.jdk.map { JavaLanguageVersion.of(it) })
```
Jolanrensen: could use `languageVersion = `, slightly cleaner :)

PR #2031, opened 2026-08-14
Updated website docs, added KDocs and tests for `valueCounts`

### LINKED ISSUE #1992: Add KDocs for `valueCounts` APIs
# Add KDocs for `valueCounts` APIs

The file `core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt` has **no KDocs at all**. Please add concise KDocs for the non-deprecated APIs, excluding deprecated / `@AccessApiOverload` overloads.

### Scope

Non-deprecated APIs in `valueCounts.kt`:

- The `ValueCount` `@DataSchema` interface (`count: Int`) — the result schema.
- `DataColumn<T>.valueCounts(sort, ascending, dropNA, resultColumn)` — returns a `DataFrame<ValueCount>`.
- `DataFrame<T>.valueCounts(sort, ascending, dropNA, resultColumn, columns)` — selector overload.
- `DataFrame<T>.valueCounts(vararg columns: String, sort, ascending, dropNA, resultColumn)` — by-name overload.

Out of scope: the deprecated `vararg columns: AnyColumnReference` / `vararg columns: KProperty<*>` overloads.

### KDocs should explain

- What `valueCounts` does: counts the occurrences of each distinct value (for a column) or each distinct row / combination of the selected columns (for a `DataFrame`), returning a new frame with the distinct values plus a count column.
- The parameters and their defaults:
  - `sort` (default `true`) — whether to sort by count;
  - `ascending` (default `false`) — sort direction, i.e. by default most frequent first;
  - `dropNA` (default `true`) — whether `NA`/`null` values are excluded from counting;
  - `resultColumn` (default `"count"`) — the name of the count column, and that it is made unique if it clashes with an existing column name (e.g. `count` → `count1`);
  - `columns` — which columns to count distinct combinations of (whole frame if `null`).
- The result shape: the value/selected columns are preserved, with the added count column of type `Int`; the `DataFrame` result schema is `ValueCount`-based.

### Test coverage assessment

- There is **no dedicated test file** for `valueCounts.kt`; it is only exercised incidentally (`samples/api/Analyze.kt`, `testSets/person/DataColumnTests.kt`).
- Optional follow-up: a focused test file covering `sort`/`ascending`, `dropNA` behavior with nulls, the `resultColumn` name-clash handling, and multi-column counting.

### Website coverage

- Documented at `valueCounts.html`; keep KDocs consistent with it.

### Acceptance criteria

- All non-deprecated `valueCounts` overloads and the `ValueCount` schema interface have concise KDocs.
- No documentation is added for the deprecated / `@AccessApiOverload` overloads.
- KDocs cover every parameter and its default, the `dropNA` and count-column-name-clash behavior, and the result shape.
- Docs are consistent with the site page `valueCounts.html`.

Allex-Nik:
Fixes #1992
Helps #898


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
```diff
@@ -10,16 +10,76 @@ import org.jetbrains.kotlinx.dataframe.annotations.Interpretable
...
 import kotlin.reflect.typeOf
 
+// region docs
+
+/**
+ * @param [sort\] If `true` (default), the result is sorted by count.
+ * Otherwise, the counted values keep the order of their first occurrence.
+ * @param [ascending\] The sorting direction. If `false` (default), the most frequent values come first.
+ * Only used when [sort\] is `true`.
+ * @param [dropNA\] If `true` (default), [`NA`][NA] values (`null`s and `NaN`s) are not counted
```
Jolanrensen: you can use `{@include [NALink]}`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
```diff
@@ -10,16 +10,76 @@ import org.jetbrains.kotlinx.dataframe.annotations.Interpretable
...
+ *
+ * The resulting [DataFrame] contains:
+ * - the distinct combinations of these values,
+ * - a new [Int] column (named [resultColumn\], default is `"count"`) with the number of occurrences of each combination.
+ *
+ * By default, the result is sorted by count in descending order (the most frequent combination first),
+ * and a row is not counted at all if any of its selected values is [`NA`][NA].
+ *
+ * See also:
+ * - [valueCounts][DataColumn.valueCounts] — counts of unique values in a single [DataColumn].
```
Jolanrensen: not 100% whether we should already put the aliases in backticks, but I can do that in my PR if this one is merged first :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
```diff
@@ -29,6 +89,41 @@ public interface ValueCount {
 
 internal val defaultCountColumnName: String = ValueCount::count.name
 
+/**
+ * Returns a [DataFrame] containing counts of unique values in this [DataColumn].
+ *
+ * The resulting [DataFrame] contains:
+ * - the column with the distinct values of the original [DataColumn]
+ * - a new [Int] column ([resultColumn]) with the number of occurrences of each value.
```
Jolanrensen, line 98: I would rewrite this part about "resultColumn" to say that it is named "count" by default unless this name is overridden by `resultColumn`. Same in `CommonValueCountsDocs`

Jolanrensen: come to think of it... if `resultColumn` is supplied, the `cast` to `DataFrame<ValueCount>` is broken... Maybe it makes sense to add an overload without `resultColumn` that returns `DataFrame<ValueCount>` and one with `resultColumn` that simply returns `AnyFrame` :) we could make a separate issue for it... but it's a quick fix imo, wdyt?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
```diff
@@ -56,6 +151,27 @@ public fun <T> DataColumn<T>.valueCounts(
...
+ * // starting with the most frequent one
+ * df.valueCounts { name and city }
+ *
+ * // Counts the unique rows of the whole dataframe,
+ * // including rows with `NA` values, in the order of their first occurrence
+ * df.valueCounts(sort = false, dropNA = false)
+ * ```
+ *
+ * @include [ValueCountsParams]
+ * @param [columns] The [ColumnsSelector] used to select the columns whose distinct value combinations
```
Jolanrensen: 'optional [ColumnsSelector]'. I'd repeat here the "if null or not supplied, all columns are selected" statement :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
```diff
@@ -78,6 +194,22 @@ public fun <T> DataFrame<T>.valueCounts(
...
+ * #### Example
+ *
+ * ```kotlin
+ * // Counts the unique combinations of the values in the "name" and "city" columns,
+ * // starting with the most frequent one
+ * df.valueCounts("name", "city")
+ * ```
+ *
+ * @include [ValueCountsParams]
+ * @param [columns] The names of the columns whose distinct value combinations are counted.
```
Jolanrensen: also optional. If unsupplied, all columns will be selected (yes, it will resolve to the other overload, but to the user will not notice this difference)

Jolanrensen: also, this param should be the first :) above ValueCountsParams

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
```diff
@@ -0,0 +1,269 @@
...
+
+        resultValues[0] shouldBe 1.0
+        (resultValues[1] as Double).isNaN() shouldBe true
+        resultValues[2] shouldBe 2.0
+        result.count.values() shouldBe listOf(2, 2, 1)
+    }
+
+    @Test
+    fun `value counts of a column with a custom result column name`() {
+        val result = df["city"].valueCounts(resultColumn = "number")
```
Jolanrensen, line 93: You are able to type `result.count` here, resulting in an exception because this column doesn't exist

docs/StardustDocs/topics/valueCounts.md
```diff
@@ -1,30 +1,84 @@
 [//]: # (title: valueCounts)
 
-<!---IMPORT org.jetbrains.kotlinx.dataframe.samples.api.Analyze-->
+<!---IMPORT org.jetbrains.kotlinx.dataframe.samples.api.ValueCountsSamples-->
 
 Return [`DataFrame`](DataFrame.md) containing counts of unique values in [`DataFrame`](DataFrame.md)
```
Jolanrensen: Returns a DataFrame containing the counts of the unique values in a DataFrame or DataColumn

docs/StardustDocs/topics/valueCounts.md
```diff
@@ -1,30 +1,84 @@
...
-* `sort: Boolean = true` — sort by count
-* `ascending: Boolean = false` — sort in ascending order
-* `dropNA: Boolean = true` — don't include counts of [`NA`](nanAndNa.md) values 
-* `columns = all` — columns to use when counting unique combinations 
+* `sort: Boolean = true` — whether to sort by count
+* `ascending: Boolean = false` — sort direction (by default most frequent first)
+* `dropNA: Boolean = true` — doesn't include counts of [`NA`](nanAndNa.md) values 
+* `resultColumn: String = "count"` — name of the column with counts
+* `columns = all` — columns to use when counting unique combinations
+
```
Jolanrensen, line 21: I'd put a small introduction here. Something like "For example:"

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/ValueCountsSamples.kt
```diff
@@ -0,0 +1,87 @@
...
+
+    private val aliceColor = RgbColor(189, 206, 233)
+
+    @Test
+    fun valueCountsDf() {
+        // SampleStart
+        df
+            // SampleEnd
+            .format().perRowCol { row, _ ->
+                if (row.name == "Alice" && row.age == 15) background(aliceColor) and textColor(black) else null
```
Jolanrensen: I wonder if it makes sense to give the `Alice, null` row a different color to annotate it's being considered too, but `dropNA` makes it be skipped. (The text could be stricken through maybe? Or maybe not since it's used in the second example, let's see what looks best :) )

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/ValueCountsSamples.kt
```diff
@@ -0,0 +1,87 @@
...
+            .saveDfHtmlSample()
+    }
+
+    @Test
+    fun valueCountsWithNA() {
+        // SampleStart
+        df.valueCounts(dropNA = false)
+            // SampleEnd
+            .format().perRowCol { row, _ ->
+                if (row.name == "Alice" && row.age == null) background(aliceColor) and textColor(black) else null
```
Jolanrensen: Ah yes and then here the `Alice, 15` row could be `aliceColor` and the `Alice, null` row could be the same color as `Alice, null` above :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
```diff
@@ -10,16 +10,76 @@ import org.jetbrains.kotlinx.dataframe.annotations.Interpretable
...
+ * and are excluded from the result.
+ * @param [resultColumn\] The name of the resulting count column. Default — `"count"`.
+ * If a column with this name is already present, the name is made unique by appending
+ * a number to it (for example, `"count"` becomes `"count1"`).
+ */
+@ExcludeFromSources
+private typealias ValueCountsParams = Nothing
+
+/**
+ * Returns a [DataFrame] containing counts of unique rows (or combinations of selected values) in this [DataFrame].
```
Jolanrensen: *the counts

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
```diff
@@ -10,16 +10,76 @@ import org.jetbrains.kotlinx.dataframe.annotations.Interpretable
...
+ * - [valueCounts][DataColumn.valueCounts] — counts of unique values in a single [DataColumn].
+ * - [countDistinct][DataFrame.countDistinct] — the number of distinct rows, without their counts.
+ * - [distinct][DataFrame.distinct] — the distinct rows themselves, without their counts.
+ * - [count][DataFrame.count] — the total number of rows in this [DataFrame].
+ *
+ * For more information: {@include [DocumentationUrls.ValueCounts]}
+ *
+ * All summary statistics: {@include [DocumentationUrls.Statistics]}
+ *
+ * ### This ValueCounts Overload
```
Jolanrensen: *Value Counts (like a title) or *`valueCounts` referring to the operation name, the name of the function

Jolanrensen (changes requested):
Overall good :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
Allex-Nik: "[`NA`][NA]" is resolved to the same `internal typealias NA = Nothing` with its docs, I haven't seen any difference :) 
Is using `{@include [NALink]}` better?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
Allex-Nik: I've added the backticks to the aliases :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
Allex-Nik: We can do that. It also should not break the DataFrame IDEA plugin as far as I see. The only thing that bothers me is that providing `resultColumn` is not the only way to make the name of the column with counts not "count": if the name of the receiver is the same as the name of the column with counts (so the new column becomes `count1`), the cast will also break, and another overload will not fix it, right? It's a rare case though

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
Jolanrensen: it indeed resolves the same :) It's just a reusable link. If we ever decide to change the name, the links that are reused will change too

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
Jolanrensen: Damn, you're right. No, only the compiler plugin can fix that :)

Jolanrensen: It seems compiler plugin support is missing for both `DataColumn<T>.valueCounts` and `DataFrame<T>.valueCounts(vararg columns: String...)`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
Allex-Nik: Ah, I see, thanks! I've replaced "[`NA`][NA]" with `{@include [NALink]}`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
Allex-Nik: Yes, I can see the support only for `DataFrame<T>.valueCounts(columns: ColumnsSelector<T, *>?...`. Should we create an issue about the missing support? It seems possible in principle to add it for these overloads, but maybe there is something why it can't be done?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
Allex-Nik: Done

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
Allex-Nik: Done

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
Allex-Nik: Yes, sadly :(
If we open an issue for compiler plugin support and this problem will be fixed, I think the proper access to the columns of the result will be tested on the compiler plugin side and it will not require any changes in these tests in the DataFrame repo. 
Or do you think we should change anything here?

docs/StardustDocs/topics/valueCounts.md
Allex-Nik: Fixed, thanks

docs/StardustDocs/topics/valueCounts.md
Allex-Nik: Done

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/ValueCountsSamples.kt
Allex-Nik: I agree, fixed it, it's better now :)

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/ValueCountsSamples.kt
Allex-Nik: Done

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
Jolanrensen: Yes, please do :) Adding support for the String overload should be pretty straightforward nowadays. Support for the DataColumn overload might require us to fetch the name of the column.

@koperagen can the compiler plugin do that? Fetch the name of the receiver column when calling `DataColumn<T>.valueCounts(): DataFrame<R>`?

In any case, we can refine the returned DataFrame type with the right name for the resultColumn.

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
Jolanrensen: on the compiler side should be fine indeed :)

Jolanrensen (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/valueCounts.kt
koperagen: Support for string overload will be done as a part of big PR that enables support for all of those that already have ColumnsSelector interpreter
as for DataColumn, generally no column name tracking is performed. Will consider it

PR #2032, opened 2026-08-15
Added KDocs for sum.kt

### LINKED ISSUE #1985: Add KDocs for non-deprecated `sum` APIs
# Add KDocs for non-deprecated `sum` APIs

The file `core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt` holds the public `sum` APIs, and the coverage sheet marks the group `sum/*` as missing KDocs. None of these functions are currently documented (there is only a `TODO KDocs` comment block at the top of the file). Please add concise KDocs for the current non-deprecated APIs only, excluding deprecated and binary-compatibility (`@AccessApiOverload`, and the whole `// region binary compatibility` of `DeprecationLevel.HIDDEN` overloads) overloads.

### Scope

APIs needing KDocs, by receiver:

- `DataColumn.*` — `sum`, `sumOf`
- `DataRow.*` — `rowSum`, `rowSumOf`
- `DataFrame.*` — `sum`, `sumFor`, `sumOf`
- `Grouped.*` — `sum`, `sumFor`, `sumOf`
- `Pivot.*` — `sum`, `sumFor`, `sumOf`
- `PivotGroupBy.*` — `sum`, `sumFor`, `sumOf`

### KDocs should explain

  - `sum` — the sum of the values (over the whole column / row / frame);
  - `sumFor` — a `DataRow`/`DataFrame` with a separate sum per selected column;
  - `sumOf` — the sum of a value computed per row/element via the given `expression`;
  - the grouped / pivoted variants — sum computed per group / per pivot cell.
- The common numeric behavior (already captured in the `TODO KDocs` block, lines 35–42):
  - supported for all primitive number types; for mixed primitive number types the values are unified before summing;
  - `null`s are filtered out;
  - the return type is always the same as the input type and never `null`, **except** `Byte` and `Short`, which are returned as `Int`;
  - empty input yields `0` in the corresponding number type.
- The `skipNaN` parameter: how `NaN` values in floating-point columns are treated.
- The `separate` parameter for the `Pivot` / `PivotGroupBy` variants, and the optional `name` / `resultName` parameter for the `Grouped` variants.
- Note the `rowSumOf` overloads are split per concrete primitive type (`Short`/`Byte`/`Int`/`Long`/`Float`/`Double`) plus a `KType` overload, because a single `reified T : Number?` is not possible — a shared KDoc (e.g. via `@include`) is fine.

### Acceptance criteria

- Concise KDocs are added to all non-deprecated `sum` / `sumFor` / `sumOf` / `rowSum` / `rowSumOf` overloads in `sum.kt`.
- No documentation is added for deprecated / `@AccessApiOverload` overloads, nor for anything in the `// region binary compatibility` block.
- KDocs cover the `skipNaN` parameter, the "same-as-input type, never null" return contract (with the `Byte`/`Short` → `Int` exception), empty-input → `0`, and null-filtering.
- KDocs clearly distinguish the value-returning (`sum`, `sumOf`, `rowSum*`) variants from the per-column aggregation (`sumFor`) variants.
- Docs are consistent with the site docs at `sum.html`.
- The `TODO KDocs` comment block at the top of the file is removed once its content has been folded into the actual KDocs.

Jolanrensen:
Fixes #1985 

Wrote KDocs for sum.kt with the help of Claude AI. I also moved some common stuff from the MinMaxDocs to CommonStatisticsDocs for reuse later :)

> Regarding [#2032 (comment)](https://github.com/Kotlin/dataframe/pull/2032#discussion_r4070524117) and [#2032 (comment)](https://github.com/Kotlin/dataframe/pull/2032#discussion_r4047060686)
> 
> I redid the fix properly:
> 
> * `AggregatorInputHandler` now requires a `valueTypeIfEmpty: KType`
> * This is set to `Nothing` by default but can be overridden by each input handler and thus any aggregator.
> * For `mean`, and `std`, this is set to `Double`, as `Double` empty input -> `Double.NaN`
> * For `sum` this is set to the input type for `rowSumOf<T>()` or the column type if it's called on a single column. Else, I set it to `Double`. This makes sure the `0` value of the expected type will be used in case there are no values.


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -33,38 +38,687 @@ import kotlin.reflect.KType
...
+     */
+    @ExcludeFromSources
+    typealias ZeroOnEmptySnippet = Nothing
+
+    /**
+     * {@comment Note about the behavior on empty input for the modes with multiple results. KDoc-snippet.}
+     *
+     * Result cells for which there is nothing to sum
+     * (for instance, because the input was empty or contained only `null` values)
+     * simply become `0` in the result type (defaulting to [Double] if there is no number type to speak of).
```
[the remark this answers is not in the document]
Jolanrensen: good point

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -115,6 +885,26 @@ public inline fun <reified T : Double> DataRow<*>.rowSumOf(
...
+ * df[0].rowSumOf(typeOf<Int>())
+ * ```
+ * @param [type] The type of the values to sum. Only columns of this type are taken into account.
+ * @include [SumDocs.SkipNanParam]
+ * @return The sum of the values of the given [type] in this row.
+ * @throws IllegalArgumentException if [type] is not a primitive number type or [Number] itself.
+ */
 public fun DataRow<*>.rowSumOf(type: KType, skipNaN: Boolean = skipNaNDefault): Number {
     require(type.isPrimitiveOrMixedNumber()) {
         "Type $type is not a primitive number type. Mean only supports primitive number types."
```
Allex-Nik, line 910: I think we can fix "mean" while we are at it

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -77,36 +731,152 @@ public inline fun <C, reified V : Number?> DataColumn<C>.sumOf(
...
+ * ```kotlin
+ * // The sum of all `Double` values in the first row, ignoring `NaN` values
+ * df[0].rowSumOf<Double>(skipNaN = true)
+ * ```
+ * @include [SumDocs.SkipNanParam]
+ * @return The sum of the [Double] values in this row.
+ */
 @Suppress("FINAL_UPPER_BOUND")
 @JvmName("rowSumOfDouble")
 public inline fun <reified T : Double> DataRow<*>.rowSumOf(
```
Allex-Nik, line 882: Would it be beneficial to add a note explaining how it works with nested columns and the restrictions in this regard?

For example, if I do

```kotlin
val df = dataFrameOf(
        "measurements" to columnOf(
            "age" to columnOf(20, 25),
            "weight" to columnOf(15, 20)
        ),
    )
    val result = df[0].rowSumOf<Int>()
```

I get 
```
Exception in thread "main" java.lang.IllegalArgumentException: Types [] are not all primitive numbers, only those are supported in sum.
```

And if I do 

```kotlin
val df = dataFrameOf(
        "measurements" to columnOf(
            "age" to columnOf(20, 25),
            "weight" to columnOf(15, 20)
        ),
        "number" to columnOf(1, 2)
    )
    val result = df[0].rowSumOf<Int>()
```

Then I get 1.

So apparently it only looks at top-level value columns.

But I think a user can easily bump into unexpected behavior when they use a df with ColumnGroup, for example.

There is a similar thing with `DataFrame<T>.sum` (but it returns an empty row instead of throwing an exception), and maybe with other overloads

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -126,18 +916,68 @@ public fun DataRow<*>.rowSumOf(type: KType, skipNaN: Boolean = skipNaNDefault):
 // endregion
 
 // region DataFrame
+
+/**
+ * Returns the sum of the values of each suitable column of this [DataFrame] separately.
+ *
+ * @include [SumDocs.AllSuitableColumnsSnippet]
+ * @include [SumDocs.SupportedTypesSnippet]
+ * @include [SumDocs.ZeroCellOnEmptySnippet]
```
Allex-Nik, line 925:
```kotlin
val df = dataFrameOf(
        "number" to columnOf(null, null)
    )
val result = df.sum()
```

returns an empty row rather than a row with a 0

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -310,15 +1373,73 @@ public inline fun <T, reified R : Number?> Grouped<T>.sumOf(
...
+ * df.pivot { city }.sumFor { age and weight }
+ * // The same, but with the results grouped by aggregated column instead of by city
+ * df.pivot { city }.sumFor(separate = true) { age and weight }
+ * ```
+ * @include [SumDocs.SeparateParam]
+ * @include [SumDocs.SkipNanParam]
+ * @include [SumDocs.AggregateColumnsSelectorParam]
+ * @return A single [DataRow] with the sum of each selected column per [pivot] group.
+ */
 public fun <T, R : Number?> Pivot<T>.sumFor(
```
Allex-Nik, line 1425: The example seems to not be rendered here and in some other functions (I marked them below).
Adding a missing backtick in `DocumentationUrls` fixes it. See the comment there

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -310,15 +1373,73 @@ public inline fun <T, reified R : Number?> Grouped<T>.sumOf(
...
+ * ```kotlin
+ * // For each city, the sum of the "age" values and the sum of the "weight" values
+ * df.pivot { city }.sumFor("age", "weight")
+ * ```
+ * @include [SumDocs.ColumnNamesParam]
+ * @include [SumDocs.SeparateParam]
+ * @include [SumDocs.SkipNanParam]
+ * @return A single [DataRow] with the sum of each selected column per [pivot] group.
+ */
 public fun <T> Pivot<T>.sumFor(
```
Allex-Nik, line 1443: The example is not rendered

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -341,11 +1462,35 @@ public fun <T, C : Number?> Pivot<T>.sumFor(
...
+ * ```kotlin
+ * // For each city, the sum of all values in the "age" and "weight" columns
+ * df.pivot { city }.sum { age and weight }
+ * ```
+ * @include [SumDocs.SkipNanParam]
+ * @include [SumDocs.ColumnsSelectorParam]
+ * @return A single [DataRow] with, per [pivot] group, the sum of all the values
+ *   in the selected columns.
+ */
 public fun <T, C : Number?> Pivot<T>.sum(
```
Allex-Nik, line 1477: The example is not rendered

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -341,11 +1462,35 @@ public fun <T, C : Number?> Pivot<T>.sumFor(
...
+ * ```kotlin
+ * // For each city, the sum of all values in the "age" and "weight" columns
+ * df.pivot { city }.sum("age", "weight")
+ * ```
+ * @include [SumDocs.ColumnNamesParam]
+ * @include [SumDocs.SkipNanParam]
+ * @return A single [DataRow] with, per [pivot] group, the sum of all the values
+ *   in the selected columns.
+ */
 public fun <T> Pivot<T>.sum(vararg columns: String, skipNaN: Boolean = skipNaNDefault): DataRow<T> =
```
Allex-Nik, line 1494: The example is not rendered

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -370,15 +1545,73 @@ public inline fun <T, reified R : Number?> Pivot<T>.sumOf(
...
+ * ```kotlin
+ * // Per city and last name, the sum of the "age" values and the sum of the "weight" values
+ * df.pivot { city }.groupBy { name.lastName }.sumFor { age and weight }
+ * ```
+ * @include [SumDocs.SeparateParam]
+ * @include [SumDocs.SkipNanParam]
+ * @include [SumDocs.AggregateColumnsSelectorParam]
+ * @return A [DataFrame] with the sum of each selected column per group.
+ */
 public fun <T, R : Number?> PivotGroupBy<T>.sumFor(
```
Allex-Nik, line 1597: The example is not rendered

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -370,15 +1545,73 @@ public inline fun <T, reified R : Number?> Pivot<T>.sumOf(
...
+ * ```kotlin
+ * // Per city and last name, the sum of the "age" values and the sum of the "weight" values
+ * df.pivot { city }.groupBy { name.lastName }.sumFor("age", "weight")
+ * ```
+ * @include [SumDocs.ColumnNamesParam]
+ * @include [SumDocs.SeparateParam]
+ * @include [SumDocs.SkipNanParam]
+ * @return A [DataFrame] with the sum of each selected column per group.
+ */
 public fun <T> PivotGroupBy<T>.sumFor(
```
Allex-Nik, line 1615: The example is not rendered

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -401,11 +1634,33 @@ public fun <T, C : Number?> PivotGroupBy<T>.sumFor(
...
+ * @set [SumDocs.PivotGroupBySumSnippet.EXAMPLE]
+ * ```kotlin
+ * // Per city and last name, the sum of all values in the "age" and "weight" columns
+ * df.pivot { city }.groupBy { name.lastName }.sum { age and weight }
+ * ```
+ * @include [SumDocs.SkipNanParam]
+ * @include [SumDocs.ColumnsSelectorParam]
+ * @return A [DataFrame] with, per group, the sum of all the values in the selected columns.
+ */
 public fun <T, C : Number?> PivotGroupBy<T>.sum(
```
Allex-Nik, line 1648: The example is not rendered

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -401,11 +1634,33 @@ public fun <T, C : Number?> PivotGroupBy<T>.sumFor(
...
+ * @set [SumDocs.PivotGroupBySumSnippet.EXAMPLE]
+ * ```kotlin
+ * // Per city and last name, the sum of all values in the "age" and "weight" columns
+ * df.pivot { city }.groupBy { name.lastName }.sum("age", "weight")
+ * ```
+ * @include [SumDocs.ColumnNamesParam]
+ * @include [SumDocs.SkipNanParam]
+ * @return A [DataFrame] with, per group, the sum of all the values in the selected columns.
+ */
 public fun <T> PivotGroupBy<T>.sum(vararg columns: String, skipNaN: Boolean = skipNaNDefault): DataFrame<T> =
```
Allex-Nik, line 1664: The example is not rendered

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/DocumentationUrls.kt
```diff
@@ -323,6 +323,16 @@ public interface DocumentationUrls {
     /** [See "Row statistics" on the documentation website.]({@include [Url]}/rowstats.html) */
     public typealias RowStatistics = Nothing
 
+    /** [See "Number Unification" on the documentation website.]({@include [Url]}/numberunification.html) */
+    public typealias NumberUnification = Nothing
```
Allex-Nik, line 327: Rendering of examples works if we add a missing backtick to `PivotAggregationStatistics` above

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -33,38 +38,687 @@ import kotlin.reflect.KType
...
+        // The note about `NaN` handling. Filled in by default; must be set to nothing for the
+        // overloads of a fixed integer type, as those have no `skipNaN` parameter to refer to.
+        typealias NAN_NOTE = Nothing
+    }
+
+    /**
+     * {@comment Note about the behavior on empty input for the modes with a single result. KDoc-snippet.}
+     *
+     * When there is nothing to sum, for instance, when the input is empty or contains only `null` values,
+     * the result is `0` in the result type (defaulting to [Double] if there is no number type to speak of).
```
Allex-Nik, line 113:
```kotlin
val df = dataFrameOf("name" to columnOf("Alice", "Bob", "Charlie", "Charlie"))
val result = df[0].rowSum()
```
throws
```
Exception in thread "main" java.lang.IllegalArgumentException: Types [] are not all primitive numbers, only those are supported in sum.
```

The same holds for `val result = df[0].rowSumOf<Int>()`

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -193,6 +1083,23 @@ public fun <T, C : Number?> DataFrame<T>.sum(
...
+ * @set [SumDocs.DataFrameSumSnippet.EXAMPLE]
+ * ```kotlin
+ * // The sum of all values in the "age" and "weight" columns together
+ * df.sum("age", "weight")
+ * ```
+ * @include [SumDocs.ColumnNamesParam]
+ * @include [SumDocs.SkipNanParam]
+ * @return The sum of all the values in the selected columns.
+ */
 public fun <T> DataFrame<T>.sum(vararg columns: String, skipNaN: Boolean = skipNaNDefault): Number =
```
Allex-Nik, line 1103: It can throw an exception:

```kotlin
val df = dataFrameOf("name" to columnOf("Alice", "Bob", "Charlie", "Charlie"))
val result = df.sum("name")
```

```
Exception in thread "main" java.lang.IllegalArgumentException: Type ValueType(kType=kotlin.String, needsFullConversion=false) is not a subtype of Number?, only primitive numbers are supported in sum.
```

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -298,6 +1331,36 @@ public fun <T, C : Number?> Grouped<T>.sum(
...
+ * ### Example
+ * ```kotlin
+ * // For each city, the sum of the weight-to-age ratios, in a column called "totalRatio"
+ * df.groupBy { city }.sumOf("totalRatio") { (weight ?: 0) / age }
+ * ```
+ *
+ * @param [resultName] The name of the resulting column. If `null` (the default), `"sum"` is used.
+ * @include [SumDocs.SkipNanParam]
+ * @include [SumDocs.ExpressionParam]
+ * @return A new [DataFrame] with the group keys and a single sum per group.
```
Allex-Nik, line 1362: Throws 
```
Exception in thread "main" org.jetbrains.kotlinx.dataframe.api.DuplicateColumnPathInsertException: Can not insert column `sum` because column with this path already exists in DataFrame
```

if "sum" exists and we group by it

Not sure if we need to document it though :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -33,38 +38,687 @@ import kotlin.reflect.KType
...
+ * for each element of this [DataColumn], as an [Int].
+ *
+ * @include [SumDocs.DataColumnSumOfSnippet] {@set [SumDocs.SupportedTypesSnippet.NAN_NOTE]}
+ * @set [SumDocs.DataColumnSumOfSnippet.EXAMPLE]
+ * ```kotlin
+ * // The sum of all halved values in the "amount" column of `Byte`s, as an `Int`
+ * df.amount.sumOf { (it / 2).toByte() }
+ * ```
+ * @return The sum of the values [expression] returns, as an [Int].
+ */
```
Allex-Nik, line 703: Do we need the "All primitive number types are supported" paragraph in overloads with the fixed type?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -33,38 +38,687 @@ import kotlin.reflect.KType
...
+     * @comment Version of [SelectingColumns] with correctly filled in examples
+     * @include [SelectingColumns] {@include [SetSumForOperationArg]}
+     */
+    typealias SumForSelectingOptions = Nothing
+
+    /**
+     * @include [SumDocs.ZeroOnEmptySnippet]
+     *
+     * See also:
+     * - [`sumOf`][DataColumn.sumOf] — the sum of the values a selector returns for each element.
```
Allex-Nik, line 175: Would "an expression returns" be better?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -77,36 +731,152 @@ public inline fun <C, reified V : Number?> DataColumn<C>.sumOf(
...
+ * @set [SumDocs.ReifiedRowSumOfSnippet.EXAMPLE]
+ * ```kotlin
+ * // The sum of all `Int` values ("age" and "weight") in the first row
+ * df[0].rowSumOf<Int>()
+ * ```
+ * @return The sum of the [Int] values in this row.
+ */
 @Suppress("FINAL_UPPER_BOUND")
 @JvmName("rowSumOfInt")
 public inline fun <reified T : Int> DataRow<*>.rowSumOf(_kClass: KClass<Int> = Int::class): Int =
```
Allex-Nik, line 821: Can we say that the result defaults to Double in this case?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -33,38 +38,687 @@ import kotlin.reflect.KType
...
+ * df.age.sum()
+ * // The sum of all weights in the "weight" `Double?` column, ignoring `null` values
+ * df.weight.sum()
+ * ```
+ *
+ * @include [SumDocs.SkipNanParam]
+ * @return The sum of the values in this column.
+ */
 @Suppress("UNCHECKED_CAST")
 @JvmName("sumNumber")
```
Allex-Nik, line 671:
```kotlin
val df = dataFrameOf(
        "name" to columnOf("Alice", "Bob", "Charlie", "Charlie"),
        "age" to columnOf<Number>(1, 2.0, 3, 4),
    )
val result = df.age.take(0).sum()
```
also throws
```
Exception in thread "main" java.lang.IllegalArgumentException: Types [] are not all primitive numbers, only those are supported in sum.
```
even though docs say that when there is nothing to sum, for instance, when the input is empty or contains only null values, the result is 0

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -33,38 +38,687 @@ import kotlin.reflect.KType
...
+ * ```kotlin
+ * // The total length of all first names in the "name"/"firstName" column
+ * df.name.firstName.sumOf { it.length }
+ * ```
+ * @include [SumDocs.SkipNanParam]
+ * @return The sum of the values [expression] returns.
+ */
 @Suppress("UNCHECKED_CAST")
 @JvmName("sumOfNumber")
 @OverloadResolutionByLambdaReturnType
```
Allex-Nik, line 724: Not sure if it needs to be changed, but I stumbled over the phrase "All primitive number types are supported" here, because it's true for the result of the expression, but the column can have other types

Allex-Nik (commented):
Added some comments with suggestions. Some of them are applicable to other overloads as well. When I checked something manually and encountered a problem, I added the comment mostly only to the overload where I found the problem.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
Jolanrensen: that's because a `Nothing?` column is not a suitable column. Any suitable column will become 0

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
Jolanrensen: hmm yeah I think it makes sense to mention column groups indeed

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
Jolanrensen: indeed! It's not a selector

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
Jolanrensen: uhhm, yeah we could mention it for the resultName param

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/DocumentationUrls.kt
Jolanrensen: nice :D was already picked up by the grammar nazi bot on master

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
Jolanrensen: I'd say yes, because the overload is selected by the inferred type. When you're still writing the function or reading the KDocs before deciding to use it, it shouldn't matter which of the overloads you're reading. Does that make sense?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
Jolanrensen: hmm, no that indeed makes no sense... Looks like I cannot reuse that snippet

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
Allex-Nik: Yes, agree

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
Jolanrensen: Okay this is different indeed. It's a suitable column, it should return 0, definitely

Jolanrensen: I believe this is a bug. Something in the type calculation. Inferring types in `NumberInputHandler` defaults to `typeOf<Number>()`  if there are no values, which is not a primitive type, so it fails. Seems the best way to fix it is to be able to set a default :)

Jolanrensen: The same bug appears for `.std()` and `.mean()`, I'll fix it quickly here

Jolanrensen:
@Allex-Nik Very good review, thanks!

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
Jolanrensen: I'll clarify that the "input" we're talking about is the result of the expression. That way it should be clearer we're talking about that and not the column

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -126,18 +961,68 @@ public fun DataRow<*>.rowSumOf(type: KType, skipNaN: Boolean = skipNaNDefault):
 // endregion
 
 // region DataFrame
+
+/**
+ * Returns the sum of the values of each suitable column of this [DataFrame] separately.
+ *
+ * @include [SumDocs.AllSuitableColumnsSnippet]
```
Allex-Nik, line 973: Shall we add the note about ColumnGroups and flattening to this overload as well?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -274,6 +1339,19 @@ public fun <T, C : Number?> Grouped<T>.sum(
...
+ * ```kotlin
+ * // For each city, the sum of all values in the "age" and "weight" columns,
+ * // in a column called "total"
+ * df.groupBy { city }.sum("age", "weight", name = "total")
+ * ```
+ * @include [SumDocs.ColumnNamesParam]
+ * @include [SumDocs.ResultColumnNameParam]
+ * @include [SumDocs.SkipNanParam]
+ * @return A new [DataFrame] with the group keys and a single sum per group.
+ */
```
Allex-Nik, line 1359: Seems like this overload (maybe the whole ResultColumnNameParam?) also needs mentioning DuplicateColumnPathInsertException

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -361,6 +1554,36 @@ public fun <T, C : Number?> Pivot<T>.sum(
...
+ * ### Example
+ * ```kotlin
+ * // For each city, the sum of the weight-to-age ratios
+ * df.pivot { city }.sumOf { (weight ?: 0) / age }
+ * ```
+ *
+ * @include [SumDocs.SkipNanParam]
+ * @include [SumDocs.ExpressionParam]
+ * @return A single [DataRow] with, per [pivot] group, the sum of the expression's results.
+ */
```
Allex-Nik, line 1593: Shall we add "The result of the [expression\] is considered the 'input' of this operation." to this overload as well?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/CommonMinMaxDocs.kt
```diff
@@ -17,11 +15,13 @@ import org.jetbrains.kotlinx.dataframe.api.with
  *    Holds all KDoc-snippets that the `min` and `max` operations have in common.
  *    Both `MinDocs` and `MaxDocs` inherit from this interface, so the snippets can be
  *    included from either of them, like `{@include [MaxDocs.SkipNaNParam]}`.
```
Allex-Nik: SkipNanParam

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -33,38 +39,715 @@ import kotlin.reflect.KType
...
+    @ExcludeFromSources
+    typealias ZeroCellOnEmptySnippet = Nothing
+
+    /**
+     * {@comment Note about the behavior on empty input for the Pivot functions. KDoc-snippet.}
+     *
+     * Result cells for which there exists a group, but there is nothing to sum
+     * (for instance, because the group was empty or contained only `null` values)
+     * simply become `0` in the result type (defaulting to [Double] if there is no number type to speak of).
+     * For more information about the resulting types:
```
Allex-Nik, line 148: The new line disappeared

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
```diff
@@ -370,15 +1593,73 @@ public inline fun <T, reified R : Number?> Pivot<T>.sumOf(
...
+/**
+ * Aggregates this [PivotGroupBy] by computing the sum of the values of
+ * each suitable column separately, per group.
+ *
+ * Returns a [DataFrame] where each cell contains the sum of each suitable column
+ * of the group corresponding to that [pivot] key (column) and [groupBy] key (row).
+ *
+ * @include [SumDocs.AllSuitableColumnsSnippet]
+ * @include [SumDocs.SupportedTypesSnippet]
+ * @include [SumDocs.ZeroCellOnEmptyPivotSnippet]
```
Allex-Nik, line 1612: Should `ZeroCellOnEmptyPivotSnippet` be here? And in `PivotGroupBy.sumOf`?

Jolanrensen:
Regarding https://github.com/Kotlin/dataframe/pull/2032#discussion_r4070524117 and https://github.com/Kotlin/dataframe/pull/2032#discussion_r4047060686

I redid the fix properly:
- `AggregatorInputHandler` now requires a `valueTypeIfEmpty: KType`
- This is set to `Nothing` by default but can be overridden by each input handler and thus any aggregator.
- For `mean`, and `std`, this is set to `Double`, as `Double` empty input -> `Double.NaN`
- For `sum` this is set to the input type for `rowSumOf<T>()` or the column type if it's called on a single column. Else, I set it to `Double`. This makes sure the `0` value of the expected type will be used in case there are no values.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/sum.kt
Jolanrensen: I'll put it inside AllSuitableColumnsSnippet

Allex-Nik (approved):
Cool! Nice work :)

PR #2037, opened 2026-08-20
Fix JSON serialization for all-null value column

### LINKED ISSUE #2035: toJson() drops row objects when a real "value" column contains only nulls
### Description

`DataFrame.toJson()` incorrectly serializes a multi-column DataFrame as an array of `null` values when a column named `value` contains only nulls, even though other columns contain valid data.

### Minimal reproduction

```kotlin
import org.jetbrains.kotlinx.dataframe.api.dataFrameOf
import org.jetbrains.kotlinx.dataframe.io.toJson

fun main() {
    val df =
        dataFrameOf(
            "value" to listOf<Double?>(null, null),
            "name" to listOf("Alice", "Bob"),
        )

    println(df.toJson())
}
```

### Actual result

```json
[null,null]
```

### Expected result

```json
[
  {
    "value": null,
    "name": "Alice"
  },
  {
    "value": null,
    "name": "Bob"
  }
]
```

### Environment

- Kotlin: 2.2.21
- Kotlin DataFrame: 1.0.0-Beta5
- JDK: 21
- Gradle: 9.4.1
- OS: Windows 11

chimission:
Fix https://github.com/Kotlin/dataframe/issues/2035

### DIFF
--- a/dataframe-json/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
+++ b/dataframe-json/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
@@ -289,10 +289,14 @@ internal fun AnyFrame.extractValueColumn(): DataColumn<*>? {
             if (valueCol.kind() != ColumnKind.Value) {
                 null
             } else {
-                // Check that no other column contains data when this frame represents scalar values.
+                val hasNonNullValues = rows().any { valueCol[it] != null }
                 val isValidValueColumn = rows().all { row ->
-                    allColumns.all { col ->
-                        col.name == valueCol.name || col[row] == null
+                    if (valueCol[row] != null || !hasNonNullValues) {
+                        allColumns.all { col ->
+                            col.name == valueCol.name || col[row] == null
+                        }
+                    } else {
+                        true
                     }
                 }
                 if (isValidValueColumn) {
@@ -320,15 +324,11 @@ internal fun AnyFrame.extractArrayColumn(): DataColumn<*>? {
             if (arrayCol.kind() == ColumnKind.Group) {
                 null
             } else {
-                // check that value in this column is not null only when other values are null
+                val hasNonNullValues = rows().any { arrayCol[it] != null }
                 val isValidArrayColumn = rows().all { row ->
-                    if (arrayCol[row] != null) {
+                    if (arrayCol[row] != null || !hasNonNullValues) {
                         allColumns.all { col ->
-                            if (col.name != arrayCol.name) {
-                                col[row] == null
-                            } else {
-                                true
-                            }
+                            col.name == arrayCol.name || col[row] == null
                         }
                     } else {
                         true
--- a/dataframe-json/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
+++ b/dataframe-json/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
@@ -20,9 +20,10 @@ import org.jetbrains.kotlinx.dataframe.AnyFrame
 import org.jetbrains.kotlinx.dataframe.ColumnsContainer
 import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.api.FormattedFrame
+import org.jetbrains.kotlinx.dataframe.api.all
+import org.jetbrains.kotlinx.dataframe.api.allNulls
 import org.jetbrains.kotlinx.dataframe.api.indices
 import org.jetbrains.kotlinx.dataframe.api.isList
-import org.jetbrains.kotlinx.dataframe.api.rows
 import org.jetbrains.kotlinx.dataframe.api.schema
 import org.jetbrains.kotlinx.dataframe.api.take
 import org.jetbrains.kotlinx.dataframe.columns.ColumnGroup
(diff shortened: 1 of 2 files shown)


dataframe-json/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
```diff
@@ -286,22 +286,13 @@ internal fun AnyFrame.extractValueColumn(): DataColumn<*>? {
...
-                        allColumns.all { col ->
-                            if (col.name != valueCol.name) {
-                                col[row] == null
-                            } else {
-                                true
-                            }
-                        }
-                    } else {
-                        true
+                    allColumns.all { col ->
```
Jolanrensen: I believe the same logic exists in `extractArrayColumn()` maybe it should be fixed there as well?

chimission: Yes, I’ll fix that as well.

dataframe-json/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/JsonTests.kt
```diff
@@ -511,6 +513,149 @@ class JsonTests {
...
+        val scalarRows = DataFrame.readJsonStr(mixedJson)[1..3]
+
+        scalarRows.toJson() shouldBe "[1,2,3]"
+    }
+
+    @Test
+    fun `recognize generated value column in mixed json`() {
+        @Language("json")
+        val mixedJson = """[1,{"label":"record"},null]"""
+        val df = DataFrame.readJsonStr(mixedJson)
```
Jolanrensen, line 567: So... If we read DataFrame.readJsonStr here we get:
```
       label  value
     String?   Int?
        ----   ----
  0     null      1
  1   record   null
  2     null   null
```
and then `.toJson()`:
```json
[1,null,null]
```

Isn't that a bit odd? Where did `label` go?

This isn't something you changed, I understand :) But it's worth a new issue; we should look into this next

dataframe-json/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
```diff
@@ -286,19 +286,14 @@ internal fun AnyFrame.extractValueColumn(): DataColumn<*>? {
...
         ?.let { valueCol ->
-            // check that value in this column is not null only when other values are null
             if (valueCol.kind() != ColumnKind.Value) {
                 null
             } else {
-                // check that value in this column is not null only when other values are null
+                val hasNonNullValues = rows().any { valueCol[it] != null }
                 val isValidValueColumn = rows().all { row ->
-                    if (valueCol[row] != null) {
+                    if (valueCol[row] != null || !hasNonNullValues) {
```
Jolanrensen, line 294: Is it okay if I commit a tiny refactor after this commit? Your logic looks good :) But I'd like to make some changes that make the complex logic a bit easier to understand (like removing the double negative `!non-null`, use `takeIf { boolean }` instead of `let { if (boolean)... null }`, etc.)

dataframe-json/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
```diff
@@ -329,15 +324,11 @@ internal fun AnyFrame.extractArrayColumn(): DataColumn<*>? {
             if (arrayCol.kind() == ColumnKind.Group) {
                 null
             } else {
-                // check that value in this column is not null only when other values are null
+                val hasNonNullValues = rows().any { arrayCol[it] != null }
                 val isValidArrayColumn = rows().all { row ->
-                    if (arrayCol[row] != null) {
+                    if (arrayCol[row] != null || !hasNonNullValues) {
```
Jolanrensen, line 329: Actually, I believe this logic is incorrect in regard to our `readJson`:

```kt
DataFrame.readJsonStr("""[{"label":"record"},[123]]""")
```
becomes:
```
       label      array
     String?  List<Int>
        ----       ----
  0   record        [ ]
  1     null      [123]
```
not `null` but `emptyList`. Yes, this was a bug that was already there ;P

dataframe-json/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/JsonTests.kt
Jolanrensen: Created this https://github.com/Kotlin/dataframe/issues/2045

dataframe-json/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
Jolanrensen: Let's solve that as part of https://github.com/Kotlin/dataframe/issues/2048 :) you can leave it here as is

chimission: Yes. Since I wasn’t entirely sure whether the original behavior was intentional or a workaround for a particular edge case, I tried to preserve the intent of the existing code and limited the change to the clearly reproducible bug scenario.
It now seems that a more systematic fix is needed. In that case, I’ll follow [#2048](https://github.com/Kotlin/dataframe/issues/2048) for further updates.

dataframe-json/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/io/writeJson.kt
chimission: Absolutely, feel free to refactor it. I agree that simplifying the conditionals would make the logic easier to follow. Thanks for improving it! :)

Jolanrensen: I currently rewrote it to:
```kt
internal fun AnyFrame.extractValueColumn(): DataColumn<*>? {
    val allColumns = columns()

    return allColumns.filter { it.name.startsWith(VALUE_COLUMN_NAME) }
        .takeIf { isPossibleToFindUnnamedColumns }
        ?.maxByOrNull { it.name }
        ?.takeIf { it.kind() == ColumnKind.Value }
        ?.takeIf { valueCol ->
            val otherCols = allColumns - valueCol
            val isValidValueColumn =
                when {
                    // Consider an all-null column as 'value' column only if all other columns are also all-nulls.
                    valueCol.allNulls() ->
                        otherCols.all { it.allNulls() }

                    // Else, if the column does have values, then it's a 'value' column only if for each value,
                    // the corresponding cells in the other columns are all null.
                    else ->
                        this.all { row ->
                            valueCol[row] == null || otherCols.all { it[row] == null }
                        }
                }

            isValidValueColumn
        }
}
```
but I didn't commit it yet due to https://github.com/Kotlin/dataframe/issues/2035#issuecomment-5428362598

chimission: I agree with your point in [#2035 (comment)](https://github.com/Kotlin/dataframe/issues/2035#issuecomment-5428362598).

Jolanrensen:
@chimission have a look at my refactor/fixes :) do you think it solves your original problem? (besides issues #2048, #2046, and #2045 of course)

chimission:
@Jolanrensen 
Yes, it looks good to me. Thanks for the refactor and fixes! :)

PR #2042, opened 2026-08-24
Fixed dataframe schemas equality check + added tests and docs

### LINKED ISSUE #1717: Align `DataFrameSchemaImpl.equals` with `ColumnSchema.equals`
Recently i tried changing ColumnSchema.equals from
```
is Group -> schema == (otherType as Group).schema
is Frame -> schema == (otherType as Frame).schema
```
to 
```
is Group -> schema.compare((otherType as Group).schema).matches()
is Frame -> schema.compare((otherType as Frame).schema).matches()
```
Logically it should've caused this test to fail, but it didn't
`org.jetbrains.kotlinx.dataframe.codeGen.MatchSchemeTests#nested comparison with order`

Turns out `nested comparison with order` doesn't hit this code at all, `DataFrameSchemaImpl.equals` uses `ColumnSchema.compare`  instead
Task: carefully review usages of equals to not cause changes in places where it was expected to be "matches", and migrate DataFrameSchemaImpl.equals

===================================================================================

## Motivation

`DataFrameSchemaImpl.equals` currently does not consistently rely on `ColumnSchema.equals`. It first uses `DataFrameSchema.compare(...) == Matches`, then performs extra manual checks for nested `Group` and `Frame` schemas.

This makes equality semantics harder to reason about: changing `ColumnSchema.equals` may not affect `DataFrameSchemaImpl.equals`, and existing tests such as `MatchSchemeTests#nested comparison with order` can miss that.

For release, schema equality should have clear and tested semantics, because it affects schema comparison, code generation, generated accessors, and any logic that relies on exact schema identity.

## Scope

- Review usages of `DataFrameSchema.equals`, `ColumnSchema.equals`, and `compare(...).matches()`.
- Decide where exact equality is required and where structural “matches” comparison is intended.
- Update `DataFrameSchemaImpl.equals` to use `ColumnSchema.equals` consistently for exact equality.
- Keep `compare(...)` as the API for lenient/semantic schema matching.

## Definition of Done

- `DataFrameSchemaImpl.equals` no longer depends on `compare(...)` for exact equality.
- `ColumnSchema.equals` changes are reflected by `DataFrameSchemaImpl.equals`.
- Tests cover nested `Group` and `Frame` schemas, including column order.
- Tests distinguish exact equality from `compare(...).matches()`.
- No existing call sites accidentally lose intended lenient matching behavior.

AndreiKingsley:
Now `DataFrameSchemaImpl.equals` is checked using `ColumnSchema.equals`.

Fixes #1717

### DIFF
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MatchSchemeTests.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MatchSchemeTests.kt
@@ -127,7 +127,6 @@ class MatchSchemeTests {
 
     @Test
     fun `simple data schema comparison`() {
-
         val scheme1 = dataFrameOf(
             "a" to columnOf(1, 2, 3, null),
             "b" to columnOf(1.0, 2.0, 3.0, 4.0),
--- a/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MatchSchemeTests.kt
+++ b/core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MatchSchemeTests.kt
@@ -11,9 +11,10 @@ import org.jetbrains.kotlinx.dataframe.api.after
 import org.jetbrains.kotlinx.dataframe.api.cast
 import org.jetbrains.kotlinx.dataframe.api.columnOf
 import org.jetbrains.kotlinx.dataframe.api.dataFrameOf
-import org.jetbrains.kotlinx.dataframe.api.generateCode
+import org.jetbrains.kotlinx.dataframe.api.generateInterfaces
 import org.jetbrains.kotlinx.dataframe.api.move
 import org.jetbrains.kotlinx.dataframe.api.schema
+import org.jetbrains.kotlinx.dataframe.api.sumOf
 import org.jetbrains.kotlinx.dataframe.api.update
 import org.jetbrains.kotlinx.dataframe.api.with
 import org.jetbrains.kotlinx.dataframe.impl.codeGen.ReplCodeGenerator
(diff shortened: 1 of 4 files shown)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MatchSchemeTests.kt
```diff
@@ -247,4 +255,80 @@ class MatchSchemeTests {
         typed.schema().compare(movedInFrameAndGroup.schema()).matches() shouldBe true
         (typed.schema() == movedInFrameAndGroup.schema()) shouldBe false
     }
+
+    @Test
+    fun `column order inside a column group affects equality but not compare`() {
```
Jolanrensen: but not compare *matches. We might gain something like `CompareResult.Equals` in the future :)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/codeGen/MatchSchemeTests.kt
```diff
@@ -247,4 +255,80 @@ class MatchSchemeTests {
...
+
+    @Test
+    fun `equal schemas have equal hash codes`() {
+        val scheme = typed.schema()
+        val same = typed.schema()
+
+        (scheme == same) shouldBe true
+        scheme.hashCode() shouldBe same.hashCode()
+
+        val reordered = typed.move { pageInfo.resultsPerPage }.after { pageInfo.snippets }.schema()
```
koperagen, line 328: name of this test inspired me :) could you please add another case

val reordered = typed.move { pageInfo.resultsPerPage }.after { pageInfo.snippets }.schema()
val reordered1 = typed.move { pageInfo.resultsPerPage }.after { pageInfo.snippets }.schema()

check hashcode, schema equality..

koperagen (approved):

Jolanrensen (approved):
Looks way simpler and straight-to-the-point now :) And thanks for the KDocs! Really helpful :D

PR #2043, opened 2026-08-25
Fixed reading nullable nested (struct) columns from Parquet/Arrow

### LINKED ISSUE #2041: Bug in reading nullable nested columns from Parquet
## Setup

Start with a parquet file `test.parquet` with the following schema:
```
message root {
  required int64 timestamp (TIMESTAMP(NANOS,false));
  optional group record_channel_1 {
    required int32 x;
    required int32 y;
  }
  optional group record_channel_2 {
    required int32 x;
    required int32 y;
  }
}
```
and two rows like this, as printed when reading the file with pandas:
```
   timestamp  record_channel_1  record_channel_2
0 2000-01-01  {'x': 1, 'y': 2}              None
1 2000-01-02              None  {'x': 3, 'y': 4}
```

## Actual Behavior

When I read the file into a Kotlin DataFrame using `DataFrame.readParquet("test.parquet")` and format the DataFrame in a debugger, I get this:
```
            timestamp  record_channel_1  record_channel_2
        LocalDateTime    {x:Int, y:Int}    {x:Int, y:Int}
                 ----              ----              ----
  0  2000-01-01T00:00      { x:1, y:2 }      { x:3, y:4 }
  1  2000-01-02T00:00      { x:0, y:0 }      { x:3, y:4 }
```

This shows record values for two cells (record_channel_1 at 2000-01-02T00:00, and record_channel_2 at 2000-01-01T00:00) where I expect to see nulls, since the parquet file does not contain values for these cells. Accessing the DataFrame entries in other ways is consistent with this debugging output. The issue appears to be with the data in the DataFrame, and not merely a formatting bug.

## Expected Behavior

I expect something more like this, where the columns are nullable because the entire group for that column is optional, and the values are null where no value is recorded:
```
            timestamp  record_channel_1  record_channel_2
        LocalDateTime   {x:Int, y:Int}?   {x:Int, y:Int}?
                 ----              ----              ----
  0  2000-01-01T00:00      { x:1, y:2 }              null
  1  2000-01-02T00:00              null      { x:3, y:4 }
```

## Version information
DataFrame library version: 1.0.0-rc01
Kotlin Version: 2.4.10
DataFrame plugin: 2.4.10
JVM Toolchain: 21

## Test Data
Here's the parquet file I used, with the extension changed to allow me to upload it: [test.parquet](https://github.com/user-attachments/files/31393239/test.parquet.txt)

zaleslaw:
Fixes #2041 

# Fix reading nullable nested (struct) columns from Parquet/Arrow

## Problem it fixes 
When reading a Parquet or Arrow file, a nullable nested column (an "optional group",
e.g. `record { x, y }`) was read incorrectly. In rows where the group is absent, the
DataFrame showed fake data — `{x:0, y:0}`, or values leaked from another row — and the
nested columns were typed as non-null.

Cause: Arrow stores a separate "is-null" bit for the struct itself, independent of its
child arrays. The reader ignored this bit and read the child values directly. Under a
null row those child values are undefined, so we got garbage.

## What this PR does
- The Arrow reader now checks the struct's own null bit. When a group is absent in a
  row, all of its child cells become `null` and the child column types become nullable.
- It works at any depth: a null parent nulls every descendant field. A null struct
  *element inside a list* is handled the same way.
- Required groups (never null) are read exactly as before — no change.
- The bug lived in the shared Arrow conversion code, so the fix covers Parquet, Arrow
  IPC and Feather.

## How it works (short)
A `ColumnGroup` cannot be `null` per row, so the null is pushed down into the child
columns. A small recursive helper `injectNullsAt` sets the child cells to `null`:
- value column → the cell becomes `null` (type widened to nullable);
- nested group → recurse into its children;
- list-of-struct (frame) → the cell becomes an empty frame (a frame cannot hold `null`).
It dispatches on the column kind (public `ColumnGroup` / `FrameColumn` interfaces, no
impl details) and returns a column unchanged when there is nothing to null, so no data
is copied when there are no nulls.

## Files
- `arrowReadingImpl.kt` — the fix: consult the struct validity buffer, `injectNullsAt`
  helper, and the same handling for null struct elements in `readListVector`; removed a
  piece of dead code.
- `arrowReading.kt` — KDoc note on `readParquet` about the behavior.
- `docs/.../Parquet.md` — updated (the old "Struct not supported" note was stale).
- Tests: new `ArrowNullableStructTest.kt` (all the nullable-struct cases), a shared
  `arrowTestUtils.kt` helper, the committed fixture
  `nullable_nested_struct.parquet` + `TestFiles.md` describing it.
- No public API change (apiCheck passes).

## Tests
- In-memory Arrow IPC tests that put **non-zero hidden values** under a null parent, so
  a fix relying on zeros would be caught; one runs in `NullabilityOptions.Checking`.
- A null struct element inside a list.
- A "boss" case: a 3-level struct with value / list / list-of-struct / nested-group
  children and nulls at two levels at once.
- End-to-end Parquet: one test generates the data in code and one reads a committed
  fixture. The fixture is produced by our own code (Arrow `DatasetFileWriter`) and can
  be regenerated on any OS; DuckDB / the DataFrame writer can't produce this layout, so
  they are not used here.
- A test documenting the limitation below; plus Feather/IPC round-trip checks.
- Each test/fixture is documented with its pseudo-schema and a small data "head".

## Known limitations (not fixed here, on purpose)
- A `ColumnGroup` cannot be `null` per row, so an absent group and a present group with
  all-null children look the same (`{x:null, y:null}`). The exact type `{x:Int, y:Int}?`
  from the issue would need a larger change in core.
- Writing a struct-level null back to Arrow is still not supported (the writer marks
  every struct row as present).


dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
```diff
@@ -311,6 +309,50 @@ private fun List<Nothing?>.withTypeNullable(
...
+ *   [applyNullability] — the null is structurally mandated by the parent and must not throw under
+ *   [NullabilityOptions.Checking]).
+ *
+ * Dispatch is on `kind()` (defined on the base column type) and casts to the public [ColumnGroup] /
+ * [FrameColumn] interfaces the column actually implements, so it does not rely on any impl detail.
+ * When the [isNull] slice contains no `true`, the column is returned unchanged — nothing is reallocated
+ * (e.g. for a non-null struct element inside a list).
+ */
+private fun AnyBaseCol.injectNullsAt(isNull: BooleanArray): AnyBaseCol {
+    if (isNull.none { it }) return this
```
koperagen, line 330: Seems this check can be performed once outside the function. Now it'll do it for each child column

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
```diff
@@ -311,6 +309,50 @@ private fun List<Nothing?>.withTypeNullable(
...
+            DataColumn.createColumnGroup(
+                name = name(),
+                df = (this as ColumnGroup<*>).columns().map { it.injectNullsAt(isNull) }.toDataFrame(),
+            )
+
+        ColumnKind.Frame ->
+            DataColumn.createFrameColumn(
+                name = name(),
+                groups = (this as FrameColumn<*>).toList().mapIndexed { i, frame ->
+                    if (isNull[i]) DataFrame.empty() else frame
```
koperagen, line 342: Use DataFrame.empty(schema) to create type safe null object values

dataframe-arrow/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ArrowNullableStructTest.kt
```diff
@@ -0,0 +1,536 @@
...
+            list.startNewValue(1)
+            list.endValue(1, 2)
+            root.setRowCount(2)
+        }
+
+        val ordersColumn = DataFrame.readArrowIPC(bytes)["orders"].shouldBeInstanceOf<FrameColumn<*>>()
+        ordersColumn[0]["item"].values().toList() shouldBe listOf("A", "B")
+        ordersColumn[0]["qty"].values().toList() shouldBe listOf(1, 2)
+        // the null struct element must NOT leak its hidden values
+        ordersColumn[1]["item"].values().toList() shouldBe listOf(null, "C")
```
koperagen, line 392: might be better to assert values per row, it'll be more obvious that 
0 row is nullified

dataframe-arrow/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ArrowNullableStructTest.kt
```diff
@@ -0,0 +1,536 @@
...
+    @Test
+    fun `null struct element inside a list is nulled`() {
+        val element = Field(
+            "element",
+            FieldType.nullable(ArrowType.Struct()),
+            listOf(Field("item", FieldType.notNullable(ArrowType.Utf8()), null), Field("qty", intField, null)),
+        )
+        val orders = Field("orders", FieldType.nullable(ArrowType.List()), listOf(element))
+
+        val bytes = arrowBytes(orders) { root ->
```
koperagen, line 358: I suggest to create a separate variable and move initializing logic there
listWithNullableStructElement
Hard to scan verbose subjest under test

dataframe-arrow/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ArrowNullableStructTest.kt
```diff
@@ -0,0 +1,536 @@
...
+     * req        rec
+     * { p:7 }    { x:1, inner:{ a:100 } }
+     * { p:8 }    { x:null, inner:{ a:null } }   // rec is null (hidden x:999, a:777) -> all descendants null
+     * ```
+     */
+    @Test
+    fun `nullable nested struct nulls descendants and keeps required struct`() {
+        val required = Field("req", FieldType.notNullable(ArrowType.Struct()), listOf(Field("p", intField, null)))
+        val inner = Field("inner", FieldType.nullable(ArrowType.Struct()), listOf(Field("a", intField, null)))
+        val rec = Field("rec", FieldType.nullable(ArrowType.Struct()), listOf(Field("x", intField, null), inner))
```
koperagen, line 248: Maybe inline `inner` variable and format initialized on multiple lines
Structure should become more visible

dataframe-arrow/src/test/kotlin/org/jetbrains/kotlinx/dataframe/io/ArrowNullableStructTest.kt
```diff
@@ -0,0 +1,536 @@
...
+        assertNullableChannels(df)
+    }
+
+    /**
+     * A null parent nulls all descendant leaves (nested struct); a sibling required struct is untouched.
+     *
+     * Schema (as read):
+     * ```
+     * req: { p: Int }                       // required struct -> non-null ColumnGroup
+     * rec: { x: Int?, inner: { a: Int? } }  // nullable struct, one level deeper
```
koperagen, line 235: Frankly reading initializers of 3 variables below gives more clear picture than this comment. I believe it's misleading because it's not clear what exactly here is nullable struct? 
Code gives answer that it's `inner`. Comment doesn't

dataframe-arrow/src/main/kotlin/org/jetbrains/kotlinx/dataframe/io/arrowReadingImpl.kt
```diff
@@ -311,6 +309,50 @@ private fun List<Nothing?>.withTypeNullable(
...
+ * no per-row null mask ("a column group is never null, instead, make the columns inside nullable").
+ *
+ * - [ColumnKind.Group]: recurse so that a null parent nulls out **all** descendant leaves.
+ * - [ColumnKind.Frame] (a struct containing a list): a [FrameColumn] cannot hold `null`, so null rows
+ *   become an empty [DataFrame] — the nested list genuinely does not exist for that row.
+ * - [ColumnKind.Value]: null the cell and widen the type to nullable directly (do **not** re-run
+ *   [applyNullability] — the null is structurally mandated by the parent and must not throw under
+ *   [NullabilityOptions.Checking]).
+ *
+ * Dispatch is on `kind()` (defined on the base column type) and casts to the public [ColumnGroup] /
```
koperagen, line 324: Personal opinion. Maybe this ship has sailed already, but i still would like to see comment express things that code cannot: intentions, context. 
We must guard value of comments. If it's there, it must mean it tells you something important and you should likely read it first. 
If we still expect to read code, that is. I'd prefer to not read comments that repeat code, but do it worse (in plain English instead of code language).

zaleslaw: Fair point. I’ll remove the implementation-level walkthrough that merely mirrors the when branches.
I’d keep the Arrow-specific rationale and the ColumnGroup/FrameColumn constraints though: they explain why this transformation is necessary, rather than how the code executes it.

PR #2047, opened 2026-08-25
Documented `transpose` and `transposeTo` functions in `DataRow` API

### LINKED ISSUE #1989: Add KDocs for `transpose` / `transposeTo`
# Add KDocs for `transpose` / `transposeTo`

The file `core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/transpose.kt` has two public, undocumented functions. Please add concise KDocs for both.

### Scope

- `DataRow<*>.transpose()` — turns a single row into a two-column `DataFrame<NameValuePair<*>>`, with a `name` column (the original column names) and a `value` column (the row's values).
- `DataRow<*>.transposeTo<T>()` — the same, but converts every value to type `T`, producing a `DataFrame<NameValuePair<T>>`.

### KDocs should explain

- What is produced: one output row per column of the source row, with `name` / `value` columns.
- The difference between the two: `transpose()` keeps values as-is (`NameValuePair<*>`); `transposeTo<T>()` converts each value to `T`.
- That the resulting schema is `NameValuePair`.

### Acceptance criteria

- Both `transpose()` and `transposeTo()` have concise KDocs.
- The KDocs make the as-is vs converted-to-`T` distinction clear.

zaleslaw:
Fixes #1989

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/transpose.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/transpose.kt
@@ -4,6 +4,8 @@ import org.jetbrains.kotlinx.dataframe.AnyRow
 import org.jetbrains.kotlinx.dataframe.DataColumn
 import org.jetbrains.kotlinx.dataframe.DataFrame
 import org.jetbrains.kotlinx.dataframe.DataRow
+import org.jetbrains.kotlinx.dataframe.documentation.DocumentationUrls
+import org.jetbrains.kotlinx.dataframe.exceptions.TypeConverterNotFoundException
 import org.jetbrains.kotlinx.dataframe.impl.api.convertTo
 import org.jetbrains.kotlinx.dataframe.impl.columnName
 import org.jetbrains.kotlinx.dataframe.impl.owner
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/transpose.kt
```diff
@@ -13,14 +13,59 @@ import kotlin.reflect.typeOf
...
+ *
+ * ### Example
+ * ```kotlin
+ * // Convert the first row into a two-column "name" / "value" DataFrame,
+ * // converting all values to String
+ * df.first().transposeTo<String>()
+ * ```
+ *
+ * @param T the type every value is converted to.
+ * @return A [DataFrame] of [NameValuePair] with `name` and `value` columns,
```
koperagen, line 60: throws TypeConverterNotFoundException

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/transpose.kt
```diff
@@ -13,14 +13,59 @@ import kotlin.reflect.typeOf
...
+ */
 public fun DataRow<*>.transpose(): DataFrame<NameValuePair<*>> {
     val valueColumn = DataColumn.createByInference(NameValuePair<*>::value.columnName, values)
     val nameColumn = owner.columnNames().toValueColumn(NameValuePair<*>::name.name)
     return dataFrameOf(nameColumn, valueColumn).cast()
 }
 
+/**
+ * Converts this [DataRow] into a [DataFrame] with one row per column of the original row,
+ * converting every value to the given type [T].
```
koperagen, line 43: Might be worth to add
"if it is not [T] already"
or something along the line. 

Btw, later description seems to convey key idea better. Maybe merge it here?
* This is the same as [transpose], but instead of keeping the values as-is,
 * each value is converted to [T].

AndreiKingsley (approved):

AndreiKingsley (commented):
Would be nice to add links to DataRow API docs on website.

PR #2050, opened 2026-08-26
Add KDocs for public `reverse` APIs

### LINKED ISSUE #1980: Add KDocs for public `reverse` APIs
Add concise KDocs for public `reverse` APIs in:

`core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/reverse.kt`

Current public overloads are undocumented:

- `DataFrame<T>.reverse()`
- `DataColumn<T>.reverse()`
- `ColumnGroup<T>.reverse()`
- `FrameColumn<T>.reverse()`
- `ValueColumn<T>.reverse()`

The KDocs should briefly explain that `reverse` returns the same dataframe/column with row or value order reversed.

Reference docs page:

https://kotlin.github.io/dataframe/reverse.html

Existing tests are present in:

`core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/reverse.kt`

They cover:

- dataframe row order reversal
- value column reversal
- column group reversal

Acceptance criteria:

- Public `reverse` overloads have concise KDocs.
- KDocs are consistent with the existing docs page.
- Existing tests continue to pass.

zaleslaw:
# Add KDocs for public `reverse` APIs

Closes #1980. All five public `reverse` overloads now have KDocs. No behavior changes.

## KDoc

- Added a shared KoDEx template `CommonReverseDocs` in `api/reverse.kt`. It holds the first
  line, the website link, `@param [T]` and `@return`. Each overload only sets its own parts
  (`RECEIVER`, `UNIT`, `DETAILS`, `SEE_ALSO`, `TYPE_PARAM`). Same pattern as `CommonTakeAndDropDocs`.
- Added `DocumentationUrls.Reverse` for the `reverse.html` link.
- The text describes the contract only — what you get back, not how it is done.

What each overload now states:

| Overload | Documented contract |
|---|---|
| `DataFrame` | Only row order changes. Columns, types and schema stay the same. |
| `DataColumn` | Name, type and kind stay the same. The result is typed as `DataColumn`, so for a column group you need `asColumnGroup()`. |
| `ColumnGroup` | Name and nested structure stay. Values in all nested columns stay aligned row-wise. |
| `FrameColumn` | Only the order of the dataframes changes. Rows inside each dataframe keep their order. |
| `ValueColumn` | Name and type stay the same. |

## Tests

3 tests -> 10. Every statement in the KDocs now has a test. The three old tests are unchanged.

**The old tests did not call the overloads one may expect.** `columnOf(a, b)` returns
`DataColumn<DataRow<*>>`, not `ColumnGroup`. `columnOf(1, 2, 3)` returns `DataColumn<Int>`, not
`ValueColumn`. So `ColumnGroup.reverse()` and `ValueColumn.reverse()` were never called by any
test. The new tests build these columns explicitly with `DataColumn.createColumnGroup()` and
`DataColumn.createValueColumn()`.

New tests:

- `dataframe keeps schema and size` — schema and row count.
- `column keeps name and type`.
- `column overload keeps the column kind` — a group stays `Group`, a frame stays `Frame`, a value
  stays `Value`. The `DataColumn<*>` types in this test are required: without them the calls go to
  the other overloads and the test would check something else.
- `column group type can be restored after the column overload` — `asColumnGroup()` round-trip.
  `ColumnGroup` is not a subtype of `DataColumn` (it extends `BaseColumn` and `DataFrame`), so this
  cast is the only way to get the group type back.
- `column group reverses rows as a whole` — real `ColumnGroup` receiver. It checks the name, the
  schema and the row pairs. It compares against explicit expected pairs and not against
  `DataFrame.reverse()`: comparing two implementations can pass while both are wrong.
- `frameColumn` — the frames swap places, rows inside them keep their order.
- `valueColumn` — the `ValueColumn<Int>` type of the result is the assertion here. The compiler
  checks it.

Note on assertions: column equality (`impl/columns/Utils.kt`) compares name, type and values. One
`shouldBe` on a whole column therefore already covers name and type. Separate `name()` / `type()`
asserts are used only where there is no whole-column comparison.

## KDoc guidelines

Two notes added to `KDOC_GUIDELINES.md`, both found while writing the template:

- A multi-line `@set` value keeps the indent of its continuation lines. An indent of 4+ spaces
  becomes a code block in Markdown, so the paragraph renders as monospace.
- `@param` and `@return` must stay at the end of a KDoc. Text cannot be added after `@include`, so
  all parts that differ between overloads must be passed in as `@set` arguments.

### DIFF
--- a/KDOC_GUIDELINES.md
+++ b/KDOC_GUIDELINES.md
@@ -574,11 +574,6 @@ When using `@set` and `@get` / `$`, it's a good practice to use a reference as t
 This makes the KDoc more refactor-safe, and it makes it easier to understand which arguments
 need to be provided for a certain template.
 
-A `@set` value may span several lines, but the indentation of the continuation lines is kept
-verbatim in the processed KDoc. Keep continuation lines at the same indentation as the first line:
-a hanging indent of four or more spaces is read by Markdown as a code block, so the whole
-paragraph would silently render as monospace.
-
 A good example of this concept can be found in the
 [`AllColumnsSelectionDsl.CommonAllSubsetDocs` documentation interface](./core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/all.kt).
 This interface provides a template for all overloads of `allBefore`,
--- a/KDOC_GUIDELINES.md
+++ b/KDOC_GUIDELINES.md
@@ -498,6 +498,15 @@ private typealias Common~OperationName~Docs = Nothing
 public fun <T, C> DataFrame<T>.operation(columns: ColumnsSelector<T, C>)
 ```
 
+Keep in mind that `@param` and `@return` have to stay at the end of a KDoc.
+So when a template already ends with them, you cannot `@include` it and then append
+overload-specific text after the include — that text would land inside the template's `@return`.
+Everything that differs between the overloads has to be passed *into* the template as a `@set`
+argument instead. That is why a template shared by several sibling overloads usually has one
+argument per varying part (receiver type, "See also" section, type parameter description, and so on),
+like
+[`CommonTakeAndDropDocs`](./core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/documentation/CommonTakeAndDropDocs.kt).
+
 ### Grammar
 
 DSL Grammar (helpers usually are called simply `Grammar` and place inside the
@@ -564,6 +573,11 @@ internal interface CommonDoc {
 When using `@set` and `@get` / `$`, it's a good practice to use a reference as the key name.
 This makes the KDoc more refactor-safe, and it makes it easier to understand which arguments
 need to be provided for a certain template.
+
+A `@set` value may span several lines, but the indentation of the continuation lines is kept
+verbatim in the processed KDoc. Keep continuation lines at the same indentation as the first line:
+a hanging indent of four or more spaces is read by Markdown as a code block, so the whole
+paragraph would silently render as monospace.
 
 A good example of this concept can be found in the
 [`AllColumnsSelectionDsl.CommonAllSubsetDocs` documentation interface](./core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/all.kt).
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/reverse.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/reverse.kt
@@ -48,13 +48,13 @@ internal interface CommonReverseDocs {
 
 /**
  * @include [CommonReverseDocs]
- * {@set [CommonReverseDocs.RECEIVER] [DataFrame]}
- * {@set [CommonReverseDocs.UNIT] row}
- * {@set [CommonReverseDocs.DETAILS] Only the order of the rows changes:
- * the columns, their names and types, and thus the schema of the dataframe stay the same.}
- * {@set [CommonReverseDocs.SEE_ALSO] See also [shuffle][DataFrame.shuffle], which reorders rows randomly,
- * and [sortBy][DataFrame.sortBy], which orders rows by the values of the selected columns.}
- * {@set [CommonReverseDocs.TYPE_PARAM] The schema marker type of this [DataFrame].}
+ * @set [CommonReverseDocs.RECEIVER] [DataFrame]
+ * @set [CommonReverseDocs.UNIT] row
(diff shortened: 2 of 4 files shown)


KDOC_GUIDELINES.md
```diff
@@ -565,6 +574,11 @@ When using `@set` and `@get` / `$`, it's a good practice to use a reference as t
 This makes the KDoc more refactor-safe, and it makes it easier to understand which arguments
 need to be provided for a certain template.
 
+A `@set` value may span several lines, but the indentation of the continuation lines is kept
```
Jolanrensen, line 577: So, like:

```kt
/**
 *  {@get A}
 * {@set A   hello}
 */
```
? 
I don't think this can happen, because the argument in `@set` is trimmed

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/reverse.kt
```diff
@@ -3,17 +3,115 @@ package org.jetbrains.kotlinx.dataframe.api
...
+import org.jetbrains.kotlinx.dataframe.documentation.ExcludeFromSources
 import org.jetbrains.kotlinx.dataframe.impl.columns.asValueColumn
 import org.jetbrains.kotlinx.dataframe.indices
 
+/**
+ * {@comment Shared KDoc template for all `reverse` overloads. KDoc-snippet.}
+ * Returns a new {@get [RECEIVER]} with the same {@get [UNIT]}s in reversed order,
+ * so the last {@get [UNIT]} becomes the first one.
+ *
+ * {@get [DETAILS]}
```
Jolanrensen, line 19: When using a tag on a newline, no `{}` are needed

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/reverse.kt
```diff
@@ -3,17 +3,115 @@ package org.jetbrains.kotlinx.dataframe.api
...
+
+    // Description of the [T] type parameter
+    typealias TYPE_PARAM = Nothing
+}
+
+// region DataFrame
+
+/**
+ * @include [CommonReverseDocs]
+ * {@set [CommonReverseDocs.RECEIVER] [DataFrame]}
```
Jolanrensen, line 51: nowhere in this kdoc are `{}` necessary

zaleslaw: Will it work? Was it in the initial KDOC or KODEX guide?

KDOC_GUIDELINES.md
zaleslaw: Ah, okey, agent faced several time with the problem and decided to include it, probably need to double-check

Jolanrensen: can you give an example where this holds?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/reverse.kt
Jolanrensen: Yes, it works; block tags end when another block tag is encountered. I'm not sure how explicitly it's described, but it's how all other block tags work :)

Jolanrensen: Brackets are only needed when you want to put a tag inline

Jolanrensen: Normally you can see where a tag starts and ends using the intellij plugin of KoDEx, but it seems that has broken in the latest updates...

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/reverse.kt
```diff
@@ -3,17 +3,115 @@ package org.jetbrains.kotlinx.dataframe.api
...
 
+// endregion
+
+// region DataColumn
+
+/**
+ * @include [CommonReverseDocs]
+ * {@set [CommonReverseDocs.RECEIVER] [DataColumn]}
+ * {@set [CommonReverseDocs.UNIT] value}
+ * {@set [CommonReverseDocs.DETAILS] The column keeps its name, type and [kind][ColumnKind]:
```
Jolanrensen, line 69: oxford comma missing

Jolanrensen:
The produced KDocs look good and useful :)

Issue #1995, opened 2026-07-24
Add KDocs for `xs` (cross-section) APIs


zaleslaw:
# Add KDocs for `xs` APIs

The file `core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/xs.kt` has **no KDocs at all**. Please add concise KDocs for all four functions.

### Scope

- `DataFrame<T>.xs(vararg keyValues: Any?)` — default-key-columns overload.
- `DataFrame<T>.xs(vararg keyValues: C, keyColumns: ColumnsSelector<T, C>)` — explicit key-columns overload.
- `GroupBy<T, G>.xs(vararg keyValues: Any?)` — default-key-columns overload.
- `GroupBy<T, G>.xs(vararg keyValues: C, keyColumns: ColumnsSelector<T, C>)` — explicit key-columns overload.

### KDocs should explain

- What `xs` does (cross-section): returns the subset containing only the rows whose key columns match the given `keyValues`, in order — like fixing the leading key columns to specific values. For `DataFrame` it returns a filtered `DataFrame<T>`; for `GroupBy` it filters within the grouped structure and returns a `GroupBy<T, G>`.
- The `keyValues` → `keyColumns` positional matching: the i-th `keyValues` is matched against the i-th selected key column.
- The default key-column selection (the overloads without `keyColumns`): the first `keyValues.size` non-column-group columns at any depth (`colsAtAnyDepth().filter { !it.isColumnGroup() }.take(keyValues.size)`).
- That the matched key columns are typically dropped from the result (cross-section semantics) — verify against `xsImpl` and state the actual behavior.

### Test coverage assessment

- `DataFrame.xs` has coverage (`DataFrameTests.xs`, plus usage in `DataFrameTreeTests` and `samples/api/Access.kt`).
- `GroupBy.xs` has no obvious dedicated test — optional follow-up to add one.

### Website coverage

- Documented at `xs.html`; keep KDocs consistent with it.

### Acceptance criteria

- All four `xs` overloads have concise KDocs.
- KDocs explain the cross-section semantics, the positional `keyValues`↔`keyColumns` matching, and the default key-column selection.
- KDocs distinguish the `DataFrame` result from the `GroupBy` result.
- Docs are consistent with the site page `xs.html`.

zaleslaw:
Closed in #2052

PR #2052, opened 2026-08-27
Add KDocs for the four `xs` overloads (#1995)

zaleslaw:
**Add KDocs for the four `xs` overloads (#1995)**

`xs` returns a cross-section of a `DataFrame` or a `GroupBy`. 

All four overloads had no KDoc, and the operation is easy to get wrong: besides filtering rows it also
**removes** the key columns, and when you pass no selector it picks the key columns
for you — the first `n` columns, looking inside column groups. `GroupBy.xs` works on
two sides at once, the keys and the groups. The new docs are built around these
three points.

## What changed

| File | Change |
| :--- | :----- |
| `core/…/api/xs.kt` | KDoc for the four overloads. Shared text sits in one KoDEx topic and three snippets, so nothing is copy-pasted per overload. |
| `core/…/documentation/DocumentationUrls.kt` | New `Xs` entry with the link to `xs.html`. |
| `core/…/test/…/api/xs.kt` | New `XsTests`, 14 tests. |
| `KDOC_GUIDELINES.md` | New "Data tables in examples" subsection; note that parameter references must not be used inside KDoc snippets. |
| `KODEX_KDOC_PREPROCESSING.md` | Clearer rule on when a tag needs `{}`. |

## Examples

Every overload shows the input as a table, then the call and its result — the layout
`concatWithKeys` already uses. The data and the two `DataFrame` calls are taken from
the `xs` page on the documentation website, so the KDoc, the website and the tests
show one example instead of three different ones. The website data has a column group,
which makes the "a group is not counted, its columns are" rule visible instead of only
stated.

## Tests

Before this PR `xs` had `DataFrameTests.xs`, three scattered cases in
`DataFrameTests`/`DataFrameTreeTests`, and one `groupBy xs` test that only printed its
result. Return types, pairing order and row order were not checked at all, and
`GroupBy.xs` had no real assertions.

| Test | What it pins down |
| :--- | :---------------- |
| `default key columns are the first columns and are removed` | Return type `DataFrame<T>`, removal of the key column, original row order |
| `key values are paired with key columns by position` | The i-th value goes to the i-th column; the same values swapped match nothing |
| `a column group is not counted as a key column but its columns are` | Counting rule for nested columns |
| `explicit key columns are paired in selection order` | Selection order wins over the column order in the frame |
| `number of key values must be equal to the number of key columns` | `IllegalArgumentException`, for three overloads |
| `groupBy xs filters keys and groups and removes the key columns from both` | Return type `GroupBy<T, G>`; the key column is gone from the keys and from the groups |
| `groupBy xs default key columns are the first key columns` | Default selection starts at the first grouping key |
| `groupBy xs explicit key columns are paired in selection order` | Selection order for `GroupBy` |
| `groupBy xs key column present only in the keys` | A column missing in the groups is ignored there |
| `groupBy xs key column present only in the groups` | A column that is not a grouping key filters the groups only |
| `KDoc example - …` (4 tests) | Every result table in the KDoc is an expected value here, on the website data, so the tables cannot drift from the code |

Return types are asserted with explicit type annotations, so the compiler is the
assertion. Expected values are written out, not computed by a second implementation of
the same logic.


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/xs.kt
```diff
@@ -4,17 +4,149 @@ import org.jetbrains.kotlinx.dataframe.ColumnsSelector
...
 import org.jetbrains.kotlinx.dataframe.annotations.Interpretable
 import org.jetbrains.kotlinx.dataframe.annotations.Refine
+import org.jetbrains.kotlinx.dataframe.columns.ColumnGroup
+import org.jetbrains.kotlinx.dataframe.documentation.DocumentationUrls
+import org.jetbrains.kotlinx.dataframe.documentation.ExcludeFromSources
+import org.jetbrains.kotlinx.dataframe.documentation.SelectingColumns
 import org.jetbrains.kotlinx.dataframe.impl.api.xsImpl
 
+/**
+ * ## The Xs Operation
```
AndreiKingsley, line 14: Snippets for operation KDocs shouldn't contain headers!

AndreiKingsley: @zaleslaw what did you provide as a reference?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/xs.kt
```diff
@@ -4,17 +4,149 @@ import org.jetbrains.kotlinx.dataframe.ColumnsSelector
...
+     * where `n` is the number of given key values.
+     * Columns are counted from left to right;
+     * a [column group][ColumnGroup] is not counted itself, but the columns inside it are.
+     * If there are fewer columns than key values, an [IllegalArgumentException] is thrown.
+     */
+    @ExcludeFromSources
+    typealias DefaultKeyColumnsSnippet = Nothing
+
+    /**
+     * The key columns are the columns returned by the `keyColumns` selector,
```
AndreiKingsley, line 41: `keyColumns` can be a reference

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/xs.kt
```diff
@@ -4,17 +4,149 @@ import org.jetbrains.kotlinx.dataframe.ColumnsSelector
...
+    @ExcludeFromSources
+    typealias SOURCE = Nothing
+}
+
 // region DataFrame
 
+/**
+ * @include [XsDocs]
+ *
+ * See also [filter][DataFrame.filter], which selects rows by a condition and leaves all columns in place,
```
AndreiKingsley, line 79: Not a big deal, but why see also section is here, not in a general snippet?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/xs.kt
```diff
@@ -4,17 +4,149 @@ import org.jetbrains.kotlinx.dataframe.ColumnsSelector
...
+     * The key columns are the columns returned by the `keyColumns` selector,
+     * in the order the selector returns them — not in the order they appear in the data.
+     * The selector must return exactly one column per key value;
+     * otherwise, an [IllegalArgumentException] is thrown.
+     */
+    @ExcludeFromSources
+    typealias SelectedKeyColumnsSnippet = Nothing
+
+    /**
+     * The examples below use this [DataFrame], the same data as the `xs` page on the
```
AndreiKingsley, line 50: link to the website then?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/xs.kt
```diff
@@ -4,17 +4,149 @@ import org.jetbrains.kotlinx.dataframe.ColumnsSelector
...
+ * Key values and key columns are paired by position:
+ * the first key value is compared with the first key column,
+ * the second key value with the second key column, and so on.
+ * A row is kept when every pair is equal.
+ *
+ * The rows that are kept stay in their original order.
+ *
+ * For more information: {@include [DocumentationUrls.Xs]}
+ */
+internal interface XsDocs {
```
Jolanrensen, line 28: This is only included in other places, never referred to? Then it can have `@ExcludeFromSources` and no title

KDOC_GUIDELINES.md
```diff
@@ -111,12 +112,18 @@ has a KDoc which describes column path creation behavior.
 The whole file is excluded from sources,
 but the KDoc is included in other KDocs.
 
+Do not write parameter references like `[columns]` inside a KDoc-snippet.
```
AndreiKingsley, line 115: Not correct. We do write them, but with escaping: `[columns\]`.
This way it is not resolved into `…dataframe.columns`, but is resolved correctly after KDoc completely processed. 
So you can add this note but, please completely change a message. And then you can apply it to my comment on`XsDocs.SelectedKeyColumnsSnippet`.
@Jolanrensen didn't I lie about anything?

KDOC_GUIDELINES.md
```diff
@@ -420,6 +427,60 @@ Start the section with
 ### Examples
 ```
 
+#### Data tables in examples
```
AndreiKingsley, line 430: It's a great idea, but I need to think about standardisation to find an approach that works well for all possible operations.

AndreiKingsley: @Jolanrensen What are your thoughts on this?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/xs.kt
```diff
@@ -4,17 +4,149 @@ import org.jetbrains.kotlinx.dataframe.ColumnsSelector
...
+
+    /**
+     * The key columns are the first `n` columns of {@get [SOURCE]},
+     * where `n` is the number of given key values.
+     * Columns are counted from left to right;
+     * a [column group][ColumnGroup] is not counted itself, but the columns inside it are.
+     * If there are fewer columns than key values, an [IllegalArgumentException] is thrown.
+     */
+    @ExcludeFromSources
+    typealias DefaultKeyColumnsSnippet = Nothing
```
Jolanrensen, line 38: If `SOURCE` is an argument for this snippet, this snippet better be an `interface`, so the argument can be `DefaultKeyColumnsSnippet.SOURCE`

KDOC_GUIDELINES.md
```diff
@@ -111,6 +112,12 @@ has a KDoc which describes column path creation behavior.
 The whole file is excluded from sources,
 but the KDoc is included in other KDocs.
 
+Do not write parameter references like `[columns]` inside a KDoc-snippet.
+A snippet is not attached to a function, so the reference doesn't point to a parameter;
+when the snippet is included, KoDEx rewrites it to whatever it does resolve to
+(`[columns]` in `Remove`, for instance, becomes a link to the `…dataframe.columns` *package*).
+Write the parameter name in a code span there, and keep `[]` references for the function's own KDoc.
```
Jolanrensen, line 119: No, write the parameter like `[columns\]`, making KoDEx leave the ref in place as is. The escape character will be removed

KDOC_GUIDELINES.md
```diff
@@ -420,6 +427,60 @@ Start the section with
 ### Examples
 ```
 
+#### Data tables in examples
+
+When an operation changes the shape of the data — rows dropped, columns removed, columns regrouped —
+a call with a comment does not show the reader what actually happened. Show the data before and after
+as Markdown tables. The reference layout is
```
Jolanrensen, line 434: this doesn't always make sense in KDocs. It's big and tables don't render borders. Probably best to send users to the website in those cases... But if it's unavoidable, or completely unclear without table, we might add it.

AndreiKingsley: I agree. Such tables are worth adding for uncommon operations, such as 'xs', or 'pivot', for example.

But yes, KDocs' support for tables is poor :(.

KODEX_KDOC_PREPROCESSING.md
```diff
@@ -100,9 +100,20 @@ KoDEx uses special notations in KDocs to indicate that a certain (tag) processor
 in that place.
 These notations follow the Javadoc/KDoc `@tag content`/`{@tag content}` tag conventions.
 
-Tags without `{}` are allowed, but only at the beginning of a line, like you're used to with
-`@param`, `@return`, `@throws`, etc. If you want to use them in the middle of a line, or inside ` ``` ` blocks,
-you should use `{}`.
+When a tag starts a line, write it **without** `{}`, the same way you write `@param`, `@return`, `@throws`.
+That is the normal form; `{@tag ...}` alone on a line does exactly the same thing and only adds noise.
+Use `{}` only when the tag sits in the middle of a line, or inside a ` ``` ` block.
```
Jolanrensen, line 105: One case where it may be unavoidable is `@set [A] value`, especially if you have more text below it. In such cases, move `@set` down among other tags or do add `{}` brackets to properly scope the tag :)

KODEX_KDOC_PREPROCESSING.md
```diff
@@ -100,9 +100,20 @@ KoDEx uses special notations in KDocs to indicate that a certain (tag) processor
...
+
+A block tag runs until the next tag, so whatever you write after it — on the same line or on the lines below —
+stays in place, right after the content the tag produced. So a following paragraph is **not** a reason
+to reach for `{}`:
```
Jolanrensen, line 109: This is a bit conflicting. You say it runs to the next tag so you don't need brackets when a new paragraph starts... a paragraph is not a tag? it continues alright.

This whole doc is part of the `@incude` tag:
```
@include [CommonDocs]
This sentence stays right after the included content.
```
However, `@include` is clever enough to only replace the `@include [CommonDocs]` part with the kdocs of `CommonDocs` and it leaves the rest in place. Other tags like `@get`/`$`, `@sample`, `@sampleNoComments`, `@includeFile`, `@exportAsHtmlStart` etc. do the same. They are fed all content until the next block tag starts but they only take the input they need and leave the rest in place. This cannot be said for `@comment` and `@set KEY`... Those consider everything after it as part of their input (until the next block comment starts, of course).

You can use the IntelliJ KoDEx plugin to see how far the background changes color to see what part of the doc will be affected by each tag

AndreiKingsley (changes requested):

KDOC_GUIDELINES.md
Jolanrensen: haha you didn't lie ;P you're entirely correct

KDOC_GUIDELINES.md
Jolanrensen: - Might be worth it to move such full examples to a separate KDoc element, so people can see read it by navigating to it
- We might not need standardization if we only use these tables sparingly

Jolanrensen:
The contents, grammar and explanations are great!

KDOC_GUIDELINES.md
zaleslaw: thanks

KDOC_GUIDELINES.md
zaleslaw: True, want to keep to proper generation for rare cases, which has unclear semantic/name

KDOC_GUIDELINES.md
zaleslaw: thanks

KDOC_GUIDELINES.md
zaleslaw: I researched some context for KDocs, this is a summary

### Side note: what happens to KDoc-helpers once we publish Dokka

Not blocking this PR — but since this PR adds a KDoc-topic (`XsDocs`) that the overloads link to
rather than include, I checked what Dokka currently does with it. Two findings, both reproducible
with a single command:
 :core:dokkaGenerateHtml
 
 **1. `internal` KDoc-helpers do not leak into Dokka.** There is no page for `XsDocs`, `MinDocs` or
any other doc-helper in the generated HTML. That is the Dokka default: `documentedVisibilities` is
`VisibilityModifier.Public`, and our guidelines already require every KDoc-helper to be `internal`
or `private`. So the "topic + link" pattern is safe from that angle. The only visible effect is
that a link to a topic is emitted as an unresolved link
(`<span data-unresolved-link="…/XsDocs…">XsDocs</span>`), i.e. plain text with no navigation —
worth knowing when deciding between `@include` and a link for content readers should actually see.

If we ever want to make that explicit rather than rely on the default, it is one line:

```kotlin
dokka {
    dokkaSourceSets.configureEach {
        documentedVisibilities(VisibilityModifier.Public)
    }
}
```

2. Dokka currently generates from the un-preprocessed sources. dokkaGenerateHtml reads src/main/kotlin, not generated-sources, so all KoDEx notation ends up verbatim in the HTML: 20 literal {@include …} occurrences in xs.html, 69 in min.html. 

This is repository-wide, not specific to xs. Before Dokka can be published, it needs to be pointed at the KoDEx output the same way changeJarTask already does for sources.jar. 

After that, the snippets disappear from the input entirely (they are stripped by @ExcludeFromSources) and only the internal topics remain — at which point finding 1 is what applies.

Docs: Dokka Gradle configuration options (documentedVisibilities, default VisibilityModifier.Public; perPackageOption { suppress }), Migrate to Dokka Gradle plugin v2 (DGP v2 is the default since 2.1.0; Visibility.PUBLIC → VisibilityModifier.Public).

[Dokka Gradle configuration options](https://kotlinlang.org/docs/dokka-gradle-configuration-options.html) · [Migrate to Dokka Gradle plugin v2](https://kotlinlang.org/docs/dokka-migration.html) · [Dokka Gradle](https://kotlinlang.org/docs/dokka-gradle.html)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/xs.kt
zaleslaw: min.kt + CommonMinMaxDocs will try to follow it

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/xs.kt
zaleslaw: Switched to the min.kt + CommonMinMaxDocs pattern — that was the reference I should have followed. XsDocs is now a KDoc-topic: the overloads link to it with {@include [XsDocsLink]} instead of including it, so no header lands inside an overload's KDoc. It keeps the title and stays in sources, because a topic has to be linkable — same as MinDocs.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/xs.kt
zaleslaw: the example data links to the xs page via DocumentationUrls.Url

KDOC_GUIDELINES.md
zaleslaw: **Softened to opt-in:** by default send the reader to the website; add tables only for uncommon operations whose result a sentence does not convey (xs, pivot), and only when the task asks for them. No standardization implied. You could drop the section later if you disagree totally, I am fine.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/xs.kt
```diff
@@ -24,13 +156,106 @@ public fun <T, C> DataFrame<T>.xs(vararg keyValues: C, keyColumns: ColumnsSelect
...
+ * Both parts of a [GroupBy] are filtered: the key–group pairs, and the rows inside the groups.
+ * The key columns are removed from [keys][GroupBy.keys] and from the groups.
+ * ### This `xs` Overload
+ */
+@ExcludeFromSources
+private typealias CommonGroupByXsDocs = Nothing
+
+/**
+ * @include [CommonGroupByXsDocs]
+ * @include [XsDocs.DefaultKeyColumnsSnippet] {@set [XsDocs.SOURCE] [keys][GroupBy.keys]}
```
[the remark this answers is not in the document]
zaleslaw: Fixed. 

GroupBy.xs(vararg keyValues) now excludes the frame column that holds the groups from the default selection, so keys really is the source and the documented IllegalArgumentException is thrown. Added the missing case to number of key values must be equal to the number of key columns, asserting the message — a bare shouldThrow<IllegalArgumentException> passed before the fix as well, since Column not found: 'group' is also an IllegalArgumentException.

I excluded the groups column by path rather than filtering out all frame columns: DefaultKeyColumnsSnippet is shared with DataFrame.xs, where a frame column can legitimately be a key column, so a blanket !it.isFrameColumn() would have made the shared wording wrong for that overload.

zaleslaw:
Also fixes a bug this review surfaced: GroupBy.xs(vararg keyValues) counted the groups frame column as a key column, so the documented size check did not fire. One-line fix in the default selector plus a test.

PR #2057, opened 2026-09-01
Enable plugin evaluation of String API overloads

koperagen:
This PR enables plugin interpretation for String overloads that have CS DSL counterparts, see 
Plugin.kt:
```
Marks String API overloads (functions with vararg String parameters) with metadata to enable automatic delegation to their typed counterparts implementations in the compiler plugin. This annotation should not have default values for arguments (breaks plugin resolve).
Params:
interpreter - The interpreter name from the corresponding @Interpretable annotation on the typed API function (the one with ColumnSelectionDsl parameter)
targetArgument - The name of the ColumnSelectionDsl parameter in the typed API function
```

RawSchemaOperationsResultCorrectnessTest:
```
Making sure compiler evaluation of String API overload on raw dataframe (no schema) is correct whatever columns appear as a result are "type safe"; 
no ColumnNotFound or ClassCast exceptions will happen if property is used. 
```
**User scenario** when df has no schema, some string operations are invoked, and as a result we enrich schema

TypedSchemaOperationsResultConsistencyTest:
```
Making sure compiler evaluates String API overloads same as CS DSL overloads on typed input.
Sometimes some precision is expected to be lost, see Merge0
```
**User scenario** when dataframe itself is typed, but user chose to use String overload for simplicity => we do not loose information

I had to update kotlin to 2.4.20-RC2 so that compiler plugin understands StringApiInterpretable, to have these tests in this repo, i think it's better for visibility in this case

I plan to update this PR: clean up lint.kt and add annotations to a few remaining functions. 
Check the overall idea, clarity of tests, etc.

### DIFF
--- a/samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/plugin/stringApi/RawSchemaOperationsResultCorrectnessTest.kt
+++ b/samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/plugin/stringApi/RawSchemaOperationsResultCorrectnessTest.kt
@@ -1,0 +1,413 @@
+package org.jetbrains.kotlinx.dataframe.plugin.stringApi
+
+import org.jetbrains.kotlinx.dataframe.api.by
+import org.jetbrains.kotlinx.dataframe.api.byName
+import org.jetbrains.kotlinx.dataframe.api.cast
+import org.jetbrains.kotlinx.dataframe.api.convert
+import org.jetbrains.kotlinx.dataframe.api.cumSum
+import org.jetbrains.kotlinx.dataframe.api.distinct
+import org.jetbrains.kotlinx.dataframe.api.dropNA
+import org.jetbrains.kotlinx.dataframe.api.dropNulls
+import org.jetbrains.kotlinx.dataframe.api.excludeJoin
+import org.jetbrains.kotlinx.dataframe.api.explode
+import org.jetbrains.kotlinx.dataframe.api.fillNA
+import org.jetbrains.kotlinx.dataframe.api.fillNaNs
+import org.jetbrains.kotlinx.dataframe.api.fillNulls
+import org.jetbrains.kotlinx.dataframe.api.filterJoin
+import org.jetbrains.kotlinx.dataframe.api.flatten
+import org.jetbrains.kotlinx.dataframe.api.fullJoin
+import org.jetbrains.kotlinx.dataframe.api.gather
+import org.jetbrains.kotlinx.dataframe.api.group
+import org.jetbrains.kotlinx.dataframe.api.groupBy
+import org.jetbrains.kotlinx.dataframe.api.implode
+import org.jetbrains.kotlinx.dataframe.api.innerJoin
+import org.jetbrains.kotlinx.dataframe.api.into
+import org.jetbrains.kotlinx.dataframe.api.join
+import org.jetbrains.kotlinx.dataframe.api.leftJoin
+import org.jetbrains.kotlinx.dataframe.api.max
+import org.jetbrains.kotlinx.dataframe.api.maxFor
+import org.jetbrains.kotlinx.dataframe.api.mean
+import org.jetbrains.kotlinx.dataframe.api.meanFor
+import org.jetbrains.kotlinx.dataframe.api.median
+import org.jetbrains.kotlinx.dataframe.api.medianFor
+import org.jetbrains.kotlinx.dataframe.api.merge
+import org.jetbrains.kotlinx.dataframe.api.min
+import org.jetbrains.kotlinx.dataframe.api.minFor
+import org.jetbrains.kotlinx.dataframe.api.move
+import org.jetbrains.kotlinx.dataframe.api.moveTo
+import org.jetbrains.kotlinx.dataframe.api.moveToEnd
+import org.jetbrains.kotlinx.dataframe.api.moveToStart
+import org.jetbrains.kotlinx.dataframe.api.percentile
+import org.jetbrains.kotlinx.dataframe.api.percentileFor
+import org.jetbrains.kotlinx.dataframe.api.remove
+import org.jetbrains.kotlinx.dataframe.api.rename
+import org.jetbrains.kotlinx.dataframe.api.reorder
+import org.jetbrains.kotlinx.dataframe.api.rightJoin
+import org.jetbrains.kotlinx.dataframe.api.select
+import org.jetbrains.kotlinx.dataframe.api.split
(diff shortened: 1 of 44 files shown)


koperagen:
AI is concerned about such code:
```
// 1. have typed dataframe
interface MySchema {
  val myOtherCol: Int
  val myCol: List<Int>
}

val df = DataFrame.read().cast<MySchema>() 
// 2. apply String operation
df.distinct("myCol"), df.explode("myCol"), df.flatten("myCol"), df.join(other, "myCol") etc... 
```

Here's the breakdown of behavior:

1.0.0-rc01, 2.2.20..2.4.0
```
df.explode("myCol")

myOtherCol: Int // some useful info
myCol: List<Int> // INVALID

```

1.0.0-dev, 2.2.20..2.4.0
Code that could work in 1.0.0-rc01, breaks in dev:
df.explode("myCol").add("myOtherCol1") { myOtherCol + 1 }

```
df.explode("myCol")

<empty schema> 
```

How it will be on up to date version:
1.0.0-dev, 2.4.20+
```
df.explode("myCol")

myOtherCol: Int
myCol: Int // now schema is type safe for String operations too.
```

But it's worth to note that overall `@Refine` annotations already downgrades T to empty schema in essentially any multiplex operation combined with String API:
df.group("a", "b").into("gr")
df.gather("a", "b").valuesInto("v")
...
because all terminal operations already have Refine to work with CS DSL intermediate column selection operations

At the same time, this PR completely resolves this for 2.4.20 and onward. Starting from 2.4.20 both CS DSL and String API overloads will produce type safe results and not misleading, partially/sometimes correct DataFrame<T> -> DataFrame<T> schema preservation that is only useful in notebooks (no changes there).

Let's accepts this and recommend using 1.0.0 together with 2.4.20+ and IDE 2026.2+

AndreiKingsley (approved):
Great!

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/plugin/stringApi/commonTestData.kt
```diff
@@ -0,0 +1,81 @@
...
+import org.jetbrains.kotlinx.dataframe.annotations.DataSchema
+import org.jetbrains.kotlinx.dataframe.api.cast
+import org.jetbrains.kotlinx.dataframe.api.columnOf
+import org.jetbrains.kotlinx.dataframe.api.compileTimeSchema
+import org.jetbrains.kotlinx.dataframe.api.dataFrameOf
+import org.jetbrains.kotlinx.dataframe.api.first
+
+abstract class CommonTestData {
+    val df: DataFrame<Df> = dataFrameOf("key", "value", "other", "nullable", "nan")(
+        "a",
```
zaleslaw: hard to read the dataset

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/plugin/stringApi/RawSchemaOperationsResultCorrectnessTest.kt
```diff
@@ -0,0 +1,413 @@
...
+
+    @Test
+    fun DataFrameGroupBy() {
+        dfRaw.groupBy("key").toDataFrame() matches df.groupBy { key }.toDataFrame()
+            .convert { group }.with { it.select { key.cast<Any?>() } }
+            .select { key.cast<Any?>() and group }
+    }
+
+    @Test
+    @Ignore
```
zaleslaw: please add some comments why tests are ignored

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/plugin/stringApi/StringApiInterpretableConsistencyTests.kt
```diff
@@ -0,0 +1,175 @@
...
+            }
+
+        val stringOverloads = dataFrameApi
+            .select { annotations and name and parameters }
+            .cast<DataFrameApi>(verify = false)
+            .stringApiFunctions()
+            .split { annotationArguments }.inward("delegateInterpreter", "stringArgument", "targetArgument")
+            .rename { annotationArguments }.to("adapter")
+
+        val ignore = setOf(
```
zaleslaw: Complicated tests, please clarify the purpose

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/plugin/stringApi/StringApiInterpretableConsistencyTests.kt
```diff
@@ -0,0 +1,175 @@
...
+        val name: String,
+        val parameters: List<KoParameterDeclaration>,
+        val returnType: KoTypeDeclaration?,
+        val annotations: List<KoAnnotationDeclaration>,
+        val projectPath: String,
+        val isTopLevel: Boolean,
+        val typeParameters: List<KoTypeParameterDeclaration>,
+    )
+
+    private fun DataFrame<DataFrameApi>.interpretableFunctions(): DataFrame<InterpretableFunctions> =
```
zaleslaw: I'd like the approach to analyze the API itself with KDF, maybe worth to extract in separate utility?

zaleslaw (approved):
Please clarify my comments and could be merged

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/plugin/stringApi/commonTestData.kt
koperagen: indeed, disabled auto formatting here

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/plugin/stringApi/StringApiInterpretableConsistencyTests.kt
koperagen: Glad you liked it :) Extracted in a separate file

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/plugin/stringApi/StringApiInterpretableConsistencyTests.kt
koperagen: Updated. I agree, writing tests with dataframe logic i realized how much clarity of tests relies on some very basic rules. Imagine if we had something like checkAll here, it'd let us express assertions (and purpose) more clearly. 
"size shouldBe 0" is not exactly straightforward. Well, i hope that idea here will evolve

Issue #1951, opened 2026-07-24
Improve site docs for `append` and add a separate `appendNulls` page


zaleslaw:
The current `append` site page exists but is very brief: https://kotlin.github.io/dataframe/append.html

Please improve the `append` documentation and add a separate documentation page for `appendNulls` instead of mixing both APIs on one page.

### Scope

- Expand the `append` page with:
  - a short behavior description
  - function signature
  - input/output data tables, similar to other API pages such as `add`
  - examples for appending one row and multiple rows
  - a note that `append` returns a new `DataFrame` and does not mutate the original one
  - a performance note: repeatedly appending rows in a loop is an antipattern; prefer building data first or appending/concatenating in batches

- Add a separate `appendNulls` page with:
  - a short behavior description
  - function signature
  - input/output data tables
  - examples for `appendNulls()` and `appendNulls(numberOfRows)`
  - a note that `appendNulls(0)` returns the original dataframe
  - a note that the operation is immutable and returns a new `DataFrame` when rows are added

### Acceptance criteria

- `append` page contains data tables and examples consistent with the style of pages like `add`.
- `appendNulls` has its own dedicated site page.
- Both pages explain immutability clearly.
- `append` page mentions the loop-append performance antipattern.
- Related operations/navigation are updated so both pages are discoverable.

Related to #1950 better to complete together

Issue #1950, opened 2026-07-24
Add KDocs for `append` and `appendNulls`


zaleslaw:
`core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/append.kt` has no KDocs for `DataFrame.append` and `DataFrame.appendNulls`.

The site page exists for `append`: https://kotlin.github.io/dataframe/append.html, but the public APIs should also document their behavior directly in KDoc.

Please add concise KDocs that explain:

- `append` returns a new `DataFrame` with one or more rows appended.
- `append` assigns values row-wise, and the number of values must be a multiple of the dataframe column count.
- `appendNulls` returns a new `DataFrame` with the requested number of rows filled with `null` values.
- `appendNulls(0)` returns the original dataframe.
- Both operations do not mutate the original dataframe.
- Since these operations create a new dataframe/columns, appending rows one by one in a loop is an antipattern; users should prefer constructing data first or appending/concatenating in batches.

### Acceptance criteria

- `DataFrame.append(vararg values: Any?)` has KDoc.
- `DataFrame.appendNulls(numberOfRows: Int = 1)` has KDoc.
- KDocs are consistent with the existing site documentation.
- KDocs explicitly mention immutability and the performance caveat for repeated appends in loops.

PR #2061, opened 2026-09-04
Defined `append`s behavior on some corner cases and added tests

### LINKED ISSUE #1927: `append` fails when adding nulls to FrameColumn
To reproduce

```kt
val df = dataFrameOf("col", columnOf(emptyDataFrame()))
df.append(null) // Can not add value 'null' to FrameColumn
```

For consistency with `update`, this could also turn into a `ValueColumn<DataFrame<>?>`

### LINKED ISSUE #1952: Add Unit tests for `append`
## Test scenarios for `DataFrame.append`

Based on `core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/append.kt` and the site docs: https://kotlin.github.io/dataframe/append.html

### Basic append

- Given a dataframe with columns `name`, `age`
- When calling `df.append("Bill", 30)`
- Then the result contains one additional row
- And values are assigned by column order: `name = "Bill"`, `age = 30`

### Append multiple rows

- Given a dataframe with columns `name`, `age`
- When calling `df.append("Mike", 15, "John", 17, "Bill", 30)`
- Then three rows are appended
- And values are interpreted row-wise, not column-wise

### Invalid number of values

- Given a dataframe with `n` columns
- When calling `append` with a number of values that is not a multiple of `n`
- Then an exception is thrown
- And the error explains that a multiple of the column count is expected

### Append zero rows

- Given a dataframe with at least one column
- When calling `df.append()` with no values
- Then the result has the same row count and values as the original dataframe

### Append nullable values

- Given a dataframe with nullable-compatible columns
- When calling `df.append(null, someValue)`
- Then `null` is appended as a value in the corresponding column

### Append with compiler plugin overload

- Given a `@DataSchema` class, as shown on the site
- When the compiler plugin overload is used, e.g. `df.append(Person("Bill", 30))`
- Then the row is appended according to the schema
- This should be covered in plugin/integration tests if the overload is generated there

### Empty dataframe edge case

- Given a dataframe with zero columns
- When calling `append(...)`
- Then behavior should be explicitly tested/documented, because current implementation checks `values.size % ncol`, which is problematic when `ncol == 0`

Allex-Nik:
Fixes #1952
Fixes #1927 

- Currently `append` fails when one appends a `null` to a `FrameColumn`. Changes from this PR make `append` add an empty frame instead of `null`, keeping the type of the column `FrameColumn`.
- Make `append` return the original dataframe early when `append` does not have any arguments.
- Make `append` with arguments fail with an informative message when the receiver dataframe has no columns.
- Add tests for `append`

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/append.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/append.kt
@@ -9,8 +9,13 @@ import org.jetbrains.kotlinx.dataframe.nrow
 // region DataFrame
 
 public fun <T> DataFrame<T>.append(vararg values: Any?): DataFrame<T> {
+    if (values.isEmpty()) return this
+
     val ncol = ncol
-    check(values.size % ncol == 0) {
+    require(ncol > 0) {
+        "Cannot append values to a DataFrame with no columns"
+    }
+    require(values.size % ncol == 0) {
         "Invalid number of arguments. Multiple of $ncol is expected, but actual was: ${values.size}"
     }
     val newRows = values.size / ncol
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
@@ -123,7 +123,9 @@ internal fun <T, C> DataColumn<C>.updateImpl(
 }
 
 /**
- * Replaces all values in column asserting that new values are compatible with current column kind
+ * Replaces all values in column asserting that new values are compatible with current column kind.
+ *
+ * For a [FrameColumn], `null` values are replaced by empty dataframes with the column's schema.
  */
 internal fun <T> DataColumn<T>.updateWith(values: List<T>): DataColumn<T> =
     when (this) {
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
@@ -128,10 +128,14 @@ internal fun <T, C> DataColumn<C>.updateImpl(
 internal fun <T> DataColumn<T>.updateWith(values: List<T>): DataColumn<T> =
     when (this) {
         is FrameColumn<*> -> {
-            values.forEach {
-                require(it is AnyFrame) { "Can not add value '$it' to FrameColumn" }
+            val frameSchema by schema
+            val groups = values.map {
+                when (it) {
+                    null -> DataFrame.empty(frameSchema)
+                    is AnyFrame -> it
+                    else -> throw IllegalArgumentException("Can not add value '$it' to FrameColumn")
+                }
             }
-            val groups = (values as List<AnyFrame>)
             DataColumn.createFrameColumn(name, groups) as DataColumn<T>
         }
 
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
@@ -26,6 +26,7 @@ import org.jetbrains.kotlinx.dataframe.columns.size
 import org.jetbrains.kotlinx.dataframe.impl.columns.AddDataRowImpl
 import org.jetbrains.kotlinx.dataframe.impl.createDataCollector
 import org.jetbrains.kotlinx.dataframe.index
+import org.jetbrains.kotlinx.dataframe.nrow
 import org.jetbrains.kotlinx.dataframe.type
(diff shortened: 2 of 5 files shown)


core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/append.kt
```diff
@@ -0,0 +1,193 @@
...
+        val result = df.append(null)
+
+        result.nrow shouldBe 2
+        result["col"].kind() shouldBe ColumnKind.Frame
+        result["col"].type() shouldBe typeOf<DataFrame<*>>()
+        result["col"].values() shouldBe listOf(frame, DataFrame.empty(frame.schema()))
+    }
+
+    @Test
+    fun `append uses the DataRowSchema overload`() {
```
Allex-Nik: I am a bit unsure regarding this scenario. The [issue](https://github.com/Kotlin/dataframe/issues/1952) mentioned the scenario "Append with compiler plugin overload". The compiler plugin does not work in `core/...test/` but this overload from `DataRowSchemaApi.kt` seems to do what is expected (`df.append(Person("Bill", 30))`). And I also saw that there is a test checking this behavior on the compiler plugin side (`dataRowSchemaApi.kt`). Do you think it still makes sense to add a test to a module with compiler plugin working (or maybe move all these tests there)?

Image

koperagen: It does because  Person manually adds required interfaces, same what plugin does:
private data class Person(val name: String, val age: Int) : DataRowSchema
But yeah, indeed, it's a workaround and makes more sense to move this case to samples module
Not all tests

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/append.kt
```diff
@@ -0,0 +1,193 @@
...
+    fun `append null to frame column replaces null with empty dataframe with schema`() {
+        val frame = dataFrameOf("value")(1)
+        val df = dataFrameOf(columnOf(frame) named "col")
+
+        val result = df.append(null)
+
+        result.nrow shouldBe 2
+        result["col"].kind() shouldBe ColumnKind.Frame
+        result["col"].type() shouldBe typeOf<DataFrame<*>>()
+        result["col"].values() shouldBe listOf(frame, DataFrame.empty(frame.schema()))
```
koperagen: Please add something like this too
```
result["col"].values().checkAll { 
  shouldNotThrowAny { 
    it["value"] // check column access
  }
}
```

koperagen (commented):
Nice :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/append.kt
```diff
@@ -9,8 +9,13 @@ import org.jetbrains.kotlinx.dataframe.nrow
 // region DataFrame
 
 public fun <T> DataFrame<T>.append(vararg values: Any?): DataFrame<T> {
```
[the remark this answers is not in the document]
Allex-Nik: I am planning to do it in the next PR fixing #1950 and #1951

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
```diff
@@ -128,10 +128,14 @@ internal fun <T, C> DataColumn<C>.updateImpl(
 internal fun <T> DataColumn<T>.updateWith(values: List<T>): DataColumn<T> =
```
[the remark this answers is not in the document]
Allex-Nik: Done

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/append.kt
Allex-Nik: Done

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/append.kt
Allex-Nik: Done

koperagen (approved):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/impl/api/update.kt
```diff
@@ -123,15 +123,21 @@ internal fun <T, C> DataColumn<C>.updateImpl(
...
+            val frameSchema by schema
+            val groups = values.map {
+                when (it) {
+                    null -> DataFrame.empty(frameSchema)
+                    is AnyFrame -> it
+                    else -> throw IllegalArgumentException("Can not add value '$it' to FrameColumn")
+                }
             }
-            val groups = (values as List<AnyFrame>)
             DataColumn.createFrameColumn(name, groups) as DataColumn<T>
```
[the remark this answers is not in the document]
Allex-Nik: Done

PR #2063, opened 2026-09-04
Read Arrow/Parquet timestamps with a time zone as `Instant`

### LINKED ISSUE #926: Support Parquet UTC timestamp vectors (`TimeStampXXXTZVector`) in Arrow reader

zaleslaw:
## Read Arrow/Parquet timestamps with a time zone as `Instant`

Fixes #926.

Reading a Parquet file with a timestamp column flagged `isAdjustedToUTC = true` failed with
`NotImplementedError("reading from TimeStampXXXTZVector is not implemented")`. Such columns are
common — PyArrow, Polars and pandas write them for every time-zone-aware column — so a valid file
could not be read at all.

### What the types mean

Parquet encodes one bit of meaning that we now keep:

| Parquet logical type | Kotlin type |
|---|---|
| `Timestamp(isAdjustedToUTC = true, …)` | `kotlin.time.Instant` |
| `Timestamp(isAdjustedToUTC = false, …)` | `kotlinx.datetime.LocalDateTime` |

An `Instant` is a point on the time line. A `LocalDateTime` is a calendar-and-clock reading with no
zone, so it is not a point in time at all. Reading both as `LocalDateTime` would erase the
difference, and Kotlin's own types exist to keep it: converting between them always needs an
explicit `TimeZone`.

`MILLIS`, `MICROS` and `NANOS` are all supported and keep full precision. Arrow also has a `SECOND`
unit and allows any zone string; both work, even though no Parquet file can contain them.

A zone *name* may still be stored next to such a column — Parquet keeps only the flag, but PyArrow
writes the original zone into `ARROW:schema`, and in Arrow IPC/Feather it is part of the schema. It
is display metadata: the values are already normalized to UTC, so it is neither applied nor
returned.

### Writing

An `Instant` column is written as `Timestamp(MICROSECOND, "UTC")`, so it reads back as an `Instant`.
Before this PR it fell through to `Utf8` and was saved as a string — **this changes the bytes such
files contain.**

Two decisions here:

* **Microseconds, not nanoseconds.** An Arrow nanosecond timestamp is an `int64` count of
  nanoseconds since the epoch, so it only covers 1677–2262. Microseconds reach roughly ±292 277
  years around the epoch — not the whole `Instant` range, but every instant a file realistically
  holds — and match what pandas/Polars/PyArrow emit. The cost is that a sub-microsecond instant
  loses its last three digits, now reported as the new `ConvertingMismatch.PrecisionReduced`; pass a
  target `Schema` with `Timestamp(NANOSECOND, "UTC")` to keep them. A value that does not fit the
  target unit at all is reported as the new `ConvertingMismatch.ValueOutOfRange` and then refused
  with a `ConvertingException` (`Mode.STRICT`, the default) or written as `null` (`Mode.LOYAL`) —
  never silently wrapped around. Dropping it needs a nullable target field, so a non-nullable one is
  refused in either mode.
* **`Timestamp` conversions are pinned to UTC.** The generic converters in `core` resolve against
  `TimeZone.currentSystemDefault()`, so the same frame produced different files depending on the
  machine. Arrow and Parquet define a timestamp against `1970-01-01T00:00:00Z`, so the writer uses
  its own `convertToInstantInUtc` / `convertToLocalDateTimeInUtc` for all three source shapes that
  need a zone: local date-times, dates, and numbers read as epoch millis.

### Public API

Additive only, `apiDump` updated:

* `ConvertingMismatch.PrecisionReduced`
* `ConvertingMismatch.ValueOutOfRange`
* `ARROW_PARQUET_DEFAULT_BATCH_SIZE` is now `public` (the docs sample had a hand-copied duplicate
  that had already drifted out of date — `1000` against the real `32768`)

Both mismatches are new subclasses of the public sealed `ConvertingMismatch`, so an exhaustive `when`
over it in user code needs a new branch — source-incompatible, binary-compatible.

### Files

| File | What changed |
|---|---|
| `arrowReadingImpl.kt` | read the four `TimeStamp*TZVector`s; `Duration` null guard |
| `arrowTypesMatching.kt` | `Instant` → Arrow field, shared unit helpers |
| `ArrowWriterImpl.kt` | one branch for all eight timestamp vectors; UTC-pinned conversions; exact `Long` boundary; out-of-range values reported through the mismatch subscriber and gated on the mode; close the vector when infill throws |
| `ConvertingMismatch.kt` | `PrecisionReduced`, `ValueOutOfRange` |
| `arrowReading.kt` | KDoc on every read overload, public constant |
| `api/dataframe-arrow.api` | `apiDump` |
| `ArrowTimestampTzTest.kt`, `ArrowTypeMappingTest.kt` | new; `ArrowKtTest.kt` extended |
| `timestamps_tz.py`, 3 fixtures, `TestFiles.md` | generator, committed fixtures, their documentation |
| `dataSources/ApacheArrow.md`, `dataSources/Parquet.md`, `setup/Modules.md`, `samples/io/Parquet.kt` | type mapping tables, timestamps and time zones, Korro sample |
| `dataframe-arrow/AGENTS.md`, `AGENTS.md`, `plugins/AGENTS.md`, `plugins/CLAUDE.md` | writer rules for agents; plugin-folder status — see the note below |

### Tests

None of this was covered before: the reader threw on these vectors and the writer had no timestamp
path at all. `ArrowTimestampTzTest` has 17 tests, `ArrowTypeMappingTest` 5.

| Behaviour | What it pins down |
|---|---|
| Reading zone-tagged timestamps | all four units through IPC and Feather, any zone string, the three `NullabilityOptions`, null and pre-1970 rows |
| Zone-less timestamps | stay `LocalDateTime`, side by side with a zoned column in one file |
| PyArrow/Polars interop | committed fixtures, plus a Feather twin from the same script because the `arrow-dataset` JNI library is not available everywhere |
| Writing is environment-independent | the round trip under four shifted default zones, one column per source shape — reverting the UTC fix makes it fail |
| Writer edge cases | precision loss is reported; an out-of-range instant reaches the mismatch subscriber and is then refused (`Mode.STRICT`, or any non-nullable field) or written as `null` (`Mode.LOYAL`); the `Long.MIN_VALUE` boundary of a unit round-trips; a failed write leaks no Arrow buffers |
| Arrow ↔ Kotlin mapping | the whole table as published in `ApacheArrow.md`, including that the write side has **no** `List` mapping — a `FrameColumn` degrades to `Utf8` at top level and fails inside a `ColumnGroup` — so code and docs cannot drift apart |

The Parquet fixtures are written by Polars/PyArrow on purpose, so the tests prove interop rather
than round-tripping our own encoder. `src/test/generators/timestamps_tz.py` regenerates them and
records the exact package versions.

Also fixed on the way: a nullable `Duration` column threw `NullPointerException`, because its reader
called `getObject(i).toKotlinDuration()` without the null guard every neighbouring reader has.

### Note for the reviewer

* **Scope left out:** only `Timestamp` fields are UTC-pinned. The `Date(DAY)` and `Time(unit)`
  branches still use `core`'s plain converters and stay zone-dependent — same bug, reachable only
  with an explicit target `Schema`. The root cause is in `core` and is tracked as #2072, so I left
  it there rather than widening this PR.
* **Unrelated changes I did include:** `AGENTS.md`, `plugins/AGENTS.md` and `plugins/CLAUDE.md`
  document that three of the six modules under `plugins/` are disabled and absent from
  `settings.gradle.kts`. Found while tracing the compiler plugin for this PR — the root `AGENTS.md`
  presented `symbol-processor` and `dataframe-gradle-plugin` as live tooling. Say the word and I
  will move them out into their own PR.
* **Behaviour change since the first revision of this branch:** an out-of-range instant used to
  abort the write with a bare `IllegalArgumentException` in every mode. It now goes through
  `mismatchSubscriber` and only aborts where the mode (or a non-nullable field) says it must —
  `Mode.LOYAL` degrades to `null` as documented.
* **Documented as unsupported, not implemented:** the write table no longer claims
  `FrameColumn` → `List`. List vectors are still unimplemented in the writer (the `TODO` in
  `ArrowWriterImpl.infillVector`), so a top-level frame column is saved as `Utf8` and a nested one
  fails on allocation; both are now pinned by a test and stated on the page. Implementing list
  vectors is a separate PR — say the word and I will open an issue.
* **Limitation I shipped:** `Mode.LOYAL` cannot degrade an out-of-range value into a non-nullable
  field, because the emitted file would contradict its own schema (`NullabilityOptions.Checking`
  refuses to read it back). It is refused there instead of widening the field — widening a field the
  user declared non-nullable seemed the worse surprise.
* **Finding, not fixed here:** the `zone` overloads of `convertToLocalDateTime` in `core` have a
  copy-paste KDoc — "interpret the … timestamp as a **time**", carried over from `toLocalTime`.
* **Assumption I shipped:** `MICROSECOND` as the default unit for an `Instant` column. It silently
  drops sub-microsecond digits (reported, not thrown). `NANOSECOND` would keep them but cap the
  range at 1677–2262, which seemed the worse trade.


koperagen (approved):
Nice!

dataframe-arrow/src/test/generators/timestamps_tz.py
Jolanrensen, line 1: What is this python doing here

PR #2065, opened 2026-09-08
Add KDoc and tests for `ColumnAccessor.nullable()`

### LINKED ISSUE #1996: Add KDoc for `ColumnAccessor.nullable`
# Add KDoc for `ColumnAccessor.nullable`

The single public function in `core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/ColumnAccessorApi.kt` has no KDoc. Add one.

### Scope

- `ColumnAccessor<T>.nullable(): ColumnAccessor<T?>` — a compile-time cast that retypes a column accessor's element type as nullable.

### Recommended KDoc (ready to use)

```kotlin
/**
 * Marks this [ColumnAccessor]'s element type as nullable, converting a
 * [ColumnAccessor]`<T>` into a [ColumnAccessor]`<T?>`.
 *
 * Purely a compile-time cast: no values are changed or checked. Use it when you
 * declare a typed accessor for a column that can actually contain `null`s, so that
 * reading through it gives you `T?` instead of `T` and the compiler forces you to
 * handle the `null` case.
 *
 * ### Example
 * ```kotlin
 * // A column of scores where some rows may be missing:
 * val score by column<Double>().nullable()   // ColumnAccessor<Double?>
 *
 * df.filter { score() != null && score()!! > 0.5 }
 * ```
 *
 * @receiver A typed accessor for a column of `T`.
 * @return The same accessor retyped as [ColumnAccessor]`<T?>`.
 * @see castToNullable
 */
public inline fun <reified T> ColumnAccessor<T>.nullable(): ColumnAccessor<T?> = cast()
```

### Acceptance criteria

- `ColumnAccessor.nullable()` has a concise KDoc stating it is a compile-time cast (no data change/check), when to use it, and includes an example.
- The `@see castToNullable` cross-reference is kept only if it resolves and is useful for navigation; otherwise drop it.

zaleslaw:
# Add KDoc for `ColumnAccessor.nullable()`

Closes #1996

`ColumnAccessor<T>.nullable()` was the only public declaration in `ColumnAccessorApi.kt`
and had no documentation at all.

The function is easy to misread as a data operation. It is a plain unchecked cast: it
returns the very same accessor object with the value type widened to `T?`, looks up no
column and reads no value. Two things are worth knowing and were not written down anywhere:

- `ColumnAccessor` is covariant, so a `ColumnAccessor<Double>` is already accepted where a
  `ColumnAccessor<Double?>` is expected. What the call changes is the declared type of the
  accessor itself.
- Without the cast, a `null` from the column arrives as a `null` in the non-nullable type
  `T` and only fails later, at the first use of the value. That is the reason the function
  exists, so the KDoc now says it.

## Files

`core/.../api/ColumnAccessorApi.kt` — the KDoc, following the structure of `first.kt`.

`core/.../test/.../api/ColumnAccessorApi.kt` — new `ColumnAccessorApiTests`, one test per
documented claim.

`KDOC_GUIDELINES.md` — a note in the "Examples section": keep deprecated API out of KDoc
examples, and where to find the deprecation constants.

## Example

The example shows a `concat` with a frame that has no `score` column, because that is how
the nulls appear in practice. The same scenario is asserted by the test named
`KDoc example nullable`, so the example cannot silently rot.

## Tests

`nullable()` had no test of its own before; its only usage was incidental, inside a `concat`
test.

Nine tests, one per claim: the KDoc example; a `null` read through a non-nullable accessor
(it returns `null` and throws `NullPointerException` only at the first use); the return type
and object identity; the name and the path for a plain, a nested and a column-group
accessor; a missing column and a column without nulls; an already nullable accessor; and the
return type of `castToNullable` on an accessor, which the "See also" section refers to.

Where the return type *is* the contract, it is pinned by an explicit type annotation, so the
compiler is the assertion. All expected values are written out literally.

## Open questions

**Should this function be deprecated instead?** Almost everything that reads through an
accessor is already `@Deprecated(DEPRECATED_ACCESS_API)`: `df[accessor]`, `row[accessor]`,
`accessor()`, `df.getColumn(accessor)`. The only supported reading path left is
`accessor.getValue(row)`, which is why the example uses it. If the Column Accessors API is
going away, `nullable()` should probably go with it, and this KDoc documents an API on its
way out. Please tell me which way you want it.

**`inline` / `reified` look unnecessary.** `cast()` here resolves to the non-reified
`ColumnAccessor<*>.cast()`, and a plain `fun <T> ColumnAccessor<T>.nullable()` compiles
fine, call sites included. I left the signature untouched, because it is public and out of
scope for a KDoc issue.

**`@param [T]`.** The guidelines ask to document type parameters, so I did, but this one is
fully implied by the receiver type. If we do not want such `@param`s, the rule in
`KDOC_GUIDELINES.md` should be changed as well — today, the rule and the practice disagree
(16 uses across `api/`).

### DIFF
--- a/KDOC_GUIDELINES.md
+++ b/KDOC_GUIDELINES.md
@@ -420,18 +420,19 @@ For complex operations, write a **complete** example with all steps.
 
 For methods with Columns Selection DSL, add several examples with different CS DSL selection methods.
 
-Keep deprecated API out of the examples, even when documenting a legacy function. Around the Column
-Accessors API, for instance, almost everything that reads through an accessor (`df[accessor]`,
-`row[accessor]`, `accessor()`) is deprecated, while `accessor.getValue(row)` is not — check the
-annotations before copying a call from an old test. The deprecation messages are constants in
-[`util/deprecationMessages.kt`](./core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt)
-(`DEPRECATED_ACCESS_API`, `DEPRECATED_DATA_ROW_COLUMN_REFERENCE_GET`), so a grep for the constant
-shows the whole affected surface at once.
+Keep deprecated API out of the examples, even when documenting a legacy function — check for
+`@Deprecated` before copying a call from an old test.
 
 Start the section with
 
 ```
 ### Examples
+```
+
+or, when the KDoc has exactly one example, with
+
+```
+### Example
 ```
 
 #### Data tables in examples
--- a/KDOC_GUIDELINES.md
+++ b/KDOC_GUIDELINES.md
@@ -420,6 +420,14 @@ For complex operations, write a **complete** example with all steps.
 
 For methods with Columns Selection DSL, add several examples with different CS DSL selection methods.
 
+Keep deprecated API out of the examples, even when documenting a legacy function. Around the Column
+Accessors API, for instance, almost everything that reads through an accessor (`df[accessor]`,
+`row[accessor]`, `accessor()`) is deprecated, while `accessor.getValue(row)` is not — check the
+annotations before copying a call from an old test. The deprecation messages are constants in
+[`util/deprecationMessages.kt`](./core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/util/deprecationMessages.kt)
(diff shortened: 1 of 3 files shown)


AndreiKingsley (changes requested):
Please deprecate this instead as was discussed.

PR #2066, opened 2026-09-08
Add KDocs, tests and site updates for `map.kt` APIs

### LINKED ISSUE #1941: Add KDocs for `map.kt` APIs
# Add missing KDocs for public APIs in `map.kt`

## Motivation

`core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/map.kt` contains several public APIs without KDocs. Some of them are marked with `@Interpretable`, so their expected behavior should be documented explicitly for users and for consistency with other DataFrame operations.

The main source of confusion is that these APIs share the `map` naming but return different shapes:

- `DataColumn.map` / `mapIndexed` return a `DataColumn`
- `DataFrame.map` returns a `List`
- `mapToColumn` creates a computed `DataColumn`
- `mapToFrame` builds a new `DataFrame`
- `GroupBy.map` / `mapToRows` / `mapToFrames` operate on grouped data

KDocs should make those differences clear without duplicating full website documentation.

## Scope

Add or improve KDocs for:

- `DataFrame.map`
- `DataFrame.mapToColumn`
- `ColumnsContainer.mapToColumn`
- `DataFrame.mapToFrame`
- `GroupBy.map`
- `GroupBy.mapToRows`
- `GroupBy.mapToFrames`

Also check and document or improve:

- `DataColumn.map`
- `DataColumn.map(type: KType, ...)`
- `DataColumn.mapIndexed`
- `DataColumn.mapIndexed(type: KType, ...)`

Explicit `KType` overloads should explain how the provided type is used for the resulting column.

## Acceptance Criteria

- Public APIs in `map.kt` have concise KDocs.
- `@Interpretable` APIs have documented behavior and parameters.
- Explicit `KType` overloads explain how the provided type is used.
- KDocs clearly state what each API maps over and what it returns.
- Existing examples/docs for `map`, `mapToColumn`, and `mapToFrame` remain consistent with the new KDocs.

zaleslaw:
## `map` KDocs: correct the column-kind rule and stop showing the wrong result

Closes #1941 

Ten public functions share the name `map`. The one thing that is genuinely hard to get right
about them is what kind of column `DataColumn.map` produces: the kind is chosen from the type
argument (or the explicit `type`) **as given** — before `infer`, and regardless of the computed
values. A previous revision of this page said the opposite ("follows the type of the computed
values"), which is wrong in both directions: `map<_, Any> { someFrame }` gives a `ValueColumn`,
not a `FrameColumn`, and `infer` never changes a kind — it only adjusts the type of a
`ValueColumn`. When the two disagree the call can fail outright: a `ValueColumn` may never have
a `DataFrame` type, so `Infer.Type` over computed dataframes, or a nullable `DataFrame` type,
throws `IllegalArgumentException`.

| File | What changed |
|---|---|
| `core/…/api/map.kt` | The column-kind and `infer` rule; `See also` lists with the difference to each neighbour; examples taken from the website; the schema caveat on `mapToRows` |
| `core/…/documentation/DocumentationUrls.kt` | The `map` links grouped into `Map { … }`, as `MinMax`/`Drop`/`Fill` |
| `core/api/core.api` | `apiDump` for that nested interface |
| `core/…/test/api/map.kt` | 8 tests for the rule above and for the claims added here |
| `core/…/samples/api/Modify.kt` | `mapOnGroupBy` no longer claims a rendered output; `mapToFrames` renders its own frame column |
| `docs/…/topics/map.md` | The same rule; `prev()?.newValue()`; the key–group pair defined where it is used |
| `docs/…/topics/groupBy.md` | `mapper` in the grammar; the page's own terminology |
| `docs/…/AGENTS.md` | When a sample may carry `@TransformDataFrameExpressions` |

### Examples

The examples now use the dataset and the calls from the `map` page, so the KDoc, the website and
the tests show one example instead of three. `mapToFrames` groups by first name rather than city:
with `city` every group has at most two rows, so `take(2)` dropped nothing and the example
demonstrated nothing.

### Tests

| Test | What it pins down |
|---|---|
| `the kind of the new column follows R and not the computed values` | The declared type decides the kind, computed rows under `Any` stay a `ValueColumn` |
| `R has to fit the computed values` | `Infer.Type` over dataframes, and a nullable frame type, both throw |
| `map returns a column group / frame column / value column …` (3) | One kind per test |
| `a nullable row type still gives a column group` | Nullability changes the frame case but not the row case |
| `Infer None / Nulls / Type` on `map`, and `Infer None` on `mapToColumn` (5) | Which of the two the type comes from |
| `KDoc example - mapToFrames can drop rows inside the groups` | The example's claim that only one group loses a row |
| `KDoc example - …` (5 others) | Every example in the KDoc and on the page |

The kind of the new column, the `KType` overloads, `mapIndexed`, `mapToFrame` and the `GroupBy`
functions had no tests before. Where the return type is the contract, the tests declare it
explicitly and let the compiler assert it; the two `Infer`-independent throw cases are asserted
with `shouldThrow`, since one of the two checks behind them only runs with
`kotlin.dataframe.debug=true`.

### Note for the reviewer

`core:samplesTest` goes from 27 to 28 failures. The new one is `mapOnGroupBy`: it returns a
`List`, which the expressions converter cannot render, and the alternative is worse — with the
annotation it passes but the page then shows the intermediate `groupBy` table under an example
whose result is a list (that is what this PR removes, in two places). `mapToFramesOnGroupBy`
avoids the trade-off by rendering its frame column itself, as `JoinWith` does.


zaleslaw:
@Allex-Nik did I add data snippets correctly?

Allex-Nik:
> @Allex-Nik did I add data snippets correctly?

Not exactly. 

Tables for `mapToFramesOnGroupBy` and `mapToRowsOnGroupBy` are not shown on the website page because the corresponding html-samples are not added to the `_shadow_resources.md` file. The `gradle updateShadowResources` command should be run when new samples are added.

But the most important thing is that the samples are added to the `core` module, and we add new samples to the `samples` module and the old samples should be moved there by the end of the 1.0 milestone (#898). Since the original `map` page had few samples (and only one with an iframe) and we cannot have samples for one page in two modules, I think we could move the existing samples to the `samples` module and add the new samples to this module as well (as we do for `groupBy`, `pivot`, `countDistinct`, `between`, `join`) because anyway we should do it soon. Additionally, adding samples to the `samples` module would allow to show tables for the `mapOnColumn`, `mapIndexedOnColumn`, and `mapToFramesOnGroupBy` without any workarounds.

docs/StardustDocs/topics/groupBy.md
```diff
@@ -13,7 +13,7 @@ Each row in a `GroupBy` corresponds to a keys/group combination.
 ```text
 groupBy(moveToTop = true) { columns }
       [ transformations ]
-      reducer | aggregator | pivot
+      reducer | aggregator | pivot | mapper
```
Allex-Nik, line 16: We should update the `groupBy` KDoc accordingly (Grammar and descriptions)

docs/StardustDocs/topics/groupBy.md
```diff
@@ -25,6 +25,8 @@ reducer = .minBy { column } | .maxBy { column } | .medianBy { rowExpression } |
 
 aggregator = .count() | .countDistinct() | .concat() | .concatWithKeys() | .toDataFrame() | .into([column]) [{ rowExpression }] | .values { valueColumns } | .aggregate { aggregations } | .<stat> [ { columns } ]
 
+mapper = .map { GroupWithKey -> value } | .mapToRows { GroupWithKey -> DataRow? } | .mapToFrames { GroupWithKey -> DataFrame }
```
Allex-Nik, line 28: Since `mapper` is separate from `aggregator`, `reducer`, and `transformations`, should we maybe make mapping a separate section on this page (with the descriptions that are added below and a couple of samples with tables)?

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/map.kt
```diff
@@ -21,6 +24,106 @@ import kotlin.reflect.KProperty
...
+ *   [DataColumn] that is not part of that [DataFrame].
+ * - [mapToFrame][DataFrame.mapToFrame] — returns a new [DataFrame] made of the columns described in an [AddDsl].
+ * - [map][GroupBy.map] — goes over the key–group pairs of a [GroupBy] and returns a [List].
+ * - [mapToRows][GroupBy.mapToRows] — goes over the key–group pairs of a [GroupBy] and returns a [DataFrame].
+ * - [mapToFrames][GroupBy.mapToFrames] — goes over the key–group pairs of a [GroupBy] and returns a [FrameColumn].
+ *
+ * Each result keeps the order of the values, rows, or key–group pairs it was computed from.
+ *
+ * See also:
+ * - [add][DataFrame.add] — computes a new column in the same way, and returns the [DataFrame] with that column in it.
```
Allex-Nik, line 54: Maybe it's clear enough already, but I would make it more explicit that the returned dataframe keeps the initial columns. Something like "...with that column added to it"

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/map.kt
```diff
@@ -50,6 +205,34 @@ public inline fun <T, R> DataColumn<T>.map(
...
+ * - {@include [MapDocsLink]} — an overview of the whole `map` family.
+ * - [map][DataColumn.map] — the same, without the positions.
+ * - the overload with an explicit `type` argument — when the type of the new column
+ *   is only known at runtime.
+ *
+ * ### Example
+ *
+ * ```kotlin
+ * // The first names, numbered: "1. Alice", "2. Bob", ...
+ * df.name.firstName.mapIndexed { i, firstName -> "\${i + 1}. \$firstName" }
```
Allex-Nik: The resulting column will contain strings: "${i + 1}. $firstName", not "1. Alice" etc.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/map.kt
```diff
@@ -152,16 +465,104 @@ public inline fun <T> DataFrame<T>.mapToFrame(body: AddDsl<T>.() -> Unit): DataF
...
+ * - {@include [MapDocsLink]} — an overview of the whole `map` family.
+ * - [mapToRows][GroupBy.mapToRows] — the same, but every result is a single row instead of a whole [DataFrame].
+ * - [updateGroups][GroupBy.updateGroups] — the same computation over the groups,
+ *   but the result is a [GroupBy] with its key columns, and not a bare [FrameColumn].
+ *
+ * ### Example
+ *
+ * ```kotlin
+ * // The two oldest people with each first name, as a frame column:
+ * // only the group of "Charlie" has a third person to leave out
```
Allex-Nik, line 557: If we don't show the dataset, maybe we don't need this line

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/map.kt
```diff
@@ -152,16 +465,104 @@ public inline fun <T> DataFrame<T>.mapToFrame(body: AddDsl<T>.() -> Unit): DataF
...
+ * See also:
+ * - {@include [MapDocsLink]} — an overview of the whole `map` family.
+ * - [mapToRows][GroupBy.mapToRows] — the same, but every result is a single row instead of a whole [DataFrame].
+ * - [updateGroups][GroupBy.updateGroups] — the same computation over the groups,
+ *   but the result is a [GroupBy] with its key columns, and not a bare [FrameColumn].
+ *
+ * ### Example
+ *
+ * ```kotlin
+ * // The two oldest people with each first name, as a frame column:
```
Allex-Nik, line 556: "including their first names"?

docs/StardustDocs/topics/map.md
```diff
@@ -2,12 +2,30 @@
...
+| [`map`](#map-on-datacolumn) / [`mapIndexed`](#map-on-datacolumn) | values of a [`DataColumn`](DataColumn.md) | a [`DataColumn`](DataColumn.md) of the same size |
+| [`map`](#map-on-groupby) | key–group pairs of a [`GroupBy`](groupBy.md) | `List` |
+| [`mapToRows`](#map-on-groupby) | key–group pairs of a [`GroupBy`](groupBy.md) | a [`DataFrame`](DataFrame.md) |
+| [`mapToFrames`](#map-on-groupby) | key–group pairs of a [`GroupBy`](groupBy.md) | a [`FrameColumn`](DataColumn.md#framecolumn) |
+
+Each result keeps the order of the values, rows, or key–group pairs it was computed from.
+
+## map
+
+Maps the rows of a [`DataFrame`](DataFrame.md) into a `List` with one element per row.
```
Allex-Nik, line 28: Applicable if samples are moved to `samples`: Since we use the same `df` for every sample, shall we show this df as a table (iframe) with a note that the samples below use this dataset? So that users can see how the operation affected the dataset

docs/StardustDocs/topics/map.md
```diff
@@ -51,9 +78,18 @@ df.mapToColumn("year of birth") { 2021 - "age"<Int>() }
```
Allex-Nik, line 78: We could add an iframe (table) for `mapToColumn`. As well as for `mapOnColumn`, `mapIndexedOnColumn`

docs/StardustDocs/topics/map.md
```diff
@@ -93,3 +129,116 @@ df.mapToFrame {
...
+
+<!---FUN mapToFramesOnGroupBy-->
+
+```kotlin
+// The two oldest people with each first name, as a frame column:
+// only the group of "Charlie" has a third person to leave out
+df.groupBy { name.firstName }.mapToFrames { group.sortByDesc { age }.take(2) }
+```
+
+<inline-frame src="resources/org.jetbrains.kotlinx.dataframe.samples.api.Modify.mapToFramesOnGroupBy.html" width="100%"/>
```
Allex-Nik, line 235: I think it would be easier to read if we unfold this table

docs/StardustDocs/AGENTS.md
```diff
@@ -27,6 +27,41 @@ Project marker: `project.ihp`; instance profile / table-of-contents: `d.tree`
...
+is compiled and run on every build — a hand-written block is the way a hallucinated or long-dead API reaches the
+website. Only `text`-fenced pseudo-grammar (`operation { columnMapping }: DataFrame`) is written by hand.
+
+To add one:
+
+1. **Find which module owns the topic.** `:core`'s korro block takes `topics/*.md` + `topics/concepts/*.md`
+   wholesale; `:samples` has an explicit `include(...)` allow-list in `samples/build.gradle.kts`. If the page
+   already has an `<!---IMPORT ...-->` line, the class it names tells you the owner. Don't split one topic
+   across both modules — both korro tasks would write the same file.
+2. **Add a `@Test @TransformDataFrameExpressions fun` to that module's sample class**, with the body wrapped in
```
Allex-Nik, line 42: @TransformDataFrameExpressions is added only in the `core` module

docs/StardustDocs/AGENTS.md
```diff
@@ -27,6 +27,41 @@ Project marker: `project.ihp`; instance profile / table-of-contents: `d.tree`
...
+   `// SampleStart` / `// SampleEnd`. New pages should go to `:samples` (migration #898); an existing `:core`
+   page keeps its samples next to its siblings.
+3. **Put `<!---FUN funName-->` / `<!---END-->` in the topic** and run korro to fill it in
```
Allex-Nik, line 45: If the sample is in the `samples` module, we also add the iframe line manually right after `<!---END-->`.
To produce `html`s, run samples as tests

docs/StardustDocs/AGENTS.md
```diff
@@ -27,6 +27,41 @@ Project marker: `project.ihp`; instance profile / table-of-contents: `d.tree`
...
+   (`./gradlew core:korro`, or `samples:korro`). Suffix the function `_properties` / `_strings` to get tabs.
+4. **Run it with `DATAFRAME_SAVE_OUTPUTS=1`** if you want the rendered result: korro then also injects an
+   `<inline-frame>` and writes the matching `resources/snippets/*.html`.
+   **Only annotate a sample with `@TransformDataFrameExpressions` when its last expression is a `DataFrame`
```
Allex-Nik, line 49: I might be wrong, but it seems to me that we don't need to add instructions related to adding samples to `core` at all (this is `core`-specific)

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/map.kt
```diff
@@ -34,4 +42,350 @@ class MapTests {
...
+
+        // the values themselves are put into the column unchanged, without any conversion
+        asNumber.values() shouldBe listOf(15, 45, 20, 20)
+        asNumber.values().map { it::class } shouldBe List(4) { Int::class }
+    }
+
+    @Test
+    fun `map with an explicit type and Infer Type uses that type as an upper bound only`() {
+        // Number is only the upper bound here; the actual type comes from the values
+        age.map(typeOf<Number>(), Infer.Type) { it }.type() shouldBe typeOf<Int>()
```
Allex-Nik, line 153: If we replace "Number" with "String" or "Boolean", the test still passes. So maybe "Number" is rather ignored? Might be relevant for the docs as well

core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/map.kt
```diff
@@ -34,4 +42,350 @@ class MapTests {
...
+    // the same data and the same calls as on the `map` page of the documentation website
+    private val df = dataFrameOf(
+        "firstName" to listOf("Alice", "Bob", "Charlie"),
+        "lastName" to listOf("Cooper", "Dylan", "Daniels"),
+        "age" to listOf(15, 45, 20),
+        "city" to listOf("London", "Dubai", "Moscow"),
+    ).group("firstName", "lastName").into("name")
+
+    @Test
+    fun `KDoc example - map over rows returns a list in row order`() {
```
Allex-Nik, line 187: There is also an option to reuse samples for the KDoc examples :)

https://jetbrains.slack.com/archives/C06GSHDKJVB/p1788012174146399

Allex-Nik (commented):
Left some comments :)

PR #2068, opened 2026-09-08
Added KDocs, tests, site updates for `indices`

### LINKED ISSUE #1965: Add KDoc, tests, and docs for `indices`
## Add KDoc, tests, and docs for `indices`

`core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/indices.kt` contains two small public APIs without KDocs:

```kotlin
public fun DataFrame<*>.indices(): IntRange = 0 until rowsCount()

public inline fun <T> DataFrame<T>.indices(filter: RowFilter<T>): List<Int>
```

There is no dedicated unit test file:
`core/src/test/kotlin/org/jetbrains/kotlinx/dataframe/api/indices.kt`

### What it does

`indices()` returns the row index range of a dataframe: `0 until rowsCount()`.

`indices { ... }` returns the indices of rows that satisfy the given row filter.

### Scope

- Add concise KDocs for both `indices` overloads.
- Add one minimal dedicated unit test.
- Add a short mention or example to the site docs, probably in the indexing / row access section.

### Test scenario

- Given a dataframe with 3 rows.
- `df.indices()` returns `0..2`.
- `df.indices { condition }` returns only indexes of matching rows.
- Empty dataframe returns an empty range/list.

### Acceptance criteria

- `DataFrame.indices()` and `DataFrame.indices { ... }` have KDocs.
- There is at least one direct unit test for both overloads.
- Site docs mention `indices` with a small example.

zaleslaw:
`indices()` returns the row indices of a `DataFrame` — the whole range, or, with a row filter, only the
indices of the matching rows. Both overloads were public but undocumented, absent from the website and
untested. The easy thing to get wrong: the filter overload returns `List<Int>` while the no-arg one returns
`IntRange`, and the result is meant to be fed back into a slice — which nothing in the API hinted at.

## Files

| File | What changed |
|---|---|
| `core/.../api/indices.kt` | KDoc for both overloads, plus a shared fragment with the website link and "See also" |
| `core/.../documentation/DocumentationUrls.kt` | `Indices` URL entry |
| `docs/.../topics/indexing.md`, `access.md` | new "Row indices" section; the page now shows its dataset |
| `core/.../samples/api/Access.kt` → `samples/.../IndexingSamples.kt` | the page's `getCell` samples moved to `:samples`, with assertions they lacked |
| `samples/build.gradle.kts`, `_shadow_resources.md` | register the topic and its two iframes |
| `docs/StardustDocs/AGENTS.md` | two korro gotchas: a topic must not span two modules; a new iframe needs `updateShadowResources` |
| `core/src/test/.../api/indices.kt` | new test class |

## Examples

The website examples use the standard ten-row `peopleDf`, so `city == "Moscow"` gives `[2, 6]` and the reader
can check it against the table now rendered at the top of the page. The KDoc example is a literal code block
rather than a `@sample`: Gradle resolves a `@sample` into `:samples`, but the IntelliJ KoDEx plugin does not,
and hovering the function then renders a stack trace instead of the docs.

## Tests

Nothing was covered before — neither overload, nor the column-level `indices`. Return types are asserted
through explicit type annotations, since `IntRange` vs `List<Int>` is part of the contract.

| Behavior | What it pins down |
|---|---|
| Full range | every row index, as an `IntRange` |
| Filtered indices | the documented example and the string-API form, in row order, as a `List<Int>` |
| `RowFilter` contract | the row arrives as both `this` and `it` |
| Empty results | empty frame and no-match both give an empty range / list, not a failure |
| Column overload | the function and the property agree; one index per row for a column group, one per cell for a frame column |

## Note for the reviewer

- **The `RowFilter` description is inlined, not `@include`d.** `SelectingRows.RowFilterSnippet` names the
  parameter `predicate`; `indices` calls it `filter`. Reusing the snippet would mean either making it
  parameterisable — a change to a file shared by eight operations — or renaming a public parameter. Three
  duplicated lines looked cheaper than either. `SelectingRows.kt` is untouched.
- **`AnyCol.indices` / `AnyBaseCol.indices()` are still undocumented.** Tested here, not documented, to keep
  the diff on one operation. KDoc them too, or a ticket — your call.
- **`getCell` changed module.** Adding the new section would have left one page importing from both `:core`
  and `:samples`, which blocks iframes and works against #898. `getCell` was used by this page only, so the
  move was two functions — but it puts an unrelated section in this diff. Alternative: a second
  `<!---IMPORT-->` and no table at the top.


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/indices.kt
```diff
@@ -1,13 +1,68 @@
...
+ */
+@ExcludeFromSources
+private typealias CommonIndicesDocs = Nothing
+
+/**
+ * Returns the indices of all rows in this [DataFrame],
+ * from `0` to the number of rows minus one,
+ * or an empty range if this [DataFrame] has no rows.
+ *
+ * {@include [CommonIndicesDocs]}
```
Jolanrensen, line 34: no braces necessary

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/indices.kt
```diff
@@ -1,13 +1,68 @@
...
+ *  - [rowsCount][DataFrame.rowsCount] — returns the number of rows in this [DataFrame].
+ *  - [getRows][DataFrame.getRows] — returns the rows at the given indices, as a new [DataFrame].
+ *  - [filter][DataFrame.filter] — returns the rows that satisfy a condition, instead of their indices.
+ */
+@ExcludeFromSources
+private typealias CommonIndicesDocs = Nothing
+
+/**
+ * Returns the indices of all rows in this [DataFrame],
+ * from `0` to the number of rows minus one,
```
Jolanrensen, line 31: This seems very obvious, but okay, why not :)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/indices.kt
```diff
@@ -1,13 +1,68 @@
...
+ * The [filter] is a [RowFilter] — a lambda that receives each [DataRow] as both `this` and `it`
+ * and is expected to return a [Boolean] value.
+ *
+ * It allows you to define conditions using the row's values directly,
+ * including through [extension properties][AccessApis.ExtensionPropertiesApi]
+ * for convenient and type-safe access.
+ *
+ * For more information, {@include [DocumentationUrls.DataRow.RowFilter]}
+ *
+ * {@include [CommonIndicesDocs]}
```
Jolanrensen, line 54: no braces

Jolanrensen: image

It looks a bit ugly with this duplicate "For more information". Even more so because one uses ":" and one ","

Issue #1630, opened 2025-12-10
Improve 1.0 Migration guide


AndreiKingsley:
Need to add

* Info about modules
* Info about Compiler Plugin (and improve https://kotlin.github.io/dataframe/migration-from-plugins.html)
* Info about changes in work with JDBC
* Most probably, something else

AndreiKingsley:
https://github.com/Kotlin/dataframe/pull/1631#discussion_r2610259941

zaleslaw:
@AndreiKingsley, please create a sub-issue

AndreiKingsley:
+ need to restructure (have a separate page for each of these parts)

zaleslaw:
Let's add an migration example 0/15->1.0 like Titanic

PR #2077, opened 2026-09-11
Document deprecation of `BigDecimal` and `BigInteger` statistics

AndreiKingsley:
Helps #1630

### DIFF
--- a/docs/StardustDocs/topics/cumSum.md
+++ b/docs/StardustDocs/topics/cumSum.md
@@ -30,3 +30,8 @@ df.groupBy { city }.cumSum { weight }.concat()
 ```
 
 <!---END-->
+
+> `java.math.BigDecimal` and `java.math.BigInteger` are not supported.
+> Count statistics manually with Kotlin standard library methods
+> and Java big numbers arithmetics or [`convert`](convert.md) them to primitive types.
+> {style="warning"}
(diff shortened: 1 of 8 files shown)


zaleslaw:
What if we add a simple workaround, as an example, with 1-2 statistics on how to handle it? As a recommendation

Jolanrensen:
The AI review is correct, only aggregations that reduce to a different type (by unifying numbers) don't support big numbers. They consider `Double` to be the most complex type (so `mean`, `sum`, `std`, etc.). Aggregations that select a value by order support any mutually-comparable type, including big numbers (so `medianBy`, `min`, `minBy`, etc.) `median`/`percentile` of `Number` indeed don't support big numbers and they will fail, so for that the message is okay.

Maybe we could write it like "big numbers are not explicitly supported anymore" hinting that in some cases it might still work, but if it doesn't, users just have to convert to primitives.

AndreiKingsley:
@Jolanrensen fixed

docs/StardustDocs/topics/cumSum.md
```diff
@@ -30,3 +30,35 @@ df.groupBy { city }.cumSum { weight }.concat()
...
+> or [`convert`](convert.md) the column to a primitive type first.
+> {style="warning"}
+
+For a `BigDecimal` column `amount`, the exact cumulative sum can be computed with Java `BigDecimal` arithmetic:
+
+<!---FUN cumSumBigNumbersManually-->
+
+```kotlin
+// exact cumulative sum, computed with Java `BigDecimal` arithmetic
+df.amount.toList().runningReduce(BigDecimal::add)
```
Jolanrensen: shouldn't we turn this back into a column to get the full workaround?

Jolanrensen (approved):

PR #2078, opened 2026-09-11
Added KDocs for `api/schema.kt` 

### LINKED ISSUE #1981: Add minimal KDocs for schema APIs
# Add minimal KDocs for schema APIs

## Motivation

`core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/schema.kt` exposes small public schema-inspection APIs, but most of them have no KDoc.

These functions are simple wrappers, but users still need a minimal explanation of what schema is returned and when to use each function:

- `DataFrame.schema()` extracts the runtime schema of a DataFrame.
- `DataRow.schema()` returns the schema of the owning DataFrame.
- `GroupBy.schema()` returns the schema after representing the grouped data as a DataFrame.
- `DataFrame.compileTimeSchema()` returns the schema known through the typed DataFrame/compiler-plugin path.

## Scope

Add minimal KDocs for:

```kotlin
DataRow<*>.schema()
DataFrame<*>.schema()
GroupBy<*, *>.schema()
DataFrame<T>.compileTimeSchema(ordered: Boolean = true)
```

## Acceptance Criteria

- Public functions in `schema.kt` have concise KDocs.
- KDocs distinguish runtime schema from compile-time schema.
- `DataRow.schema()` explains that it uses the row owner’s schema.
- `GroupBy.schema()` explains that it derives schema via `toDataFrame().schema()`.
- `compileTimeSchema(ordered)` documents the meaning of `ordered`.
- KDocs stay minimal and do not duplicate website-level documentation.

### Variant A: Very minimal

```kotlin
/** Returns the schema of the DataFrame that owns this row. */
public fun DataRow<*>.schema(): DataFrameSchema = owner.schema()

/** Returns the runtime schema of this DataFrame. */
@RequiredByIntellijPlugin
public fun DataFrame<*>.schema(): DataFrameSchema = extractSchema()

/** Returns the schema of this grouped data as represented by [toDataFrame]. */
public fun GroupBy<*, *>.schema(): DataFrameSchema = toDataFrame().schema()

/**
 * Returns the compile-time schema of this typed DataFrame.
 *
 * @param ordered if `true`, orders columns to match the runtime schema for easier comparison;
 * otherwise uses the column order produced by the compiler plugin.
 */
public inline fun <reified T> DataFrame<T>.compileTimeSchema(
    ordered: Boolean = true,
): DataFrameSchema = compileTimeSchemaImpl(if (ordered) schema() else null, T::class)
```

### Variant B: Slightly more explanatory

```kotlin
/**
 * Returns the schema of the DataFrame that owns this row.
 *
 * This is equivalent to asking the row owner for its runtime schema.
 */
public fun DataRow<*>.schema(): DataFrameSchema = owner.schema()

/**
 * Extracts the runtime schema of this DataFrame.
 *
 * The returned schema describes the columns currently present in the DataFrame.
 */
@RequiredByIntellijPlugin
public fun DataFrame<*>.schema(): DataFrameSchema = extractSchema()

/**
 * Returns the schema of this [GroupBy] after converting it to a DataFrame representation.
 */
public fun GroupBy<*, *>.schema(): DataFrameSchema = toDataFrame().schema()

/**
 * Returns the compile-time schema known for this typed DataFrame.
 *
 * This is useful for comparing the schema inferred from the DataFrame type with the runtime schema.
 *
 * @param ordered if `true`, columns are ordered like in the runtime schema for easier diffs;
 * otherwise columns keep the order used by the compiler plugin representation.
 */
public inline fun <reified T> DataFrame<T>.compileTimeSchema(
    ordered: Boolean = true,
): DataFrameSchema = compileTimeSchemaImpl(if (ordered) schema() else null, T::class)
```

zaleslaw:
Closes #1981.

`api/schema.kt` exposes four schema-inspection functions and none of them had a KDoc.

The thing that is easy to get wrong is that a `DataFrame` carries **two** schemas that do
not have to agree: the runtime one, read from the columns it holds, and the compile-time
one, read from the type argument `T` — they drift apart after a `cast` to a schema the data
does not have. 

The second trap is the `ordered` parameter of `compileTimeSchema`: with
`ordered = false` the column order is *not* the declaration order of `T`, it is the primary
constructor order for a `data class` marker and the reflection order otherwise.

## Files

| File | What changed |
|---|---|
| `core/…/api/schema.kt` | KDocs for the four functions; two `@ExcludeFromSources` KoDEx snippets for the shared runtime-vs-compile-time text and the shared example dataset |
| `core/…/schema/ColumnSchema.kt` | KDocs for the sealed class, its three per-kind subclasses and `compare` |
| `core/…/documentation/DocumentationUrls.kt` | new `Schema` entry for the `schema` page |
| `core/…/api/groupBy.kt`, `core/…/schema/DataFrameSchema.kt` | cross-links to the schema functions |
| `core/src/test/…/api/schema.kt` | tests for all four functions |
| `core/src/test/…/schema/ColumnSchemaTests.kt` | new — tests for `ColumnSchema` |
| `docs/StardustDocs/topics/schema.md` | one sentence reworded |

## Examples

The dataset and the calls are taken from the `schema` page on the website, so the page, the
KDocs and the tests all show the same example. Every expected output in the KDocs is an
asserted value in a `KDoc example …` test.

## Tests

Only `columns order test` existed before, and it compared the two schemas against each other
without pinning either. Expected values are explicit literals, never one implementation
compared to another.

| Behaviour | What it pins down |
|---|---|
| Runtime schema source | It comes from the columns and ignores `T` |
| Frame columns | Only the columns all the nested dataframes share; empty dataframes take nothing away |
| Row schema | Describes the whole dataframe, is equal for every row, and is the group's schema for a group row |
| `GroupBy` schema | Key columns first, then a frame column named `group` |
| `compileTimeSchema` order | `ordered = true` follows the columns and puts a column the runtime schema lacks first; `ordered = false` follows the constructor for a `data class` marker and reflection for an interface |
| `compileTimeSchema` source | Follows `T`, differs from runtime when `T` does not fit, and is empty without a marker |
| `ColumnSchema` | One subclass per column kind, nested schemas, and `compare` across kinds and by type |

## Note for the reviewer

- Scope: the ticket names only `api/schema.kt`. `ColumnSchema.kt`, `DataFrameSchema.kt`,
  `groupBy.kt` and `ColumnSchemaTests.kt` go beyond it (~105 of 465 added lines). 
- Outside this PR: `docs/StardustDocs/topics/schema.md` covers only `DataFrame.schema()` and
  `GroupBy.schema()`. It says nothing about the compile-time schema, `DataRow.schema()` or
  `compileTimeSchema()`, yet all four KDocs now point at it with "For more information".
- Outside this PR: that page links `DataFrameSchema` to `schemas.md`, which documents the
  `@DataSchema` annotation, not the `DataFrameSchema` type. Pre-existing; link left as it was.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/schema.kt
+++ b/core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/schema.kt
@@ -17,16 +17,11 @@ import org.jetbrains.kotlinx.dataframe.schema.DataFrameSchema
 internal interface SchemaDocs {
 
     /**
-     * A [DataFrame] has two schemas that do not have to agree.
-     * [schema][DataFrame.schema] gives the **runtime** one, read from the columns the [DataFrame]
-     * holds; [compileTimeSchema][DataFrame.compileTimeSchema] gives the **compile-time** one,
+     * A [DataFrame] has two schemas that do not have to agree:
+     * the **runtime** one, read from the columns it holds, and the **compile-time** one,
      * read from the type argument `T` of `DataFrame<T>`.
-     * The two differ when the type of a [DataFrame] does not fit its columns,
-     * for example after a [cast] to a schema the data does not have;
-     * comparing the two is how you notice.
-     *
-     * The runtime schema is also reachable from a [DataRow] — [schema][DataRow.schema] —
-     * and from a [GroupBy] — [schema][GroupBy.schema].
+     * They differ when the type does not fit the columns, for example after a [cast]
+     * to a schema the data does not have; comparing them is how you notice.
      *
      * Use [print][DataFrameSchema.print] to write a schema out as a tree,
      * and [compare][DataFrameSchema.compare] to relate two schemas to each other.
@@ -37,8 +32,7 @@ internal interface SchemaDocs {
     /**
      * {@comment The input of the `schema` examples below. KDoc-snippet.
      *    Every output that follows it is an expected value in `SchemaKDocExampleTests`.}
-     * The examples below use the [DataFrame] from the
-     * [`schema` page on the documentation website]({@include [DocumentationUrls.Url]}/schema.html):
+     * The examples below use the same [DataFrame] as the `schema` page on the documentation website:
      * a `name` [column group][ColumnGroup] holding `firstName` and `lastName`,
      * and the columns `age`, `city`, `weight` and `isHappy`.
      */
@@ -60,7 +54,6 @@ internal interface SchemaDocs {
  * For more information: {@include [DocumentationUrls.Schema]}
  *
  * ### Example
- *
  * @include [SchemaDocs.ExampleDataSnippet]
  *
  * ```kotlin
@@ -69,6 +62,8 @@ internal interface SchemaDocs {
  * ```
  *
  * @return The [DataFrameSchema] of the [DataFrame] this row is part of.
+ * @see [DataFrame.schema]
+ * @see [DataFrame.compileTimeSchema]
  */
 public fun DataRow<*>.schema(): DataFrameSchema = owner.schema()
 
@@ -90,7 +85,6 @@ public fun DataRow<*>.schema(): DataFrameSchema = owner.schema()
  * For more information: {@include [DocumentationUrls.Schema]}
  *
  * ### Example
- *
  * @include [SchemaDocs.ExampleDataSnippet]
  *
  * ```kotlin
@@ -110,6 +104,9 @@ public fun DataRow<*>.schema(): DataFrameSchema = owner.schema()
  * ```
  *
  * @return The [DataFrameSchema] of this [DataFrame].
+ * @see [DataFrame.compileTimeSchema]
+ * @see [DataRow.schema]
+ * @see [GroupBy.schema]
  */
 @RequiredByIntellijPlugin
 public fun DataFrame<*>.schema(): DataFrameSchema = extractSchema()
@@ -130,7 +127,6 @@ public fun DataFrame<*>.schema(): DataFrameSchema = extractSchema()
  * For more information: {@include [DocumentationUrls.Schema]}
  *
  * ### Example
- *
  * @include [SchemaDocs.ExampleDataSnippet]
  *
  * ```kotlin
@@ -152,6 +148,8 @@ public fun DataFrame<*>.schema(): DataFrameSchema = extractSchema()
  * ```
  *
  * @return The [DataFrameSchema] of this [GroupBy] as a [DataFrame].
+ * @see [GroupBy.toDataFrame]
+ * @see [DataFrame.schema]
  */
 public fun GroupBy<*, *>.schema(): DataFrameSchema = toDataFrame().schema()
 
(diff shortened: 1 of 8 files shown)


core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/schema.kt
```diff
@@ -2,34 +2,207 @@ package org.jetbrains.kotlinx.dataframe.api
...
+     * Use [print][DataFrameSchema.print] to write a schema out as a tree,
+     * and [compare][DataFrameSchema.compare] to relate two schemas to each other.
+     */
+    @ExcludeFromSources
+    typealias SchemaSourcesSnippet = Nothing
+
+    /**
+     * {@comment The input of the `schema` examples below. KDoc-snippet.
+     *    Every output that follows it is an expected value in `SchemaKDocExampleTests`.}
+     * The examples below use the same [DataFrame] as the `schema` page on the documentation website:
```
Jolanrensen: this sounds like it needs a link

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/schema.kt
```diff
@@ -2,34 +2,207 @@ package org.jetbrains.kotlinx.dataframe.api
...
+ * For more information: {@include [DocumentationUrls.Schema]}
+ *
+ * ### Example
+ * @include [SchemaDocs.ExampleDataSnippet]
+ *
+ * ```kotlin
+ * df.schema()
+ * ```
+ *
+ * Written out, a column group is shown by indentation and a frame column by `*`:
```
Jolanrensen: might be useful if the example below actually had a frame column too

Jolanrensen: like in GroupBy

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/schema/ColumnSchema.kt
```diff
@@ -92,6 +116,22 @@ public sealed class ColumnSchema {
...
+     * For [Value] the [types][type] are compared, for [Group] and [Frame] their
+     * [DataFrameSchema]s. How strict that comparison is, is decided by the [comparisonMode];
+     * see [ComparisonMode] for what each mode means.
+     *
+     * @param [other] The column schema to compare this one with.
+     * @param [comparisonMode] The [mode][ComparisonMode] to compare the column schemas by.
+     * @return a [CompareResult] that indicates whether this column schema compared to [other] is
+     *   [matching][CompareResult.Matches], [derived][CompareResult.IsDerived],
+     *   [superset][CompareResult.IsSuper], or [incomparable][CompareResult.None].
+     */
```
Jolanrensen, line 146: best to refer to `equals` too. Which is like "matches and also equals in order". This may be added to `ComparisonResult` in the future too.

Jolanrensen (changes requested):

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/schema.kt
zaleslaw: Agreed. Added a separate short example with a frame column after the group one, instead of putting one into the shared dataset snippet — pinned in SchemaKDocExampleTests.

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/schema/ColumnSchema.kt
zaleslaw: Added — compare now states it ignores column order, whereas equals requires the same order, plus an `@see`. Left CompareResult untouched for the future change you mentioned.

PR #2079, opened 2026-09-11
Added KDocs for `single` plus site changes and tests

### LINKED ISSUE #1982: Add KDocs for non-deprecated `single` APIs
# Add KDocs for non-deprecated `single` APIs

The file `core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/single.kt` holds public `single` / `singleOrNull` APIs. The value-level (`DataColumn`) and row-level (`DataFrame`) overloads currently have no KDocs, while their `ColumnsSelectionDsl` (`singleCol`) counterparts in the same file are already documented via `CommonSingleDocs`.

Please add concise KDocs for the current non-deprecated APIs only, consistent with the already-documented sibling files `first.kt` and `last.kt`. Deprecated and binary-compatibility overloads are out of scope.

### Scope

APIs needing KDocs (in `single.kt`, regions `// region DataColumn` and `// region DataFrame`):

- `DataColumn.*` — `single()`
- `DataFrame.*` — `single()`, `singleOrNull()`, `single(predicate)`, `singleOrNull(predicate)`

Already documented / out of scope: the `ColumnsSelectionDsl` and `singleCol` overloads, and all deprecated / binary-compatibility overloads.

### KDocs should explain

  - `DataColumn.single()` — the single value contained in the column.
  - `DataFrame.single()` — the single [row][DataRow] of the frame.
  - `single(predicate)` — the single row that satisfies the given row predicate.
  - the `*OrNull` variants — the same result, but `null` instead of throwing.
- The difference between the throwing variants and the `*OrNull` variants.
- Empty / non-unique behavior:
  - non-null variants throw `NoSuchElementException` when there is no matching element and `IllegalArgumentException` when there is more than one;
  - `singleOrNull` returns `null` both when empty and when there is more than one element.
- For the predicate overloads: the predicate is a row expression evaluated per row, and exactly one row must match.
- `See also` cross-links to the related `first` / `last` / `singleOrNull` operations, matching the "See also" blocks used in `first.kt` / `last.kt`.

### Acceptance criteria

- Concise KDocs are added to all non-deprecated `DataColumn` / `DataFrame` `single` and `singleOrNull` overloads in `single.kt`.
- No documentation is added for deprecated or binary-compatibility overloads.
- KDocs state the return value and the exact empty / non-unique behavior (`NoSuchElementException`, `IllegalArgumentException`, or `null`).
- Style and `See also` cross-references are consistent with the already-documented `first.kt` / `last.kt`.

zaleslaw:
Closes #1982 

`single`/`singleOrNull` are the "exactly one" accessors, and they were the last of the
`first`/`last`/`single` family with no KDoc, no website page for the `DataColumn` overload, and no
tests outside the column-selection DSL. What is easy to get wrong is that the four overloads fail in
three different ways — `NoSuchElementException` on an empty frame, `IllegalArgumentException` on more
than one row, `null` for the `OrNull` pair — and that `DataColumn.single()` legitimately returns
`null` when its one value is `null`. This PR documents that and pins all of it.

## Files

| File | What changed |
|---|---|
| `core/…/api/single.kt` | KDoc for all four overloads; predicate type aligned with `first` (`RowExpression<T, Boolean>` → `RowFilter<T>`, same typealias) |
| `core/…/api/first.kt`, `last.kt` | reciprocal `See also [single]` |
| `core/…/documentation/DocumentationUrls.kt` | three website links |
| `core/…/documentation/ExpressionsGivenRow.kt`, `SelectingRows.kt` | typo `Fore more` → `For more` (see note) |
| `core/src/test/…/api/single.kt` | new `DataColumn` and `DataFrame` regions |
| `docs/…/topics/single.md` | exact exception types, Korro examples, summaries |
| `docs/…/topics/operations/column/singleOnColumn.md` (new), `d.tree`, `columnOperations.md` | new column page, registered and linked |
| `samples/…/api/SingleSamples.kt`, `…/api/column/SingleOnColumnSamples.kt` (new), `samples/build.gradle.kts` | asserting Korro samples; `include("single.md")` |
| `docs/…/resources/api/single/*.html`, `topics/_shadow_resources.md` | generated iframes |
| `docs/StardustDocs/AGENTS.md` | how a new topic under a glob-included folder is picked up |

## Examples

The rows page uses the shared `peopleDf`, the column page a 4-row frame; every `// returns …`
comment in a sample is backed by a `shouldBe` in the same test, so the numbers cannot drift.

## Tests
`single` had no tests beyond the DSL. Types and expected values are written out explicitly so the
overload under test is unambiguous.

| Test | What it pins down |
|---|---|
| `single on DataColumn`, `… empty`, `… more than one value`, `… a single null value` | the column contract, including that a lone `null` is returned, not thrown on |
| `single`/`singleOrNull on DataFrame`, `… empty`, `… more than one row` | the no-argument pair |
| `single on empty DataFrame throws`, `… more than one row throws` | the two hand-written exception messages, including the `Use singleOrNull` hint |
| predicate variants: match, no match, several matches, empty frame | which of the two exceptions each case raises, and where `null` replaces it |
| predicate over a nested column, a nullable column, `it`, and the String access API | the reach the included `RowFilter` snippet promises |


docs/StardustDocs/topics/single.md
```diff
@@ -1,7 +1,99 @@
...
+```
+
+<!---END-->
+<inline-frame src="./resources/singleDf.html" width="100%" height="500px"></inline-frame>
+
+<!---FUN single-->
+
+```kotlin
+val dylan = df
+    .filter { age == 45 } // one row is left after filtering
```
Allex-Nik, line 36: We could add the String API version as well.

docs/StardustDocs/topics/single.md
```diff
@@ -1,7 +1,99 @@
...
 
-Returns the single [row](DataRow.md) that matches the given [condition](DataRow.md#row-conditions), or `null` if there is no or more than one matching row.
+Returns the single [row](DataRow.md) in this [`DataFrame`](DataFrame.md),
+or `null` if the [`DataFrame`](DataFrame.md) is empty or has more than one row.
+
+<!---FUN singleOrNull-->
+
+```kotlin
+val noOne = df
+    .filter { age > 50 } // df is empty after filtering
```
Allex-Nik, line 73: We could add the String API version as well

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/single.kt
```diff
@@ -33,25 +36,127 @@ import kotlin.reflect.KProperty
 
 // region DataColumn
 
+/**
+ * Returns the single value in this [DataColumn].
+ *
+ * For more information: {@include [DocumentationUrls.SingleOnColumn]}
+ *
+ * See also [firstOrNull][DataColumn.firstOrNull], that returns `null` instead of throwing
```
Allex-Nik, line 44: `firstOrNull` also does not require the column to have exactly one value. Looking at the next overload, maybe `singleOrNull` was meant, but it does not exist on `DataColumn`. 

Would it be better to list these operations one by one (or in pairs like first/last, take/takeLast) and add short explanations to each entry regarding what they do? Something like "returns the first/last element in the column" for first/last, etc.

Maybe `firstOrNull` could even be omitted in this particular overload (optionally)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/single.kt
```diff
@@ -33,25 +36,127 @@ import kotlin.reflect.KProperty
...
 // region DataFrame
 
+/**
+ * Returns the single [row][DataRow] in this [DataFrame].
+ *
+ * For more information: {@include [DocumentationUrls.Single]}
+ *
+ * See also [singleOrNull][DataFrame.singleOrNull], that returns `null` instead of throwing,
+ * and [first][DataFrame.first], [last][DataFrame.last],
+ * that do not require the [DataFrame] to contain exactly one matching row.
```
Allex-Nik, line 69: Might be obvious, but I would still explicitly tell "... that return the first/last row" because this is also an important difference (also applicable to other overloads)

core/src/main/kotlin/org/jetbrains/kotlinx/dataframe/api/single.kt
```diff
@@ -33,25 +36,127 @@ import kotlin.reflect.KProperty
...
+ *
+ * @return A [DataRow] containing the single row in this [DataFrame].
```
Allex-Nik, line 71: "The single [DataRow] in this [DataFrame]"? (also applicable to other overloads)
Probably leaving as is is also fine

docs/StardustDocs/topics/operations/column/singleOnColumn.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+<!---END-->
+<inline-frame src="./resources/singleOnColumnDf.html" width="100%" height="500px"></inline-frame>
+
+<!---FUN singleOnColumn-->
+
+```kotlin
+df
+    .filter { name == "Bob" } // one row is left after filtering
```
Allex-Nik, line 36: We could add the String API version here as well

docs/StardustDocs/topics/single.md
```diff
@@ -1,7 +1,99 @@
...
+```kotlin
+df
+```
+
+<!---END-->
+<inline-frame src="./resources/singleDf.html" width="100%" height="500px"></inline-frame>
+
+<!---FUN single-->
+
+```kotlin
```
Allex-Nik, line 34: We can show an iframe for this sample in the same way it is done for the sample above. See the comments for `SingleSamples.kt`

The same applies to `FUN singleCondition`

docs/StardustDocs/topics/single.md
```diff
@@ -1,7 +1,99 @@
...
+val dylan = df
+    .filter { age == 45 } // one row is left after filtering
+    .single()
+```
+
+<!---END-->
+
```
Allex-Nik, line 41: Once `single.html` is generated,

```
<inline-frame src="./resources/single.html" width="100%" height="500px"></inline-frame>" 
```

can be added here

samples/src/test/kotlin/org/jetbrains/kotlinx/dataframe/samples/api/SingleSamples.kt
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Test
+    fun single() {
+        // SampleStart
+        val dylan = df
+            .filter { age == 45 } // one row is left after filtering
+            .single()
+        // SampleEnd
+        dylan.name.lastName shouldBe "Dylan"
```
Allex-Nik, line 28: We can do it like this:
```kotlin
    .also { it.name.lastName shouldBe "Dylan" }
    .toDataFrame()
    .saveDfHtmlSample()
```
Then the iframe for this row can be generated by running this test.

AGENTS.md
```diff
@@ -84,8 +84,13 @@ Public KDocs use KoDEx notations (`{@include [X]}`, `@set`/`@get`/`$`, `@sample`
   `**/generated-dataschema-accessors/**`) just doubles every hit and bloats context — do code work from
   `src/main/kotlin/…`. The one reason to open a generated file is the KDoc: it is the **fully-expanded**
   documentation (all `{@include …}`/`@set`/`@get` resolved, URLs inlined). If you specifically need a symbol's
-  final rendered docs, open that single generated file deliberately — don't explore. A CI bot regenerates and
-  auto-commits all of these on `master` after merge — you don't run/commit generation yourself.
+  final rendered docs, open that single generated file deliberately — don't explore. A CI bot runs
+  `processKDocsMain korro syncExampleFolders` and auto-commits `*/generated-sources`, the accessors,
```
Allex-Nik, line 88: I checked `generated-dataschema-accessors` in `core` and `dataframe-geo` - they were committed manually. @Jolanrensen the CI bot doesn't regenerate and auto-commit the accessors, right?

Allex-Nik (commented):
I think we could add a couple more iframes to the existing samples as shown in the comments, add missing String API versions. 

There are some other suggestions in the comments.

Otherwise looks good, I like the docs!

AGENTS.md
Jolanrensen: Indeed, they cannot be regenerated anymore since we deprecated our KSP/Gradle plugin

Jolanrensen: But it's not mentioned anywhere here right?

Allex-Nik: The file says: "Every module built with KoDEx ... has a tracked `generated-sources/` copy,
  plus checked-in accessors under `**/src/generated-dataschema-accessors/` ... A CI bot runs
  `processKDocsMain korro syncExampleFolders` and auto-commits `*/generated-sources`, the accessors, ..."

PR #2098, opened 2026-09-21
Fix Access APIs samples

Allex-Nik:
Helps #898 

- Removed `ApiLevels.kt` (and its generated accessors) with unused duplicated samples regarding `titanic.csv`: these samples were copied to `AccessApis.kt` before.
- The samples regarding `titanic.csv` didn't work with the `titanic.csv` dataset that we provide on the website because of a different delimiter and differences in column names and types. Even though these samples are marked as `@Ignored` and it does not cause errors when running tests, these samples are demonstrated to the users, and on the webpage we suggest users download the dataset. So if users do it and try our samples with the dataset, their code will fail. Therefore, I corrected the samples so that they work with the dataset.


AndreiKingsley (approved):
Nice!

AndreiKingsley:
Maybe worth to add `titanic.csv` as website resource.
#2106 should help

PR #2100, opened 2026-09-22
[Air] Remove redundant {} braces around line-starting @include tags in min.kt/max.kt

Jolanrensen:
## Summary

In `min.kt` and `max.kt`, many KDoc `@include` tags were written as `{@include [...]}` even though they start their own line. Per [KODEX_KDOC_PREPROCESSING.md](../blob/master/KODEX_KDOC_PREPROCESSING.md), a tag that starts a line should be written **without** `{}` — the braces there change nothing and only add noise.

This PR removes the redundant braces for those line-starting `@include` tags (140 in each file).

## What was intentionally left untouched
- `@comment` tags — they consume everything fed to them, so their `{}` scoping is required.
- `@set` tags and any tags that appear **mid-line** or inside backticks/code blocks — braces are still required there.

## Verification
Ran `./gradlew :core:processKDocsMain` on both the original (master) sources and the edited sources and diffed the regenerated KDoc output for `min.kt` and `max.kt`: the output is **byte-for-byte identical**, confirming the change is purely cosmetic in the source and does not affect generated documentation.

The `generated-sources` are intentionally not committed here (they are auto-committed by the CI bot).


Jolanrensen:
Thanks air, very cool

